PR #10, opened 2023-11-07
Add support for AIE translator

yzhang93:


makslevental:
Just curious: what's the plan here? To emit AIE dialect and then use `aie-translate` to emit LLVM IR? Or to actually use passes in AIE?

MaheshRavishankar (commented):
So we dont have tests here yet right?

MaheshRavishankar:
> Just curious: what's the plan here? To emit AIE dialect and then use `aie-translate` to emit LLVM IR? Or to actually use passes in AIE?

This is just an off-ramp we are using to connect to some other flows. We have this in the short term to check for correctness and build up some functionality that the other flows expect (and will eventually be needed). Long term we should just be able to generate LLVM IR and compile it using LLVM to native executable.... there are a few steps to get us there, but thats what we are hoping for.

yzhang93:
> So we dont have tests here yet right?

Just added the tests. It passed if you run `ctest -R amd-aie`.

MaheshRavishankar (approved):

PR #11, opened 2023-11-08
Fix the issue when gemm inputs are from subview ops

yzhang93:


compiler/plugins/target/AMD-AIE/Translation/AIESerializer.cpp
```diff
@@ -354,6 +354,14 @@ FailureOr<std::string> getOperandStr(Value operand,
     return failure();
   }
 
+  while (!scope.symbolTable.count(operand)) {
```
MaheshRavishankar, line 357: This looks similar to the `walkSubViews`. Can we just call that method?

compiler/plugins/target/AMD-AIE/Translation/AIESerializer.cpp
```diff
@@ -627,7 +635,7 @@ LogicalResult AccelSerializer::processOperation(scf::ForOp forOp,
   }
 
   for (Operation &op : *forOp.getBody()) {
-    if (failed(processOperation(&op, subScope))) {
+    if (failed(processOperation(&op, scope))) {
```
MaheshRavishankar, line 638: This change shouldnt be needed? subScope should have everything scope has + induction variable.

MaheshRavishankar (changes requested):

compiler/plugins/target/AMD-AIE/Translation/AIESerializer.cpp
yzhang93: I have to use `scope` to make indentation correct for the contents within the for loop.

MaheshRavishankar: But scope wont have the induction variable... surprising that the indentation goes off if you use subscope... 
If its just an indentation issue, could you change this to subScope again. Maybe something else is off with the indentation, but thats cosmetic.

MaheshRavishankar (approved):
Approved based on using subscope again.

PR #13, opened 2023-11-10
Add AIR transform and AIE dialect

stephenneuendorffer:
This should be enough to run the air-to-aie pass.


compiler/plugins/target/AMD-AIE/aie/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+
+iree_cc_library(
+    NAME
+        AIEXDialectIR
+    SRCS
+      ${IREE_MLIR_AIE_SOURCE_DIR}/lib/Dialect/AIEX/IR/AIEXDialect.cpp
+    DEPS
+    ::defs
+    ::AIEXDialectGen
+    ::AIEXOpsGen
```
makslevental, line 112: Just curious: is this sorcery or just that CMake allows `::` in target identifiers?

stephenneuendorffer: I believe cmake allows it in identifiers and with some cmake magic, iree uses this to explicitly reference things elsewhere in the build hierarchy

stellaraccident: They are technically "ALIAS" targets and a small bit of string munging for the :: prefix case. CMake being CMake, they are a thing all their own vs unified with everything else. One power they have is that CMake will hard error if you refer to one that does not exist (i.e. vs trying to link to a library on the system with the name).

PR #20, opened 2023-11-15
Fix dyn_cast compiler deprecation error

newling:
I hit the deprecation error with the following setup:

iree, iree-amd-aie: both on main branch, pulled afternoon of Nov 14


nirvedhmeshram:
This commit is causing an error for me
```
/home/nirvedh/iree/third_party/llvm-project/llvm/include/llvm/Support/Casting.h:64:57: error: no member named 'classof' in 'mlir::AffineBinaryOpExpr'
  static inline bool doit(const From &Val) { return To::classof(&Val); }
  ```
  works after reverting..

nirvedhmeshram:
oh so the problem was I was on an older branch of IREE from early in the week, once I pulled latest IREE it works with this commit and fails otherwise, thanks for the fix!

PR #22, opened 2023-11-18
Add XRT device and wire it up with the driver, add event semaphore

nirvedhmeshram:
- Add device
- Add some dump information and driver query functions
- Add event semaphore and creates it from device, although I dont think these will be used by the runtime in the near future
- (we will add a blocking sync call in queue execute in the near future when we implement queue functions next, provide thread per queue in the mid future and eventually do a full async execution

### DIFF
--- a/runtime/src/iree-amd-aie/driver/xrt/CMakeLists.txt
+++ b/runtime/src/iree-amd-aie/driver/xrt/CMakeLists.txt
@@ -22,10 +22,6 @@ iree_add_all_subdirs()
 
 find_package(XRT REQUIRED)
 
-set(xrt_core_LIBRARY XRT::xrt_core)
-set(xrt_coreutil_LIBRARY XRT::xrt_coreutil)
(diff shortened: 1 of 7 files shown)


runtime/src/iree-amd-aie/driver/xrt/event_semaphore.h
```diff
@@ -0,0 +1,28 @@
+// Copyright 2021 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#ifndef IREE_AMD_AIE_DRIVER_XRT_SEMAPHORE_H_
```
antiagainst, line 7: Makes sure the header guard follows the file path.

antiagainst (changes requested):

antiagainst (approved):

PR #29, opened 2023-11-28
[ci] Fetch submodules and always update cache.

stellaraccident:


.github/workflows/ci.yml
```diff
@@ -44,7 +44,7 @@ jobs:
       - name: Install deps
         run: |
           apt update
-          apt install libboost-all-dev
+          apt install -y libboost-all-dev
```
makslevental, line 47: just a btw: there's a [marketplace action for a this](https://github.com/Xilinx/mlir-air/blob/wheels_1/.github/workflows/distro.yml#L102-L108) that's more "portable".

stellaraccident: I am going to spend no time on this. Boost must go.

stellaraccident: (less snark: I'll just add it to the docker image that is behind this)

stellaraccident: Reverted this change (it is in the GH actions runner image now)

PR #30, opened 2023-11-28
Update pad based TD script to generate the expected input for AIE lowering pipeline

MaheshRavishankar:
This removes the hand-edited IR that was used as input for the AIE lowering pipeline. Instead, the test is made to start from the input to translate executables and use the transform dialect script to lower to the AIE pipeline.


nirvedhmeshram (approved):
LGTM!

PR #36, opened 2023-11-30
Add passes to lower AIE Device code to LLVM.

MaheshRavishankar:
This does not get us all the way to binary, but the sample lowers to LLVM successfully.
Next step is to generate the actual binaries.


stephenneuendorffer (approved):

MaheshRavishankar:
> Qq: can this be fed to peano only or chess as well?

I hope both. I havent gotten there yet. Thats the next thing to do.

stephenneuendorffer:
> Qq: can this be fed to peano only or chess as well?

We can feed to both,  Mainly we target Peano, and insert workarounds for Chess.  The Chess workarounds are:
1) IR textual format workarounds, due to different LLVM versions
2) Intrinsic differences, so we link with a small library to redirect intrinsics to what Chess expects.

erwei-xilinx (approved):

PR #38, opened 2023-12-01
[AMD-AIE] Update mlir-air and mlir-aie submodules

Abhishek-Varma:
Updates `mlir-air` and `mlir-aie` submodules.
1. `mlir-air` contains the latest set of commits which deals with the AIR Conversion/Transform PassOptions change which @MaheshRavishankar added.
2. `mlir-aie` could've been a "clean" submodule mimicking upstream branch itself - but the delta is just one change that I added here: [nod-ai/mlir-aie/iree_fixes](https://github.com/nod-ai/mlir-aie/tree/iree_fixes).


nirvedhmeshram (approved):
LGTM, might be nice to fix the upstream mlir-aie error as a next step (there should be some switch not using all enums?) cc @makslevental

PR #39, opened 2023-12-01
[XRT] Add native executable

nirvedhmeshram:
A flatbuffer format that allows multiple entry points for each xclbin is added here. Each entry point will have a xrt::kernel associated with it which will be extracted from the xclbin. Similarly, each entry point will have a assembly instruction stream that needs to be passed to the kernel call that is loaded in the flatbuffer as well.

The parameters required for executable include xrt::kernel and xrt::bo C++ objects with the asm instructions. Since these cant be saved in the POD struct we save pointers to these and support exception handling of this memory allocation process.

### DIFF
--- a/runtime/src/iree-amd-aie/driver/xrt/native_executable.cc
+++ b/runtime/src/iree-amd-aie/driver/xrt/native_executable.cc
@@ -1,0 +1,257 @@
+// Copyright 2023 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree-amd-aie/driver/xrt/native_executable.h"
+
+#include <stddef.h>
+
+#include "iree-amd-aie/schemas/xrt_executable_def_reader.h"
+#include "iree-amd-aie/schemas/xrt_executable_def_verifier.h"
+#include "iree/base/api.h"
+#include "iree/base/internal/flatcc/parsing.h"
+
+typedef struct iree_hal_xrt_native_executable_t {
+  // Abstract resource used for injecting reference counting and vtable; must be
+  // at offset 0.
+  iree_hal_resource_t resource;
+
+  iree_allocator_t host_allocator;
+
+  iree_host_size_t entry_point_count;
+  iree_hal_xrt_kernel_params_t entry_points[];
+} iree_hal_xrt_native_executable_t;
+
+namespace {
+extern const iree_hal_executable_vtable_t iree_hal_xrt_native_executable_vtable;
+}  // namespace
+
+static iree_hal_xrt_native_executable_t* iree_hal_xrt_native_executable_cast(
+    iree_hal_executable_t* base_value) {
+  IREE_HAL_ASSERT_TYPE(base_value, &iree_hal_xrt_native_executable_vtable);
+  return (iree_hal_xrt_native_executable_t*)base_value;
+}
+
+// Verifies the structure of the flatbuffer so that we can avoid doing so during
+// runtime.
+//
+// There are still some conditions we must be aware of (such as omitted names on
+// functions with internal linkage), however we shouldn't need to bounds check
+// anything within the flatbuffer after this succeeds.
+static iree_status_t iree_amd_aie_hal_xrt_native_executable_flatbuffer_verify(
+    iree_const_byte_span_t flatbuffer_data) {
+  if (!flatbuffer_data.data || flatbuffer_data.data_length < 16) {
+    return iree_make_status(
+        IREE_STATUS_INVALID_ARGUMENT,
+        "flatbuffer data is not present or less than 16 bytes (%zu total)",
+        flatbuffer_data.data_length);
+  }
+
+  // Run flatcc generated verification. This ensures all pointers are in-bounds
+  // and that we can safely walk the file, but not that the actual contents of
+  // the flatbuffer meet our expectations.
+  int verify_ret = iree_amd_aie_hal_xrt_ExecutableDef_verify_as_root(
+      flatbuffer_data.data, flatbuffer_data.data_length);
+  if (verify_ret != flatcc_verify_ok) {
+    return iree_make_status(IREE_STATUS_INVALID_ARGUMENT,
+                            "flatbuffer verification failed: %s",
+                            flatcc_verify_error_string(verify_ret));
+  }
+
+  iree_amd_aie_hal_xrt_ExecutableDef_table_t executable_def =
+      iree_amd_aie_hal_xrt_ExecutableDef_as_root(flatbuffer_data.data);
+
+  flatbuffers_string_vec_t entry_points_vec =
+      iree_amd_aie_hal_xrt_ExecutableDef_entry_points_get(executable_def);
+  size_t entry_point_count = flatbuffers_string_vec_len(entry_points_vec);
+  for (size_t i = 0; i < entry_point_count; ++i) {
+    if (!flatbuffers_string_len(
+            flatbuffers_string_vec_at(entry_points_vec, i))) {
+      return iree_make_status(IREE_STATUS_INVALID_ARGUMENT,
+                              "executable entry point %zu has no name", i);
+    }
+  }
+
+  flatbuffers_string_t xclbin =
+      iree_amd_aie_hal_xrt_ExecutableDef_xclbin_get(executable_def);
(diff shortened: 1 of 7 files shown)


nirvedhmeshram:
@stellaraccident I also wanted to check your thoughts on the organization of the schemas directory I am adding in this PR. We will need this accessed in the compiler plugin as well. From what I remember we want the runtime/compiler plugin isolated but do we just include this directory by giving some relative path in the compiler when we need to use it or is there a more elegant solution?

stellaraccident:
Compiler can depend on runtime common code, but shouldn't depend on "tricky" things: shared schemas are perfectly fine. It would be good to keep the property of being able to build the compiler without having xrt setup: the other backends don't have such a thick dependency that would preclude even getting CMake to configure, so we haven't had to slice it this way yet.

This might call for two top level directories: runtime and runtime_xrt or something. Or runtime_common and runtime.

runtime/src/iree-amd-aie/schemas/xrt_executable_def.fbs
```diff
@@ -0,0 +1,37 @@
...
+file_identifier "XRTR";
+file_extension "xrtr";
+
+// Source code location denoted by a file name and line within that file.
+table FileLineLocDef {
+  filename:string;
+  line:int32;
+}
+
+// assembly instructions
```
antiagainst, line 19: Capitalize the first word and end it with `.`. Applies to other comments in this file--we need to be consistent regarding the comment style.

runtime/src/iree-amd-aie/schemas/xrt_executable_def.fbs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+table ExecutableDef {
+  // A map of entry point ordinals to string names as used by XRT
+  entry_points:[string];
+
+  // XCLBIN string of the module.
+  xclbin:string;
+
+  // assembly instructions to run for each kernel
```
antiagainst, line 31: What is the size requirement of the array? It should be the same as `entry_points`? It would be nice to be clear here.

runtime/src/iree-amd-aie/driver/xrt/api.h
```diff
@@ -16,6 +16,38 @@
 extern "C" {
 #endif  // __cplusplus
 
+//===----------------------------------------------------------------------===//
+// iree_hal_xrt_device_params_t
+//===----------------------------------------------------------------------===//
+
+typedef enum iree_hal_xrt_command_dispatch_type_e {
```
antiagainst, line 23: Do we actually have these two modes? I think we can only do serialized dispatch in XRT?

runtime/src/iree-amd-aie/driver/xrt/native_executable.h
```diff
@@ -0,0 +1,48 @@
...
+#include "xrt/xrt_device.h"
+#include "xrt/xrt_kernel.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif  // __cplusplus
+
+// Object and launch parameters for a compute kernel.
+typedef struct iree_hal_xrt_kernel_params_t {
+  xrt::kernel* kernel;
```
antiagainst, line 24: Can we put comments to explain what `kernel` and `instr`are? And how their lifetime is managed? (Given we are playing with C++ under the hood..)

runtime/src/iree-amd-aie/driver/xrt/native_executable.cc
```diff
@@ -0,0 +1,257 @@
...
+                            "flatbuffer verification failed: %s",
+                            flatcc_verify_error_string(verify_ret));
+  }
+
+  iree_amd_aie_hal_xrt_ExecutableDef_table_t executable_def =
+      iree_amd_aie_hal_xrt_ExecutableDef_as_root(flatbuffer_data.data);
+
+  flatbuffers_string_vec_t entry_points_vec =
+      iree_amd_aie_hal_xrt_ExecutableDef_entry_points_get(executable_def);
+  size_t entry_point_count = flatbuffers_string_vec_len(entry_points_vec);
```
antiagainst, line 68: Should we check that `entry_point_count` is not zero?

runtime/src/iree-amd-aie/driver/xrt/native_executable.cc
```diff
@@ -0,0 +1,257 @@
...
+  if (flatbuffers_string_len(xclbin) == 0) {
+    return iree_make_status(IREE_STATUS_INVALID_ARGUMENT, "no xclbin present");
+  }
+
+  iree_amd_aie_hal_xrt_AsmInstDef_vec_t asm_instr =
+      iree_amd_aie_hal_xrt_ExecutableDef_asm_instrs_get(executable_def);
+  if (iree_amd_aie_hal_xrt_AsmInstDef_vec_len(asm_instr) != entry_point_count) {
+    return iree_make_status(
+        IREE_STATUS_INVALID_ARGUMENT,
+        "number of entry points and asm instructions mismatched");
```
antiagainst, line 88: Can we also print out the number of entry points the number asm instructions to be clear?

runtime/src/iree-amd-aie/schemas/xrt_executable_def.fbs
```diff
@@ -0,0 +1,37 @@
...
+// assembly instructions
+table AsmInstDef {
+  asm_inst:[uint32];
+}
+
+table ExecutableDef {
+  // A map of entry point ordinals to string names as used by XRT
+  entry_points:[string];
+
+  // XCLBIN string of the module.
```
antiagainst, line 28: Can you add some description of what `xclbin` and `asm_instrs` mean for XRT in general? It's not straightforward to tell from the names.

runtime/src/iree-amd-aie/driver/xrt/native_executable.cc
```diff
@@ -0,0 +1,257 @@
...
+      iree_amd_aie_hal_xrt_ExecutableDef_entry_points_get(executable_def);
+  iree_host_size_t entry_count = flatbuffers_string_vec_len(entry_points_vec);
+
+  flatbuffers_string_t xclbin_fb =
+      iree_amd_aie_hal_xrt_ExecutableDef_xclbin_get(executable_def);
+
+  iree_amd_aie_hal_xrt_AsmInstDef_vec_t asm_instrs_vec =
+      iree_amd_aie_hal_xrt_ExecutableDef_asm_instrs_get(executable_def);
+
+  iree_host_size_t entry_point_count =
```
antiagainst, line 121: Isn't this the same as `entry_count`?

runtime/src/iree-amd-aie/driver/xrt/native_executable.cc
```diff
@@ -0,0 +1,257 @@
...
+        iree_amd_aie_hal_xrt_AsmInstDef_vec_at(asm_instrs_vec, i);
+    flatbuffers_uint32_vec_t asm_inst =
+        iree_amd_aie_hal_xrt_AsmInstDef_asm_inst_get(asminst_def);
+    uint32_t num_instr = flatbuffers_uint32_vec_len(asm_inst);
+    std::unique_ptr<xrt::kernel> kernel;
+    std::unique_ptr<xrt::bo> instr;
+    try {
+      auto kernel = std::make_unique<xrt::kernel>(context, entry_name);
+      auto instr = std::make_unique<xrt::bo>(
+          device, num_instr * sizeof(uint32_t), XCL_BO_FLAGS_CACHEABLE,
```
antiagainst, line 166: Please document here regarding the `XCL_BO_FLAGS_CACHEABLE` and group id 0.. It's a great source of confusion. We'd want others checking the code to be able to understand why such magic symbols / numbers.

runtime/src/iree-amd-aie/driver/xrt/native_executable.cc
```diff
@@ -0,0 +1,257 @@
...
+    }
+    uint32_t* instr_buffer = instr->map<uint32_t*>();
+    for (int j = 0; j < num_instr; j++) {
+      instr_buffer[j] = flatbuffers_uint32_vec_at(asm_inst, j);
+    }
+    iree_hal_xrt_kernel_params_t* params = &executable->entry_points[i];
+    params->kernel = kernel.release();
+    params->instr = instr.release();
+    params->num_instr = num_instr;
+    params->layout = executable_params->pipeline_layouts[i];
```
antiagainst, line 183: We need to retain the `params->layout` to make sure it's still available while this executable is live?

antiagainst (changes requested):

antiagainst (approved):

PR #42, opened 2023-12-05
Add core elf-generation methods. 

MaheshRavishankar:
The AIE Core elf generation needs mlir-aie installation. So this PR
adds a flag `iree-amd-aie-mlir-aie-install-dir` that needs to point to
the MLIR-AIE installation path. This is meant to be a temporary
situation for initial binary generation prototyping.

Refactor some of the serialization logic to allow for cloning of the
ModuleOp. This is required since the same initial module is used in
multiple binary generation flows.


Abhishek-Varma (approved):
LGTM!

nirvedhmeshram (approved):
LGTM!

PR #43, opened 2023-12-05
Transform -> C++

Abhishek-Varma:
[DRAFT][WIP]

This is the initial effort to pull through the "Transform -> C++".

The current revision has all the necessary pieces (ROUGH DRAFT) that I'd need to slowly churn out a final C++ equivalent of the transform dialect script we have.

I'm not sure if this repository is the ideal place for this support to be in or it should be in IREE, but there are bigger fishes to catch right now -> at least get through with each invocation step properly.


Abhishek-Varma:
Not required.

PR #44, opened 2023-12-07
Finishing up xclbin generation.

MaheshRavishankar:
With this PR the binary translation adds all the functionality needed to generate the two binary pieces that IREE needs to generate. To get here  a new option, `--iree-amd-aie-vitis-install-dir`, was needed to point to the Vitis installation directory to pick up the libraries and binaries needed for the final xclbin generation.
This effectively now is capable of replicating the flow in aiecc.py using the following command

```
SAMPLES_DIR=<iree-amd-aie source dir>/tests/samples
iree-compile  --iree-hal-target-backends=amd-aie  \
 ${SAMPLES_DIR}/matmul_fill_static_i32.mlir \
  --iree-codegen-transform-dialect-library=${SAMPLES_DIR}/matmul_fill_spec_pad.mlir \
  --iree-amd-aie-peano-install-dir=<peano installation directory> \
  --iree-amd-aie-mlir-aie-install-dir=<mlir-aie installation directory> \
  --iree-amd-aie-vitis-install-dir=<vitis installation directory> \
  --iree-hal-dump-executable-files-to=$PWD  --iree-amd-aie-show-invoked-commands
```

### DIFF
--- a/compiler/plugins/target/AMD-AIE/aie/AIETransformPasses.cpp
+++ b/compiler/plugins/target/AMD-AIE/aie/AIETransformPasses.cpp
@@ -17,6 +17,7 @@ void registerAIETransformPasses() {
   registerAIEAssignBufferAddresses();
   registerAIECanonicalizeDevice();
   registerAIECoreToStandard();
+  registerAIERoutePathfinderFlows();
   registerAIELocalizeLocks();
   registerAIENormalizeAddressSpaces();
   registerAIEObjectFifoRegisterProcess();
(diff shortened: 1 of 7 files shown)


MaheshRavishankar:
This PR is currently failing at this step https://github.com/nod-ai/iree-amd-aie/pull/44/files#diff-2a27c2baf66b75bcf426d927b55cbf5f3b5d7bfa69b9b7492ab0d3672805bac4R565 . The diagnosis is there are some expected temporary files that the IREE compilation flow isnt preserving. So that command is failing.

nirvedhmeshram (approved):
I know we have a long way to go in terms of making sure this generalize but since we can make an xclbin with this, lets check this in so that everyone has the same starting point.

PR #48, opened 2023-12-11
Describe need to run sync_deps.py in README instructions 

newling:


README.md
```diff
@@ -10,22 +10,33 @@ with this repository have an IREE build setup in an `iree-build` directory that
...
 
 ```
 git submodule update --init
 ```
 
 ## Enabling in IREE
 
+Currently (12/12/2023) we are waiting for an MLIR fix in the transform dialect,
+and so cannot build with the latest IREE (and its LLVM submodule). To pin IREE
+and its submodules to a amd-aie compatible version, run
```
nirvedhmeshram, line 23: maybe above three lines should be an issue since we want end users to run sync_deps.py regardless of the current transform dialect issue, so that the project builds for them consistently.

nirvedhmeshram (approved):
minor comment about moving the waiting for mlir fix into an issue rather than README. Thanks for this!

README.md
newling: Ok, I thought this was just a temporary fix. I will update to say that running sync_deps.py is the expected workflow for this plugin

PR #49, opened 2023-12-11
Replace nod.ai/mlir-aie/iree_fixes with xilinx/mlir-aie/main 

newling:
There are no longer any fixes required to build iree-amd-aie, and all the (2!) tests pass with xilinx/mlir-aie/main


nirvedhmeshram (approved):
Nice!

PR #50, opened 2023-12-11
Flatbuffer executable serialization (xlcbin, lx6 instructions)

newling:
Write the xclbin and lx6 instructions to a flatbuffer, which eventually ends up in the .vmfb artefact generated by the compiler. The design for the AIE backend here is based on the CUDA plugin's design. 

This needs further testing, once the ends are connected. I can see the xclbin and lx6 control code instructions are inserted into the MLIR module, and the .vmfb file is generated, but the ultimate test requires using the runtime. In other words the compilation and artefact generation seems fine:

```
./iree-compile -o test.vmfb --mlir-print-ir-after=iree-hal-serialize-executables ... 
```

but the ultimate test requires the runtime, which still has a few missing pieces on the command buffer side

```
./iree-run-module --device=xrt --module=../scripts/scripts_2/test.vmfb --input=8x16xi32 --input=16x8xi32
```

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
@@ -17,6 +17,7 @@
 #include "iree-amd-aie/Target/XclBinGeneratorKit.h"
 #include "iree-amd-aie/Transforms/Passes.h"
 #include "iree/compiler/Codegen/Dialect/IREECodegenDialect.h"
+#include "iree/compiler/Utils/FlatbufferUtils.h"
 #include "llvm/Bitcode/BitcodeWriter.h"
 #include "llvm/Support/raw_ostream.h"
 #include "mlir/Conversion/AffineToStandard/AffineToStandard.h"
@@ -35,6 +36,8 @@
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Target/LLVMIR/ModuleTranslation.h"
 #include "mlir/Transforms/Passes.h"
+#include "runtime/plugins/AMD-AIE/iree-amd-aie/schemas/xrt_executable_def_builder.h"
+#include "runtime/plugins/AMD-AIE/iree-amd-aie/schemas/xrt_executable_def_reader.h"
 
 // Forward declaration of some translate methods from AIE. THis is done
 // here since the headers in MLIR-AIE repo are in a place that is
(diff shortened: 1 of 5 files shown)


nirvedhmeshram:
Ya I had a similar experience initially, I think you have to include the schema in your own cmake library target like this
https://github.com/nod-ai/iree-amd-aie/blob/main/runtime/src/iree-amd-aie/driver/xrt/CMakeLists.txt#L52
and then it should build those headers. Let me know if that doesnt work

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -834,12 +837,22 @@ LogicalResult generateXCLBin(MLIRContext *context, ModuleOp moduleOp,
 LogicalResult AIETargetBackend::serializeExecutable(
     const SerializationOptions &serOptions,
     IREE::HAL::ExecutableVariantOp variantOp, OpBuilder &executableBuilder) {
+  // Create a flatbuffer containing xclbin and lx6 instructions, which gets
+  // insertered into the IR as an op attribute. The design here is copied from
+  // the CUDA plugin implementation, see
+  // iree/compiler/plugins/target/CUDA/CUDATarget.cpp
```
nirvedhmeshram, line 843: No need for this comment as it is standard practice.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -834,12 +837,22 @@ LogicalResult generateXCLBin(MLIRContext *context, ModuleOp moduleOp,
...
+  FlatbufferBuilder builder;
+  iree_amd_aie_hal_xrt_ExecutableDef_start_as_root(builder);
+
```
nirvedhmeshram, line 846: I think it would read better if we do all the flatbuffer related activity at once in the end, so lets move this at the very end?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -864,6 +877,10 @@ LogicalResult AIETargetBackend::serializeExecutable(
                               StringRef(dumpString.data(), dumpString.size()));
   }
 
+  // Serialize lx6 control instructions into flatbuffer.
+  auto ipuInstrsRef = builder.createInt32Vec(ipuInstrs);
+  iree_amd_aie_hal_xrt_ExecutableDef_asm_instrs_add(builder, ipuInstrsRef);
+
```
nirvedhmeshram, line 883: This can also move to the end

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -911,9 +928,26 @@ LogicalResult AIETargetBackend::serializeExecutable(
                               toolkit, ostream))) {
       return moduleOp.emitOpError() << "failed to generate XCLbin";
     }
+
+    // Serialize xclbin into flatbuffer.
+    llvm::StringRef xclbinStringView(xclbin.begin(), xclbin.size());
+    auto xclbinStringRef = builder.createString(xclbinStringView);
+    iree_amd_aie_hal_xrt_ExecutableDef_xclbin_add(builder, xclbinStringRef);
```
nirvedhmeshram, line 935: can be moved outside this scope

nirvedhmeshram (changes requested):
Just a minor refactoring request, looking good overall.

nirvedhmeshram (commented):
LGTM!

nirvedhmeshram (approved):

PR #52, opened 2023-12-12
Bundle the me_basic.o entry point with IREE

spaceotter:
For now, me_basic.o is a magic binary that provides the entry point for execution on the AIE cores. It cannot be replicated (yet) by llvm-aie. So this patch fetches the binary checked in to mlir-aie and bundles it with the IREE compiler. It isn't necessary to have Vitis or a full bulid of mlir-aie for this to work.


MaheshRavishankar:
I dont think we need this.

PR #55, opened 2023-12-15
[XRT] Add command buffer

nirvedhmeshram:
With this commit we can start testing to see if we can launch kernels. Some changes are made to the pipeline layout to better suit the needs of XRT API.


nirvedhmeshram:
after testing I have uncovered issues such as block_pool not being optional but necessary for internal resources and that I didnt actual create the command bvuffer from device, I will fix these issues and come back with something that works

PR #56, opened 2023-12-15
Add dependency to flatbuffer building

nirvedhmeshram:
Lets add this dependency since we do actually depend on it.


newling (approved):
Hopefully this'll fix the intermittent build issue

PR #61, opened 2023-12-22
Decompose iree_linalg_ext.pack and unpack ops to MLIR-AIR

erwei-xilinx:
- Decompose pack and unpack operations to a series of memref.expand_shape, memref.transpose, memref.subview and air.dma_memcpy_nd ops.
- Fold the series of memref data layout transformations into offsets, sizes and strides fields in air.dma_memcpy_nd.
- Have MLIR-AIR lower the air.dma ops through air.channel ops to AIE BDs.


newling:
Can all pack ops ( https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorpack-tensorpackop ) be decomposed as

```
pack(x) -> transpose(expand_shape(pad(x)))
```

is that part of what is happening here? 

I think some lit tests would be quite useful for me to fully understand what the pass does.

MaheshRavishankar (changes requested):
+1 to lit tests! I am reviewing the details on my other machine

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,873 @@
...
+  if (!rankReducedOuterDimsPerm.empty())
+    applyPermutationToVector<int64_t>(perm, rankReducedOuterDimsPerm);
+
+  // The tile always ends up as the inner most dims after packing.
+  perm.append(innerDims);
+
+  return perm;
+}
+
+class GeneralizePackOpPattern
```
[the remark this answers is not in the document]
hanhanW: ohhhhh are we adding more burdens on iree_linalg_ext.pack/unpack...? Can we do that at tensor level?

erwei-xilinx:
> Can all pack ops ( https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorpack-tensorpackop ) be decomposed as
> 
> ```
> pack(x) -> transpose(expand_shape(pad(x)))
> ```
> 
> is that part of what is happening here?
> 
> I think some lit tests would be quite useful for me to fully understand what the pass does.

Yes, and in addition there could also be a subview op which transforms the data layout to do tiling. Sure let me work on a lit test.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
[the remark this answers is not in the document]
erwei-xilinx: Might be tricky. ATM AIR ops operate on memrefs only. Might work if we decompose tensor pack/unpack to tensor transpose/reshape etc, then bufferize to memref ops. But the tensor pack decompose seems to generate an extra tensor empty which we shall not bufferize to memrefs...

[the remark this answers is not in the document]
hanhanW: I dont have context about AIR, and I don't want to bring up a long discussion about how this should be set up in my mind. It sounds weird to me if the tensor.empty can't be replaced with an existing buffer (like what it is used in linalg_ext.pack/unpack).

In IREE CPU backends, we are making progress to get rid of the decomposition. Instead, we will vectorize them directly.

The tensor version is doing something different from what you want as well. It generates linalg.transpose, but not tensor.transpose.

There are no patterns to break down linalg_ext.pack/unpack on memref levels. All the transforms only work on tensors and they are upstreamed to MLIR repo, exept LowerToScalarLoops. You will likely need to copy most of logics from [lowerPack and lowerUnPack](https://github.com/llvm/llvm-project/blob/1f61507401995b434a0720d3f9869659466d9b31/mlir/lib/Dialect/Linalg/Transforms/Transforms.cpp#L219-L480) and generate memref ops instead. They have better mapping about all the permutation things.

hanhanW:
> Can all pack ops ( https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorpack-tensorpackop ) be decomposed as
> 
> ```
> pack(x) -> transpose(expand_shape(pad(x)))
> ```

The answer is yes, if there all the inner tile sizes are static. We have very well support for tensor.pack/unpack decomposition at tensor level. We only bail out if any of inner tile sizes is dynamic: https://github.com/llvm/llvm-project/blob/1f61507401995b434a0720d3f9869659466d9b31/mlir/lib/Dialect/Linalg/Transforms/Transforms.cpp#L221-L229

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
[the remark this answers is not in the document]
erwei-xilinx: Thanks for your comments.

> There are no patterns to break down linalg_ext.pack/unpack on memref levels.

You are right, and that's also what I observed, and what I did here is exactly as you mentioned, memref version of lowerPack and lowerUnPack which generates memref transpose, subview and expans_shape ops instead. In fact, if you search for the "lowerPack" in the code, you would see traces of where the logic comes from.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,873 @@
+// Copyright 2023 The IREE Authors
```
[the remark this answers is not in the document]
erwei-xilinx: Updated.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,873 @@
...
+
+//===----------------------------------------------------------------------===//
+// Pass
+//===----------------------------------------------------------------------===//
+
+namespace mlir::iree_compiler::AMDAIE {
+
+namespace {
+
+/// Returns a pad op if padding value is set. Otherwise, returns the
```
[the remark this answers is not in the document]
erwei-xilinx: Updated.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,873 @@
...
+/// source directly. The method assumes that the `packOp` has static shapes.
+/// TODO: padding
+static Value getPackOpInputOrPaddedSource(OpBuilder &builder,
```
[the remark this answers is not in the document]
erwei-xilinx: Agreed. Dropping all TODO notes for padding.

erwei-xilinx:
> +1 to lit tests! I am reviewing the details on my other machine

@MaheshRavishankar @newling Lit test added.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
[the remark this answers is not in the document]
newling: Ok, I see where the name `stripMinedShape` comes from now!

So is it not feasible to reuse the tensor version of this, before bufferization? I'm still mostly in the dark here, but an extra tensor.empty sounds like an orthogonal problem for another pass to eliminate..

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,864 @@
...
+                     [](OpFoldResult tile) { return tile.is<Value>(); })) {
+      return rewriter.notifyMatchFailure(
+          packOp, "require inner tile sizes being static");
+    }
+
+    auto innerDimsPos = packOp.getInnerDimsPos();
+    int64_t srcRank = packOp.getInputRank();
+    auto destShape = packOp.getOutputType().getShape();
+    if (llvm::any_of(innerDimsPos, [destShape](int64_t index) {
+          return destShape[index] != 1;
```
MaheshRavishankar, line 127: I dont think this is enough. The outer dimensions could be permuted as well. AFAICS you are looking for cases where the outer pack dimension is 1. So you will need to account for `outer_dims_perm` as well?

MaheshRavishankar (changes requested):
I am reading through this change... its fairly big, and it seems like there are two separate passes in here. Can we break that up? One just to lower the pack operation to the dma memcpy op and the other to "canonicalize" it. The canonicalization could just be done separately as a canonicalization of dma memcpy ops?

I will need to work this out a little bit more to understand whats happening here. but a few comments below.

erwei-xilinx:
>  Can we break that up? One just to lower the pack operation to the dma memcpy op and the other to "canonicalize" it. The canonicalization could just be done separately as a canonicalization of dma memcpy ops?

I think that would be possible. The pass consists of two parts, where the first part lowers `iree_linalg_ext.pack/unpack` to a series of memref ops, and the second part condenses them into wrap and stride lists for `air.dma_memcpy_nd`.

I can move the second part into a new conversion pass in AIR, if that would make sense?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
erwei-xilinx: The `outer_dims_perm` will be handled in `lowerPack` pattern. that happens after this `GeneralizePackOpPattern`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,864 @@
...
+}
+
+SmallVector<OpFoldResult> getMixedSizes(OpBuilder &builder, Location loc,
+                                        Value value) {
+  auto memrefType = llvm::cast<MemRefType>(value.getType());
+  SmallVector<OpFoldResult> result;
+  for (int64_t i = 0; i < memrefType.getRank(); ++i)
+    result.push_back(getMixedSize(builder, loc, value, i));
+  return result;
+}
```
hanhanW, line 54: Does these upstream methods work?

https://github.com/llvm/llvm-project/blob/02f95b77515fe18ed1076b94cbb850ea0cf3c77e/mlir/include/mlir/Dialect/MemRef/IR/MemRef.h#L66-L68

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,864 @@
...
+  // Generalization patterns for outer unit dims have higher priority because
+  // they do not generate reshape ops.
+  {
+    RewritePatternSet patterns(ctx);
+    patterns.add<GeneralizePackOpPattern, GeneralizeUnPackOpPattern>(ctx);
+    if (failed(applyPatternsAndFoldGreedily(getOperation(),
+                                            std::move(patterns)))) {
+      return signalPassFailure();
+    }
+  }
```
hanhanW, line 766: I haven't looked into details yet. The reason that we have them for tensor version is that we want to get rid of tensor.collpase_shape/expand_shape ops. They could block vectorization and bufferization analysis. Given that you are working on memref's world. Is it okay to just leverage them to LowerPackPattern and LowerUnPackPattern? Perhaps memref.collapse_shape/expand_shape ops are probably okay in your use cases, have you considered that?

MaheshRavishankar:
> > Can we break that up? One just to lower the pack operation to the dma memcpy op and the other to "canonicalize" it. The canonicalization could just be done separately as a canonicalization of dma memcpy ops?
> 
> I think that would be possible. The pass consists of two parts, where the first part lowers `iree_linalg_ext.pack/unpack` to a series of memref ops, and the second part condenses them into wrap and stride lists for `air.dma_memcpy_nd`.
> 
> I can move the second part into a new conversion pass in AIR, if that would make sense?

If you are doing that, the review here would be easier....

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
erwei-xilinx: Thanks for the pointer. Yes the upstream method works. PR updated.

erwei-xilinx:
> If you are doing that, the review here would be easier....

I have removed the second half of the PR, and will create a separate PR to MLIR-AIR with the second half.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,505 @@
...
+    SmallVector<OpFoldResult> writeSizes =
+        memref::getMixedSizes(rewriter, loc, packOp.getOutput());
+
+    memorySpace = packOp.getOutputType().cast<MemRefType>().getMemorySpace();
+    auto writeType = MemRefType::get(destShape, elemType, nullptr, memorySpace);
+    auto output_subview = rewriter.create<memref::SubViewOp>(
+        loc, writeType, packOp.getOutput(), writeOffsets, writeSizes,
+        writeStrides);
+
+    SmallVector<Value, 2> mt;
```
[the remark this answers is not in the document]
erwei-xilinx: Thanks for the comment. Will change "mt" to "emptyVec" to be more descriptive. Those are basically the dependency list (empty because dep analysis is in a later pass), offsets, sizes and strides fields for both src and dst. They are left empty for now and they will be populated in the AIR conversion pass in https://github.com/Xilinx/mlir-air/pull/395. 

Fundamentally this `air.dma` op being created here is simply a `memref.copy` with the more forgiving memref type checks. By creating `air.dma` instead of `memref.copy` I could skip over some `memref.expand_shape` and `memref.collapse_shape` (or rank-reducing `memref.subview`) here.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,505 @@
...
+    memorySpace = unpackOp.getOutputType().cast<MemRefType>().getMemorySpace();
+    auto writeType = MemRefType::get(
+        destShape, elemType,
+        unpackOp.getOutputType().cast<MemRefType>().getLayout(), memorySpace);
+
+    auto output_subview = rewriter.create<memref::SubViewOp>(
+        loc, writeType, unpackOp.getOutput(), writeOffsets, writeSizes,
+        writeStrides);
+
+    SmallVector<Value, 2> mt;
```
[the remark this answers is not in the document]
erwei-xilinx: Will update the variable name.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,505 @@
...
+    RewritePatternSet patterns(ctx);
+    patterns.add<GeneralizePackOpPattern, GeneralizeUnPackOpPattern>(ctx);
+    if (failed(applyPatternsAndFoldGreedily(getOperation(),
+                                            std::move(patterns)))) {
+      return signalPassFailure();
+    }
+  }
+
+  // Do not convert pack and unpack ops if outer dims are expected to always be
+  // tiled to one.
```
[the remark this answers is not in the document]
erwei-xilinx: Sry the comments were put here by mistake. Updated to more informative comments.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
erwei-xilinx: Thanks for your comment. Yes I did consider memref.collapse_shape but I coundn't get it to work with the test case that I have, as it was complaining about "invalid source layout map or collapsing non-contiguous dims". Separating the `GeneralizePackOpPattern` from `LowerPack` helped us to quickly generate working examples with the application we are looking at.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,505 @@
...
+      } else {
+        readSizes.push_back(rewriter.getIndexAttr(inputShape[i]));
+      }
+      readShape.push_back(inputShape[i]);
+    }
+
+    Type elemType = packOp.getInputType().getElementType();
+    Attribute memorySpace =
+        packOp.getInputType().cast<MemRefType>().getMemorySpace();
+    auto readType = MemRefType::get(
```
MaheshRavishankar, line 147: I am not even sure why this subview is needed here? This is essentially taking the full slice of the underlying object?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,505 @@
...
+};
+
+class GeneralizeUnPackOpPattern
+    : public OpRewritePattern<IREE::LinalgExt::UnPackOp> {
+ public:
+  using OpRewritePattern<IREE::LinalgExt::UnPackOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(IREE::LinalgExt::UnPackOp unpackOp,
+                                PatternRewriter &rewriter) const override {
+    int64_t srcRank = unpackOp.getInputRank();
```
MaheshRavishankar, line 197: I think you can common out a lot of this implementation by moving the core logic to a common function. `unpack` and `pack` are just reverse of each other, so just flipping the shapes of the input/output should work out the same way.

MaheshRavishankar (changes requested):
Left a few more comments. I think the general pack/unpack lowering looks fine. I am not sure about the rank-reduced one.

hanhanW:
> Left a few more comments. I think the general pack/unpack lowering looks fine. I am not sure about the rank-reduced one.

I think the only difference is that one is using tensor/memref.collapse/expand_shape, and the other is using rank-reduced subview/extract_slice. GeneralizePackOpPattern and GeneralizeUnPackOpPattern are for corner cases. You can use rank-reduced subview/extract_slice when outer dims are all ones. In the past, they have different support coverage, but today they are well supported for any permutation. If you add creating rank-reduced subview logic to lowerPack and lowerUnPack, you will no longer need GeneralizePackOpPattern and GeneralizeUnPackOpPattern.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,505 @@
...
+  applyPermutationToVector(packedToStripMinedShapePerm, outerPositionPerm);
+
+  // 3. Compute the stripMinedShape: this is the packed shape before any outer
+  // or inner permutations have been applied.
+  SmallVector<int64_t> stripMinedShape(packedMemrefType.getShape());
+  applyPermutationToVector(stripMinedShape, packedToStripMinedShapePerm);
+
+  // 4. Expand from the padded result to the stripMinedShape.
+  auto reshapeOp = rewriter.create<memref::ExpandShapeOp>(
+      loc, stripMinedShape, packOp.getInput(), packingMetadata.reassociations);
```
hanhanW, line 352: To be more specific, you can check if outer dims are all ones or not here. If so, you create rank-reduced subview. If not, you create memref::ExpandShapeOp. Then you are able to drop the other pattern.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,505 @@
...
+// assuming rank reduction of unit outer dims.
+static SmallVector<int64_t> getPackUnpackRankReducedPerm(
+    ArrayRef<int64_t> shape, ArrayRef<int64_t> innerDimsPos,
+    ArrayRef<int64_t> outerDimsPerm) {
+  SmallVector<int64_t> rankReducedOuterDimsPerm;
+  SmallVector<int64_t> outerDims;
+  SmallVector<int64_t> innerDims;
+  int64_t dim = 0;
+  int64_t unpackedRank = shape.size();
+  for (auto i : llvm::seq<unsigned>(0, unpackedRank)) {
```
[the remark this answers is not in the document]
erwei-xilinx: `unpackedRank` is not always equal to `outerDimsPermRank + innerDimsPosRank`. If taking into consideration the code to infer the ranks then it may not necessarily be simpler than normalizing the two perms and applying them.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,864 @@
...
+    SmallVector<OpFoldResult> readSizes;
+    SmallVector<int64_t> readShape;
+    for (auto i : llvm::seq<unsigned>(0, srcRank)) {
+      if (dimAndTileMapping.count(i)) {
+        readShape.push_back(getConstantIntValue(dimAndTileMapping[i])
+                                .value_or(ShapedType::kDynamic));
+        readSizes.push_back(dimAndTileMapping[i]);
+        continue;
+      }
+      if (ShapedType::isDynamic(inputShape[i])) {
```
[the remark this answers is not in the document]
erwei-xilinx: I need to look at `memref.dim`. I think the main blocker here is that the downstream MLIR-AIR currently doesn't have applications driving a concrete dynamic shape support.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
erwei-xilinx: Agreed. Removed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
erwei-xilinx: Agreed. Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,506 @@
...
+// Pass
+//===----------------------------------------------------------------------===//
+
+namespace mlir::iree_compiler::AMDAIE {
+
+namespace {
+
+/// Returns the padding value if set. Otherwise, returns the source directly.
+/// The method assumes that the `packOp` has static shapes.
+static Value getPackOpInputOrPaddedSource(OpBuilder &builder,
```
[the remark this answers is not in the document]
erwei-xilinx: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
erwei-xilinx: Agreed. Moved inference of transform perms to a common function.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
erwei-xilinx: Thanks for the explanation. I did a code refactor today and dropped the `GeneralizePack(UnPack)OpPattern` functions.

MaheshRavishankar (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,363 @@
...
+  // Ditto for the outer dims.
+  SmallVector<int64_t> perm = outerDims;
+
+  rankReducedOuterDimsPerm =
+      getPackUnpackNormalizedPerm(unpackedRank, rankReducedOuterDimsPerm);
+  if (!rankReducedOuterDimsPerm.empty())
+    applyPermutationToVector<int64_t>(perm, rankReducedOuterDimsPerm);
+
+  // The tile always ends up as the inner most dims after packing.
+  perm.append(innerDims);
```
MaheshRavishankar, line 75: Ok, I think I follow the logic now. I think you need to state that is assuming that all the "outer dimensions" created for the packed dimensions are assumed to be 1. The name isnt self explanatory.

That said, the `outerDims` is normalized, and the `innerDims` is normalized. How does the final permutation vector work out to be just an "append" of the two...

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
```diff
@@ -0,0 +1,363 @@
...
+        loc, stripMinedShape, packOp.getInput(),
+        packingMetadata.reassociations);
+
+    // 4. Transpose stripMinedShape to packedShape.
+    transpPerm = invertPermutationVector(packedToStripMinedShapePerm);
+  } else {
+    tile = packOp.getInput();
+    // Transpose the tile to match the inner tile order.
+    auto inputShape = packOp.getInputType().getShape();
+    transpPerm = getPackUnpackRankReducedPerm(inputShape, innerDimsPos,
```
MaheshRavishankar, line 163: Instead of doing this, isnt it easier to just do the `if` part for both the rank-reduced and non-rank-reduced part, and for the rank-reduced case (where all the outer packed dimensions are 1) just drop those dimensions in the permutation vector and then normalize it....

MaheshRavishankar (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/pack_to_air.mlir
```diff
@@ -0,0 +1,100 @@
+// RUN: iree-opt --pass-pipeline="builtin.module(hal.executable(hal.executable.variant(iree-amdaie-decompose-pack-unpack-to-air)))" %s | FileCheck %s
+
+#executable_target_amdaie_xclbin_fb = #hal.executable.target<"amd-aie", "amdaie-xclbin-fb", {target_arch = "chip-tbd"}>
```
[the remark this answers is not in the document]
erwei-xilinx: Sure, lit test file edited.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/pack_to_air.mlir
```diff
@@ -0,0 +1,100 @@
+// RUN: iree-opt --pass-pipeline="builtin.module(hal.executable(hal.executable.variant(iree-amdaie-decompose-pack-unpack-to-air)))" %s | FileCheck %s
```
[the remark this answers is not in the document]
erwei-xilinx: Added four small tests corresponding to both cases for both pack and unpack.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
erwei-xilinx: Agreed. We could just normalize the `innerDimPos` and use that as permutation vector. Updated.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/DecomposeLinalgExtPackUnPackToAIR.cpp
erwei-xilinx: Dropping this method altogether and using the normalized `innerDimPos` as permutation vector instead.

MaheshRavishankar (approved):
Thanks. This looks good now!

PR #62, opened 2023-12-29
Adapt to changes for next IREE integration post 20231228.

stellaraccident:
A patch like this will be necessary if syncing to an IREE at/past https://github.com/openxla/iree/pull/16011.

Syncing to c35d8e91dfd8cc724520ec652031a3279b5edd69


stellaraccident:
This has some other issues, so I'll just leave it here for the team to pick up when they are ready to move forward.

stellaraccident:
Actually - fixed it. Going to land.

PR #64, opened 2024-01-05
[AMD-AIE] Add a pass to tile and fuse TileInterface ops using scf.forall

Abhishek-Varma:
-- This commit adds the initial effort for tiling + fusing TileInterface ops using scf.forall by creating a new pass `--iree-amdaie-tile-and-fuse`.
-- It creates a new concise transform dialect script and adds a `--iree-amdaie-tile-and-fuse` test case to demo the generalised feature.

Co-authored-by: Nirvedh Meshram <nmeshram@amd.com>

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
@@ -31,8 +31,7 @@ namespace mlir::iree_compiler::AMDAIE {
 
 namespace {
 
-/// Lowers the computation within the workgroup count region for the ops
-/// that are handled by default.
+/// Lower WorkgroupCountFromSliceOp with all 1's.
 static LogicalResult lowerWorkgroupCount(RewriterBase &rewriter,
                                          func::FuncOp entryPointFn) {
   FailureOr<IREE::HAL::ExecutableExportOp> exportOp =
@@ -54,7 +53,7 @@ static LogicalResult lowerWorkgroupCount(RewriterBase &rewriter,
   }
   if (workgroupCountOps.empty()) {
     // If there are no default handled `flow.dispatch.workgroup_count`
-    // operation, do nothing. do nothing.
+    // operation, do nothing.
     return success();
   }
   if (!llvm::hasSingleElement(workgroupCountOps)) {
@@ -386,8 +385,6 @@ void AMDAIETileAndFusePass::runOnOperation() {
       return signalPassFailure();
     }
 
-    // Check if workgroup count lowering is needed and if it is then perform it
-    // with a workgroup of (1, 1, 1)
     if (failed(lowerWorkgroupCount(rewriter, funcOp))) {
       funcOp->emitOpError("failed to lower workgroup count region");
       return signalPassFailure();
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
@@ -33,11 +33,8 @@ namespace {
 
 /// Lowers the computation within the workgroup count region for the ops
 /// that are handled by default.
-static LogicalResult lowerWorkgroupCount(
-    RewriterBase &rewriter, func::FuncOp entryPointFn,
-    ArrayRef<OpFoldResult> workgroupCount, ArrayRef<int64_t> tileSizes,
-    ArrayRef<int64_t> staticLoopRanges, ArrayRef<int64_t> interchange,
-    ArrayRef<unsigned> partitionedLoops, int maxWorkgroupParallelDims) {
+static LogicalResult lowerWorkgroupCount(RewriterBase &rewriter,
+                                         func::FuncOp entryPointFn) {
   FailureOr<IREE::HAL::ExecutableExportOp> exportOp =
       getEntryPoint(entryPointFn);
   if (failed(exportOp)) {
@@ -48,31 +45,32 @@ static LogicalResult lowerWorkgroupCount(
   if (!body) {
     return exportOp->emitOpError("unexpected empty workgroup count region");
   }
-  SmallVector<Operation *> countOps;
+  SmallVector<IREE::Flow::DispatchWorkgroupCountFromSliceOp> workgroupCountOps;
   for (Operation &op : *body) {
-    if (isa<IREE::Flow::DispatchWorkgroupCountFromSliceOp,
-            IREE::Flow::DispatchWorkgroupCountFromDagRootOp>(&op)) {
-      countOps.push_back(&op);
+    if (isa<IREE::Flow::DispatchWorkgroupCountFromSliceOp>(&op)) {
+      workgroupCountOps.push_back(
+          cast<IREE::Flow::DispatchWorkgroupCountFromSliceOp>(&op));
     }
   }
-  if (countOps.empty()) {
+  if (workgroupCountOps.empty()) {
     // If there are no default handled `flow.dispatch.workgroup_count`
     // operation, do nothing. do nothing.
     return success();
   }
-  if (!llvm::hasSingleElement(countOps)) {
+  if (!llvm::hasSingleElement(workgroupCountOps)) {
     return exportOp->emitOpError(
-        "unexpected multiple flow.dispatch.workgroup_count_from_dag_root "
-        "operations "
-        "in body");
+        "unexpected multiple flow.dispatch.workgroup_count_slice_op "
+        "operations in body");
   }
-  return TypeSwitch<Operation *, LogicalResult>(countOps[0])
-      .Case<IREE::Flow::DispatchWorkgroupCountFromSliceOp>([&](auto countOp) {
-        return lowerWorkgroupCountFromSliceOp(rewriter, countOp, entryPointFn,
-                                              workgroupCount,
-                                              maxWorkgroupParallelDims);
-      })
-      .Default([&](Operation *) { return success(); });
+  OpBuilder::InsertionGuard g(rewriter);
+  rewriter.setInsertionPoint(workgroupCountOps[0]);
+  Location loc = workgroupCountOps[0].getLoc();
+  SmallVector<OpFoldResult> results;
+  results.resize(workgroupCountOps[0].getNumResults(),
+                 rewriter.getIndexAttr(1));
+  rewriter.replaceOp(workgroupCountOps[0],
+                     getValueOrCreateConstantIndexOp(rewriter, loc, results));
+  return success();
 }
 
 //===----------------------------------------------------------------------------------===//
@@ -177,6 +175,10 @@ std::optional<scf::SCFFuseProducerOfSliceResult> tileAndFuseProducerOfSlice(
                                            tileAndFuseResult->tiledValues[0],
                                            tileAndFuseResult->tiledOps};
 }
+
+//===----------------------------------------------------------------------------------===//
+// End methods copied over from core to implement tile and fuse with scf.forall.
+//===----------------------------------------------------------------------------------===//
 
 /// Starting from `op` walk all operands backwards to find all
 /// potentially fusable operations, i.e. operations that implement
@@ -325,9 +327,6 @@ LogicalResult applyTileAndFuse(RewriterBase &rewriter, Operation *rootOp,
 
   return success();
 }
-//===----------------------------------------------------------------------------------===//
-// End methods copied over from core to implement tile and fuse with scf.forall.
-//===----------------------------------------------------------------------------------===//
 
 /// This pass starts with the last TilingInterface operation, tiles the op and
 /// fuses its producers recursively. The `tilingLevel` must be specified. It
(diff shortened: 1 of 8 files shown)


tests/samples/matmul_fill_spec_pad_new.mlir
```diff
@@ -0,0 +1,129 @@
...
+//   ${IREE_DIR}/build/tools/iree-opt \
+//     ${IREE_AMD_AIE_DIR}/tests/samples/matmul_fill_static.mlir \
+//     --iree-hal-target-backends=amd-aie \
+//     --iree-abi-transformation-pipeline \
+//     --iree-flow-transformation-pipeline \
+//     --iree-stream-transformation-pipeline \
+//     --iree-hal-configuration-pipeline | \
+//   ${IREE_DIR}/build/tools/iree-opt \
+//      --pass-pipeline='builtin.module(hal.executable(hal.executable.variant(iree-hal-translate-target-executable-variants{target=amd-aie})))' \
+//      --iree-codegen-transform-dialect-library=${IREE_AMD_AIE_DIR}/tests/samples/matmul_fill_spec_pad.mlir
```
nirvedhmeshram, line 15: This should be matmul_fill_spec_pad_new.mlir?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -21,7 +21,17 @@ namespace mlir::iree_compiler::AMDAIE {
...
 void buildAMDAIETransformPassPipeline(OpPassManager &pm) {
   addCommonTargetExecutablePreprocessingPasses(pm);
+  // TODO: Current we don't include C++ equivalent of Transform
+  //       dialect scripts. We are thus guarding their inclusion
+  //       with a bool `useCPlusPlusTransformPasses` which would
+  //       be used during development efforts. Once the C++ passes
+  //       are ready, we will include these passes by default and
+  //       take away the guarding.
+  bool useCPlusPlusTransformPasses = false;
+  if (useCPlusPlusTransformPasses)
```
nirvedhmeshram, line 31: IIUC you want to set this to true when using the matmul_fill_spec_pad_new.mlir script? If thats the case I think what you want is a global option that you can set from the command as true, here is an example, 
https://github.com/openxla/iree/blob/main/compiler/src/iree/compiler/Dialect/HAL/Target/LLVMCPU/LLVMCPUTarget.cpp#L56-L59
so then when running with pad_new.mlir you can set it to true

tests/samples/matmul_fill_spec_pad_new.mlir
nirvedhmeshram: You can also add `useCPlusPlusTransformPasses` as a flag mentioned in another comment

tests/samples/tile_and_fuse.mlir
```diff
@@ -0,0 +1,60 @@
...
+    } -> tensor<?x?xf32>
+  return %2 : tensor<?x?xf32>
+}
+//      CHECK: @matmul_bias_add
+//      CHECK:   scf.forall
+// CHECK-SAME:   {
+//      CHECK:       linalg.fill
+//      CHECK:       linalg.matmul
+//      CHECK:       linalg.generic
+//      CHECK:   }
```
nirvedhmeshram, line 60: Nit: Add newline at end of file

nirvedhmeshram, line 60: Nit: Add newline at end of file

nirvedhmeshram (commented):
I will let Mahesh review on the core logic, but some minor comments.

yzhang93 (commented):
I haven't compared this to the original pad transform dialect script. Could you summarize what has been changed?

tests/samples/matmul_fill_spec_pad_new.mlir
```diff
@@ -0,0 +1,129 @@
...
+      transform.apply_patterns.iree.fold_fill_into_pad
+      transform.apply_patterns.scf.for_loop_canonicalization
+      transform.apply_patterns.canonicalization
+    } : !transform.any_op
+    transform.iree.apply_licm %func : !transform.any_op
+    transform.iree.apply_cse %func : !transform.any_op
+    transform.yield
+  }
+
+  transform.named_sequence @__transform_main(%variant_op: !transform.any_op {transform.read_only}) {
```
yzhang93, line 32: I haven't compared this to the original pad transform dialect script. Could you summarize what has been changed?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
```diff
@@ -0,0 +1,337 @@
...
+  LLVM_DEBUG(llvm::dbgs() << "consumerOp: " << consumerOp << "\n");
+  LLVM_DEBUG(llvm::dbgs() << "tilingLevel: " << tilingLevel << "\n");
+  // TODO: Currently hardcoding the tile size to make the tile and fuse run for
+  //       at least the first level. So, we can generalize this by adding tile
+  //       level, similar to LLVMCPUTileAndFuse.cpp.
+  SmallVector<OpFoldResult> tileSizes = getAsIndexOpFoldResult(context, {8, 8});
+  auto options = scf::SCFTilingOptions().setTileSizes(tileSizes);
+  options.setMapping(
+      {gpu::GPUBlockMappingAttr::get(context, gpu::MappingId::DimY),
+       gpu::GPUBlockMappingAttr::get(context, gpu::MappingId::DimX)});
```
yzhang93, line 320: This might be ok for now, but ideally we'll need some AIE specific attributes similar to the gpu ones.

Abhishek-Varma:
> I haven't compared this to the original pad transform dialect script. Could you summarize what has been changed?

Sure, the change is only the [first level of tiling and fusing](https://github.com/nod-ai/iree-amd-aie/blob/main/tests/samples/matmul_fill_spec_pad.mlir#L36-L44) where the tile_size=8x8.
In the new transform dialect script we were able to get rid of these exact lines and in turn depend on the C++ pass.

tests/samples/tile_and_fuse.mlir
```diff
@@ -0,0 +1,60 @@
...
+        linalg.yield %3 : f32
+    } -> tensor<?x?xf32>
+  return %2 : tensor<?x?xf32>
+}
+//      CHECK: @matmul_bias_add
+//      CHECK:   scf.forall
+// CHECK-SAME:   {
+//      CHECK:       linalg.fill
+//      CHECK:       linalg.matmul
+//      CHECK:       linalg.generic
```
yzhang93, line 59: What is the specific part of this test? Could you elaborate the result IR?

Abhishek-Varma: I didn't understand. Do you mean to suggest to have the entire result IR captured in the lit test? I don't think that should be necessary though.

I have taken the base test from [iree/LLVMCPU/tile_and_fuse.mlir](https://github.com/openxla/iree/blob/main/compiler/src/iree/compiler/Codegen/LLVMCPU/test/tile_and_fuse.mlir#L25-L34) and they too aren't capturing the entire resulting IR.

This is just to demonstrate that indeed the pass will take up the last TileInterface operation in the IR, and start tiling+fusing and will indeed just have one scf.forall (I can perhaps add a stricter CHECK-NEXT statements to ensure that ending with the return statement).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
Abhishek-Varma: Yep, you're correct. We'll be revisiting this all iteratively. Currently we're trying to mimic exactly what the [pad transform dialect script gets us](https://github.com/nod-ai/iree-amd-aie/blob/1e2aa24c856ce33fcef3d749e98c4ac2aff407ac/tests/samples/matmul_fill_spec_pad.mlir#L39). Similar thing for the pack transform dialect script, for now.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -20,4 +20,14 @@ def AMDAIELowerExecutableTarget :
   let constructor = "mlir::iree_compiler::AMDAIE::createAMDAIELowerExecutableTargetPass()";
 }
 
+def AMDAIETileAndFuse :
+    Pass<"iree-amdaie-tile-and-fuse", ""> {
+  let summary = "Pass to tile and fuse TilingInterface operations.";
```
newling, line 25: This summary doesn't really explain what the pass does.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
```diff
@@ -0,0 +1,337 @@
...
+  }
+  return scf::SCFFuseProducerOfSliceResult{fusableProducer,
+                                           tileAndFuseResult->tiledValues[0],
+                                           tileAndFuseResult->tiledOps};
+}
+
+/// Starting from `op` walk all operands backwards to find all
+/// potentially fusable operations, i.e. operations that implement
+/// the `TilingInterface`.
+static void collectTiledAndFusedOps(Operation *rootOp,
```
newling, line 131: I hope the plan is to re-use the code in IREE, to avoid code duplication.

tests/samples/tile_and_fuse.mlir
nirvedhmeshram: Yes actually we want to keep the CHECK's general enough so that they dont break with minor changes like ops moving around.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
nirvedhmeshram: Yes this will be refactored once Mahesh finishes upstream changes to allow scf for_all tiling

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
nirvedhmeshram: This is lifted from here
https://github.com/openxla/iree/blob/main/compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.td#L128
These summary's are high level mentioning what it accomplishes which in this case is tile and fuse, how its done should be explained in the cpp file doing it.

tests/samples/matmul_fill_spec_pad_new.mlir
nirvedhmeshram: added a comment explaining that

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
```diff
@@ -0,0 +1,403 @@
...
+    // operation, do nothing. do nothing.
+    return success();
+  }
+  if (!llvm::hasSingleElement(countOps)) {
+    return exportOp->emitOpError(
+        "unexpected multiple flow.dispatch.workgroup_count_from_dag_root "
+        "operations "
+        "in body");
+  }
+  return TypeSwitch<Operation *, LogicalResult>(countOps[0])
```
MaheshRavishankar, line 69: Instead of this for now you can just replace the operation with all results being value 1 for now.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
```diff
@@ -0,0 +1,403 @@
...
+        return lowerWorkgroupCountFromSliceOp(rewriter, countOp, entryPointFn,
+                                              workgroupCount,
+                                              maxWorkgroupParallelDims);
+      })
+      .Default([&](Operation *) { return success(); });
+}
+
+/// Clones the operation and updates the destination if the operation
+/// implements the `DestinationStyleOpInterface`.
+static Operation *cloneOpAndUpdateDestinationArgs(RewriterBase &rewriter,
```
MaheshRavishankar, line 80: For all methods that care copied over from core can you just demaracte them with 
```
//===----------------------------------------------------------------------------------===//
// Methods copied over from core to implement tile and fuse with scf.forall.
// TODO(MaheshRavishankar) Replace them with core methods when upstream
// can handle this natively
//===----------------------------------------------------------------------------------===//
```

and

//===----------------------------------------------------------------------------------===//
// End methods copied over from core to implement tile and fuse with scf.forall.
//===----------------------------------------------------------------------------------===//
```

**Review by MaheshRavishankar — CHANGES_REQUESTED** [review]
Nice! Thanks @Abhishek-Varma and @nirvedhmeshram . I will take the AI of replacing it with core methods when they are ready (maybe a week or so away). Just a few minor changes.

File: compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp (hunk 9 above)
Comment (Abhishek-Varma) [reply compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp:69]: I didn't understand this point.

Can you please elaborate? Also, the intention behind the same.

Comment (nirvedhmeshram) [reply compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp:69]: I can add this, @Abhishek-Varma the intention is that the workgroup count lowering is dummy for the aie backend as we dont have this concept in our programming model, so this logic that I blindly lefted from the cpu backend can be greatly simplified.

**Comment by newling** [comment]
@MaheshRavishankar  Is there a reason bufferization done in the transform dialect (matmul_fill_spec_pad.mlir)? Can it be done as a separate pass in Passes.cpp?

File: compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp [hunk 11]
```diff
@@ -0,0 +1,402 @@
...
+  }
+  auto forallLoop = cast<scf::ForallOp>(tilingResult->loops[0]);
+  yieldedValuesToOrigValues.append(rootOp->result_begin(),
+                                   rootOp->result_end());
+  // A map from untiled value to scf.forall shared_outs. The shared_outs is used
+  // for DPS init operand if they use the same init operands.
+  llvm::DenseMap<Value, Value> mapToSharedOuts;
+
+  // WAR for `if` ops generating `scf.if` operations.
+  if (auto rootPadOp = dyn_cast<tensor::PadOp>(rootOp)) {
```
MaheshRavishankar, line 258: The pad stuff seems like a distraction for now... If you want to remove it that would be fine. We might need it later if we want to handle pad, but we can punt on it for now.

MaheshRavishankar (approved):

PR #65, opened 2024-01-08
[AMD-AIE] Add initial transform and padding strategy for Matmul + cleanup

Abhishek-Varma:
-- This commit introduces a pass which would pad a MatmulOp, for now, and would rewrite/transform its input operands for 1st level.
-- It also introduces a clean up pass.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPadAndBufferize.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPadAndBufferize.cpp
@@ -1,0 +1,172 @@
+// Copyright 2023 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree-amd-aie/Transforms/PassDetail.h"
+#include "iree-amd-aie/Transforms/Passes.h"
+#include "mlir/Dialect/Bufferization/IR/Bufferization.h"
+#include "mlir/Dialect/Linalg/Transforms/TilingInterfaceImpl.h"
+#include "mlir/Dialect/Linalg/Transforms/Transforms.h"
+#include "mlir/Dialect/Linalg/Utils/Utils.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/Dialect/MemRef/Transforms/Transforms.h"
+#include "mlir/Dialect/SCF/Transforms/TileUsingInterface.h"
+#include "mlir/Dialect/SCF/Transforms/Transforms.h"
+#include "mlir/IR/Iterators.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+#define DEBUG_TYPE "iree-amdaie-pad-and-bufferize"
+
+namespace mlir::iree_compiler::AMDAIE {
+
+namespace {
+
+static LogicalResult applyBufferizeToAllocation(RewriterBase &rewriter,
+                                                Operation *op,
+                                                Attribute memorySpace) {
+  linalg::BufferizeToAllocationOptions options;
+  options.memcpyOp =
+      linalg::BufferizeToAllocationOptions::MemcpyOp::MaterializeInDestination;
+  options.allocOp = linalg::BufferizeToAllocationOptions::AllocOp::MemrefAlloc;
+  options.bufferizeDestinationOnly = true;
+  options.emitDealloc = true;
+
+  // Bufferize ops.
+  Value buffer =
+      linalg::bufferizeToAllocation(rewriter, options, op, memorySpace);
+  if (!buffer) {
+    LLVM_DEBUG(llvm::dbgs() << "----- failed to bufferize operation -----\n");
+    return failure();
+  }
+  return success();
+}
+
+static LogicalResult applyPadAndConvertToDPS(
+    RewriterBase &rewriter, linalg::MatmulOp linalgTarget,
+    linalg::LinalgPaddingOptions &options) {
+  linalg::LinalgOp paddedOp;
+  SmallVector<Value> replacements;
+  SmallVector<tensor::PadOp> newPadOps;
+  if (failed(rewriteAsPaddedOp(rewriter, linalgTarget, options, paddedOp,
+                               replacements, newPadOps))) {
+    LLVM_DEBUG(llvm::dbgs() << "----- failed to pad op -----\n");
+    return failure();
+  }
+
+  // We need to perform our own replacement here because this API is still
+  // used in patterns that "pad and hoist", for which the replacement values
+  // need to be different.
+  // TODO: clean this up and stop "pad and hoist" behavior more globally now
+  // that we have more composable abstractions.
+  rewriter.replaceOp(linalgTarget, replacements);
+  // We rewrite each operand of the linalgOp (currently, MatmulOp) into DSP.
+  for (auto newPadOp : newPadOps) {
+    rewriter.setInsertionPointAfter(newPadOp);
+    if (failed(linalg::rewriteInDestinationPassingStyle(rewriter, newPadOp))) {
+      LLVM_DEBUG(llvm::dbgs() << "----- failed to rewrite in DPS -----\n");
+      return failure();
+    }
+  }
+
+  return success();
+}
+
+class AMDAIEPadAndBufferizePass
+    : public AMDAIEPadAndBufferizeBase<AMDAIEPadAndBufferizePass> {
+ public:
+  void getDependentDialects(DialectRegistry &registry) const override {
+    registry.insert<bufferization::BufferizationDialect, tensor::TensorDialect,
+                    linalg::LinalgDialect>();
+  }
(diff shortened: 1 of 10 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPadAndBufferize.cpp
```diff
@@ -0,0 +1,260 @@
...
+  StringAttr allocOp;
+  bool bufferizeDestinationOnly;
+  bool emitDealloc;
+};
+
+static LogicalResult applyBufferizeToAllocation(
+    RewriterBase &rewriter, Operation *op,
+    _BufferizeToAllocationOptions _options) {
+  linalg::BufferizeToAllocationOptions options;
+  if (_options.memCpyOp == "bufferization.materialize_in_destination") {
```
yzhang93, line 47: I think it might be better to move this to somewhere else, e.g. Utils.cpp? This can be reused by other passes/pipelines.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -17,10 +17,31 @@
...
+  // `useCPlusPlusTransformPasses` which would be used during development
+  // efforts. Once the C++ passes are ready, we will include these passes by
+  // default and take away the guarding.
+  if (clUseCPlusPlusTransformPasses) {
+    pm.addPass(createAMDAIETileAndFusePass(1));
+    pm.addPass(createAMDAIEPadAndBufferizePass());
+    pm.addPass(createCleanupPass());
+    pm.addPass(createCSEPass());
+    // pm.addPass(createAMDAIETileAndFusePass(2));
+  }
```
yzhang93, line 43: Should we reorganize `buildAMDAIETransformPassPipeline` and add padding pass in a way similar to LLVMCPU pipelines? https://github.com/openxla/iree/blob/73f5947ff170fbf1ea8e415c019673cb7dd06aea/compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp#L280

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPadAndBufferize.cpp
```diff
@@ -0,0 +1,260 @@
...
+/// in order to adapt to LinalgPadding options. Once we address upstream
+/// methods by having APIs, we can better address this instead using the
+/// upstream struct. I could've used the upstream struct here, but I chose
+/// not to in order to first adapt to minimal changes required for the C++
+/// methods to work.
+struct _LinalgPaddingOptions {
+  SmallVector<Attribute> paddingValues;
+  ArrayAttr paddingDimensions;
+  ArrayAttr packPaddings;
+  StringAttr copyBackOp;
```
MaheshRavishankar, line 92: Still reviewing this, but I think we dont need to try to use Attributes here... The attributes are used by the transform dialect operation since it has to be a serializable op. We can use a more direct data structure here, like a list of integers or a BitVector, etc. Even for the copy back op. We always use the same copy back op, so we dont need that generality in our options.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPadAndBufferize.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+/// TODO(avarma): For now we are creating a temporary struct options
+/// in order to adapt to BufferizeToAllocation options. Once we address upstream
+/// methods by having APIs, we can better address this instead using the
+/// upstream struct. I could've used the upstream struct here, but I chose
+/// not to in order to first adapt to minimal changes required for the C++
+/// methods to work.
+struct _BufferizeToAllocationOptions {
+  Attribute memorySpace;
+  StringAttr memCpyOp;
```
MaheshRavishankar, line 37: I think apart from memorySpace, all other options can be dropped. We always use the same choice for these.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPadAndBufferize.cpp
```diff
@@ -0,0 +1,260 @@
...
+  StringAttr allocOp;
+  bool bufferizeDestinationOnly;
+  bool emitDealloc;
+};
+
+static LogicalResult applyBufferizeToAllocation(
+    RewriterBase &rewriter, Operation *op,
+    _BufferizeToAllocationOptions _options) {
```
MaheshRavishankar, line 45: Lets make this simpler and just use the pat we need for now.

MaheshRavishankar (changes requested):
Thanks! Just a few small comments, and simplifications. After that it should be good though!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
Abhishek-Varma: Fair ask. Although, for similar justification I'd request taking these reorganisation after we have an e2e C++ for at least the Matmul Pad flow. Would help better once we have the complete picture ready.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPadAndBufferize.cpp
```diff
@@ -0,0 +1,172 @@
...
+          if (op.getLoopIteratorTypes().empty() || !isa<linalg::MatmulOp>(op))
+            return WalkResult::advance();
+          matmulOp = cast<linalg::MatmulOp>(op);
+          return WalkResult::interrupt();
+        });
+    if (!matmulOp) {
+      LLVM_DEBUG(llvm::dbgs() << "----- skip, no matmul op -----\n");
+      return;
+    }
+    for (auto operand : matmulOp->getOperands()) {
```
MaheshRavishankar, line 154: Wait, we shouldnt be bufferizing yet. We need to do more levels tile and distribute on tensors right?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPadAndBufferize.cpp
```diff
@@ -0,0 +1,172 @@
...
+void AMDAIEPadAndBufferizePass::runOnOperation() {
+  MLIRContext *context = &getContext();
+  IREE::HAL::ExecutableVariantOp variantOp = getOperation();
+  ModuleOp innerModule = variantOp.getInnerModule();
+  for (func::FuncOp funcOp : innerModule.getOps<func::FuncOp>()) {
+    linalg::MatmulOp matmulOp;
+    funcOp->walk<WalkOrder::PostOrder, ReverseIterator>(
+        [&](TilingInterface op) {
+          // Find the next matmul op if it does not have loops.
+          if (op.getLoopIteratorTypes().empty() || !isa<linalg::MatmulOp>(op))
```
MaheshRavishankar, line 96: Maybe make this an `LinalgOp`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Cleanup.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+/// COPIED FROM IREE TRANSFORM.
+/// Fold `tensor.pad(cst, tensor.extract*(linalg.fill(cst)))` into
+/// `linalg.fill(cst, empty)` when the padding constant and the fill constant
+/// are the same.
+/// This seems generally desirable as a folding but may be too intrusive, so we
+/// only apply it selectively for now.
+// TODO: atm hardcoded on linalg.fill but we could take any result of any
+// generic that yields a constant in that result.
+struct FoldFillIntoPad : public OpRewritePattern<tensor::PadOp> {
```
MaheshRavishankar, line 67: Isnt there a pattern to do this upstream already?

MaheshRavishankar: Found it. Here https://github.com/llvm/llvm-project/blob/9ed30012fb4f43de42ef2f265fe384d9d0b0edf2/mlir/lib/Dialect/Linalg/IR/LinalgOps.cpp#L601

MaheshRavishankar (changes requested):

MaheshRavishankar (approved):
Thanks!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPadAndBufferize.cpp
nirvedhmeshram: Hmmm, this was never actually done and is the reason for issues here, https://github.com/nod-ai/iree-amd-aie/pull/206
@Abhishek-Varma did you make a stab at it and ran into issues, or just didnt have a testr case yet so skipped. I am asking becuase I will try to do it tomorrow so just pre-fetching what you might know.

nirvedhmeshram: You can disregard was pretty trivial to add it, will push a commit with this change by EOD.

PR #66, opened 2024-01-08
[RFC] Supporting convolution

newling:
I won't land this. This is just my proposed plan to getting convolution working through IREE.

### DIFF
--- a/RFCs/convolution/handling_convolution.md
+++ b/RFCs/convolution/handling_convolution.md
@@ -1,0 +1,53 @@
+# RFC: Convolution support (resnet-50 and beyond)
+
+There are a few approaches to supporting convolution through IREE with AIE target. This is what I think is the order of easiest (with worst performance) to most difficult (with best performance).
+
+
+## 1 Send conv to CPU
+
+If a dispatch has a linalg.convolution in it, then that dispatch goes to CPU. With the approach proposed by Mahesh on discord:
+
+https://discord.com/channels/973663919757492264/1104195883307892837/1192962050540974310
+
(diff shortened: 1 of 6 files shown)


RFCs/convolution/handling_convolution.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+I have heard that this is the approach that the DPU team used, at some point, with success. 
+
+There are 2 PRs out for this: https://github.com/nod-ai/iree-amd-aie/pull/80 and https://github.com/nod-ai/iree-amd-aie/pull/78
+
+
+## 3 Direct support of convolution 
+
+We should eventually be able to lower linalg.conv directly to AIE. It was mentioned in the meeting this morning that AIE has special instruction-level support for convolution. As a first step, the code offloaded from IREE can be scalar, this way mlir-air and mlir-aie do not need to support convolution in any way (mlir-air will just see a linalg.generic op). As a second step, see if the (tiled) convolution can be propagated down through mlir-aie to target the best instructions.
```
MaheshRavishankar, line 44: We should look into seeing if the CPU path works on AIE path... AFAIK, once it gets to `vector.contract` it should connect to rest of the lowering in MLIR-AIE (we might not have pulled that in yet into IREE)

MaheshRavishankar (commented):
Thanks! 

Sorry it took me a while to look at this. It didnt show up on my notifcations since I wasnt added as a reviewer. Please do so that it comes on my radar.

newling:
Closing for now, I have another PR with a transform dialect script: https://github.com/nod-ai/iree-amd-aie/pull/157

PR #70, opened 2024-01-12
[AMD-AIE] Add final e2e C++ pass pipeline constructs for Matmul Pad flow

Abhishek-Varma:
This PR contains 3 commits :-

**Commit 1 (Level 2 - Tile + Fuse + Pad + Bufferize + Cleanup)**
-- This commit adds a level based Tile+Pad logic and adds a fix
   to tile-and-fuse pass.
-- The fix to the tile-and-fuse addressed the following issues :-
   a. The pass was made to pick up the last TilingInterface op and because
      of this linalg.copy was being pulled in as the root op.
   b. While trying to pull in untiled producers, couple of linalg.copy also
      made it to the list. Although this seemed correct, this isn't what
      we get via the transform dialect script since it directly fuses a
      linalg.fill.
      
**Commit 2 (Level 3 - Tile using Scf.for + Fuse + Pad + Bufferize + Cleanup)**
-- This commit majorly involves adding a distinctive utility for tiling via scf.for
   and scf.forall - which will be replaced once upstream core method gets
   merged.
   
**Commit 3 (Finale - Bufferize)**
-- This commit adds `IREEComprehensiveBufferizePasses` to be run as
   part of the C++ pass pipeline.
-- It shifts the test cases pertaining to checking individual C++
   passes to Transforms/test.
-- It also adds a few nit changes to `--iree-amaie-pad-and-bufferize`.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -27,6 +28,40 @@ static llvm::cl::opt<bool> clUseCPlusPlusTransformPasses(
```
nirvedhmeshram, line 28: lets change the default value back to false as we dont want other developers flow to change till we refactor this in a different pipeline.

tests/samples/matmul_fill_spec_pad.mlir
```diff
@@ -1,135 +0,0 @@
-// This script shows an example lowering matmul through IREE for a special accelerator.
```
nirvedhmeshram, line 1: lets not delete it? we may want to keep it for those who want to experiment with it.

tests/samples/matmul_fill_static_i32.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: iree-compile --iree-hal-target-backends=amd-aie --compile-to=executable-sources %s | iree-opt --pass-pipeline="builtin.module(hal.executable(hal.executable.variant(iree-hal-translate-target-executable-variants{target=amd-aie})))" --iree-codegen-transform-dialect-library=%S/matmul_fill_spec_pad_new.mlir | FileCheck %s
+// RUN: iree-compile --iree-hal-target-backends=amd-aie --compile-to=executable-sources %s | iree-opt --pass-pipeline="builtin.module(hal.executable(hal.executable.variant(iree-hal-translate-target-executable-variants{target=amd-aie})))" | FileCheck %s
```
nirvedhmeshram, line 1: we want to keep the support for transform dialect as well, so lets not change this script.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEEliminateEmptyTensor.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Dialect/SCF/Transforms/Patterns.h"
+#include "mlir/Dialect/SCF/Transforms/TileUsingInterface.h"
+#include "mlir/Dialect/SCF/Transforms/Transforms.h"
+#include "mlir/Dialect/Utils/StaticValueUtils.h"
+#include "mlir/IR/Iterators.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "mlir/Transforms/LoopInvariantCodeMotionUtils.h"
```
nirvedhmeshram, line 32: Please check which header files are not needed and then you can remove them.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPadAndBufferize.cpp
```diff
@@ -44,6 +44,45 @@ static LogicalResult applyBufferizeToAllocation(RewriterBase &rewriter,
   return success();
 }
 
+static LogicalResult formLinalgPaddingOptions(
```
nirvedhmeshram, line 47: may be the name should be ` formMatmulPaddingOptions` and the name of the pass should be `...BufferizeMatmul.cpp` ?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
```diff
@@ -327,26 +376,41 @@ void AMDAIETileAndFusePass::runOnOperation() {
...
+    // a given tilinglevel, similar to LLVMCPUTileAndFuse.cpp.
+    SmallVector<OpFoldResult> tileSizes;
+    // TODO(avarma): Have a global CONSTANT defining tiling stages and the
+    // tiling strategy.
+    if (tilingLevel == 1) {
+      tileSizes = getAsIndexOpFoldResult(context, {8, 8});
+    } else if (tilingLevel == 2) {
+      tileSizes = getAsIndexOpFoldResult(context, {4, 4});
+    } else if (tilingLevel == 3) {
+      tileSizes = getAsIndexOpFoldResult(context, {0, 0, 4});
```
nirvedhmeshram, line 389: what does a tile size of 0 mean?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
```diff
@@ -327,26 +376,41 @@ void AMDAIETileAndFusePass::runOnOperation() {
...
+      // llvm::outs()<<funcOp<<"==============\n\n";
```
nirvedhmeshram, line 390: remove debug print comment?

nirvedhmeshram (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPadAndBufferize.cpp
MaheshRavishankar, line 47: Instead of interleaving the logic this way, can we separate out the logic using the padding level as a hihg-level switch.

```
static FailureOr<LinalgPaddingOptions> getFirstLevelLinalgPaddingOptions() {
....
}

static FailureOr<LinalgPaddingOptions> getSecondLevelLinalgPaddingOptions() {

}
...

static FailureOr<LinalgPaddingOptions> getLinalgPaddingOptions(int64_t level) {
  if (level == 0) {
   return getFirstLevelPaddingOptions();
  }
  if (level == 1) {
    return getSecondLevelPaddingOptions();
  } 
   ...
  return failure();
}
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPadAndBufferize.cpp
```diff
@@ -44,6 +44,45 @@ static LogicalResult applyBufferizeToAllocation(RewriterBase &rewriter,
...
+    IRRewriter &rewriter, linalg::MatmulOp &matmulOp, int64_t paddingLevel,
+    linalg::LinalgPaddingOptions &options) {
+  SmallVector<Attribute> paddingValues;
+  for (auto operand : matmulOp->getOperands()) {
+    auto type = dyn_cast<RankedTensorType>(operand.getType());
+    if (!type) {
+      matmulOp->emitOpError("failed to lower workgroup count region");
```
MaheshRavishankar, line 54: This error message seems out of place.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPadAndBufferize.cpp
```diff
@@ -151,8 +171,19 @@ void AMDAIEPadAndBufferizePass::runOnOperation() {
       LLVM_DEBUG(llvm::dbgs() << "----- skip, no matmul op -----\n");
       return;
     }
-    for (auto operand : matmulOp->getOperands()) {
+    auto i64Type = rewriter.getI64Type();
+    for (unsigned i = 0, n = matmulOp->getNumOperands(); i < n; i++) {
```
MaheshRavishankar, line 175: THis is a bit hard to read.. again refactor this to switch on `paddingLevel`.

```
FailureOr<SmallVector<Value>> getOperandsToBufferize(int64_t paddingLevel, Operation *matmul) {
  if (paddingLevel == 1) {
     return {...}
  }
  if (paddingLevel == 2) {
    return {..}
  }
}

FailureOr<SmallVector<Value>> bufferizeOperands = getOperandsToBufferize(...);
if (failed(bufferizeOperands)) {
  ...
}
for (auto operands : bufferizeOperands.value()) {
  if (failed(applyBufferizeToAllocation)) {
    ..
  }
}

```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -35,12 +70,35 @@ void buildAMDAIETransformPassPipeline(OpPassManager &pm) {
...
+    pm.addPass(createCSEPass());
+    pm.addPass(createAMDAIETileAndFusePass(3));
+    pm.addPass(createCleanupPass());
+    pm.addPass(createCanonicalizerPass());
+    pm.addPass(createCSEPass());
+    pm.addPass(createAMDAIEPadAndBufferizePass(3));
+    pm.addPass(createCleanupPass());
+    pm.addPass(createCanonicalizerPass());
+    pm.addPass(createCSEPass());
+    pm.addPass(createAMDAIEEliminateEmptyTensorPass());
```
MaheshRavishankar, line 94: Isnt the `EliminateEmptyTensorsPass` already run as part of addIREEComprehensiveBufferizePasses. Do we need to run this pass at all 

https://github.com/openxla/iree/blob/914f30641f3c41338f8a7820711743ca49a2a76f/compiler/src/iree/compiler/Codegen/Common/IREEComprehensiveBufferizePass.cpp#L249

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEEliminateEmptyTensor.cpp
```diff
@@ -0,0 +1,103 @@
+// Copyright 2023 The IREE Authors
```
MaheshRavishankar, line 1: Follow up from comment below. This seems like a copy of what is done here https://github.com/openxla/iree/blob/914f30641f3c41338f8a7820711743ca49a2a76f/compiler/src/iree/compiler/Codegen/Common/IREEComprehensiveBufferizePass.cpp#L147 and should already be run as part of `IREEComprehensiveBufferizePasses` added to the pass pipeline. I think this file can be removed.

tests/samples/tile_and_fuse_using_scf_for.mlir
```diff
@@ -0,0 +1,75 @@
+// RUN: iree-opt --pass-pipeline='builtin.module(hal.executable(hal.executable.variant(iree-amdaie-tile-and-fuse{tiling-level=3})))' --split-input-file %s | FileCheck %s --check-prefix=TILE-LEVEL-3
```
MaheshRavishankar, line 1: Actually these are just checking each pass. These can just go into `compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/tests/` . If there are any other similar tests that were added to the samples folder please move them to the location above.

MaheshRavishankar (changes requested):

nirvedhmeshram (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
```diff
@@ -288,6 +321,20 @@ LogicalResult applyTileAndFuse(RewriterBase &rewriter, Operation *rootOp,
...
 
+LogicalResult applyTileAndFuseUsingSCF(RewriterBase &rewriter,
+                                       Operation *rootOp,
+                                       DominanceInfo &dominanceInfo,
+                                       scf::SCFTilingOptions options,
+                                       int64_t tilingLevel) {
+  if (tilingLevel == 3) {
+    return applyTileAndFuseUsingSCFFor(rewriter, rootOp, dominanceInfo,
+                                       options);
+  } else {
```
yzhang93, line 332: Can we separate TileAndFuseUsingSCFFor without using tilingLevel as 3? One reason to do so is in packing pipeline, we have tiling organized as scf.forall - scf.for - scf.forall. It would be weird to use tilingLevel 1, 3 and 2...

Abhishek-Varma: I'm not sure if I understood the requirement (and thus the solution) properly but adding a few points to consider. Let me know your thoughts.

I guess as a temporary workaround we should indeed club it as 1 -> 3 -> 2, for now since all these are in their very initial stages.

I can do the following :-
```
Pack =>
tileFuse(1)
tileFuse()
tileFuse(2)

Pad =>
tileFuse(1)
tileFuse(2)
tileFuse()
```
The above would be one way to avert the numbering weirdness - but it's just not tackling the real issue and is still a temporary workaround.

I'd therefore still suggest to go ahead with the current PR because of the following reasons :-
1. @nirvedhmeshram and I are implementing set/get functionality for `lowering_config` + `translation_info` - this would help in having a separate pipeline for pad and pack based approaches along with the expected tiling sizes at each level. One of the constraints for the pack based approach should be addressed by this.
2. @MaheshRavishankar mentioned that with his upstream PR we would be able to better control whether to use scf.for or scf.forall and also club them with tiling level while invoking the pass itself as flag/template - and these constructs would soon evolve in a better shape iteratively. The other constraint of clubbing 1->2->3 or 1->3->2 should be addressed by this incrementally.
3. Since this is a first complete e2e C++ pipeline - any other implementation/fixes we make would be on top of this so we should at least have this revision (working for pad based flow) in `main`.

yzhang93:
> @yzhang93 the `hal.executable.export public` function name and the `func.func` name have to match, in your case they are different as `matmul_example_dispatch_0_matmul_16x2048x2048_i8xi8xi32` and `matmul_example_dispatch_0_matmul_16x256x256_i8xi8xi32` respectively this is the workgroup lowering part that we are doing in this pass not directly related to tiling thats complaining about the entry point name (the hal.executable) name not matching the func.func name I believe.

Thanks @nirvedhmeshram! Yes, that's the problem. I didn't notice the function name is different.

MaheshRavishankar (approved):
THis looks good! Thanks @Abhishek-Varma

PR #72, opened 2024-01-15
Add bufferization to compiler pass pipeline 

newling:
This allows compilation to proceed further. Testing on the simple example:

```
export IREE_BUILD_DIR=~/iree_build
$IREE_BUILD_DIR/tools/iree-compile --iree-hal-target-backends=amd-aie simple_abs.mlir --mlir-print-ir-before-all --print-module-scope > dump.mlir
```

Before this patch, this failed in the --air-dependency pass, which assumes input is bufferized. With the addition of bufferization in the main pipeline in this PR, compilation proceeds to the --airrt-to-ipu. 

I guess we could remove bufferization from the transform dialect tests now, although there is no harm in bufferizing twice I assume?

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
@@ -65,6 +65,8 @@ void addMLIRAIRAIELoweringPasses(OpPassManager &passManager) {
   passManager.addPass(createEraseHALDescriptorTypeFromMemRefPass());
   passManager.addPass(createAMDAIEBridgeToAIRPass());
   passManager.addPass(memref::createFoldMemRefAliasOpsPass());
+  passManager.addPass(
+      mlir::iree_compiler::createIREEComprehensiveBufferizePass());
 
   {
     xilinx::air::ParallelToHerdOptions options;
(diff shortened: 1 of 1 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -65,6 +65,8 @@ void addMLIRAIRAIELoweringPasses(OpPassManager &passManager) {
   passManager.addPass(createEraseHALDescriptorTypeFromMemRefPass());
   passManager.addPass(createAMDAIEBridgeToAIRPass());
   passManager.addPass(memref::createFoldMemRefAliasOpsPass());
+  passManager.addPass(
+      mlir::iree_compiler::createIREEComprehensiveBufferizePass());
```
Abhishek-Varma, line 69: Hi @newling - this perhaps won't be required since [e2e PR](https://github.com/nod-ai/iree-amd-aie/pull/70) is taking care of adding `createIREEComprehensiveBufferizePass` [here](https://github.com/nod-ai/iree-amd-aie/blob/afcccb558bf26d93b6a8ee4e363140ff949a40dc/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp#L56-L63).

CC: @MaheshRavishankar

MaheshRavishankar (changes requested):
We probably dont need this.

PR #73, opened 2024-01-15
[XRT] fix bugs when building with tracing

nirvedhmeshram:
I was not building with tracing flag before. Abhishek found some bugs when building with it. This PR fixes it. Verified that the matmul still runs with it.


Abhishek-Varma (approved):
Thanks for the quick fix @nirvedhmeshram !

antiagainst (approved):

PR #74, opened 2024-01-15
Add packing and transpose pass

yzhang93:


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPackAndTranspose.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+namespace {
+
+struct PackConfig {
+  SmallVector<OpFoldResult> packedSizes;
+  SmallVector<int64_t> transposePackIndices;
+  SmallVector<int64_t> unpackEmpty;
+  SmallVector<SmallVector<int64_t>> innerPerm;
+  SmallVector<SmallVector<int64_t>> outerPerm;
+};
```
Abhishek-Varma, line 24: Add a brief comment about the individual fields.

tests/samples/pack_and_transpose_level2.mlir
```diff
@@ -0,0 +1,67 @@
+// RUN: iree-opt --pass-pipeline='builtin.module(hal.executable(hal.executable.variant(iree-amdaie-pack-and-transpose{pack-level=2})))' --split-input-file %s | FileCheck --check-prefix=CHECK-2 %s
+
+#map = affine_map<(d0) -> (d0 * 16)>
+#map1 = affine_map<(d0) -> (d0 * 64)>
+#map2 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d2, d3, d5)>
+#map3 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d2, d1, d5, d4)>
```
Abhishek-Varma, line 6: Just confirming - I see we have created two separate files for the two different `pack-level`s.

Is the lit test crashing if we put the two in the same file with the exact set of check prefixes? If yes, is there a way we can put a guard on the code to avoid the crash and instead have the two test cases in the same file?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPackAndTranspose.cpp
```diff
@@ -0,0 +1,165 @@
...
+#include "mlir/Dialect/Linalg/Transforms/Transforms.h"
+#include "mlir/Pass/Pass.h"
+
+#define DEBUG_TYPE "iree-amdaie-pack-and-transpose"
+
+namespace mlir::iree_compiler::AMDAIE {
+
+namespace {
+
+struct PackConfig {
```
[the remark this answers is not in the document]
yzhang93: Yes, there's no such utility available upstream.

tests/samples/pack_and_transpose_level2.mlir
Abhishek-Varma: I believe just bailing out cleanly from the C++ code with a `return signalPassFailure()` should do - this way it wouldn't crash and have a clean exit.

Perhaps @MaheshRavishankar can field this better.

yzhang93: Sorry, I didn't make it clear. It won't cause a crash, but will throw an expected error, e.g., `error: 'linalg.matmul' op requires number of packed sizes match the number of loops (6 vs 3)` with a wrong tiling level. I just couldn't find a way to test the expected error with prefix.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPackAndTranspose.cpp
yzhang93: Done

MaheshRavishankar (approved):
This looks good to me! Thanks Vivian for prefetching some of the comments and addressing them beforehand.

PR #78, opened 2024-01-17
implicit convolution (a single top-level im2col)

newling:
This is the simplest form of implicit convolution we can do. It performs the transformation before any tiling, and so the data rearrangement and duplicating (9x duplication for a 3x3 kernel window)  happens in L3 (DDR) before any matmul. 

Note that batch size = 1 converts convolution to matmul, but batch size > 1 converts convolution to linalg.generic which may require more work to support on AIE. 

Thanks @qedawkins for pointing me to the preprocessing pipeline!

### DIFF
--- a/tests/samples/CMakeLists.txt
+++ b/tests/samples/CMakeLists.txt
@@ -11,6 +11,7 @@ iree_lit_test_suite(
     "matmul_fill_static_i32.mlir"
     "pad_and_bufferize.mlir"
     "tile_and_fuse.mlir"
+    "implicit_convolution.mlir"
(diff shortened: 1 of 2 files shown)


newling:
@MaheshRavishankar ok. I've tried to make get it in the overflow in #80 (work in progress). I won't merge this PR if I can get #80 working.

newling:
I'm closing this for now, while I look at a direct convolution approach. 
I think I also prefer #80 (add im2col directly to compiler pipeline, rather than a user provided flag)

PR #80, opened 2024-01-17
Attempt at adding the implicit convolution to the direct compiler flow

newling:
In https://github.com/nod-ai/iree-amd-aie/pull/78 I showed how to add a compiler flag in a lit test to do implicit convolution. 

In this PR I try to do it directly by registering a pass. 

It kind of works, except it is making the pass run irrespective of the backend, which is not what I want (see the 2 tests: 1 is for rocm backend, which should not be running the pass for implicit convolution). 

@qedawkins do you know how I do this? i.e. make the pass pipeline modification only for amd-aie backend.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/PluginRegistration.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/PluginRegistration.cpp
@@ -40,7 +40,8 @@ struct AMDAIESession
   }
 
   void extendPreprocessingPassPipeline(OpPassManager &pm) override {
-    pm.addPass(iree_compiler::Preprocessing::createConvertConv2DToImg2ColPass());
+    pm.addPass(
+        iree_compiler::Preprocessing::createConvertConv2DToImg2ColPass());
   }
 };
 
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/PluginRegistration.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/PluginRegistration.cpp
@@ -11,8 +11,10 @@
 #include "iree-amd-aie/Target/AIETarget.h"
 #include "iree-amd-aie/Transforms/Passes.h"
 #include "iree/compiler/Dialect/HAL/Target/TargetRegistry.h"
+#include "iree/compiler/Pipelines/Options.h"
 #include "iree/compiler/PluginAPI/Client.h"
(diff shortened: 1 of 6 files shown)


qedawkins:
I've never used it myself, but there is a way to provide a callback for preprocessing passes as a callback: https://github.com/openxla/iree/blob/390eeb07ff095eb67a9f114601f1e453951c94e7/compiler/src/iree/compiler/Preprocessing/Passes.cpp#L68

Basically when invoking the IREE compile entry point you can add (in C++) hooks for extending the pass pipeline in various places: https://github.com/openxla/iree/blob/390eeb07ff095eb67a9f114601f1e453951c94e7/compiler/src/iree/compiler/Pipelines/Pipelines.h#L32

Here is the callback to add for preprocessing: https://github.com/openxla/iree/blob/390eeb07ff095eb67a9f114601f1e453951c94e7/compiler/src/iree/compiler/PluginAPI/Client.h#L82

qedawkins:
(also I haven't looked closely at the configuration of the compiler plugin here to know that this will actually work for you, just that it seems like what you're grasping for? Stella would probably know better on how these pipeline extensions are expected to work.)

newling:
Thanks @qedawkins. Yeah I think 
https://github.com/openxla/iree/blob/390eeb07ff095eb67a9f114601f1e453951c94e7/compiler/src/iree/compiler/PluginAPI/Client.h#L82 is where I'm currently plugging into IREE. I'm still trying to understand the call chain, and I asked a question on Discord about this.

MaheshRavishankar:
Could we just make the decision of running im2col be passed in as a command line option... Whatever is "packaging" IREE can use that command line option?

newling:
> Could we just make the decision of running im2col be passed in as a command line option... Whatever is "packaging" IREE can use that command line option?

Right, that's what #78 is though? Maybe I misinterpreted your comment there. I was imagining something like this, so that the thing "packaging" doesn't need to do anything extra.

MaheshRavishankar:
> > Could we just make the decision of running im2col be passed in as a command line option... Whatever is "packaging" IREE can use that command line option?
> 
> Right, that's what #78 is though? Maybe I misinterpreted your comment there. I was imagining something like this, so that the thing "packaging" doesn't need to do anything extra. 

Yeah. What is in #78  works. I approved it there. That doesn't actually need to be tested either. It is tested in IREE

newling:
I'm closing this for now, which I investigate a direct convolution approach.

PR #84, opened 2024-01-18
NFC: General cleanup of IREE-AMD-AIE code and tests.

MaheshRavishankar:
Most of the passes added recently dont need to run on the
`hal.executable.variant`. They could be run on `func.func`. This
removes a lot of boiler plate from the tests. As a consequence the
workgroup count region handling needs to be moved out of the tile and
fuse pass since it has to be run in a pass rooted on a
`hal.executable.variant`. Move just the workgroup count handling to a
separate pass.

Also organize passes better in header and tablegen files (make them
alphabetical).

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerWorkgroupCount.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerWorkgroupCount.cpp
@@ -1,0 +1,59 @@
+// Copyright 2024 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#include "iree-amd-aie/Transforms/PassDetail.h"
+#include "iree-amd-aie/Transforms/Passes.h"
+#include "iree/compiler/Dialect/Flow/IR/FlowOps.h"
+#include "iree/compiler/Dialect/HAL/IR/HALOps.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Support/LogicalResult.h"
+
+namespace mlir::iree_compiler::AMDAIE {
+
+namespace {
+class AMDAIELowerWorkgroupCountPass
+    : public AMDAIELowerWorkgroupCountBase<AMDAIELowerWorkgroupCountPass> {
+  void runOnOperation() override;
+};
+}  // namespace
+
+/// Lower workgroup count operations to all 1s.
+static LogicalResult lowerWorkgroupCount(
+    RewriterBase &rewriter, IREE::HAL::ExecutableExportOp exportOp) {
+  Block *body = exportOp.getWorkgroupCountBody();
+  if (!body) {
+    return exportOp.emitOpError("unexpected empty workgroup count region");
+  }
+  OpBuilder::InsertionGuard g(rewriter);
+  for (auto workgroupCountOp : llvm::make_early_inc_range(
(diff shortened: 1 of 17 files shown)


MaheshRavishankar:
Please review only the last commit. The first commit is part of a separate PR.

MaheshRavishankar:
Also @Abhishek-Varma , @nirvedhmeshram  and @yzhang93 , this commit is basically things I should have caught on review. So I went ahead and cleaned things up. So for future please ensure all commits follow similar conventions.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -14,24 +14,38 @@ def AMDAIEBridgeToAIR : Pass<"iree-amdaie-bridge-to-air", ""> {
   let constructor = "mlir::iree_compiler::AMDAIE::createAMDAIEBridgeToAIRPass()";
 }
 
+def AMDAIECleanup :
+    Pass<"iree-cleanup", "func::FuncOp"> {
```
Abhishek-Varma, line 18: `iree-cleanup` -> `iree-amdaie-cleanup` ?
This can be skipped because, I believe, we plan on getting rid of this pass and just keeping `canonicalize + cse` combo.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerWorkgroupCount.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include "iree-amd-aie/Transforms/PassDetail.h"
+#include "iree-amd-aie/Transforms/Passes.h"
+#include "iree/compiler/Dialect/Flow/IR/FlowOps.h"
+#include "iree/compiler/Dialect/HAL/IR/HALOps.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Support/LogicalResult.h"
+
+namespace mlir::iree_compiler::AMDAIE {
+
```
Abhishek-Varma, line 15: 1. Better to add a `DEBUG_TYPE` ? Although it's needed when guarding a debug statement with `LLVM_DEBUG`, it helps know the actual pass name that can be invoked via command-line.

2. Also, would it be favourable to add a lit test for this now that a separate pass is created ? (We anyway don't have a separate pass testing the cleanup pass so perhaps we can skip this - just adding a query)

Abhishek-Varma (approved):
Thanks for this @MaheshRavishankar ! Will keep these conventions in mind hereon.

The comments are non-blocking, so we can still go ahead and have this merged.

yzhang93 (approved):
Thanks @MaheshRavishankar for cleaning up the codes! This now looks very neat. Just one minor thing, we may want to update the copyright for recent patches to `//Copyright 2024 The IREE Authors`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
MaheshRavishankar: Ok sounds good. This pass still does loop hoisting... that might be useful to keep. Ill rename the pass name anyway.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerWorkgroupCount.cpp
MaheshRavishankar: I think I already added  a lit test?

Abhishek-Varma: Ouch. My bad! Just checked again. Sounds good.

PR #86, opened 2024-01-18
Use new tool in mlir-aie

spaceotter:
Replaces the complex code to get the xclbin file with a call to a single tool.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -7,19 +7,23 @@
 #include "iree-amd-aie/Target/AIETarget.h"
 
 #include <filesystem>
```
MaheshRavishankar, line 9: Are we use `filesystem` anywhere now. We can probably remove that?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -130,813 +135,118 @@ class AIETargetBackend final : public IREE::HAL::TargetBackend {
...
+  SmallVector<std::string> entryPointNames;
+  for (auto exportOp : variantOp.getExportOps()) {
+    entryPointNames.emplace_back(exportOp.getSymName());
   }
 
-  // Create a temporary work directory needed for subsequent commands.
-  FailureOr<std::string> workDir =
-      Artifact::createTemporaryDirectory(variantOp.getName());
-  if (failed(workDir)) {
+  if (entryPointNames.size() == 0) {
```
MaheshRavishankar, line 183: Nit: You can just have `if (entryPointsNames.size()  == 1) { return ... "expected single entry point");}` for now.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -130,813 +135,118 @@ class AIETargetBackend final : public IREE::HAL::TargetBackend {
...
+    return moduleOp.emitOpError("Entry point missing");
+  } else if (entryPointNames.size() > 1) {
+    return moduleOp.emitOpError("Multiple entry points NYI");
+  }
+
+  SmallVector<StringRef> CmdArgs{aie2xclbin,
```
MaheshRavishankar, line 189: Nit: `s/CmdArgs/cmdArgs`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -130,813 +135,118 @@ class AIETargetBackend final : public IREE::HAL::TargetBackend {
...
-                              StringRef(dumpString.data(), dumpString.size()));
+    moduleOp.print(inputMlirOut->os(), OpPrintingFlags().useLocalScope());
+    inputMlirOut->keep();
   }
 
-  XclBinGeneratorKit toolkit(options.peanoInstallDir, options.vitisInstallDir,
-                             options.showInvokedCommands);
+  SmallString<128> aie2xclbin(options.mlirAieInstallDir);
+  llvm::sys::path::append(aie2xclbin, "bin", "aie2xclbin");
+  SmallString<128> XCLBinPath(workDir);
```
MaheshRavishankar, line 171: Lets keep a consistent naming convention. All variable names start with lower case. So use `xclBinPath` and `ipuInstPath`, etc.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -130,813 +135,118 @@ class AIETargetBackend final : public IREE::HAL::TargetBackend {
...
-    OpBuilder::InsertionGuard g(executableBuilder);
-    executableBuilder.setInsertionPoint(moduleOp);
-    auto clonedModuleOp =
-        dyn_cast<ModuleOp>(executableBuilder.clone(*moduleOp.getOperation()));
-    if (failed(generateCoreElfFiles(moduleOp, objFile.value(), workDir.value(),
-                                    getOptions(), toolkit))) {
-      return failure();
+  std::vector<uint32_t> ipuInstrs;
+
+  std::ifstream instr_file(static_cast<std::string>(IPUInstPath));
```
MaheshRavishankar, line 234: I am not sure why we are doing this? I would assume that we could just open it, read it as a file and just embed it without having to do any processing here.

MaheshRavishankar (changes requested):
Very Nice! Great milestone! Just have a few comments.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
spaceotter: The function I call in the tool writes hexadecimal.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
MaheshRavishankar: I think MLIR does something differently. Thats what we tend to follow. Here too there is a mix with some local variables capitalized and some not. I think not capitalized is the convention.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
MaheshRavishankar: I thought even the binary IREE writes to vmfb in hex. @nirvedhmeshram can you comment on this.

nirvedhmeshram: I think the file is a string of hex values, and we are converting it to a vector of uint32_t for the vmfb here. So it seems required, unless the tool can allow to keep the vector in memory and we never go to disk.

MaheshRavishankar (approved):

PR #87, opened 2024-01-18
Fix XRT runtime driver

spaceotter:
The XRT runtime driver was missing a cmake library dependency after the IREE update.


nirvedhmeshram (approved):
Thanks!

PR #88, opened 2024-01-18
Fixes to AIR dialect (and minor IREE fix).

MaheshRavishankar:
This seems to fix an error that is seen locally, but isnt showing up in CI.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/BridgeToAIRPass.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/BridgeToAIRPass.cpp
@@ -43,7 +43,7 @@ class SCFForAllToParallelOp : public OpRewritePattern<scf::ForallOp> {
     // Fixup the terminator
     OpBuilder::InsertionGuard g(rewriter);
     rewriter.setInsertionPointToEnd(&parallelOp.getRegion().front());
-    rewriter.replaceOpWithNewOp<scf::ReduceOp>(
+    rewriter.replaceOpWithNewOp<scf::YieldOp>(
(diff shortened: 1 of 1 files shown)


Abhishek-Varma (approved):
Thanks @MaheshRavishankar !

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/BridgeToAIRPass.cpp
```diff
@@ -43,7 +43,7 @@ class SCFForAllToParallelOp : public OpRewritePattern<scf::ForallOp> {
     // Fixup the terminator
     OpBuilder::InsertionGuard g(rewriter);
     rewriter.setInsertionPointToEnd(&parallelOp.getRegion().front());
-    rewriter.replaceOpWithNewOp<scf::ReduceOp>(
+    rewriter.replaceOpWithNewOp<scf::YieldOp>(
```
spaceotter, line 46: One of the recent changes is that scf::ParallelOp now has to be terminated by scf::ReduceOp. I don't know how this is still working.

spaceotter (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/BridgeToAIRPass.cpp
spaceotter: You are getting a build error in mlir-air that was fixed upstream https://github.com/Xilinx/mlir-air/blob/7f69c06f631eeec6482e8300a07eed4866a68a43/mlir/lib/Conversion/AIRLoweringPass.cpp#L755

PR #89, opened 2024-01-18
[AMD-AIE] Allow iree-amdaie-tile-and-use pass to switch between scf.for* via flag

Abhishek-Varma:
-- This commit adds a pass flag `--use-scf-for` (default `false`) to
   `--iree-amdaie-tile-and-fuse` pass in order to allow it to switch
   between scf.for OR scf.forall based on the corresponding pipeline's
   invocations.


MaheshRavishankar (approved):
Its ok to do this, but I think it might be better to use the newer form of the base classes generated from tablegen for iree-amd-aie. Is that an AI that you can take up?

Abhishek-Varma:
> Its ok to do this, but I think it might be better to use the newer form of the base classes generated from tablegen for iree-amd-aie. Is that an AI that you can take up?

Sure, I will take that up as a follow up PR along with the newer approach for [memorySpace](https://github.com/nod-ai/iree-amd-aie/pull/90#issuecomment-1901113107) on Monday.

PR #90, opened 2024-01-19
[AMD-AIE] Separate pad-and-bufferize into two passes

Abhishek-Varma:
-- This commit separates `--iree-amdaie-pad-and-bufferize` into two
   individual passes : `--iree-amdaie-pad` and
   `--iree-amdaie-bufferize-to-allocation`.

Co-authored-by: Vivian Zhang <vivizhan@amd.com>


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
```diff
@@ -0,0 +1,124 @@
...
+  Value buffer =
+      linalg::bufferizeToAllocation(rewriter, options, op, memorySpace);
+  if (!buffer) {
+    LLVM_DEBUG(llvm::dbgs() << "----- failed to bufferize operation -----\n");
+    return failure();
+  }
+  return success();
+}
+
+// TODO(avarma): This is a temporary workaround until we have PaddingStrategy
```
newling, line 40: nit: if you used an enum class for this it'd make the code easier to understand. Is there a reason using ints is easier?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/bufferize_to_allocation_pad.mlir
```diff
@@ -0,0 +1,127 @@
...
+
+func.func @matmul_static(%arg0 : tensor<8x16xi32>, %arg1 : tensor<16x8xi32>) -> tensor<8x8xi32> {
+  %c0 = arith.constant 0 : index
+  %c0_i32 = arith.constant 0 : i32
+  %5 = tensor.empty() : tensor<8x8xi32>
+  %6 = scf.forall (%iv0, %iv1) = (0, 0) to (8, 8) step (8, 8) shared_outs(%arg2 = %5) -> (tensor<8x8xi32>) {
+    %extracted_slice_0 = tensor.extract_slice %arg0[%iv0, 0] [8, 16] [1, 1] : tensor<8x16xi32> to tensor<8x16xi32>
+    %extracted_slice_2 = tensor.extract_slice %arg1[0, %iv1] [16, 8] [1, 1] : tensor<16x8xi32> to tensor<16x8xi32>
+    %11 = tensor.empty() : tensor<8x8xi32>
+    %extracted_slice_3 = tensor.extract_slice %11[%iv0, %iv1] [8, 8] [1, 1] : tensor<8x8xi32> to tensor<8x8xi32>
```
newling, line 12: extracted_slice_3 isn't used, extracted_slice_4 is a no-op? 

It's not super simple IR: maybe add a sentence at the start of funcs about what the func does, where it's from, and what is being tested (to help make it easier understand what is being tested).

newling (commented):
What is our overall approach to bufferization, is all bufferization happening in this pass? It seems like we're not using IREE comprehensive bufferization approach. Is the current approach going to work for arbitrary linalg operations?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/bufferize_to_allocation_pack.mlir
```diff
@@ -0,0 +1,98 @@
+// RUN: iree-opt --pass-pipeline='builtin.module(func.func(iree-amdaie-bufferize-to-allocation{memory-space=1}))' --split-input-file %s | FileCheck %s --check-prefix=CHECK-1
+// RUN: iree-opt --pass-pipeline='builtin.module(func.func(iree-amdaie-bufferize-to-allocation{memory-space=2}))' --split-input-file %s | FileCheck %s --check-prefix=CHECK-2
+
```
Abhishek-Varma, line 3: @kumardeepakamd this is the pass which we are pulling out and adding it separately since both pad and pack based approach use it.

[Here](https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/pad_and_bufferize.mlir#L89-L112) is the current way this takes place in top-of-main for matmul pad based and the e2e for matmul pad based flow works on top-of-main.

yzhang93:
> What is our overall approach to bufferization, is all bufferization happening in this pass? It seems like we're not using IREE comprehensive bufferization approach. Is the current approach going to work for arbitrary linalg operations?

This is not comprehensive/one-shot bufferization. This pass bufferizes the targeted operation and materializes the result in a new allocation. We'll still need comprehensive bufferization at the end of the pipeline.

newling:
> > What is our overall approach to bufferization, is all bufferization happening in this pass? It seems like we're not using IREE comprehensive bufferization approach. Is the current approach going to work for arbitrary linalg operations?
> 
> This is not comprehensive/one-shot bufferization. This pass bufferizes the targeted operation and materializes the result in a new allocation. We'll still need comprehensive bufferization at the end of the pipeline. Take a look at the tests in this PR and it will help to understand what it does.

Makes sense now, thanks. So comprehensive bufferize doesn't do the right thing bufferizing this targeted op?

yzhang93:
> > > What is our overall approach to bufferization, is all bufferization happening in this pass? It seems like we're not using IREE comprehensive bufferization approach. Is the current approach going to work for arbitrary linalg operations?
> > 
> > 
> > This is not comprehensive/one-shot bufferization. This pass bufferizes the targeted operation and materializes the result in a new allocation. We'll still need comprehensive bufferization at the end of the pipeline. Take a look at the tests in this PR and it will help to understand what it does.
> 
> Makes sense now, thanks. So comprehensive bufferize doesn't do the right thing bufferizing this targeted op?

I would say they are two separate things. They are also two separate passes in llvm-project.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/bufferize_to_allocation_pad.mlir
yzhang93: Yes, this is taken taken directly from the output of padding pass without canonicalization to mimic the transform dialect transformations. I also feel some of the test cases are tedious and we may need to cleanup the test cases in the following patch.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
yzhang93: We use int values so that it has good correspondence with the paddingLevel. I've modified this part to make it clear to read.

MaheshRavishankar (approved):

MaheshRavishankar:
Meta comment. I think it makes sense to start using better constructs for memory space. Something akin to this https://github.com/llvm/llvm-project/blob/f5e58a038033c8736fed91a7e89a5faad462abcc/mlir/include/mlir/Dialect/GPU/IR/GPUBase.td#L84 . I think we can start using these at the tensor level and fix the bufferization to convert these attributes into some `IntegerAttr`s. I think this is because the AIR flow uses the `IntegerAttr` directly. We could do this using something like this https://github.com/llvm/llvm-project/pull/78484/files . 

@Abhishek-Varma is this something that I can put on your plate. its not super high priority, but would be good to have. If someone else wants to take it up please respond.

Abhishek-Varma:
> Meta comment. I think it makes sense to start using better constructs for memory space. Something akin to this https://github.com/llvm/llvm-project/blob/f5e58a038033c8736fed91a7e89a5faad462abcc/mlir/include/mlir/Dialect/GPU/IR/GPUBase.td#L84 . I think we can start using these at the tensor level and fix the bufferization to convert these attributes into some `IntegerAttr`s. I think this is because the AIR flow uses the `IntegerAttr` directly. We could do this using something like this https://github.com/llvm/llvm-project/pull/78484/files .
> 
> @Abhishek-Varma is this something that I can put on your plate. its not super high priority, but would be good to have. If someone else wants to take it up please respond.

Sure, I can take that up as a follow-up PR @MaheshRavishankar on Monday.

PR #91, opened 2024-01-19
fix single-core gemm 

frafranz:
The translation to mlir for this case was already working fine, but during the translation to TIR wrong parentheses appeared.
This is now fixed in AIESerializer.cpp and the single-core gemm works.


yzhang93:
Thanks for fixing the issue. Please add tests so we can go forward with this patch.

frafranz:
I now added tests to distinguish the single and multi-core case.

frafranz:
I further noticed https://github.com/nod-ai/iree-amd-aie/pull/67/files#diff-3e38ea9ec1ab9b80ae0566b2215a827cc5849e52943640235bfa9ccf13712120R719 in your current PR for double-buffering translator support. It seems that this is also direct at the issue I fixed here, let me know how you want to proceed with it @yzhang93

frafranz:
Besided, having this case fixed allows us to delete this comment, right? https://github.com/nod-ai/iree-amd-aie/blob/f8cf19bd748fe244411423e3a0913c799f87c409/tests/samples/matmul_fill_spec_pack_peel.mlir#L125 Since peeling now also works in the single-core case.

MaheshRavishankar (approved):

frafranz:
Closing this PR, since the changes were now addressed in the more recent https://github.com/nod-ai/iree-amd-aie/pull/127

PR #93, opened 2024-01-22
[AMD-AIE] Use newer form of the pass base class generated

Abhishek-Varma:
-- This commit refactors all the passes of iree-amd-aie plugin
   to use the newer form of the pass base class generated via
   tablegen.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.h
```diff
@@ -32,42 +33,41 @@ void addPadBasedPassPipeline(OpPassManager &passManager,
...
 /// Create a pass to lower workgroup count region of entry point operations.
-std::unique_ptr<OperationPass<IREE::HAL::ExecutableVariantOp>>
-createAMDAIELowerWorkgroupCountPass();
+std::unique_ptr<Pass> createAMDAIELowerWorkgroupCountPass();
 
 /// Create a pass to pack and transpose the linalg op.
+std::unique_ptr<Pass> createAMDAIEPackAndTransposePass();
 std::unique_ptr<OperationPass<func::FuncOp>> createAMDAIEPackAndTransposePass(
-    int64_t packLevel = 1);
+    const AMDAIEPackAndTransposeOptions &options);
```
nirvedhmeshram, line 57: can default values to options not be provided with these refactor?

Abhishek-Varma: The `.td` will take care of the default values. The beauty of using the newer base class is that even if we add/remove options, we don't have to change the function declarations and invocations - just the newer/old option parameter needs to be added/removed right before invoking the pass.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
```diff
@@ -63,19 +62,20 @@ static FailureOr<SmallVector<Value>> getOperandsToBufferize(
 }
 
 class AMDAIEBufferizeToAllocationPass
-    : public AMDAIEBufferizeToAllocationBase<AMDAIEBufferizeToAllocationPass> {
+    : public impl::AMDAIEBufferizeToAllocationBase<
```
nirvedhmeshram, line 65: I am not sure, why you need the namespace impl:: , I do not see this any in-tree passes that are also using the newer form e.g https://github.com/openxla/iree/blob/e82cfa99b3dc6447fe926329c12414bce75c0507/compiler/src/iree/compiler/Codegen/LLVMCPU/ExpandF16OpToF32Pass.cpp#L54

MaheshRavishankar: AFAIK, nothing in the `Codegen/` directory is using the new form. In the new form the `*Base` class is within the `impl` namespace (you can take a look at the `Passes.h.inc` generated).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
```diff
@@ -120,10 +120,14 @@ void AMDAIEBufferizeToAllocationPass::runOnOperation() {
 
 }  // namespace
 
+std::unique_ptr<Pass> createAMDAIEBufferizeToAllocationPass() {
```
MaheshRavishankar, line 123: I think you can combine these two to
```
std::unique_ptr<Pass> create*Pass(AMDAIE*Options options = {});
```

The default initializer should set the fields to the default values.

MaheshRavishankar (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
Abhishek-Varma: Really nice! Thanks.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.h
```diff
@@ -32,42 +33,35 @@ void addPadBasedPassPipeline(OpPassManager &passManager,
...
 /// Create a pass to lower workgroup count region of entry point operations.
-std::unique_ptr<OperationPass<IREE::HAL::ExecutableVariantOp>>
-createAMDAIELowerWorkgroupCountPass();
+std::unique_ptr<Pass> createAMDAIELowerWorkgroupCountPass();
 
 /// Create a pass to pack and transpose the linalg op.
-std::unique_ptr<OperationPass<func::FuncOp>> createAMDAIEPackAndTransposePass(
-    int64_t packLevel = 1);
+std::unique_ptr<Pass> createAMDAIEPackAndTransposePass(
+    const AMDAIEPackAndTransposeOptions &options = {});
```
MaheshRavishankar, line 54: Nit: I am not sure what happens if you initialize a reference this way. It maybe fine, but I have seen this segfault sometimes. Drop the `const &` here and everywhere.

MaheshRavishankar (changes requested):

MaheshRavishankar (approved):

PR #94, opened 2024-01-22
Add pass that calls upstream datalayout propagation patterns

nirvedhmeshram:
This pass is not currently in a pipeline. But just adding it if we want it when and if we enable fusion.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/CMakeLists.txt
```diff
@@ -44,6 +44,7 @@ iree_cc_library(
     "AMDAIEPad.cpp"
     "AMDAIEPeelForLoop.cpp"
     "AMDAIETileAndFuse.cpp"
+    "AMDAIEPropagateDataLayout.cpp"
```
Abhishek-Varma, line 47: Alphabetically needs to go one level up.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPropagateDataLayout.cpp
```diff
@@ -0,0 +1,45 @@
...
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree-amd-aie/Transforms/PassDetail.h"
+#include "iree-amd-aie/Transforms/Passes.h"
+#include "mlir/Dialect/Linalg/Transforms/Transforms.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+#define DEBUG_TYPE "iree-amdaie-propagate-pad"
```
Abhishek-Varma, line 13: Can we have a common pass name?

Currently, I see that :-
DEBUG_TYPE = `iree-amdaie-propagate-pad`
.td = `iree-amdaie-propagate-data-layout`
Test file name = `datalayout_propagation`

It'd be nice to have a common name - helps debugging/navigating eventually.

Abhishek-Varma (changes requested):
This is nice. But a few comments to address before getting this in.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPropagateDataLayout.cpp
```diff
@@ -0,0 +1,45 @@
...
+  AMDAIEPropagateDataLayoutPass() = default;
+  AMDAIEPropagateDataLayoutPass(const AMDAIEPropagateDataLayoutPass &pass){};
+  void runOnOperation() override;
+};
+
+void AMDAIEPropagateDataLayoutPass::runOnOperation() {
+  MLIRContext *context = &getContext();
+  RewritePatternSet patterns(context);
+  linalg::populateDataLayoutPropagationPatterns(
+      patterns, [](Operation *op) { return true; });
```
MaheshRavishankar, line 35: Do we want to just propagate everything by default.... We also need to call this from multiple places. Its ok to land and iterate, but this probably needs a bit more control.

MaheshRavishankar (approved):

Abhishek-Varma (approved):
Thanks! LGTM!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPropagateDataLayout.cpp
nirvedhmeshram: Ya it works for the test case I tried but may have to tweak based on what we see in the wild. Will keep this in mind.

PR #99, opened 2024-01-23
Sync the IPU instruction sequence before execution

spaceotter:
The AIE device is not cache coherent, so the IPU instructions need to be flushed explicitly.


nirvedhmeshram (commented):
Thanks for the fix, can you please check if we can insert the sync over here
https://github.com/nod-ai/iree-amd-aie/blob/main/runtime/src/iree-amd-aie/driver/xrt/native_executable.cc#L195
This way if the dispatch is called multiple times we are not unecessarily syncing.

nirvedhmeshram (approved):
Thanks!

PR #106, opened 2024-01-26
Add lowering strategy pass

jtuyls:
@nirvedhmeshram @yzhang93 @MaheshRavishankar 

This PR adds `AMDAIEAddLoweringStrategyPass` that adds tiling configurations based on a `tiling-strategy` option. Please let me know whether this is a good approach for enabling different tiling strategies, for example for `ukernel` vs `direct-codegeneration`? 

@nirvedhmeshram If you want I can use this PR to update the code here to rely on `AMDAIEAddLoweringStrategyPass` instead, like I think you meant in this TODO: https://github.com/nod-ai/iree-amd-aie/blob/aa32917f1db27ae97ebfaa9b176b534fcb4720f4/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerExecutableTarget.cpp#L107 Let me know.


nirvedhmeshram:
> @nirvedhmeshram @yzhang93 @MaheshRavishankar
> 
> This PR adds `AMDAIEAddLoweringStrategyPass` that adds tiling configurations based on a `tiling-strategy` option. Please let me know whether this is a good approach for enabling different tiling strategies, for example for `ukernel` vs `direct-codegeneration`?
> 
> @nirvedhmeshram If you want I can use this PR to update the code here to rely on `AMDAIEAddLoweringStrategyPass` instead, like I think you meant in this TODO:
> 
> https://github.com/nod-ai/iree-amd-aie/blob/aa32917f1db27ae97ebfaa9b176b534fcb4720f4/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerExecutableTarget.cpp#L107
> 
> Let me know.

Yes updating the code there would be great. Thanks!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -34,10 +34,24 @@ def AMDAIECleanup :
...
+    Option<"tilingStrategy", "tiling-strategy", "std::string", /*default=*/"",
+      "The tiling strategy to be used">  
+  ];
+}
+
 def AMDAIELowerExecutableTarget :
     Pass<"iree-amdaie-lower-executable-target", "mlir::iree_compiler::IREE::HAL::ExecutableVariantOp"> {
   let summary = "Perform lowering of executable target using one of the IREE::HAL::DispatchLoweringPassPipeline";
   let constructor = "mlir::iree_compiler::AMDAIE::createAMDAIELowerExecutableTargetPass()";
+  let options = [
```
MaheshRavishankar, line 51: We can just drop this, and just call the `AIEAddLoweringStrategy` before in our pass-pipeline. So we need to set it only one place.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -34,10 +34,24 @@ def AMDAIECleanup :
       "mlir::iree_compiler::AMDAIE::createAMDAIECleanupPass()";
 }
 
+def AMDAIEAddLoweringStrategy :
+    Pass<"iree-amdaie-add-lowering-strategy", "mlir::iree_compiler::IREE::HAL::ExecutableVariantOp"> {
+  let summary = "Add lowering strategy configurations to be used";
+  let constructor = "mlir::iree_compiler::AMDAIE::createAMDAIEAddLoweringStrategyPass()";
+  let options = [
+    Option<"tilingStrategy", "tiling-strategy", "std::string", /*default=*/"",
```
MaheshRavishankar, line 42: A `Stringef` option and string matching is better avoided. Can we just use a boolean option, and call it "usePackingStrategy" cause thats what we are using here.

MaheshRavishankar (changes requested):
Thansk Jorn! This is actually much better than what I was discussing with @yzhang93 . Just a few nits, but otherwise this looks good.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
jtuyls: Ok, I will change to boolean. As for the naming, I think `usePackingStrategy` is a bit narrow as the unique feature here is the usage of a fixed/handcoded kernel/library function. So, I would prefer `useMicroKernelStrategy` or `useMLOpsLibStrategy` (MLOpsLib is the library of handcoded kernel functions). What do you think?

yzhang93: Follow up this discussion, I think eventually we may have more tiling strategies. So maybe use an int option? And add explanations for what each value represent.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
jtuyls: I made this change and added the AIEAddLoweringStrategy pass to the pass pipeline: https://github.com/nod-ai/iree-amd-aie/pull/106/files#diff-42f0d0bb098689f25ee68e8f05ec6c2eaa89ce41a6394d7d99c1d1c912943b38R111

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
jtuyls: @yzhang93 I agree that we likely need more tiling strategies so an int (or better enum?) would be great. @MaheshRavishankar mentioned yesterday that could be done in a follow-up PR? I made the change now to use a boolean.

MaheshRavishankar:
This looks good to me. I am landing this change. cc @yzhang93 and @Abhishek-Varma

MaheshRavishankar (approved):

MaheshRavishankar:
Landed using #115

PR #107, opened 2024-01-26
Add pass to fuse fill op into forall loop

yzhang93:


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/fuse_fill_into_forall.mlir
```diff
@@ -0,0 +1,68 @@
+// RUN: iree-opt --pass-pipeline='builtin.module(func.func(iree-amdaie-fuse-fill-into-forall))' %s | FileCheck %s
+
+#map = affine_map<(d0) -> (d0 * 16)>
+#map1 = affine_map<(d0) -> (d0 * 64)>
+#map2 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d2, d3, d5)>
+#map3 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d2, d1, d5, d4)>
+#map4 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d1, d3, d4)>
+func.func @matmul_example_dispatch_0_matmul_16x256x256_i8xi8xi32() {
```
MaheshRavishankar, line 8: This example is fine, but this is much more verbose than it needs to be. A simple example of a `scf.forall` + `linalg.fill` outside would have done the trick. :)

MaheshRavishankar (approved):
Ok, this is OK to land, but the pass landed here might get folded away soon. Its good to have this for the time being though.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/fuse_fill_into_forall.mlir
yzhang93: Make sense! I'll spend some time cleaning up some tests next week.

PR #111, opened 2024-01-29
Use type of example's payload

newling:
matmul_fill_static.mlir is a f32 matmul, so the padding values in the transform dialect should be f32, not i32 (the instructions in the transform dialect script matmul_fill_spec_pad.mlir are for the file matmul_fill_static.mlir)


tests/samples/matmul_fill_static.mlir
```diff
@@ -1,9 +1,23 @@
...
-    %rhs : tensor<256x512xf32>) -> tensor<128x512xf32> {
-  %empty = tensor.empty() : tensor<128x512xf32>
-  %cst = arith.constant 0.0 : f32
-  %fill = linalg.fill ins(%cst : f32) outs(%empty : tensor<128x512xf32>) -> tensor<128x512xf32>
-  %2 = linalg.matmul ins(%lhs, %rhs : tensor<128x256xf32>, tensor<256x512xf32>)
-      outs(%fill : tensor<128x512xf32>) -> tensor<128x512xf32>
-  return %2 : tensor<128x512xf32>
+// Example using a transform dialect script.
+
+// RUN: iree-opt %s   --iree-hal-target-backends=amd-aie \
```
MaheshRavishankar, line 3: You can use `iree-compile` to make a more succinct command line https://github.com/nod-ai/iree-amd-aie/blob/a42ddaca69c1068f242c08cd204232f3078b2040/tests/samples/matmul_fill_static_i32.mlir#L1

tests/samples/matmul_fill_static.mlir
```diff
@@ -1,9 +1,23 @@
...
+// RUN:               --iree-abi-transformation-pipeline \
+// RUN:              --iree-flow-transformation-pipeline \
+// RUN:            --iree-stream-transformation-pipeline \
+// RUN:                --iree-hal-configuration-pipeline \
+// RUN: | cat -  $(dirname %s)/matmul_fill_spec_pad.mlir \
```
MaheshRavishankar, line 8: Why `cat`?

tests/samples/matmul_fill_static.mlir
```diff
@@ -1,9 +1,23 @@
...
+// RUN: | iree-opt  --iree-transform-dialect-interpreter \
```
MaheshRavishankar, line 9: Interesting, I think this approach expects the transform dialect script to be in the same file. I havent seen this usage mode.

MaheshRavishankar (changes requested):
Thanks, just a few minor questions.

tests/samples/matmul_fill_static.mlir
newling: Yes, that's why I used `cat`.

tests/samples/matmul_fill_static.mlir
newling: I liked this approach because 

1) it passes with different tensor sizes, as it does not lower to mlir-aie (for example changing to M=1024 results in a failure lowering to mlir-aie). 

2) I wanted a way to see the output of running the pass with transform dialect script, before lowering all the down to mlir-aie. 

Neither of these is a good reason to use this approach in a test, so I will change to the `iree-compile` approach. 

I guess I can use --mlir-print-ir-after=. with `iree-compile` to see the IR after transform dialect has run. 

@MaheshRavishankar - do you know if there is a way to see the IR after each of the individual passes described in the transform IR? i.e. after each of the `tile_using_forall`, `pad`, `bufferize_to_allocation` transform ops have modified the IR?

tests/samples/matmul_fill_static.mlir
```diff
@@ -1,9 +1,22 @@
...
-  %2 = linalg.matmul ins(%lhs, %rhs : tensor<128x256xf32>, tensor<256x512xf32>)
-      outs(%fill : tensor<128x512xf32>) -> tensor<128x512xf32>
-  return %2 : tensor<128x512xf32>
+// Example using a transform dialect script.
+
+// Compile to mlir-aie:
+
+// RUN: iree-compile --iree-hal-target-backends=amd-aie --compile-to=executable-sources %s | iree-opt --pass-pipeline="builtin.module(hal.executable(hal.executable.variant(iree-hal-translate-target-executable-variants{target=amd-aie})))" --iree-codegen-transform-dialect-library=%S/matmul_fill_spec_pad.mlir | FileCheck %s
+
+// To compile to just after the transform dialect pass: // RUN: iree-opt %s   --iree-hal-target-backends=amd-aie --iree-abi-transformation-pipeline --iree-flow-transformation-pipeline --iree-stream-transformation-pipeline --iree-hal-configuration-pipeline | cat - %S/matmul_fill_spec_pad.mlir | iree-opt --iree-transform-dialect-interpreter | FileCheck %s
```
MaheshRavishankar, line 7: These two `RUN` commands need to combined into one. The first `iree-opt` in this `RUN` command and the first `iree-compile` in the first `RUN` command are essentially the same.

For the second command to run the transform dialect script maybe follow what is here https://github.com/openxla/iree/blob/eaefd3ff964f14e700b1c468aa2b4f20a4401028/tests/transform_dialect/cpu/matmul.mlir#L18 (ignore the first `iree-opt` command, it is doing what you are doing here when it doesnt need to.. bonus points for fixing that as well :P )

newling: I updated it, but I'm not 100% sure this is what you have in mind...

MaheshRavishankar (approved):
Interesting! I think I understand what you are trying to do

newling:
This PR is stale and I think no longer relevant

PR #113, opened 2024-01-30
[AMD-AIE] Add support for C++ Pack pipeline

Abhishek-Varma:
This PR adds the c++ pass pipeline to lower linalg operations using pack-based approach.  Since we now need to support three separate efforts now 
1) Pad-based lowering
2) Simple Pack-based lowering (now set as default)
3) Pack-based lowering using more advanced optimizations like peeling and double buffering.

This PR now adds an enum to use to switch between the two pipelines. The flag `iree-amdaie-use-pipeline` can be used to pick between the three.

Also removed is the use of  the CPU pipeline enums from core as a proxy. Unfortunately, the enums for the pass pipelines setup in core dont make it extensible in a plugin. This needs to be addressed, but for now, the "None" value is used, and instead a pack based approach is set as default. This could be made robust, but will require upstream changes.


Abhishek-Varma:
Hi all.

I tried rebasing it over [Update mlir-aie PR](https://github.com/nod-ai/iree-amd-aie/pull/112) but it doesn't generate `.vmfbs`.
Leads to :-
```
iree/runtime/src/iree/base/internal/file_io.c:199: NOT_FOUND; failed to open file 'test_pad_cpp.vmfb'; loading bytecode module at 'test_pad_cpp.vmfb'; loading modules and dependencies; creating run context
```

Without that commit I've tested this branch and it successfully generates vmfbs with correct results for :-
1. Transform Pad.
2. Pad.
3. Pack.

Pack based pipeline is going to be the default case and if we want to test for Pad based approach we need to add `--iree-amdaie-pad-pipeline` flag and it'd work by setting the appropriate TranslationInfo and choosing the corresponding pipeline.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
```diff
@@ -125,6 +130,10 @@ void AMDAIETileAndFusePass::runOnOperation() {
   SmallVector<int64_t> tileSizesVal;
   if (auto loweringConfig = getLoweringConfig(consumerOp)) {
     tileSizesVal = loweringConfig.getTileSizeVals(tilingLevel);
+  } else if (isa<linalg::GenericOp>(consumerOp)) {
+    // TODO(vivian): Fix in upstream. For now this is a temp hack to get around
```
MaheshRavishankar, line 134: I flagged this on the other PR as well.... But this has been festering for a while....  This is a pretty big hammer.... I will talk to Vivian today, and me or her can get around this.

MaheshRavishankar (changes requested):
This is good. Left one comment. Another thing is to try to reconcile this with https://github.com/nod-ai/iree-amd-aie/pull/106 . That PR is how things should be structured. If it helps we can try to land that, and rebase this on top of that.

Abhishek-Varma:
> This is good. Left one comment. Another thing is to try to reconcile this with #106 . That PR is how things should be structured. If it helps we can try to land that, and rebase this on top of that.

Oh okay, Vivian mentioned that too - I see that the only reconciliation involved would be to add the `usePadPipeline` as a pass option in `.td` instead of the C++.

I can do that and retest, sure.

MaheshRavishankar:
> > This is good. Left one comment. Another thing is to try to reconcile this with #106 . That PR is how things should be structured. If it helps we can try to land that, and rebase this on top of that.
> 
> Oh okay, Vivian mentioned that too - I see that the only reconciliation involved would be to add the `usePadPipeline` as a pass option in `.td` instead of the C++.
> 
> I can do that and retest, sure.

Not really....  It is introducing a new strategy selection pass..

Abhishek-Varma:
> > Oh okay, Vivian mentioned that too - I see that the only reconciliation involved would be to add the `usePadPipeline` as a pass option in `.td` instead of the C++.
> > I can do that and retest, sure.
> 
> Not really.... It is introducing a new strategy selection pass..

Oh yes, I meant just as a delta since that PR is taking care of most of the base constructs. I guess it'd be nice to have that PR in - I can then rebase and work with that.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
yzhang93: Yes, so the plan is we don't do this in upstream, right? I'll add a workaround inside this repo. Should I just use `getCompilationInfo/setCompilationInfo` and add them inside the packing pass?

MaheshRavishankar:
cc @yzhang93 , @Abhishek-Varma  and @jtuyls 

Vivian and Jorn, I left the `PackPipeline` as a placeholder for you to add the peeling and the double buffering transformations.

MaheshRavishankar (approved):
This is ready to land now.. Ill do a few cleanups on top of this, but this is good to go.

Abhishek-Varma:
Thanks for this @MaheshRavishankar !

All tests passes - landing it.

PR #114, opened 2024-01-30
Remove some code duplication in AMDAIEPad.cpp 

newling:


Abhishek-Varma:
Hi James.

The work on this PR is effectively undoing the review comment addressed here : [Review comment for not interleaving the logic for selecting linalg.padding options based on padding-level](https://github.com/nod-ai/iree-amd-aie/pull/70#discussion_r1451026398).

I liked a few changes you made though :-
1. Using `ZeroAttr`.
2. Using `std::iota` (unless there is an `LLVM` counterpart for this - CC: @MaheshRavishankar )
3. The clean and better way of [setting linalgPaddingOptions' values](https://github.com/nod-ai/iree-amd-aie/blob/c9b27643a43ad104b98b22c56a1de35ae9054063/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPad.cpp#L50-L55).

I believe we can imbibe the above three in each of the non-interleaved versions itself and with regards to the interleaving work that this PR does - perhaps @MaheshRavishankar can comment here better on that aspect.

newling:
Thanks for taking a thorough look Abhishek. I hadn't realised that separation (and duplication) was a conscious decision. I guess the thinking is that they might need to diverge in the future, so don't tie them together now? 

Not a big deal, it's just a bit of duplication that caught my attention twice so I thought I'd do something about it.

newling:
@Abhishek-Varma I've changed the approach -- no longer 'interleaves logic' but still reduces duplication.

Abhishek-Varma:
Hi @newling 

The changes look okay - in my opinion, keeping the few changes which I mentioned earlier itself is good - not sure about adding classes and all to get rid of duplicity.

Perhaps @MaheshRavishankar can comment here better if this (creating classes and all) is something we would indeed want - depends on the future use case of the pad based approach and if we are going to be working on this front as well now that `simple-pack` is in and `pack` (a new pack pipeline) is in line.

MaheshRavishankar (changes requested):
I dont think this kind of heirarchy is needed here. We are just setting some parameter options, and its much easier to maintain if we just keep those as independent as possible, even if it means some duplication. This kind of code gets a lot of chrun, and couple all the padding levels (in the implementation) usually just makes it harder to evolve. I am just opting for simplest cognitive overhead with these kind of things (cause inheritence etc. adds additional levels of indirection).

PR #115, opened 2024-01-30
Add lowering strategy

MaheshRavishankar:
This PR adds AMDAIEAddLoweringStrategyPass that adds tiling configurations based on a tiling-strategy option. Please let me know whether this is a good approach for enabling different tiling strategies, for example for ukernel vs direct-codegeneration?


MaheshRavishankar:
This is duplicate of #106 , but rebased on ToT

PR #122, opened 2024-02-01
Add e2e matmul test

nirvedhmeshram:
This PR adds initial e2e tests if all the dependency paths are provided.


nirvedhmeshram:
I want the tests to be pulled in from the runtime plugin so that we can check if the driver is enabled before building the tests but if I move the tests to runtime plugin they build but are  not found by lit @ScottTodd

tests/matmul/generate_e2e_matmul_tests.py
```diff
@@ -0,0 +1,624 @@
...
+        op = op + (
+            f"  %acc = util.null : !hal.buffer_view\n"
+            f"  %result = call @module.{function.name}(%lhs, %rhs) : (!hal.buffer_view, !hal.buffer_view) -> !hal.buffer_view\n"
+        )
+
+    op = op + (
+        f"  %m = arith.constant {shape.m} : i64\n"
+        f"  %k = arith.constant {shape.k} : i64\n"
+        f"  %n = arith.constant {shape.n} : i64\n"
+        f"  call @matmul_test.check_matmul_results(%device, %m, %k, %n, %lhs, %rhs, %acc, %result) : (!hal.device, i64, i64, i64, !hal.buffer_view, !hal.buffer_view, !hal.buffer_view, !hal.buffer_view) -> ()\n"
```
makslevental, line 457: If only we had python bindings that enabled us to build such ir using an in-memory representations of ops and types and the module................ lol

stellaraccident: We do and use it for some things. But strings are easy :)

makslevental: I guess it says something about the bindings that strings are _still_ easier.

ScottTodd:
> I want the tests to be pulled in from the runtime plugin so that we can check if the driver is enabled before building the tests but if I move the tests to runtime plugin they build but are not found by ctest @ScottTodd

Is that with or without my patches on https://github.com/ScottTodd/iree/tree/infra-test-suites-external ? I'm still deciding how I want to resolve that: https://github.com/openxla/iree/issues/4644#issuecomment-1917558360 . I'm leaning towards pulling at least a chunk of the filtering out of those test rules and putting the onus back on leaf directories to filter as they like. In-tree that would mean that Vulkan check/matmul/etc. tests would need to check for `IREE_HAL_DRIVER_VULKAN`. Out-of-tree (like here), the tests you define would just be included without needing to inform the main project.

tests/matmul/generate_e2e_matmul_tests.py
stellaraccident: Meh, strings are always easier for this level of thing... Virtually no cognitive load and everyone can hack on them.

Remind me some day to dig up some of the early web assembly predecessor asm "string work" :)

nirvedhmeshram:
> > I want the tests to be pulled in from the runtime plugin so that we can check if the driver is enabled before building the tests but if I move the tests to runtime plugin they build but are not found by ctest @ScottTodd
> 
> Is that with or without my patches on https://github.com/ScottTodd/iree/tree/infra-test-suites-external ?

Yes I am seeing  this behavior with those patches.     Tomorrow I will try to use the shell script approach you provided as well.

ScottTodd:
> > > I want the tests to be pulled in from the runtime plugin so that we can check if the driver is enabled before building the tests but if I move the tests to runtime plugin they build but are not found by ctest @ScottTodd
> > 
> > 
> > Is that with or without my patches on https://github.com/ScottTodd/iree/tree/infra-test-suites-external ?
> 
> Yes I am seeing this behavior with those patches. Tomorrow I will try to use the shell script approach you provided as well.

Okay. The shell script is a good place to start while the CMake is still undergoing refactoring. If you do want to debug the CMake setup, I'd recommend the debugger (for CMake! yeah!) in VSCode: https://devblogs.microsoft.com/cppblog/introducing-cmake-debugger-in-vs-code-debug-your-cmake-scripts-using-open-source-cmake-debugger/ . Printf debugging with `message()` works, but the debugger lets you set breakpoints and step through while watching variables.

PR #123, opened 2024-02-01
Add a new pack pipeline with loop peeling

yzhang93:
This commit added a pack pipeline that can deal with large k dimension to fit into the memTile. Note currently the e2e flow (through MLIR-AIR/MLIR-AIE) doesn't support this pipeline, and the failure is expected. This pipeline is used to generate the FuncIR which should be the same output as running this transform dialect. https://github.com/nod-ai/iree-amd-aie/blob/main/tests/samples/matmul_fill_spec_pack_peel.mlir.

To generate FuncIR, one should run the following command:
```
./iree-compile --iree-hal-target-backends=amd-aie --compile-to=executable-sources ../../iree-amd-aie/tests/samples/pack_pipeline_funcIR.mlir  --mlir-print-ir-after=fold-memref-alias-ops | ./iree-opt --pass-pipeline="builtin.module(hal.executable(hal.executable.variant(iree-hal-translate-target-executable-variants{target=amd-aie})))" --iree-amdaie-use-pipeline=pack &>result.mlir
```

And the IR output is saved to result.mlir.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -95,7 +95,11 @@ def AMDAIEPackAndTranspose :
       "mlir::iree_compiler::AMDAIE::createAMDAIEPackAndTransposePass()";
   let options = [
     Option<"packLevel", "pack-level", "int64_t", /*default=*/"-1",
-      "Set the packing level number">
+      "Set the packing level number">,
+    Option<"usePassPipeline", "use-pass-pipeline",
```
Abhishek-Varma, line 99: Since this is `pad` specific option, just to disambiguate it with another `usePassPipeline` (which we have in LowerExecutableTarget) perhaps we rename this from `usePassPipeline` -> `padPipeline` ?

Abhishek-Varma (changes requested):
Hi @yzhang93 

This majorly looks good - Just one more comment besides the renaming : it warrants lit tests to be added to both `pack-and-transpose` and the e2e test case.

Abhishek-Varma (approved):
Removing merging blocker from my side since this is currently to generate funcIR for TVM/ukernel flow.

MaheshRavishankar (approved):
With this change, we can now delete all the transfor dialect scripts. Can we change the `tests.samples` to run the examples there with the correct pass pipeline and check the result of the transformation. The `tests/samples` needs a clean up to reflect the state of the IREE compilation phase, and this is as good a point as any.

MaheshRavishankar (changes requested):
Oops! Pressed the wrong button. Can we fix the tests here.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
yzhang93: This should be the same as the one used in LowerExecutableTarget. I'd keep this to avoid miss understanding.

yzhang93:
> With this change, we can now delete all the transfor dialect scripts. Can we change the `tests.samples` to run the examples there with the correct pass pipeline and check the result of the transformation. The `tests/samples` needs a clean up to reflect the state of the IREE compilation phase, and this is as good a point as any.

I've added a test for this pack pipeline and also deleted some unused IRs. I'd keep the transform dialect script for a bit longer until everyone's on speed with the new pipelines.

tests/samples/matmul_fill_spec_pad.mlir
```diff
@@ -4,7 +4,7 @@
 //   export IREE_DIR=${HOME}/iree/iree; \
 //   export IREE_AMD_AIE_DIR=${IREE_AMD_AIE_DIR:-${HOME}/iree/iree-amd-aie}
 //   ${IREE_DIR}/build/tools/iree-opt \
-//     ${IREE_AMD_AIE_DIR}/tests/samples/matmul_fill_static.mlir \
+//     ${IREE_AMD_AIE_DIR}/tests/samples/pad_pipeline_e2e.mlir \
```
MaheshRavishankar, line 7: Use `iree-compile --compile-to=executable-sources` here to remove the individual pipelines. (In the other examples as well)

tests/samples/matmul_fill_static.mlir
```diff
@@ -1,9 +0,0 @@
-func.func @matmul_static(%lhs : tensor<128x256xf32>,
```
MaheshRavishankar, line 1: Keep these inputs that are needed to run the transform dialect scripts. https://github.com/openxla/iree/blob/main/compiler/src/iree/compiler/Codegen/LLVMGPU/test/transform_dialect_eliminate_gpu_barriers.mlir is an example of how you can use `iree-opt` + transform dialect script to keep the examples and transform dialect script together.

MaheshRavishankar (approved):
Mostly looks OK. I want to just take the chance to santitize the tests a bit. I am approving this. Either fix this PR or santiize the tests as a follow up.

yzhang93:
> Mostly looks OK. I want to just take the chance to santitize the tests a bit. I am approving this. Either fix this PR or santiize the tests as a follow up.

Thanks! I'll add a follow up PR to clean these. Currently merge this to unblock @frafranz

Issue #118, opened 2024-01-31
pack_pipeline_e2e test is failing transiently


nirvedhmeshram:
Here is a gist showing the transient issue
https://gist.github.com/nirvedhmeshram/5deeaaa94729053794aa2dd9fbf223e0

Abhishek-Varma:
[Prefetch lowering config before Packing and set later](https://github.com/nod-ai/iree-amd-aie/pull/125) solves this issue.

Reason for the failure -  the `lowering_config` was non-deterministically not being carried forward after the first application of `pack-and-transpose` and due to cascading effect `tile-and-fuse` couldn't perform tiling from 2nd level onwards.

CC: @MaheshRavishankar @nirvedhmeshram @yzhang93

PR #125, opened 2024-02-01
[AMD-AIE] Prefetch lowering config before Packing and set later

Abhishek-Varma:
-- This commit prefetches the lowering config before applying
   packing operation and then sets the same to the final packed op.

This solves the [Transient failure of pack_pipeline_e2e.mlir](https://github.com/nod-ai/iree-amd-aie/issues/118) - the `lowering_config` was not carried forward after the first application of `pack-and-transpose` and due to cascading effect `tile-and-fuse` couldn't perform tiling from 2nd level onwards.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPackAndTranspose.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPackAndTranspose.cpp
@@ -118,7 +118,10 @@ void AMDAIEPackAndTransposePass::runOnOperation() {
     return;
   }
 
-  // Pack the operation
+  // Step 1. Before packing the operation, we will prefetch the lowering config.
+  auto config = getLoweringConfig(linalgOp);
+
+  // Step 2. Pack the operation
   IRRewriter rewriter(context);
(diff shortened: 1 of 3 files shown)


MaheshRavishankar (approved):
Nice catch! Should we also add a lit test for it.

Abhishek-Varma:
> Nice catch! Should we also add a lit test for it.

I'm not sure how to replicate that - because the pass specific thing passes anyway.
It's the e2e test which was non-deterministically failing.

nirvedhmeshram:
You could add a test like this https://github.com/nod-ai/iree-amd-aie/blob/e933de2b843e222f4351884708a35b0d3be7f7fa/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/pack_and_transpose_level1.mlir
but have a lowering config attached and check that the pack gets the same lowering config?

Abhishek-Varma:
> You could add a test like this https://github.com/nod-ai/iree-amd-aie/blob/e933de2b843e222f4351884708a35b0d3be7f7fa/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/pack_and_transpose_level1.mlir but have a lowering config attached and check that the pack gets the same lowering config?

Aah okay! I thought it already had the `lowering_config` - cool, I'll add those checks. Thanks!

PR #126, opened 2024-02-01
Add scripts for e2e Matmul CI

nirvedhmeshram:
This can be used to enable a matmul test while the CMake way of doing this catches up.

### DIFF
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -1,0 +1,153 @@
+#!/bin/bash
+# Copyright 2024 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# Manual IREE matmul test script.
+#
+# This pulls code from
+#   * https://github.com/openxla/iree/blob/main/build_tools/cmake/iree_e2e_matmul_test.cmake
+#   * https://github.com/openxla/iree/blob/main/tests/e2e/matmul/CMakeLists.txt
+#
+# Usage:
+#   1. Build IREE (or install packages). You'll need `iree-compile` to include
+#      support for the compiler target backend you wish to test and
+#      `iree-e2e-matmul-test` to include support for the runtime HAL
+#      driver/device you wish to test.
+#   2. Update the paths in this script or specify them via environment variables
+#   3. Run: `./run_matmul_tests.sh <peano_install_path> <mlir_aie_install_path> <vitis_path>`
+
+set -euo pipefail
+
+THIS_DIR="$(cd $(dirname $0) && pwd)"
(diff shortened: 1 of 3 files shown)


build_tools/ci/run_matmul_test.sh
```diff
@@ -0,0 +1,153 @@
...
+THIS_DIR="$(cd $(dirname $0) && pwd)"
+ROOT_DIR="$(cd $THIS_DIR/../.. && pwd)"
+
+OUTPUT_DIR="${IREE_MATMUL_BUILD_DIR:-build-matmul}"
+GENERATOR="${ROOT_DIR}/tests/matmul/generate_e2e_matmul_tests.py"
+IREE_PYTHON3_EXECUTABLE="${IREE_PYTHON3_EXECUTABLE:-python3}"
+
+IREE_INSTALL_BIN="${IREE_INSTALL_BIN:-${ROOT_DIR}/../iree-build-debug/tools/}"
+IREE_COMPILE_EXE="${IREE_INSTALL_BIN}/iree-compile"
+TEST_RUNNER="${IREE_INSTALL_BIN}/iree-e2e-matmul-test"
```
ScottTodd, line 33: Does the iree-amd-aie CI have these files available somewhere already? I'm wondering how easily this can be plugged into a GitHub Actions workflow. Did you want to set that up in this PR also?

tests/matmul/generate_e2e_matmul_tests.py
ScottTodd, line 1: I just forked this file directly from IREE (https://github.com/openxla/iree/blob/main/tests/e2e/matmul/generate_e2e_matmul_tests.py). Did you make any local changes, or is it a pure fork still?

nirvedhmeshram: Yes, we are currently very limited in shapes we can run, so I removed most of the arbitary shapes we have there. Currently making progress with just 1 small shape to keep things simple but we can run a few more shapes.

ScottTodd: Ah. I'm glad I asked. In that case, can you split this into two commits - one that is a pure fork and one that has the diff from your changes?

build_tools/ci/run_matmul_test.sh
nirvedhmeshram: I think these binary's can be found however, the dependencies are not installed in any CI bots and any CI bots dont have an AIE so I think this is next task for @saienduri

build_tools/ci/build-matmul/matmul_i32_i32_small_amd-aie_xrt_calls.mlir
ScottTodd, line 1: Add the build-matmul folder to .gitignore please (and delete these files from the PR)

https://github.com/nod-ai/iree-amd-aie/blob/main/.gitignore

[IREE has](https://github.com/openxla/iree/blob/main/.gitignore) these patterns that should work:
```
build/
build-*/
```

ScottTodd (approved):

PR #131, opened 2024-02-07
Refactor getPackUnpackNormalizedPerm method to ensure shape flexibility

erwei-xilinx:


Abhishek-Varma (approved):
The PR looks good to me.
Just for my understanding :-
```
Previously: invert(invert(perm)) - perhaps had the "undo" effect?

Current: invert(perm)
```
Was that itself one issue besides the handling of `rank vs perm.size()` ?

erwei-xilinx:
> Was that itself one issue besides the handling of rank vs perm.size() ?

Thanks for looking into this. Current version of `getPackUnpackNormalizedPerm` doesn't do any inversion anymore. It now tries to generate the normalized perm list directly.

Abhishek-Varma:
> > Was that itself one issue besides the handling of rank vs perm.size() ?
> 
> Thanks for looking into this. Current version of `getPackUnpackNormalizedPerm` doesn't do any inversion anymore. It now tries to generate the normalized perm list directly.

Oh okay. Thanks for this! :D

PR #132, opened 2024-02-07
[WIP] Enable `linalg.batch_matmul`

Abhishek-Varma:
This is a WIP PR.

As of now it contains 3 commits - (we should just be reviewing the 2nd one) :-
Commit 1 - @yzhang93 's fix patch for enabling lower sized `linalg.matmul`.

Commit 2 - The initial infra for `linalg.batch_matmul` <---- this changes the current matmul packing config too, so this will be addressed as part of this WIP PR + some comment cleaning up.

Commit 3 - @erwei-xilinx 's fix patch for `--iree-amdaie-decompose-pack-unpack-to-air` 3D -> 2D.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPackAndTranspose.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPackAndTranspose.cpp
@@ -42,16 +42,19 @@ static FailureOr<PackConfig> getPackConfig(
     } else if (passPipeline == AIEPassPipeline::SimplePackPipeline) {
       // Set constraints for pack size [M, N] from first level of tile sizes
       // Currently set pack size k as the input size K to avoid failure.
+      
+      // int64_t tileB = 64;
       int64_t tileM = 64;
       int64_t tileN = 64;
       if (lowerConfig) {
         auto tileSizes = lowerConfig.getTilingLevels()[0].getSizes();
-        tileM = tileSizes[0];
-        tileN = tileSizes[1];
+        // tileB = tileSizes[0];
+        tileM = tileSizes[1];
+        tileN = tileSizes[2];
       }
-      config.packedSizes = {rewriter.getI64IntegerAttr(tileM),
-                            rewriter.getI64IntegerAttr(tileN),
-                            rewriter.getI64IntegerAttr(kSize)};
+      config.packedSizes = {
+          rewriter.getI64IntegerAttr(0), rewriter.getI64IntegerAttr(tileM),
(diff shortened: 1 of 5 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -74,6 +117,8 @@ static LogicalResult setRootConfigImpl(func::FuncOp entryPointFn, Operation *op,
         // TODO (nmeshram): This is very limited for now, plan is to
         // let it first crash for all the other ops and then consiously
         // add support for them, this way we can verify our work.
+        .Case<linalg::ContractionOpInterface>(
+            [&](auto op) { return setRootConfig(entryPointFn, op); })
```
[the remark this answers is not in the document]
Abhishek-Varma: Yes, you're correct - just that I was trying to group both of these under the umbrella of `ContractionOpInterface`.
Haven't shifted linalg.matmul there - I'll address your point though.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPackAndTranspose.cpp
```diff
@@ -61,10 +64,11 @@ static FailureOr<PackConfig> getPackConfig(
     // 0 is used when unpack is empty
     config.unpackEmpty = {0};
     config.innerPerm = {{1, 0}};
-    config.outerPerm = {{0, 1}};
+    config.outerPerm = {{0, 1, 2}};
```
nirvedhmeshram, line 67: I believe this path is run for both matmul and batch matmul? If yes, I am wondering why it is okay add an extra permutation for the matmul. Perhaps the reason is that {0,1,2} is identity or no permutation at all? in which case we can drop this line all together?

nirvedhmeshram (commented):
Thanks for this @Abhishek-Varma , currently what are the errors that we are facing e2e while running this? Can you link the related issues here so that we can track it in once place. it might make sense to start a issue to track all the batch mamtul issue that then link the sub issues there.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPackAndTranspose.cpp
Abhishek-Varma: Oh no no - I am not sure if that claim would be right. So, I'll update this logic based on the `numLoops`/`numDims`/etc.

Currently just added the infra that just targets the batch_matmul - working on cleaning it up - should be complete by tomorrow.

We can currently review of the tile sizes and the packing config I'm using particularly for batch_matmul case.

Issue #96, opened 2024-01-23
OPT dispatch tracker (WIP)


nirvedhmeshram:
| Dispatch Type | Shapes | Running on AIE|
|-----------------|---------|------------------|
|Elementwise     |(8,32,64)|  -[]                    |

MaheshRavishankar:
Nice... the next step then is to take the `matmul_transpose_b` operation and `batchmatmul` operation of that size and run it through the pack based pipeline. @yzhang93 can we start using the c++ pass pipeline for this now?

yzhang93:
> Nice... the next step then is to take the `matmul_transpose_b` operation and `batchmatmul` operation of that size and run it through the pack based pipeline. @yzhang93 can we start using the c++ pass pipeline for this now?

Yes, we can start putting pieces together for pack based pipeline. As for the first step, when @erwei-xilinx 's pack/unpack lowering pass got merged, we can add an e2e example with a smaller input sizes and only parallel loops. After this works, we'll add reduction loops and peeling if needed. I have some additional comments in [this PR](https://github.com/nod-ai/iree-amd-aie/pull/100).

Abhishek-Varma:
CC: @MaheshRavishankar  @nirvedhmeshram @yzhang93 

I worked on creating a utility which would hopefully be helpful for each of us to individually test out various sizes of Matmul (can be later on extended for other individual ops) - [Testing utilities](https://gist.github.com/Abhishek-Varma/82a168ddc2295682017df17dbe595003)
NOTE: You will have to modify the paths based on your working setup.

Following is my observation based on the recent addition of Pack based pipeline without changing any config and using [pack_pipeline_e2e.mlir](https://github.com/nod-ai/iree-amd-aie/blob/main/tests/samples/pack_pipeline_e2e.mlir) as the base input IR structure to test the matmul sizes :-

NOTE: For each failure cases I have attached the logs too. Also, the below size combinations are a superset of the ones found in OPT.

1. Element types:
    Supported: `i32`
    Not supported: 
    - `i64` and `f16`: Failure at `airrt-to-ipu` - [mlir-print-ir-before-all](https://gist.github.com/Abhishek-Varma/1aebeb04ad5e6c71cff38927d03b8b51)
    - `f32` : Failure at `iree-hal-serialize-target-executables` - [mlir-print-ir-before-all](https://gist.github.com/Abhishek-Varma/3d39c3865b53b168a987619d44a64476)
    
[...]

PR #134, opened 2024-02-07
[AMD-AIE] Add initial effort for enabling matmul transpose

Abhishek-Varma:
-- This commit adds initial effort for enabling matmul transpose.
-- Matmul transpose is represented via a linalg.generic op with the
   transpose materializing as the interchange of the dimension. This
   commit therefore adds the initial infra for dealing with linalg.generic
   and caters specifically to the transpose case.

NOTE: The `--split-input-file` doesn't seem to work for some reason - should be a trivial fix - I'll look into it, but rest of the infra is going to be as it is.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
@@ -18,14 +18,14 @@
 namespace mlir::iree_compiler::AMDAIE {
 
 static LogicalResult setRootConfigForPadPipeline(func::FuncOp entryPointFn,
-                                                 linalg::MatmulOp matmulOp) {
+                                                 linalg::LinalgOp linalgOp) {
   SmallVector<int64_t> TileSizeLevel0 = {8, 8};
   SmallVector<int64_t> TileSizeLevel1 = {4, 4};
   SmallVector<int64_t> TileSizeLevel2 = {0, 0, 4};
   TileSizesListType tileSizes = {TileSizeLevel0, TileSizeLevel1,
                                  TileSizeLevel2};
   return setOpConfigAndEntryPointFnTranslation(
-      entryPointFn, matmulOp, tileSizes,
+      entryPointFn, linalgOp, tileSizes,
       IREE::Codegen::DispatchLoweringPassPipeline::None);
 }
 
@@ -63,19 +63,19 @@ static SmallVector<int64_t> getPackedSize(Operation *op) {
 }
 
 static LogicalResult setRootConfigForSimplePackPipeline(
-    func::FuncOp entryPointFn, linalg::MatmulOp matmulOp) {
+    func::FuncOp entryPointFn, linalg::LinalgOp linalgOp) {
   // ------------------------------------------------------
   // -------------- Set lowering config -------------------
   // ------------------------------------------------------
   // Assume working on a 2x2 AIE array and make sure the tile size is not larger
   // than the input size.
-  auto initType = matmulOp.getDpsInitOperand(0)->get().getType();
+  auto initType = linalgOp.getDpsInitOperand(0)->get().getType();
   auto initShape = llvm::cast<ShapedType>(initType).getShape();
   auto tileM0 = std::min((int)initShape[0], 64);
   auto tileN0 = std::min((int)initShape[1], 64);
   auto tileM1 = std::max((int)tileM0 / 2, 1);
   auto tileN1 = std::max((int)tileN0 / 2, 1);
-  auto lhsType = matmulOp.getDpsInputOperand(0)->get().getType();
+  auto lhsType = linalgOp.getDpsInputOperand(0)->get().getType();
   auto lhsShape = llvm::cast<ShapedType>(lhsType).getShape();
   auto tileK = std::min((int)lhsShape[1] / 8, 4);
 
@@ -85,7 +85,7 @@ static LogicalResult setRootConfigForSimplePackPipeline(
   TileSizesListType tileSizes = {TileSizeLevel0, TileSizeLevel1,
                                  TileSizeLevel2};
   if (failed(setOpConfigAndEntryPointFnTranslation(
-          entryPointFn, matmulOp, tileSizes,
+          entryPointFn, linalgOp, tileSizes,
           IREE::Codegen::DispatchLoweringPassPipeline::None))) {
     return failure();
   }
@@ -110,7 +110,7 @@ static LogicalResult setRootConfigForSimplePackPipeline(
       outerPerm);
   // Pack level => 2.
   // packed size for [M, N, K, m, n, k]
-  packedSizes = getPackedSize(matmulOp);
+  packedSizes = getPackedSize(linalgOp);
   // Transpose A matrix from [M K m k m0 k0] to [M K k m m0 k0]
   // Transpose B matrix from [K N k n n0 k0] to [K N n k k0 n0]
   // Transpose C matrix from [M N m n m0 n0] to [M N n m m0 n0]
@@ -129,25 +129,25 @@ static LogicalResult setRootConfigForSimplePackPipeline(
   auto packingConfigLevels =
       PackingConfigPackingLevelsAttr::get(context, packingConfigLevelsVal);
   auto config = PackingConfigAttr::get(context, packingConfigLevels);
-  setPackingConfig(matmulOp, config);
+  setPackingConfig(linalgOp, config);
   return success();
 }
 
 static LogicalResult setRootConfigForPackPipeline(func::FuncOp entryPointFn,
-                                                  linalg::MatmulOp matmulOp,
+                                                  linalg::LinalgOp linalgOp,
                                                   AIEConfig cfg) {
   // ------------------------------------------------------
   // -------------- Set lowering config -------------------
   // ------------------------------------------------------
   if (!(cfg.num_cores == 1 || cfg.num_cores == 2 || cfg.num_cores == 4))
-    return matmulOp.emitOpError("unhandled number of cores");
+    return linalgOp.emitOpError("unhandled number of cores");
   SmallVector<int64_t> TileSizeLevel0 = {16, 64 * cfg.num_cores};
   SmallVector<int64_t> TileSizeLevel1 = {0, 0, 64};
   SmallVector<int64_t> TileSizeLevel2 = {1, 1};
   TileSizesListType tileSizes = {TileSizeLevel0, TileSizeLevel1,
                                  TileSizeLevel2};
   if (failed(setOpConfigAndEntryPointFnTranslation(
-          entryPointFn, matmulOp, tileSizes,
+          entryPointFn, linalgOp, tileSizes,
           IREE::Codegen::DispatchLoweringPassPipeline::None))) {
     return failure();
   }
@@ -169,7 +169,7 @@ static LogicalResult setRootConfigForPackPipeline(func::FuncOp entryPointFn,
       outerPerm);
   // Pack level => 2.
   // packed size for [M, N, K, m, n, k]
-  packedSizes = getPackedSize(matmulOp);
+  packedSizes = getPackedSize(linalgOp);
   // Transpose A matrix from [M K m k m0 k0] to [M K k m m0 k0]
   // Transpose B matrix from [K N k n n0 k0] to [K N n k k0 n0]
   // Transpose C matrix from [M N m n m0 n0] to [M N n m m0 n0]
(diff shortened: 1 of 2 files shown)


tests/samples/simple_pack_pipeline_e2e.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: iree-compile --iree-hal-target-backends=amd-aie --compile-to=executable-sources %s | iree-opt --pass-pipeline="builtin.module(hal.executable(hal.executable.variant(iree-hal-translate-target-executable-variants{target=amd-aie})))" | FileCheck %s --check-prefix=CPP
+// RUN: iree-compile --split-input-file --iree-hal-target-backends=amd-aie --compile-to=executable-sources %s | iree-opt --pass-pipeline="builtin.module(hal.executable(hal.executable.variant(iree-hal-translate-target-executable-variants{target=amd-aie})))" | FileCheck %s --check-prefix=CPP
```
nirvedhmeshram, line 1: I wonder if giving `--split-input-file` to the iree-opt command as well will solve the issue for you, not sure how the piping the output of iree-compile to it interacts with that but pretty sure that ahs something to do with the module having both fucntions

Abhishek-Varma: I did try adding `--split-input-file` after the pipe to `iree-opt` as well. :P

It still behaved the same way. I can give it a fresh try tomorrow - but beside that I'm out of options unless I need to look into the AIETarget/AIELowerExecutable (highly unlikely but that would be the only place to check).

It's strange though since the individual tests pass - just while putting them in the same file it misbehaves.

nirvedhmeshram:
@Abhishek-Varma, I think @MaheshRavishankar mentioned we should verify the IR we are generating to make sure we dont have strided access patterns, could you please paste a gist of the --mlir-print-ir-after-all so that we can go over it, before landing this?

Abhishek-Varma:
> @Abhishek-Varma, I think @MaheshRavishankar mentioned we should verify the IR we are generating to make sure we dont have strided access patterns, could you please paste a gist of the --mlir-print-ir-after-all so that we can go over it, before landing this?

Hi @nirvedhmeshram 

I've added the [mlir-print-ir-after-all](https://gist.github.com/Abhishek-Varma/a461542c351c3aa1713d7cb42f81e873).

Also, as discussed yesterday - I've split the test case into two files.

I noted one thing - this patch requires [Vivian's patch](https://github.com/nod-ai/iree-amd-aie/commit/5f7a73bb76bd3cf783e11a1e2530215610370daf) for the `--iree-amdaie-pack-and-transpose` changes.

@yzhang93 - Can we get that commit merged? Currently that helps in the matmul/matmul_transpose cases. Else if it was a workaround, how are we planning on navigating the same? Trying to understand the missing parts involved.

nirvedhmeshram:
@MaheshRavishankar please confirm this is okay to land, as it makes progress on https://github.com/nod-ai/iree-amd-aie/issues/150

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -15,6 +15,46 @@
 
 namespace mlir::iree_compiler::AMDAIE {
 
+/// Sets the lowering configuration for a generic op implementing a
+/// transposition.
+static LogicalResult setTransposeLikeOpRootConfig(
```
MaheshRavishankar, line 20: This is confusing. I think you need to use something like `ContractionOpInterface` (https://github.com/llvm/llvm-project/blob/25cec33521ba1c67f5e939ceea277be18b2d0fb7/mlir/include/mlir/Dialect/Linalg/IR/LinalgInterfaces.td#L21) and check if the op is a "matmul" of the form you need. You might need to add some checkers similar to what the interface has to check for matmul/batch_matmul, etc.

You dont need to change the interface upstream to start with. You can just start with writing a similar matcher here and set the config using that. You can then graduate the matcher upstream.

MaheshRavishankar (changes requested):

yzhang93:
@Abhishek-Varma I've updated the simple pack pipeline in https://github.com/nod-ai/iree-amd-aie/pull/141, and hopefully it will be merged soon. Then could you rebase your code and run your tests?

tests/samples/simple_pack_pipeline_e2e.mlir
yzhang93: Is this problem solved? `--split-input-file` seems to work fine when I added it in this test https://github.com/nod-ai/iree-amd-aie/blob/main/tests/samples/simple_pack_pipeline_e2e.mlir#L1

Abhishek-Varma: Hi @yzhang93 , yes - it seems to have been resolved. I've removed the separate lit tests.

FranzHaniel: Good to hear this has been resolved! I had a closer look at how to use different test inputs and commands in one file, in case you haven't seen that I am using it in my PR https://github.com/nod-ai/iree-amd-aie/pull/159

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -67,6 +67,95 @@ static LogicalResult setRootConfigForPackPipeline(func::FuncOp entryPointFn,
...
+  auto rhsBlockArg = mulOp.getOperand(1).template dyn_cast<BlockArgument>();
+  auto outBlockArg = addOp.getOperand(0).template dyn_cast<BlockArgument>();
+  if (!lhsBlockArg || !rhsBlockArg || !outBlockArg ||
+      lhsBlockArg.getOwner() != body || rhsBlockArg.getOwner() != body ||
+      outBlockArg.getOwner() != body || lhsBlockArg.getArgNumber() != 0 ||
+      rhsBlockArg.getArgNumber() != 1 || outBlockArg.getArgNumber() != 2) {
+    return false;
+  }
+  return true;
+}
```
nirvedhmeshram, line 86: I appreciate you doing a clever thing to figure this out, but I fear this may be brittle as we scale of the codegen. Generally, at this level it is okay to look at what other backends are doing and lift the code :)
I would do something inspired from the cpu backend here
https://github.com/openxla/iree/blob/main/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp#L1499-L1523
At the same time be careful as CPU might be supporting more transposes then the transpose what you are in this PR

tests/samples/simple_pack_pipeline_e2e.mlir
```diff
@@ -44,3 +44,31 @@ func.func @matmul_large(%lhs: tensor<2048x512xi32>, %rhs: tensor<512x2048xi32>)
...
+//       CPP:      aiex.ipu.dma_memcpy_nd
+//       CPP:      aiex.ipu.dma_memcpy_nd
+//       CPP:      aiex.ipu.dma_memcpy_nd
+//       CPP:      aiex.ipu.sync
+func.func @matmul_transpose_static(%lhs : tensor<8x16xi32>,
+    %rhs : tensor<32x16xi32>) -> tensor<8x32xi32> {
+  %cst = arith.constant 0 : i32
+  %empty = tensor.empty() : tensor<8x32xi32>
+  %fill = linalg.fill ins(%cst : i32) outs(%empty : tensor<8x32xi32>) -> tensor<8x32xi32>
+  %matmul_transpose = linalg.generic {indexing_maps = [affine_map<(d0, d1, d2) -> (d0, d2)>, affine_map<(d0, d1, d2) -> (d1, d2)>, affine_map<(d0, d1, d2) -> (d0, d1)>], iterator_types = ["parallel", "parallel", "reduction"]} ins(%lhs, %rhs : tensor<8x16xi32>, tensor<32x16xi32>) outs(%fill : tensor<8x32xi32>) {
```
nirvedhmeshram, line 67: I think we want to support the named op, matmul.tranpose_b from the OPT dispatch tracker, note that it is still can be interacted as a generic in the kernelDispatch code?

nirvedhmeshram (changes requested):
Had a few more requests. Sorry for taking long to give feedback on this.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
nirvedhmeshram: Actually, on reviewing further, I dont think you want to support a linalg.generic that is doing a tranpose, I thought you wanted to support the named op, linalg.tranpose_b that uses the contractionOp interface?

tests/samples/simple_pack_pipeline_e2e.mlir
Abhishek-Varma: Maybe I'm missing something but I didn't see a `linalg.matmul_transpose_b` op in the OPT dispatch tracker's listed dispatches.

My first hunch was to raise `linalg.generic` -> `linalg.matmul_transpose_b` but I guess that can't be done because the `matmul_transpose_b` apparently has some elemental type constraint (@MaheshRavishankar please correct me if I'm wrong).

And if we go ahead with just `matmul_transpose_b`, the AffineMap checks I added would still be need to differentiate amongst the various ops implementing `ContractionOpInterface`. But that requires either the `linalg.matmul_transpose_b` to be directly present in the dispatch OR raising of `linalg.generic` (still using the matcher check) to `matmul_transpose_b` and then moving on to KernelDispatch's handle.

nirvedhmeshram: oh you are right about them being generic in OPT dispactches, I got confused becuase the liner dispatch Kumar was showing the other day had transpose b named op. Anyway, in that case my only comment is to adapt the transpose logic from the CPU backend then coming up with a new one?

Abhishek-Varma: So the [transpose check](https://github.com/openxla/iree/blob/main/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp#L1499-L1523) in CPU is about a `linalg.generic` op which just performs a transpose. That check is not adaptable for that reason besides it being a `static` function defined just within that file.

Please let me know if there's any such utility in the CPU backend that I should look into and can be reused (copy/paste) in this case.

nirvedhmeshram: Your check is also for generic so I was thinking just copy paste the code in our file?

nirvedhmeshram: ah, I thought it was doing a matmul transpose my bad! so how is the CPU dealing with a generic that does a matmul trasnpose, does it become a named op? Is it becoming a named op for us when we see it in the kernel dispatch code? I am just trying to avoid new logic (for the sake of lowering maintenance burden) so want to handle this how cpu/gpu handle it as at this level its fairly similar other than the actual config attriute we add to it.

nirvedhmeshram: I dont think the elemtent type is used in the raise-special-ops matcher
https://github.com/openxla/iree/blob/main/compiler/src/iree/compiler/GlobalOptimization/RaiseSpecialOps.cpp#L75-L85
Anyway, its good to know this IR is going to default config for CPU, lets discuss in a call today to make progress on this. Thanks for investigating.

nirvedhmeshram: I think that matcher is to match linalg.matmul(a, linalg.transpose(b)), not for raising from a generic, I dont think that raising exists right now.

Abhishek-Varma: Yes, you're correct.
I did check that pass and suggested to add the `linalg.generic` -> `linalg.matmul_transpose_a/b` to upstream IREE earlier, but seems like certain element types aren't supported via `linalg.matmul_transpose_a/b` later down the stack. And maybe that, in a way, explains why CPU pipeline chooses to multiplex this to a default config and doesn't have this structure handled in `raise-special-ops`.

I cannot recall the exact reason though - one way would be to indeed have an IR starting from `linalg.matmul_transpose_b` and check how is that handled in the stack. If it remains as a `linalg.matmul_transpose_b` for the element type the test case is trying to handle - then perhaps it'd indeed make sense to use the matcher I added here to raise `linalg.generic` to `linalg.matmul_transpose_a/b`. 

Depending on what we converge on I can work on either of the following :-
1. Update `raise-special-ops`.
2. Keep the current implementation as it is.
3. A default config which would work on all non-transpose and non-element-wise should be thought of and added. I'll have to check what CPU is doing and understand that - but seems like this would open other can of worms because the current pipeline is fixated on certain loop tiling structure.

nirvedhmeshram: It is worth checking what the CPU does for the tranpose_b named op, I think it wont apply the default config in that case, as it will treat it like a contracton op?
https://github.com/openxla/iree/blob/main/compiler/src/iree/compiler/Codegen/LLVMCPU/KernelDispatch.cpp#L1175

what is your reference for element type is the reason for it not getting raised? I cant see that anywhere.

Also, I am fine with landing this if its not getting raised, however, one thing to handle is what if the dispatch did actually present as mamtul.transpose_b, I saw that in a simple linear layer example Kumar had, and could happen if changes happen in the front end, so dont want to break in that case. However, that can be a separate PR. Let me know what you think.

Abhishek-Varma: > what is your reference for element type is the reason for it not getting raised? I cant see that anywhere.

I was citing that based on an earlier discussion I've had with @MaheshRavishankar when I was talking about the `linalg.generic` -> `linalg.matmul_transpose_a/b` - which I've verified today (see below).

> one thing to handle is what if the dispatch did actually present as mamtul.transpose_b

Yes, absolutely. I did try that in an IR that itself starts with `matmul_transpose_b` to check how CPU is handling that - hopefully now it gives us a complete overview of what's happening at KernelDispatch level - it indeed goes through ContractionOp's case. I see that there's no issue in the lowering with the element type `i32` and I verified the results during the execution of the generated vmfb too. `linalg.matmul_transpose_a/b` goes up until Vectorization phase, post which it vanishes because of vectorization for `i32`'s case.

For `f32` it gets converted to a `linalg.mmt4d` op. Results were verified here too.

You're correct about the front-end aspect - But adding support for `linalg.matmul_transpose_a/b` too right now wouldn't be too fruitful since the required element type's handling is `f32` - and it is indeed not supported by both `linalg.matmul_transpose_a/b` as well as the current AMD-AIE `linalg.matmul` itself is unable to handle `f32`.

So, we've got a lot of iterations to handle these incrementally.

nirvedhmeshram: About the last point, I think the element type support and shape support are orthogonal, but so is supporting this type of generic matmul traponse and matmul.tranpose_b op, so its fine to do this and add the tranpose_b later. However, I am not sure if Mahesh's earlier comment about using the contraction op interface is actually being done here, so I would check with him on that.

nirvedhmeshram (approved):
Just unblocking from my side.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
MaheshRavishankar: I think there is a confusion here. THis is checking for a `linalg.generic` that effectively implements `linalg.matmul_transpose_b`. From the input we might not necessarily get a named op. We might get a `generic` op which is implementing the `linalg.matmul_tranpose_b`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -159,6 +159,95 @@ static LogicalResult setRootConfigForPackPipeline(func::FuncOp entryPointFn,
...
+  // Extract dimensions for MxK * NxK -> MxN
+  AffineExpr m = map2.getResult(0);
+  AffineExpr n = map2.getResult(1);
+  AffineExpr k = map0.getResult(1);
+  auto *context = indexingMaps.getContext();
+  auto mapA = AffineMapAttr::get(AffineMap::get(3, 0, {m, k}, context));
+  auto mapB = AffineMapAttr::get(AffineMap::get(3, 0, {n, k}, context));
+  auto mapC = AffineMapAttr::get(AffineMap::get(3, 0, {m, n}, context));
+  auto maps = ArrayAttr::get(context, {mapA, mapB, mapC});
+  return indexingMaps == maps;
```
MaheshRavishankar, line 223: Need to check the iterator types as well.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -159,6 +159,95 @@ static LogicalResult setRootConfigForPackPipeline(func::FuncOp entryPointFn,
...
+      outBlockArg.getOwner() != body || lhsBlockArg.getArgNumber() != 0 ||
+      rhsBlockArg.getArgNumber() != 1 || outBlockArg.getArgNumber() != 2) {
+    return false;
+  }
+  return true;
+}
+
+/// `isMatmulTranspose` is a utility function that aims to indentify whether a
+/// linalg.generic op is a matmul transpose op.
+static bool isMatmulTranspose(linalg::GenericOp genericOp) {
```
MaheshRavishankar, line 182: Instead of using a `linalg.generic` here, you can use a `LinalgOp` directly. Then whether you have a named op or a `linalg.generic` that implements the same operation, it will use the same matcher.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -199,6 +288,9 @@ static LogicalResult setRootConfigImpl(func::FuncOp entryPointFn, Operation *op,
         // TODO (nmeshram): This is very limited for now, plan is to
         // let it first crash for all the other ops and then consiously
         // add support for them, this way we can verify our work.
+        .Case<linalg::GenericOp>([&](auto op) {
```
MaheshRavishankar, line 291: The switch cases need to be ordered from most specific to most general. If we move this `case` to a `LinalgOp` then this needs to go after the `case` for `linalg.matmul`.

MaheshRavishankar (changes requested):

MaheshRavishankar (approved):
Looks good! Thanks!

MaheshRavishankar:
Could we add some e2e tests for matmul_transpose_b ?

Abhishek-Varma:
As reported earlier [here](https://github.com/nod-ai/iree-amd-aie/pull/134#issuecomment-1933349137) as well as the [issue tracker](https://github.com/nod-ai/iree-amd-aie/issues/96#issuecomment-1931531397) - "matmul transpose" worked with the exact config of "matmul" - and therefore was able to yield similar results for the shape support.

It seems to be not be the case anymore albeit the lit tests pass - the xclbin is not getting generated anymore.

This is the IR which previously worked, but isn't anymore :-
```
func.func @matmul_transpose_static(%lhs : tensor<8x16xi32>,
    %rhs : tensor<32x16xi32>) -> tensor<8x32xi32> {
  %cst = arith.constant 0 : i32
  %empty = tensor.empty() : tensor<8x32xi32>
  %fill = linalg.fill ins(%cst : i32) outs(%empty : tensor<8x32xi32>) -> tensor<8x32xi32>
  %matmul_transpose = linalg.generic {indexing_maps = [affine_map<(d0, d1, d2) -> (d0, d2)>, affine_map<(d0, d1, d2) -> (d1, d2)>, affine_map<(d0, d1, d2) -> (d0, d1)>], iterator_types = ["parallel", "parallel", "reduction"]} ins(%lhs, %rhs : tensor<8x16xi32>, tensor<32x16xi32>) outs(%fill : tensor<8x32xi32>) {
  ^bb0(%in: i32, %in_0: i32, %out: i32):
    %19 = arith.muli %in, %in_0 : i32
    %20 = arith.addi %out, %19 : i32
    linalg.yield %20 : i32
  } -> tensor<8x32xi32>
  return %matmul_transpose : tensor<8x32xi32>
}
```

I tried reverting to exact config in [Vivian's patch](https://github.com/nod-ai/iree-amd-aie/commit/5f7a73bb76bd3cf783e11a1e2530215610370daf) and commenting out Passes.cpp related changes of [PR 141](https://github.com/nod-ai/iree-amd-aie/pull/141) but to no avail anymore - here is the [mlir-print-ir-after-all](https://gist.github.com/Abhishek-Varma/eaa370bed3e8a9d1d09ac4277d070adc) which I tried comparing with the [mlir-print-ir-after-all which worked](https://gist.githubusercontent.com/Abhishek-Varma/a461542c351c3aa1713d7cb42f81e873/raw/4e13b60ab9c24d0352a1911ea5a601610fc1daf0/matmul_transpose_log.log). Up until C++ simple-pack pipeline, it looked the same to me.

And here is the [mlir-print-ir-after-all](https://gist.github.com/Abhishek-Varma/024ec458dbf2a83b0b95208dfc24294f) which we get with this patch without reverting any changes. The same can be obtained from the CI's logs (I've updated the infra to include "matmul transpose" variants in e2e) OR by simply running this PR. The very immediate difference between this and the one which worked is the tiling config and the scf.forall's IV - therefore I instead tried to see what happens if I'm keeping the tiling/packing config same (as mentioned above).

@yzhang93 can you please take a look into this issue - perhaps earlier the same configuration worked but now we need to change it based on the matmul variant?

Abhishek-Varma:
I tried looking into the logs and it seemed like `ip_data::m_name` (the function name) has a size constraint.

When I made the function string name smaller - it worked!

You may take a look at the [updated e2e IR log for linalg.generic](https://gist.github.com/Abhishek-Varma/bf167732b2a6c03a67e368f449bcade9)

I tried doing the same on `linalg.matmul_transpose_b` - it didn't work (previously the lit test was passing and now even that isn't!) and led to :-
```
error: 'aie.connect' op ; connecting (N: 3) to (DMA: 0) on TileID(0, 1) targets same dst as another connect op; existing\
destinations: (S: 0), (DMA: 2), (N: 2), (N: 4), (N: 1), (DMA: 4), (DMA: 3), (N: 0), (DMA: 1), (N: 5), (DMA: 5), (N: 3), (DMA: 0)
```

I can dig deeper on this later to fetch a repro if this is indeed an mlir-air/mlir-aie bug - although linalg_ext/pack/unpack pass seems to be the culprit while skimming through.

For now, I would suggest getting this in since the intention of this PR was to lay down the initial base for handling matmul transpose - @nirvedhmeshram @MaheshRavishankar

nirvedhmeshram:
> I tried looking into the logs and it seemed like `ip_data::m_name` (the function name) has a size constraint.
> 
> When I made the function string name smaller - it worked!
> 
> You may take a look at the [updated e2e IR log for linalg.generic](https://gist.github.com/Abhishek-Varma/bf167732b2a6c03a67e368f449bcade9)
> 
> I tried doing the same on `linalg.matmul_transpose_b` - it didn't work (previously the lit test was passing and now even that isn't!) and led to :-
> 
> ```
> error: 'aie.connect' op ; connecting (N: 3) to (DMA: 0) on TileID(0, 1) targets same dst as another connect op; existing\
> destinations: (S: 0), (DMA: 2), (N: 2), (N: 4), (N: 1), (DMA: 4), (DMA: 3), (N: 0), (DMA: 1), (N: 5), (DMA: 5), (N: 3), (DMA: 0)
> ```
> 
> I can dig deeper on this later to fetch a repro if this is indeed an mlir-air/mlir-aie bug - although linalg_ext/pack/unpack pass seems to be the culprit while skimming through.
> 
> For now, I would suggest getting this in since the intention of this PR was to lay down the initial base for handling matmul transpose - @nirvedhmeshram @MaheshRavishankar
Oh yes this is a know problem. 
I agree we should get this in, however, please remove the CI side changes done here, it is wrong to use the generate script for anything other than matmul (at least for now) as that will never work with the existing infra since the reference implementation that the correctness is compared against here is for matmul only, https://github.com/openxla/iree/blob/main/tools/iree-e2e-matmul-test.cc#L202-L216

Abhishek-Varma:
Hi @nirvedhmeshram I've removed the CI related commit - let me know if this looks okay to be merged now.

nirvedhmeshram (approved):
LGTM!

Abhishek-Varma:
> I tried looking into the logs and it seemed like `ip_data::m_name` (the function name) has a size constraint.
> 
> When I made the function string name smaller - it worked!
> 
> You may take a look at the [updated e2e IR log for linalg.generic](https://gist.github.com/Abhishek-Varma/bf167732b2a6c03a67e368f449bcade9)
> 
> I tried doing the same on `linalg.matmul_transpose_b` - it didn't work (previously the lit test was passing and now even that isn't!) and led to :-
> 
> ```
> error: 'aie.connect' op ; connecting (N: 3) to (DMA: 0) on TileID(0, 1) targets same dst as another connect op; existing\
> destinations: (S: 0), (DMA: 2), (N: 2), (N: 4), (N: 1), (DMA: 4), (DMA: 3), (N: 0), (DMA: 1), (N: 5), (DMA: 5), (N: 3), (DMA: 0)
> ```
> 
> I can dig deeper on this later to fetch a repro if this is indeed an mlir-air/mlir-aie bug - although linalg_ext/pack/unpack pass seems to be the culprit while skimming through.
> 
> For now, I would suggest getting this in since the intention of this PR was to lay down the initial base for handling matmul transpose - @nirvedhmeshram @MaheshRavishankar

This was a red-herring due to wrong SSA use in the input IR itself.

I've updated the lit test : https://github.com/nod-ai/iree-amd-aie/pull/189

PR #136, opened 2024-02-08
Add pass to fuse pack operations into for loop

yzhang93:


Abhishek-Varma:
This looks okay to me.
I was trying to think about how to have lesser footprint - since we currently have two cases `(fill, forall)` and `(pack, for)` - I was thinking if we can instead add an option within the previous similar pass of this nature to make the logic based on a bool `useSCFFor` and accordingly we rename the pass as `--iree-amdaie-specific-fuse` (I'm clearly not good with names. :P)

Not adding a blocker on this though - just being vocal about another approach that (might?) ensure seemingly less footprint.

yzhang93:
> This looks okay to me. I was trying to think about how to have lesser footprint - since we currently have two cases `(fill, forall)` and `(pack, for)` - I was thinking if we can instead add an option within the previous similar pass of this nature to make the logic based on a bool `useSCFFor` and accordingly we rename the pass as `--iree-amdaie-specific-fuse` (I'm clearly not good with names. :P)
> 
> Not adding a blocker on this though - just being vocal about another approach that (might?) ensure seemingly less footprint.

Thanks for your suggestion and I think it's reasonable. Currently I would keep these as two separate passes, because the implementation details are different and we want particular combinations of the producer and the loops. Even though we can add them to the same pass, we'll need to have separate implementations.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoForLoop.cpp
```diff
@@ -0,0 +1,101 @@
...
+void AMDAIEFusePackIntoForLoopPass::runOnOperation() {
+  MLIRContext *context = &getContext();
+  func::FuncOp funcOp = getOperation();
+  IRRewriter rewriter(context);
+
+  // Walk through the graph in post order and find the pack op.
+  TilingInterface packOp;
+  funcOp->walk<WalkOrder::PostOrder, ReverseIterator>([&](TilingInterface op) {
+    if (isa<linalg::FillOp>(op)) {
+      packOp = op;
```
nirvedhmeshram, line 40: This is looking for a fill op not a pack op?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoForLoop.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+  LoopLikeOpInterface loops = cast<LoopLikeOpInterface>(forOp.getOperation());
+
+  // Materialize each slice of the producer in place.
+  for (auto sliceOp : sliceOps) {
+    std::optional<scf::SCFFuseProducerOfSliceResult> fusedProducer =
+        scf::tileAndFuseProducerOfSlice(rewriter, sliceOp,
+                                        MutableArrayRef(&loops, 1));
+    if (!fusedProducer) {
+      funcOp->emitOpError("Failed to fuse pack ops into for loop.");
```
nirvedhmeshram, line 90: This is misleading, the producer of slice is not necessarily a pack op, we are collecting all the sliceops in the loop and they can come from any kind of producer, and we are trying to fuse all of them here, not just pack ops

nirvedhmeshram (changes requested):
Please change the description to clarify that this pass fuses all the producers in the for loop and not just the pack ops.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoForLoop.cpp
yzhang93: Thanks for catching this bug. This should be a tensor::PackOp.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoForLoop.cpp
```diff
@@ -0,0 +1,101 @@
...
+    }
+    return WalkResult::advance();
+  });
+  if (!packOp) {
+    LLVM_DEBUG(llvm::dbgs() << "There is no pack op to be fused.\n");
+    return;
+  }
+
+  // Search the first use by a scf::ForOp user.
+  scf::ForOp forOp;
```
MaheshRavishankar, line 51: THis is kind of backwards. If you want to fuse the producers of the pack inside the loop, I am assuming you are only looking at lhs and rhs operands of the compute operation. 

What you should do instead is look for the compute op, if its operands are slices and the source of the slices are pack, fuse them in... 

Though, i have been thinking about this. A cleaner way to do this would be to just do the packing within the scf.for op and then just hoist the pack for the output operands out... conversely you should "hoist" the unpack operation out as well (but it goes after the loop). I havent fully worked this out though. I need some time to play with the transform dialect script you shared :P

MaheshRavishankar (commented):
I have some comments, but this pass is inherently strange. This is a means to and end and is not a long term thing... Lets evaluate benefit before landing it.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoForLoop.cpp
yzhang93: Thanks for taking a look into this. I'll push up another PR that using this pass to get to what AIR and AIE hardware requires. We can then evaluate the benefit of this pass. At the same time, I modified the logic a little bit to match what you described.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoForLoop.cpp
yzhang93: Thanks for pointing this out. I modified the logic a little bit, but still constrains the producer as a tensor::PackOp to avoid undesired fusion...

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoForLoop.cpp
```diff
@@ -0,0 +1,95 @@
...
+    return;
+  }
+
+  // Search the compute op and its producer slices within the For loop.
+  BlockArgument bbArg = forOp.getRegionIterArg(0);
+  SmallVector<tensor::ExtractSliceOp> sliceOps;
+  for (auto user : bbArg.getUsers()) {
+    if (auto genericOp = dyn_cast<linalg::GenericOp>(user)) {
+      for (auto [index, operand] : llvm::enumerate(genericOp.getOperands())) {
+        if (!isa<BlockArgument>(operand)) {
```
MaheshRavishankar, line 57: You can just do 
```
for () {
  auto sliceOp = dyn_cast_or_null<tensor::ExtractSliceOp>(operand.getDefiningOp());
  if (sliceOp && isa_and_non_null<tensor::PackOp>(sliceOp.getSource())) {
     sliceOps.push_back(sliceOp);
  }
}
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoForLoop.cpp
```diff
@@ -0,0 +1,95 @@
...
+  if (sliceOps.empty()) {
+    funcOp->emitOpError("There is no extract tensor slice.");
+    return signalPassFailure();
+  }
+
+  // Materialize each slice of the producer in place.
+  LoopLikeOpInterface loops = cast<LoopLikeOpInterface>(forOp.getOperation());
+  for (auto sliceOp : sliceOps) {
+    auto defOp = sliceOp.getOperand(0).getDefiningOp();
+    if (!isa<tensor::PackOp>(defOp)) {
```
MaheshRavishankar, line 76: You can drop this check completely with the above check.

MaheshRavishankar (approved):
Ok, left a comment which should make this easier. After that this looks good. APproving to unblock landing this.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoForLoop.cpp
yzhang93: Thanks. Done.

nirvedhmeshram (approved):

PR #139, opened 2024-02-09
Remove AIE(X) Passes

spaceotter:
These are used only by the external tool now, and can be removed from the plugin build.


MaheshRavishankar (approved):

MaheshRavishankar:
Nice!

jgmelber:
What do you mean by the "external tool"

PR #155, opened 2024-02-11
[AMD-AIE] Add AMDAIEDialect/Attributes + PackingConfig + Update KernelDispatch/pack-and-transpose

Abhishek-Varma:
This PR involves :-

1. Creating AMDAIEDialect.
2. Creating AMDAIEEnum/Attributes.
3. Creating PackingConfig Attribute infra.
4. Adapting KernelDispatch to set packing_config attribute.
5. Adapting --iree-amdaie-pack-and-transpose to work with the packing_config attribute set earlier.
5. Refactoring the directory structure to make maintaining codebase "easier" and aligns with the restructuring upstream.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/CMakeLists.txt
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/CMakeLists.txt
@@ -15,6 +15,7 @@ iree_cc_library(
   DEPS
     iree::compiler::Dialect::HAL::Target
     iree::compiler::PluginAPI
+    iree::target::amd-aie::IR::AMDAIEDialect
     iree::target::amd-aie::Target
     iree::target::amd-aie::Transforms
     iree::target::amd-aie::aie::AIEDialectIR
@@ -34,5 +35,6 @@ iree_compiler_register_plugin(
     ::registration
 )
 
+add_subdirectory(IR)
 add_subdirectory(Target)
 add_subdirectory(Transforms)
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.cpp
@@ -1,0 +1,126 @@
+// Copyright 2024 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree-amd-aie/IR/AMDAIEAttrs.h"
+
+#include "iree-amd-aie/IR/AMDAIEDialect.h"
+#include "llvm/ADT/TypeSwitch.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/Transform/IR/TransformOps.h"
+#include "mlir/IR/DialectImplementation.h"
+
+#define GET_ATTRDEF_CLASSES
+#include "iree-amd-aie/IR/AMDAIEAttrs.cpp.inc"
+#include "iree-amd-aie/IR/AMDAIEEnums.cpp.inc"
+
+static const char kPackingConfigAttrName[] = "packing_config";
+
+namespace mlir::iree_compiler {
+
+/// Returns an `ArrayAttr` where each element is an `IntegerAttr` of 64-bit
+/// integer type whose values is obtained from `values`.
+static ArrayAttr getIndexArrayAttr(MLIRContext *context,
+                                   ArrayRef<int64_t> values) {
+  return ArrayAttr::get(
+      context, llvm::map_to_vector(values, [&](int64_t value) -> Attribute {
+        return IntegerAttr::get(IndexType::get(context), APInt(64, value));
+      }));
+}
+
+}  // namespace mlir::iree_compiler
+
+namespace mlir::iree_compiler::AMDAIE {
+
+//===----------------------------------------------------------------------===//
+// amdaie.packing_config_level
+//===----------------------------------------------------------------------===//
+
+SmallVector<ArrayRef<int64_t>>
+PackingConfigPackingLevelAttr::getInnerPermArr() {
+  SmallVector<ArrayRef<int64_t>> res;
+  PermLevelsAttr permLevelsAttr = getInnerPerm();
+  for (auto permLevel : permLevelsAttr) {
+    res.push_back(permLevel.getPerm());
+  }
+  return res;
+}
+
+SmallVector<ArrayRef<int64_t>>
+PackingConfigPackingLevelAttr::getOuterPermArr() {
+  SmallVector<ArrayRef<int64_t>> res;
+  PermLevelsAttr permLevelsAttr = getOuterPerm();
+  for (auto permLevel : permLevelsAttr) {
+    res.push_back(permLevel.getPerm());
+  }
+  return res;
+}
+
+//===----------------------------------------------------------------------===//
+// amdaie.packing_config
+//===----------------------------------------------------------------------===//
+
+PackingConfigPackingLevelAttr PackingConfigAttr::getPackingConfigVals(
+    unsigned level) {
+  auto levels = getPackingLevels();
+  if (level >= levels.size()) return {};
+  return levels[level];
+}
+
+void AMDAIEDialect::initializeAMDAIEAttrs() {
+  addAttributes<
+#define GET_ATTRDEF_LIST
+#include "iree-amd-aie/IR/AMDAIEAttrs.cpp.inc"  // IWYU pragma: keeep
+      >();
+}
+
+}  // namespace mlir::iree_compiler::AMDAIE
+
+namespace mlir::iree_compiler {
+
+//===----------------------------------------------------------------------===//
+// Helpers for forming `amdaie.packing_config_level` attribute.
+// ===----------------------------------------------------------------------===//
+
+static AMDAIE::PermLevelAttr getPermLevelAttr(
+    MLIRContext *context, SmallVector<int64_t> permLevelVal) {
+  return AMDAIE::PermLevelAttr::get(context, permLevelVal);
+}
+
+static AMDAIE::PermLevelsAttr getPermLevelsAttr(
+    MLIRContext *context, SmallVector<SmallVector<int64_t>> permLevelsVal) {
+  SmallVector<AMDAIE::PermLevelAttr> permLevels;
+  for (auto permLevel : permLevelsVal) {
+    permLevels.push_back(AMDAIE::PermLevelAttr::get(context, permLevel));
+  }
+  return AMDAIE::PermLevelsAttr::get(context, permLevels);
+}
+
+AMDAIE::PackingConfigPackingLevelAttr getPackingConfigPackingLevelAttr(
+    MLIRContext *context, SmallVector<int64_t> &packedSizes,
+    SmallVector<int64_t> &transposePackIndices, SmallVector<bool> &unpackEmpty,
+    SmallVector<SmallVector<int64_t>> &innerPermVal,
+    SmallVector<SmallVector<int64_t>> &outerPermVal) {
+  auto innerPermAttr = getPermLevelsAttr(context, innerPermVal);
+  auto outerPermAttr = getPermLevelsAttr(context, outerPermVal);
+  return AMDAIE::PackingConfigPackingLevelAttr::get(
+      context, packedSizes, transposePackIndices, unpackEmpty, innerPermAttr,
+      outerPermAttr);
+}
+
+//===----------------------------------------------------------------------===//
+// Helpers for getting/setting `amdaie.packing_config` attribute.
+// ===----------------------------------------------------------------------===//
+
+AMDAIE::PackingConfigAttr getPackingConfig(Operation *op) {
+  return op->getAttrOfType<AMDAIE::PackingConfigAttr>(kPackingConfigAttrName);
+}
+
+void setPackingConfig(Operation *op, AMDAIE::PackingConfigAttr config) {
+  op->setAttr(kPackingConfigAttrName, config);
+}
+
+}  // namespace mlir::iree_compiler
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.h
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.h
@@ -1,0 +1,43 @@
+// Copyright 2024 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+//===- IREECodegenAttrs.h - Codegen dialect attributes --------------------===//
+//===----------------------------------------------------------------------===//
+
+#ifndef IREE_COMPILER_AMDAIE_DIALECT_PACKINGCONFIG_H_
+#define IREE_COMPILER_AMDAIE_DIALECT_PACKINGCONFIG_H_
+
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/IR/BuiltinTypes.h"
+
+// clang-format off
+#include "iree-amd-aie/IR/AMDAIEEnums.h.inc"
+#define GET_ATTRDEF_CLASSES
+#include "iree-amd-aie/IR/AMDAIEAttrs.h.inc"
+// clang-format on
+
+namespace mlir::iree_compiler {
+
+/// Helps in forming a `PackingConfigPackingLevelAttr`.
+AMDAIE::PackingConfigPackingLevelAttr getPackingConfigPackingLevelAttr(
+    MLIRContext *context, SmallVector<int64_t> &packedSizes,
+    SmallVector<int64_t> &transposePackIndices, SmallVector<bool> &unpackEmpty,
+    SmallVector<SmallVector<int64_t>> &innerPermVal,
+    SmallVector<SmallVector<int64_t>> &outerPermVal);
+
+/// Returns the packing configuration set for an operation. Returns `nullptr`
+/// if no value is set.  It expects that the attribute is stored using the
+/// identifier `packing_config`.
+AMDAIE::PackingConfigAttr getPackingConfig(Operation *op);
+
+/// Sets the packing configuration, overwriting existing attribute values.
+void setPackingConfig(Operation *op, AMDAIE::PackingConfigAttr config);
+
+}  // namespace mlir::iree_compiler
+
+#endif  // IREE_COMPILER_AMDAIE_DIALECT_PACKINGCONFIG_H_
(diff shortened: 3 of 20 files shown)


Abhishek-Varma:
I see all the lit tests passing locally. I even tried executing each of them individually.

Can someone please let me know why the following are failing in CI? (Especially the simple-pack one because I've mostly been testing on that when I was developing this PR from scratch)
```
pack_pipeline_funcIR.mlir.test (Failed)
pack_pipeline_funcIR_2core.mlir.test (Failed)
pack_pipeline_funcIR_4core.mlir.test (Failed)
simple_pack_pipeline_e2e.mlir.test (Failed)
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.h
```diff
@@ -0,0 +1,66 @@
...
+  ArrayRef<int64_t> unpackEmpty;
+  // Attributes for inner dimension permutation
+  ArrayRef<ArrayRef<int64_t>> innerPerm;
+  // Attributes for outer dimension permutation
+  ArrayRef<ArrayRef<int64_t>> outerPerm;
+};
+/// Typedef for packing config to use at different levels of packing.
+using PackingConfigListType = SmallVector<PackConfigTD>;
+using PackingConfigListTypeRef = ArrayRef<PackConfigTD>;
+struct PackConfig {
```
Abhishek-Varma, line 34: PackConfig vs PackConfigTD

This is a stopgap solution.

I wanted to make `packedSizes` as an Array of `I64Attr` -> but for some reason `I64Attr` remained unknown (I tried plausible `.h` and `.td` inclusions but to no avail).

I then tried `IntegerAttr` instead of `I64Attr` in the `.td` - it didn't lead to the previous error but the following became an issue :-

OpFoldResult is NOT Attribute.

I felt spent trying to fight this (yesterday and today) and my first intention was to at least make it work.
Hence I added this as a stopgap solution.

Let me know if you have any pointers regarding the same.
One of the major issues that I faced was the nuances involving : ArrayRef <-> SmallVector (and with the multi-dimensional vectors/arrays it piled up the issues).

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.cpp
```diff
@@ -0,0 +1,140 @@
...
+PackConfig PackingConfigAttr::getPackingConfigVals(unsigned level) {
+  Builder builder(getContext());
+  auto levels = getPackingLevels();
+  if (level >= levels.size()) return {};
+  PackConfig packConfig;
+  SmallVector<OpFoldResult> packedSizes;
+  for (int64_t packedSize : levels[level].getPackedSizes()) {
+    packedSizes.push_back(builder.getI64IntegerAttr(packedSize));
+  }
+  packConfig.packedSizes = packedSizes;
```
Abhishek-Varma, line 78: This is the consequence of the stopgap solution that I mentioned above.

My vote for keeping `PackingConfigTD` will be for the "cleanest" representation it provides without involving any of `i32`/`i64`/`tensor<axbx...xi32>`.

My vote for keeping `PackingConfig` will be because it seamlessly integrates with the `iree-amdaie-pack-and-transpose`.

A solution that I propose :-
API_1. Have one API which would just extract `packedSizes` and return the `SmallVector<OpFoldResult>`.
API_2: Have one API which would extract level based config in `ArrayRef` representation.

Adapt `iree-amdaie-pack-and-transpose` to use API_1 for packing and API_2 for transpose (ignoring the `packedSizes`  in case of the latter since it anyway doesn't seem to be using that during transpose).

Let me know your thoughts.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/pack_and_transpose_level1.mlir
```diff
@@ -2,6 +2,8 @@
 
 // CHECK: #config
 #config = #iree_codegen.lowering_config<tile_sizes = [[8, 8], [4, 4], [0, 0, 4]]>
+// TODO(avarma) - I would like a similar global hoisting of packing_config. Help?
```
Abhishek-Varma, line 5: Can someone please provide me pointers for this?

I tried looking both into `lowering_config` and `affine_map` but am unable to find the infra that gets me the following :-
```
packingConfig = #amdaie.packing_config......
func.func @myfunc(...) ->() {
.
.
%0 = %myOp {packing_config = #packingConfig}
.
.
}
```
I feel it's good to have the above construct.

Abhishek-Varma (commented):
Pre-fetching some of the major review comments - let me know your thoughts.

Also, with regards to the failing lit tests :-
1. I fired the IREE build.
2. deinit-ed submodules of iree-amd-aie.
3. init-ed submodules of iree-amd-aie again.
4. Re-built IREE.

I confirm I see ALL the lit tests pass - please help me find why there's difference in the local vs CI behavior? Am I missing updating something?

NOTE: Before pulling in the last two commits (3d60e43b28e4ece3a08ee2942da70ccf7ff0acd6 and 53abdcceaaf5e2d8671c2ec624030307634692bb) the CI and the local lit test for this PR was green but now there's a difference. (Just trying to be vocal about the delta to help shed some light into the apparent issue)

nirvedhmeshram:
> I see all the lit tests passing locally. I even tried executing each of them individually.
> 
> Can someone please let me know why the following are failing in CI? (Especially the simple-pack one because I've mostly been testing on that when I was developing this PR from scratch)
> 
> ```
> pack_pipeline_funcIR.mlir.test (Failed)
> pack_pipeline_funcIR_2core.mlir.test (Failed)
> pack_pipeline_funcIR_4core.mlir.test (Failed)
> simple_pack_pipeline_e2e.mlir.test (Failed)
> ```

The tests are passing locally for me as well on your branch, I tried a dymmy commit to test the CI and that passed, so I think the way forward is to add `-v` here
https://github.com/nod-ai/iree-amd-aie/blob/main/build_tools/ci/build_test_cpp.sh#L90
and see why exactly theses tests fail in CI

Abhishek-Varma:
Thanks @nirvedhmeshram for confirming! This is indeed weird. :/
I've added a dummy commit as per your suggestion to check why CI is failing.

I was typing the below since the CI lit test failures, in a way, negates the proof-of-concept for this PR :-

Attaching the log for all the lit tests which passes locally -> [here](https://gist.github.com/Abhishek-Varma/b964e6705d0eb33430bd7e156f426372)

We can observe that `packing_config` is getting attached and the lowering is indeed successful - I even tried looking into the `CHECK` statements but couldn't find the reason behind the difference in the claims of CI vs local run.

nirvedhmeshram:
> Thanks @nirvedhmeshram for confirming! This is indeed weird. :/ I've added a dummy commit as per your suggestion to check why CI is failing.
> 
> I was typing the below since the CI lit test failures, in a way, negates the proof-of-concept for this PR :-
> 
> Attaching the log for all the lit tests which passes locally -> [here](https://gist.github.com/Abhishek-Varma/b964e6705d0eb33430bd7e156f426372)
> 
> We can observe that `packing_config` is getting attached and the lowering is indeed successful - I even tried looking into the `CHECK` statements but couldn't find the reason behind the difference in the claims of CI vs local run.

my bad I think the flag you want to add is `--output-on-failure` instead of `-v`

nirvedhmeshram:
I was able to reproduce the failures, locally with this cmake command to build it, I think our local debug builds were hiding the issue
https://github.com/nod-ai/iree-amd-aie/blob/main/build_tools/ci/build_test_cpp.sh#L58-L76

compiler/plugins/target/AMD-AIE/iree-amd-aie/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ iree_cc_library(
   DEPS
     iree::compiler::Dialect::HAL::Target
     iree::compiler::PluginAPI
+    iree::target::amd-aie::IR::AMDAIEDialect
```
stephenneuendorffer, line 18: It's unclear to me what the goal of this dialect is... and I see a huge opportunity for confusion with the existing dialects.  Can you elaborate?

stephenneuendorffer: Another naming option might be "XDNA"?

Abhishek-Varma: Hi @stephenneuendorffer .

Like `lowering_config` is attached to `iree_codegen` dialect, the goal was to do the same with the new `packing_config` attribute. Since we're building this as a plugin, I cannot directly use `iree_codegen` dialect.

And since `packing_config` is only used by amd-aie, it made sense to have it attached to a dialect specific to this stage (which can be renamed to `XDNA` as you suggested).

stephenneuendorffer: Thanks, I understand :)

stephenneuendorffer: Is it worth a comment that this should eventually be merge into `iree_codegen`?

Abhishek-Varma: That's a fair ask - although I'm not sure if we would indeed want `packing_config` as part of `iree_codegen` dialect eventually since no other backend seems to be using a pack based lowering.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.td
```diff
@@ -0,0 +1,98 @@
...
+
+#ifndef IREE_COMPILER_AMDAIE_DIALECT_IREEAMDAIEATTRS
+#define IREE_COMPILER_AMDAIE_DIALECT_IREEAMDAIEATTRS
+
+include "iree-amd-aie/IR/AMDAIEDialect.td"
+include "mlir/IR/AttrTypeBase.td"
+include "mlir/IR/EnumAttr.td"
+
+def AMDAIE_MemSpace_L1 : I32EnumAttrCase<"L1", 1, "l1">;
+def AMDAIE_MemSpace_L2 : I32EnumAttrCase<"L2", 2, "l2">;
```
yzhang93, line 15: This should be pointed out earlier. I think here "1" and "2" are used for bufferize to allocation, and should match the annotation in AIR. In AIR, "1" is for shared memory and "2" is for local memory. So it's not correct to put 1 as L1 and 2 as L2.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.h
yzhang93: I see your reason for setting PackConfig vs PackConfigTD, however we should try to remove one of these. I can help take a look at the issue.

yzhang93 (changes requested):
Thanks for your effort refactoring the PackConfig codes. I think this PR should be further polished. I'll probably try your branch and then add comments / directly modify on it.

Abhishek-Varma:
> I was able to reproduce the failures, locally with this cmake command to build it, I think our local debug builds were hiding the issue https://github.com/nod-ai/iree-amd-aie/blob/main/build_tools/ci/build_test_cpp.sh#L58-L76

Hi @nirvedhmeshram .

Thank you so much for this pointer! This was quite a clever way to reproduce the CI's claims locally.

So, after building the IREE with `build_test_cpp.sh`'s command, I was able to debug this (in a very painstaking way since it was a `Release` build and for some reason, after a few runs, it stopped giving me stack traces as well) :-
`ArrayRef` was misbehaving. The first reason was it didn't let me form the attribute in the first place. And then I saw that it was forming "garbage" values (eg: If we send `[0, 4, 8]`, it'd be forming `[234234123124, 325235246123, 23523523452354]`).

I wonder why this wasn't caught by the `Debug` build itself because the [proof-of-concept] I attached earlier didn't suffer from this problem and everything was working fine!

Perhaps it isn't because `Debug`/`Release` and maybe some other `CMake` flag that's enabling the same.

I'll try keeping everything the same and toggle to `Debug` instead and see if it still crashes without [my fix](https://github.com/nod-ai/iree-amd-aie/pull/155/commits/b13b04213b93d17f6ab0520d1db3423b218d0050) - else it'd be a nightmare to "debug".

Abhishek-Varma:
> Thanks for your effort refactoring the PackConfig codes. I think this PR should be further polished. I'll probably try your branch and then add comments / directly modify on it.

Sure, that'd be great! This definitely needs polish and majorly around the `PackConfig` vs `PackConfigTD` - especially now that they both just differ on only one member variable's declaration. I tried using `Attribute`/`OpFoldResult`/`Value` in the `.td` and elsewhere but seems like that's either not "hash-able" or we have an error related to `FieldParser`.

Feel free to add separate commits on this branch.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.h
yzhang93: I added a commit using one PackConfig struct.

Abhishek-Varma:
Thanks @yzhang93 for helping to keep only one struct now. Very neat!


> I'll try keeping everything the same and toggle to Debug instead and see if it still crashes without [my fix](https://github.com/nod-ai/iree-amd-aie/pull/155/commits/b13b04213b93d17f6ab0520d1db3423b218d0050) - else it'd be a nightmare to "debug".


I just fired the IREE build directory and with the same set of flags + `Debug` build, I tried removing the fix and it didn't crash.
I'm not sure what's the delta between a `Release` and a `Debug` build.

Resorting to use `Debug` build nonetheless and still keeping the fix since that makes CI happy.

Speaking of which - the CI is either stuck or slow since it is at `/home/github/actions-runner/_work/iree-amd-aie/iree-amd-aie/iree-install/bin/iree-run-module --device=xrt --module=output.vmfb --input=8x16xi32=2 --input=16x8xi32=3`.

yzhang93 (commented):
LGTM. Remove my change request. It's better to have @MaheshRavishankar take a look also.

yzhang93 (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.td
```diff
@@ -0,0 +1,100 @@
...
+def AMDAIE_PackingConfigPackingLevelAttr :
+  AttrDef<AMDAIE_Dialect, "PackingConfigPackingLevel", []>
+{
+  let mnemonic = "packing_config_level";
+  let parameters = (ins
+    ArrayRefParameter<"int64_t",
+        "Expected packed sizes for specified iterator dimensions">:$packedSizes,
+    ArrayRefParameter<"int64_t",
+        "Indices of pack operations need to be transposed">:$transposePackIndices,
+    ArrayRefParameter<"int64_t",
```
MaheshRavishankar, line 55: I am not entirely clear why there is an indicator. Pack and unpack should always go together?

Maybe @yzhang93

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDialect.cpp
```diff
@@ -0,0 +1,17 @@
+// Copyright 2021 The IREE Authors
```
MaheshRavishankar, line 1: Change copyright everywhere to be 2024.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDialect.h
```diff
@@ -0,0 +1,24 @@
...
+#include "mlir/IR/OpDefinition.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Support/LLVM.h"
+#include "mlir/Support/TypeID.h"
+
+// clang-format off: must be included after all LLVM/MLIR eaders
+#include "iree-amd-aie/IR/AMDAIEDialect.h.inc"  // IWYU pragma: keep
+// clang-format on
+
+namespace mlir::iree_compiler {}  // namespace mlir::iree_compiler
```
MaheshRavishankar, line 22: Unnecessary?

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.h
MaheshRavishankar: I think we dont need this struct here. The same information is carried by the `Attribute`. The `Attribute` is also a C++ struct. Its definition should have accessors you need to get the information. We dont need to unpack the attribute into this struct.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+}  // namespace mlir::iree_compiler
+
+namespace mlir::iree_compiler::AMDAIE {
+
+//===----------------------------------------------------------------------===//
+// amdaie.packing_config
+//===----------------------------------------------------------------------===//
+
+PackingConfigAttr PackingConfigAttr::get(
```
MaheshRavishankar, line 42: If we drop the C++ struct, this back and forth isnt necessary. We can just use the attribute directly.

MaheshRavishankar (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.td
yzhang93: This is to indicate if the transpose target is a unpack op or a pack op. If the target is a pack op, then we need to set the upackOp as null to pass to this function https://github.com/llvm/llvm-project/blob/411554a3535e55a1436ccda80064d7a91814dc27/mlir/include/mlir/Dialect/Linalg/Transforms/Transforms.h#L1139

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.cpp
```diff
@@ -0,0 +1,127 @@
+// Copyright 2023 The IREE Authors
```
MaheshRavishankar, line 1: Nit: 2024

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.h
```diff
@@ -0,0 +1,64 @@
...
+#define IREE_COMPILER_AMDAIE_DIALECT_PACKINGCONFIG_H_
+
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/IR/BuiltinTypes.h"
+
+namespace mlir::iree_compiler {
+
+// struct PackConfig {
```
MaheshRavishankar, line 20: Remove these comments.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.td
```diff
@@ -0,0 +1,103 @@
+// Copyright 2023 The IREE Authors
```
MaheshRavishankar, line 1: Nit: 2024?

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.td
MaheshRavishankar: Shouldnt this just be a boolean then?

MaheshRavishankar (approved):
Just a few nits, but this is good to go. Thanks!

PR #157, opened 2024-02-12
Transform dialect for conv2d

newling:
Prototype for 2-d convolution. This does not lower to AIE, but is a starting point to illustrate what the llvm-cpu backend approach for convolution looks like when targeting AIE. 

I would like to check this in and get back to lowering it further after vector.contract and other things are shown to work (this transform dialect script converts a convolution into a vector.contract).


tests/samples/conv_fill_spec_pad.mlir
```diff
@@ -0,0 +1,194 @@
...
+                                                        {transform.readonly}) {
+    %f0 = transform.structured.match ops{["func.func"]} in %variant_op
+         : (!any) -> !any
+    %f1 = transform.apply_registered_pass
+         "iree-codegen-decompose-convolution-to-lower-dim-ops"
+         to %f0 : (!any) -> !any
+    transform.yield
+  }
+
+  // This is currently not used, we use the transform dialect vectorization op
```
MaheshRavishankar, line 34: Maybe worth removing this.

MaheshRavishankar (approved):
Nice! Thanks @newling

PR #159, opened 2024-02-13
fix AIESerializer for large values of M

FranzHaniel:
For values of M larger than 16, the matrix multiplication is processed in separate chunks along the M dimension.

The pack-pipeline treats these shapes correctly and yields correct mlir output, however the AIESerializer needed to be adapted in order for the TVM translation to be runnable on device. This required the ivMap from ScopeInfo to be available in each SubScope. This PR also contains some minor cleanup for the offset strings, avoiding e.g. the addition of 0.

A new test verifies the correctness of the translation in these cases.


tests/samples/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ iree_lit_test_suite(
     "pack_pipeline_funcIR.mlir"
     "pack_pipeline_funcIR_2core.mlir"
     "pack_pipeline_funcIR_4core.mlir"
```
yzhang93, line 13: I feel there are too many tests here for this task. Can you move some of them into one file?

FranzHaniel: Good point, I'll take care of that!

nirvedhmeshram: Just a head-up you might run into a problem with splitting the single file into multiple tests described over here https://github.com/nod-ai/iree-amd-aie/pull/134

frafranz: Thanks for the warning, yes I have noticed this issue with splitting as well but so far did not have the time to find a solution. Will give it another try ;)

yzhang93: @nirvedhmeshram I don't get why it's not working with --split-input-file flag? It seems to work fine when I added it in this test https://github.com/nod-ai/iree-amd-aie/blob/main/tests/samples/simple_pack_pipeline_e2e.mlir#L1

FranzHaniel: @yzhang93 the difference is that in the test you linked, you run the same command on two different inputs which I have also used in previous tests, but the tests for 2 cores and 4 cores check different commands (the flag num-cores changes)

FranzHaniel: I integrated these three tests into one now!

yzhang93: Thanks for making the tests into one file! Since you only change the M input size, I don't think you need to add this test "pack_pipeline_funcIR_4core_256_256_256.mlir". Just trying to keep minimal number of tests in this folder.

tests/samples/CMakeLists.txt
```diff
@@ -9,8 +9,7 @@ iree_lit_test_suite(
     lit
   SRCS
     "pack_pipeline_funcIR.mlir"
-    "pack_pipeline_funcIR_2core.mlir"
-    "pack_pipeline_funcIR_4core.mlir"
+    "pack_pipeline_funcIR_4core_256_256_256.mlir"
```
yzhang93, line 12: I think you can remove this test and then this PR is ready to merge.

yzhang93 (approved):

tests/samples/CMakeLists.txt
frafranz: I agree, removed it now.

tests/samples/CMakeLists.txt
frafranz: Done

PR #162, opened 2024-02-14
[WIP] pass to convert linalg to vector dialect 

newling:
Currently, this pass only converts linalg.matmul to the vector dialect, where the reduction size of the matmul is 4 (8) for 32-bit (16-bit) types. It only adds this pass to the "pad" based pipeline.

Should we rather use  `--iree-codegen-generic-vectorization` or were imagining an AIE specific pass along these lines @MaheshRavishankar ? 

I need to figure out why the linalg.matmul is converted to vector.multi_reduction and not vector.contract. I assume we must emit vector.contract @jsetoain ?

If I test with bfloat16, I hit that error which I think others have seen, not sure what the status is (@nirvedhmeshram ?) :  

```
1185: <stdin>:20:16: error: 'aiex.ipu.dma_memcpy_nd' op must be used with memref type with element width 32.
1185:           %7 = linalg.matmul ins(%3, %4 : tensor<8x16xbf16>, tensor<16x8xbf16>) outs(%6 : tensor<8x8xbf16>) -> tensor<8x8xbf16>
```

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEVectorization.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEVectorization.cpp
@@ -1,0 +1,110 @@
+// Copyright 2024 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <memory>
+
+#include "iree-amd-aie/Transforms/Passes.h"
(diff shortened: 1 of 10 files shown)


newling:
> You can use the IREE passes directly. To convert from `multi_reduction` to `contract` you need these patterns https://github.com/llvm/llvm-project/blob/main/mlir%2Finclude%2Fmlir%2FDialect%2FVector%2FTransforms%2FVectorRewritePatterns.h#L104
> 
> I think the generic vectorization lass from iree has an option to allow you to generate contract by effectively calling these patterns


I've made a new PR to use iree's vectorization pass: https://github.com/nod-ai/iree-amd-aie/pull/163 -- a later AIR pass fails if we use this more "comprehensive" vectorization

Issue #120, opened 2024-01-31
Accumulate matmul is currently failing


nirvedhmeshram:
Here is the crash I am seeing for matmul with accumulate
https://gist.github.com/nirvedhmeshram/32a76d8676b8d4d3ff6e7bbc7a10d05d
related PR: https://github.com/nod-ai/iree-amd-aie/pull/119

erwei-xilinx:
I think this is related: https://github.com/nod-ai/iree-amd-aie/issues/124#issuecomment-1925571148

PR #163, opened 2024-02-14
[WIP] Use iree's generic vectorization pass

newling:
Alternative to https://github.com/nod-ai/iree-amd-aie/pull/162 

Rather than manually convert on an op-by-op basis, use IREE's vectorization pass. 

vector.contract appears as expected, but lowering fails in an air pass -- An MLIR assertion failure in air-collapse-herd:

```
mlir::IRObjectWithUseList<mlir::OpOperand>::~IRObjectWithUseList() [OperandType = mlir::OpOperand]: Assertion `use_empty() && "Cannot destroy a value that still has uses!"' failed.
```

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
@@ -27,7 +27,7 @@ static LogicalResult applyBufferizeToAllocation(RewriterBase &rewriter,
   options.memcpyOp =
       linalg::BufferizeToAllocationOptions::MemcpyOp::MaterializeInDestination;
   options.allocOp = linalg::BufferizeToAllocationOptions::AllocOp::MemrefAlloc;
-  options.bufferizeDestinationOnly = true;
+  options.bufferizeDestinationOnly = false;
   options.emitDealloc = true;
 
   // Bufferize ops.
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
@@ -76,6 +76,13 @@ static void addAMDAIEBufferizePasses(OpPassManager &pm) {
   addIREEComprehensiveBufferizePasses(pm, allocationFn, memCpyFn);
 }
 
+static void addAMDAIEVectorizePasses(OpPassManager &pm) {
+  mlir::iree_compiler::GenericVectorizationPassOptions opts{};
+  opts.generateContract = true;
+  pm.addNestedPass<func::FuncOp>(
+      mlir::iree_compiler::createGenericVectorizationPass(opts));
+}
+
 void addPadBasedPassPipeline(OpPassManager &pm, TilingConfig &tilingConfig) {
   auto &modulePassManager = pm.nest<ModuleOp>();
   modulePassManager.addNestedPass<func::FuncOp>(createAMDAIECleanupPass());
@@ -117,6 +124,7 @@ void addPadBasedPassPipeline(OpPassManager &pm, TilingConfig &tilingConfig) {
     pm.addPass(createCanonicalizerPass());
     pm.addPass(createCSEPass());
   }
+  addAMDAIEVectorizePasses(modulePassManager);
   addAMDAIEBufferizePasses(modulePassManager);
   modulePassManager.addNestedPass<func::FuncOp>(createAMDAIECleanupPass());
   pm.addPass(createCanonicalizerPass());
--- a/tests/samples/CMakeLists.txt
+++ b/tests/samples/CMakeLists.txt
@@ -12,6 +12,7 @@ iree_lit_test_suite(
     "pack_pipeline_funcIR_2core.mlir"
     "pack_pipeline_funcIR_4core.mlir"
     "pad_pipeline_e2e.mlir"
+    "pad_pipeline_vectorized.mlir"
     "simple_pack_pipeline_e2e.mlir"
     "pad_pipeline_conv2d.mlir"
   TOOLS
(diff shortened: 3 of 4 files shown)


nirvedhmeshram:
A similar error is also causing this issue
https://github.com/nod-ai/iree-amd-aie/issues/120
It may be worth figuring out what needs to change, to allow air to lower these kind of IRs

newling:
I can "fix" this by preventing linalg::CopyOp from being vectorized. i.e. in ` GenericVectorizationPass::runOnOperation()` if I add

```
      if (isa<linalg::CopyOp>(op)){
        // Skip vectorization of copy operations.
        continue;
      }
```

over [here](https://github.com/openxla/iree/blob/e83ce9e225eefd9245cd38d37d843a915fbaa034/compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp#L164) 

This prevents CopyOp's from being vectorized [here](https://github.com/openxla/iree/blob/e83ce9e225eefd9245cd38d37d843a915fbaa034/compiler/src/iree/compiler/Codegen/Common/GenericVectorization.cpp#L190)

```
    (void)linalg::vectorize(rewriter, op, vectorSizes, scalableVecDims,
                            vectorizeGatherAccesses);
```

### Why does this help? 

With the padding approach, we have the passes which append IR like 


```
%extracted_slice_0 = tensor.extract_slice %4 ...  to tensor<4x32x3x3xf32>
```

with IR like 

```
%13 = bufferization.alloc_tensor() : tensor<4x32x3x3xf32>
%alloc_2 = memref.alloc() : memref<4x32x3x3xf32, 1>
%14 = bufferization.to_tensor %alloc_2 ... : memref<4x32x3x3xf32, 1>
%15 = linalg.copy ins(%extracted_slice_0 : tensor<4x32x3x3xf32>)
             outs(%14 : tensor<4x32x3x3xf32>) -> tensor<4x32x3x3xf32>
```

This is the mechanism used to specify that the tensor should be in memory space 1 (i.e. on the AIE memtile). 

The vectorization pass, when run on linalg.copy, is doing some kind of unwanted canonicalization and removing the linalg.copy, thus removing all trace of allocation on memtile. 

Next steps:

1) Figure out why linalg.copy is being removed in vectorization, figure out a clean way of avoiding this.
1) If the IR isn't something AIR can handle, it should pick this up earlier and provide useful information.

newling:
As a reproducer, consider simple_test.mlir:

```
module {
  func.func @extract_slice_and_copy(%arg0: tensor<8x16xf32>) -> tensor<8x16xf32> {
    %extracted_slice = tensor.extract_slice %arg0[0, 0] [8, 16] [1, 1] : tensor<8x16xf32> to tensor<8x16xf32>
    %alloc = memref.alloc() : memref<8x16xf32, 1 : i32>
    %0 = bufferization.to_tensor %alloc restrict writable : memref<8x16xf32, 1 : i32>
    %1 = linalg.copy ins(%extracted_slice : tensor<8x16xf32>) outs(%0 : tensor<8x16xf32>) -> tensor<8x16xf32>
    return %1 : tensor<8x16xf32>
  }
}
```

Running `iree-opt --canonicalize  simple_test.mlir` does not change the IR. 

Running `iree-opt --iree-codegen-generic-vectorization  simple_test.mlir`  transforms to 

```
module {
  func.func @extract_slice_and_copy(%arg0: tensor<8x16xf32>) -> tensor<8x16xf32> {
    return %arg0 : tensor<8x16xf32>
  }
}
```

And running `iree-opt --iree-codegen-generic-vectorization  simple_test.mlir` with the patch in IREE to not vectorize linalg.copy leaves the IR unchanged.

newling:
What should happen when this gets canonicalized? 
```
module {
func.func @extract_slice_and_copy(%arg0: tensor<8x16xf32>) -> tensor<8x16xf32> {
  %cst = arith.constant 0.000000e+00 : f32
  %c0 = arith.constant 0 : index
  %alloc = memref.alloc() : memref<8x16xf32, 1 : i32>
  %0 = bufferization.to_tensor %alloc restrict writable : memref<8x16xf32, 1 : i32>
  %1 = vector.transfer_read %arg0[%c0, %c0], %cst {in_bounds = [true, true]} : tensor<8x16xf32>, vector<8x16xf32>
  %2 = vector.transfer_write %1, %0[%c0, %c0] {in_bounds = [true, true]} : vector<8x16xf32>, tensor<8x16xf32>
  return %2 : tensor<8x16xf32>
}
}
```
Currently this is what happens:
```
module {
  func.func @extract_slice_and_copy(%arg0: tensor<8x16xf32>) -> tensor<8x16xf32> {
    return %arg0 : tensor<8x16xf32>
  }
}
```
which I guess is valid. I suppose it's a matter of "use `bufferization.to_tensor` at your own risk" ?

newling:
To put another way, the problem is that linalg.copy doesn't fold, but vector.transfer_read and vector.transfer_write do. So the memory space trick we use in iree-amd-aie:

```
              %alloc_2 = memref.alloc() : memref<16x8xi32, 1>
              %13 = bufferization.to_tensor %alloc_2 restrict writable : memref<16x8xi32, 1>
              %14 = linalg.copy ins(%extracted_slice_0 : tensor<16x8xi32>) outs(%13 : tensor<16x8xi32>) -> tensor<16x8xi32>
```

works fine, but when we vectorize linalg.copy, the resulting vector.transfer_read/vector.transfer_write get folded away and consumers of %14 just consume %extracted_slice_0.

newling:
Update: I can work around the folding issue by modifying the bufferize-to-allocation pass to use `options.bufferizeDestinationOnly = false;` instead of `options.bufferizeDestinationOnly = true`. 

Why does this work? A copy between memrefs lowers to vector.transfer_read/transfer_write between memref which is not folded. Basically it expands the set of ops in reference semantics by a little bit, thus reducing the folding. 

This enables lowering to proceed all to the air ping-pong transformation, where it then fails (region missing a terminator or something). 

Digging deeper, I suspect that the problem is now emerging in the `CopyToDma` pass. It does not convert the vector transfer operations to dma operations.

newling:
Closing in favour of https://github.com/nod-ai/iree-amd-aie/pull/172

PR #164, opened 2024-02-14
Ensure cleanup and avoid conflict between jobs

spaceotter:
This gives the deployed xclbins a unique name based on the current workflow run. This avoids a potential collision where parallel jobs could overwrite the deployed kernel of another job, thereby un-deploying it. Then, a cleanup step that runs even if the job is cancelled erases any kernels that were deployed using the unique name scheme.

### DIFF
--- a/build_tools/ci/pad_test.sh
+++ b/build_tools/ci/pad_test.sh
@@ -5,7 +5,8 @@ set -xe
 TESTDIR="$1"
 
 BASE_DIR=`realpath "$(dirname $0)/../.."`
-IREE_DIR="${2:-${BASE_DIR}/../iree}"
+IREE_DIR="$2"
+MLIR_AIE_VERSION="$3"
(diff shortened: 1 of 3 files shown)


build_tools/ci/pad_test.sh
```diff
@@ -19,7 +19,7 @@ cd "$TESTDIR"
 
 python3 -m venv sandbox
 source sandbox/bin/activate
-MLIR_AIE=mlir_aie-0.0.1.2024021121+e6874fb
+MLIR_AIE=mlir_aie-0.0.1.2024021415+7ee44b7
```
nirvedhmeshram, line 22: how do you feel about making this an argument, since we may have multiple scripts, it will be easy to track from ci.yml that way?

spaceotter: Yeah we can do that

nirvedhmeshram (approved):
LGTM!

PR #169, opened 2024-02-15
Bump pinned IREE version and update mlir-air and mlir-aie submodules

nirvedhmeshram:


yzhang93 (approved):

MaheshRavishankar (approved):
@erwei-xilinx we want to stop bumping mlir-air as well. Essentially check pointing what AIR commit we use. Do we stop doing that after this bump?

erwei-xilinx:
> @erwei-xilinx we want to stop bumping mlir-air as well. Essentially check pointing what AIR commit we use. Do we stop doing that after this bump?

Yeah that's fine with me. This hash of MLIR-AIR contains my most recent works in MLIR-AIR.

PR #170, opened 2024-02-16
Add initial E2E correctness tests

nirvedhmeshram:
These tests are compared against a reference matmul implementation with pseudo random initialization. The test sizes are currently limited. In upstream IREE similar tests are build using cmake and run using ctest, however that is currently not possible for a plugin so the script added here is an alternate way of doing it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,7 +100,7 @@ jobs:
       - name: Create venv and install dependencies
         run: |
           python3 -m venv .venv
-          source myenv/bin/activate
+          source .venv/bin/activate
           pip install https://github.com/Xilinx/mlir-aie/releases/download/latest-wheels/mlir_aie-0.0.1.2024021415+7ee44b7-py3-none-manylinux_2_35_x86_64.whl
           pip install -r tests/matmul/requirements.txt
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,13 +97,22 @@ jobs:
           mkdir iree-install
           tar -xf iree-dist.tar -C iree-install
 
+      - name: Create venv and install dependencies
+        run: |
+          python3 -m venv .venv
+          source myenv/bin/activate
+          pip install https://github.com/Xilinx/mlir-aie/releases/download/latest-wheels/mlir_aie-0.0.1.2024021415+7ee44b7-py3-none-manylinux_2_35_x86_64.whl
+          pip install -r tests/matmul/requirements.txt
+
       - name: Pad matmul test
         run: |
-          bash build_tools/ci/pad_test.sh test1 iree-install mlir_aie-0.0.1.2024021415+7ee44b7
+          source .venv/bin/activate
+          bash build_tools/ci/pad_test.sh test1 iree-install
 
       - name: E2E correctness matmul test
         run: |
-          bash build_tools/ci/run_matmul_test.sh test2 iree-install mlir_aie-0.0.1.2024021415+7ee44b7
+          source .venv/bin/activate
+          bash build_tools/ci/run_matmul_test.sh test2 iree-install
 
       - name: Clean up
         if: ${{ always() }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,7 @@ jobs:
           # installs.
           rm -f iree-install/bin/clang*
           rm -f iree-install/bin/llvm-link*
-          tar cf iree-dist.tar -C iree-install .
+          tar cf iree-dist.tar -C iree-install . -C ../iree-build tools/iree-e2e-matmul-test
 
       - name: Upload artifacts
         uses: actions/upload-artifact@v4
@@ -103,7 +103,7 @@ jobs:
 
       - name: E2E correctness matmul test
         run: |
-          bash build_tools/ci/run_matmul_test.sh test2 iree-install iree-build mlir_aie-0.0.1.2024021415+7ee44b7
+          bash build_tools/ci/run_matmul_test.sh test2 iree-install mlir_aie-0.0.1.2024021415+7ee44b7
 
       - name: Clean up
         if: ${{ always() }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,7 +103,7 @@ jobs:
 
       - name: E2E correctness matmul test
         run: |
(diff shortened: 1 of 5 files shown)


build_tools/ci/run_matmul_test.sh
```diff
@@ -120,12 +144,28 @@ function run_matmul_test() {
...
       -o "${OUTPUT_DIR}/${name}_matmuls.vmfb"
   ${IREE_COMPILE_EXE} \
       "${OUTPUT_DIR}/${name}_calls.mlir" \
       --iree-hal-target-backends=${target_backend} \
       -o "${OUTPUT_DIR}/${name}_calls.vmfb"
 
+  # Extract function names from the mlir file
+  function_names=$(grep -oP '@\K\S+(?=\()' ${OUTPUT_DIR}/${name}_matmuls.mlir)
+
+  # Iterate over each function name and sign the corresponding XCLBIN
```
ScottTodd, line 158: Not sure what this is doing / why this is needed... but this looks like something that the tools should do on their own, rather than need some extra scripting. Fine for now but wouldn't want this to spread to many other files.

The `iree-hal-dump-*` flags are for developer use and should not be load bearing for a deployment workflow.

build_tools/ci/run_matmul_test.sh
```diff
@@ -120,12 +144,28 @@ function run_matmul_test() {
       --iree-amd-aie-peano-install-dir=${peano_install_path} \
       --iree-amd-aie-mlir-aie-install-dir=${mlir_aie_install_path} \
       --iree-amd-aie-vitis-install-dir=${vitis_path} \
+      --iree-hal-dump-executable-files-to=$PWD \
+      --iree-hal-dump-executable-intermediates-to=$PWD \
```
ScottTodd, line 148: The `intermediates` flag is redundant here. `files` is a meta-flag that sets all the other flags: https://iree.dev/developers/general/developer-tips/#dumping-executable-files

build_tools/ci/run_matmul_test.sh
```diff
@@ -39,6 +56,13 @@ TEST_RUNNER="${IREE_INSTALL_BIN}/iree-e2e-matmul-test"
 echo "Python version: $("${IREE_PYTHON3_EXECUTABLE}" --version)"
 echo "iree-compile version: $("${IREE_COMPILE_EXE}" --version)"
 mkdir -p ${OUTPUT_DIR}
+cd ${OUTPUT_DIR}
+
+python3 -m venv sandbox
+source sandbox/bin/activate
```
ScottTodd, line 62: Does iree-amd-aie have an established naming convention for venv directories? In IREE we suffix them with `.venv` (just `.venv` on its own with no prefix is the default used in many places)

build_tools/ci/run_matmul_test.sh
nirvedhmeshram: Ya this is a signing process needed for running on AIE hardware at the moment, the estimate it is that we wont be needing it in about a month.

erieaton-amd: If it were merged into the run-module tool, either that must be run as root or it has to perform operations in a directory that is normally configured with root permissions.
Nobody likes this part, you don't need root powers to run things on GPU, but this is where we're at...
I proposed creating a separate tool for this purpose, but it's not a priority since apparently this requirement is expected to go away.

ScottTodd (approved):

.github/workflows/ci.yml
```diff
@@ -59,7 +59,7 @@ jobs:
           # installs.
           rm -f iree-install/bin/clang*
           rm -f iree-install/bin/llvm-link*
-          tar cf iree-dist.tar -C iree-install .
+          tar cf iree-dist.tar -C iree-install . -C ../iree-build tools/iree-e2e-matmul-test
```
erieaton-amd, line 62: Shouldn't you make sure this is added to the install in the cmake code?

erieaton-amd (changes requested):

.github/workflows/ci.yml
nirvedhmeshram: It was decided that we dont want it in install
https://discord.com/channels/689900678990135345/1169307503633383664/1208082521720488016

build_tools/ci/run_matmul_test.sh
```diff
@@ -39,6 +56,13 @@ TEST_RUNNER="${IREE_INSTALL_BIN}/iree-e2e-matmul-test"
 echo "Python version: $("${IREE_PYTHON3_EXECUTABLE}" --version)"
 echo "iree-compile version: $("${IREE_COMPILE_EXE}" --version)"
 mkdir -p ${OUTPUT_DIR}
+cd ${OUTPUT_DIR}
+
+python3 -m venv sandbox
```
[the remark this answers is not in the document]
ScottTodd: Good point - could install deps once then let the test steps reuse that same environment.

.github/workflows/ci.yml
ScottTodd: We chatted about this over on [IREE's Discord server here](https://discord.com/channels/689900678990135345/1169307503633383664/1208069599178924082). `iree-e2e-matmul-test` is a dev tool that should not be included in the user install. The options here were to either include it in the archive here or [re]build it from source in the next job.

build_tools/ci/run_matmul_test.sh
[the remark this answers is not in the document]
ScottTodd: Can you link to some logs?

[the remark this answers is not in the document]
nirvedhmeshram: https://github.com/nod-ai/iree-amd-aie/actions/runs/7934375875/job/21665336705?pr=170

I am suspecting this is failing because the MLIR_AIE_INSTALL path is not correct which I get from the venv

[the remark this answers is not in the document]
nirvedhmeshram: Here is a successful one before venv sharing
https://github.com/nod-ai/iree-amd-aie/actions/runs/7933813811/job/21663559947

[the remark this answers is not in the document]
ScottTodd: Bah GitHub Actions troubleshooting... Maybe try `python -m pip install` instead of `pip install`, to make sure  you use pip from the venv?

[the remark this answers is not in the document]
nirvedhmeshram: Thanks, just had to find the path to the venv correctly.

erieaton-amd (approved):

PR #171, opened 2024-02-16
[AMD-AIE] Update GEMM OPT tests

Abhishek-Varma:
-- 2 of the 5 tests were wrong/invalid - this commit rectifies that.
-- It also adds the corresponding f32 counterpart to completely cover
   all GEMMs seen in the OPT dispatches.

GEMM of OPT has three blockers :-
1. Op support. (a `linalg.generic` that performs matmul transpose AND a `linalg.batch_matmul`)
2. Shape support.
3. Element type support (`f32`).

My intention was to keep a self-contained GEMM check particularly for OPT but the previous test only satisfied `2`.
So while trying to add others, I bumped into errors in SSA access in the test case which wasn't caught due to `XFAIL`. Since this mistake from my end would've caused a red-herring later, I'm rectifying the same in this PR.

I've verified both the files now : `iree-opt ~/iree-amd-aie/tests/OPT/failing_tests/gemm_i32.mlir --split-input-file` (same for `gemm_f32.mlir`) - and I confirm all are valid ("verified" by MLIR) and covers all cases for OPT GEMM thoroughly.

NOTE: One of the test cases added a `linalg.generic` whereas we have `linalg.batch_matmul` for that shape - so this warranted a changed in the corresponding `CHECK` statement.

Apologies for the inconvenience - a genuine copy/paste error.

### DIFF
--- a/tests/OPT/failing_tests/CMakeLists.txt
+++ b/tests/OPT/failing_tests/CMakeLists.txt
@@ -8,6 +8,7 @@ iree_lit_test_suite(
   NAME
     lit
   SRCS
+    "gemm_f32.mlir"
     "gemm_i32.mlir"
   TOOLS
     ${IREE_LLD_TARGET}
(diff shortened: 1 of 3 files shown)


nirvedhmeshram (approved):
LGTM!

PR #172, opened 2024-02-16
Vectorization pass, with e2e bf16 test

newling:
This new pass is based on IREE's GenericVectorization pass.  This new pass differs in that 

(i) it includes only what is necessary to get a simple matmul example lowered end-to-end
(ii) it explicitly does not vectorize linalg.copy and linalg.fill (see code comments for reasons)
(iii) it explicitly does not vectorize operations on 32/64-bit types (see code comments). 

An end-to-end test for bfloat16 is enabled in the matmul test script. (m=n=64, k=16). Two other tests are (1) with iree-compile, a test to check for presence of vector.contract and (2) with iree-opt, running just this new pass to check (ii) and (iii) above. 

There is follow-up work to do:

1) check support for integer types (AIE can do "tensor core" matmul for various int types)
2) check vectorization pipeline for convolution
3) check vectorization where result type is bf16 (current test has result type of matmul f32)
4) check what the issue is with vectorizing linalg.fill 
5) vectorization of elementwise operations
...


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -21,7 +21,18 @@ static LogicalResult setRootConfigForPadPipeline(func::FuncOp entryPointFn,
                                                  linalg::MatmulOp matmulOp) {
   SmallVector<int64_t> TileSizeLevel0 = {8, 8};
   SmallVector<int64_t> TileSizeLevel1 = {4, 4};
-  SmallVector<int64_t> TileSizeLevel2 = {0, 0, 4};
+
+  auto dtype =
+      matmulOp->getOperand(0).getType().cast<ShapedType>().getElementType();
+
+  // TODO: find a better place to register these vector sizes.
```
MaheshRavishankar, line 28: This is as good a place as any :)

tests/samples/pad_pipeline_vectorized.mlir
```diff
@@ -0,0 +1,22 @@
+// RUN: iree-compile --iree-hal-target-backends=amd-aie --compile-to=executable-sources %s | iree-opt --pass-pipeline="builtin.module(hal.executable(hal.executable.variant(iree-hal-translate-target-executable-variants{target=amd-aie})))" --iree-amdaie-use-pipeline=pad | FileCheck %s
```
MaheshRavishankar, line 1: You dont need to start all the way from `iree-compile` . You should be able to write a lit test just for this pass and this does not go into `tests/samples`. That pass test should go into https://github.com/nod-ai/iree-amd-aie/tree/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test .

MaheshRavishankar (changes requested):
Nice! overall looks good to me.

MaheshRavishankar (commented):
Could we rename the PR please. The name is confusing me.

build_tools/ci/run_matmul_test.sh
```diff
@@ -42,7 +42,7 @@ if [ "$#" -lt 2 ] || [ "$#" -gt 6 ]; then
             "\n     6) <vitis-install-dir>        (optional)" \
             "\n Example, dependent on environment variables:" \
             "\n     ./run_matmul_test.sh  " \
-            "results_dir_tmp  \$MLIR_AIE_INSTALL_DIR  \$IREE_INSTALL_DIR  " \
```
newling, line 45: Fixes typo in a previous commit

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/vectorization.mlir
```diff
@@ -0,0 +1,70 @@
...
+// is vectorized to a vector.contract operation.
+// CHECK-LABEL: func @matmul_m4_n4_k8_bf16_f32
+func.func @matmul_m4_n4_k8_bf16_f32(%arg0: tensor<4x8xbf16>, %arg1: tensor<8x4xbf16>, %arg2: tensor<4x4xf32>) -> tensor<4x4xf32> {
+  // CHECK-DAG: %[[cst:.*]] = arith.constant 0.000000e+00 : f32
+  // CHECK-DAG: %[[c0:.*]] = arith.constant 0 : index
+  // CHECK-DAG: %[[cst_0:.*]] = arith.constant 0.000000e+00 : bf16
+  // CHECK-DAG: vector.transfer_read{{.*}} tensor<4x8xbf16>, vector<4x8xbf16>
+  // CHECK-DAG: vector.transfer_read{{.*}} tensor<8x4xbf16>, vector<8x4xbf16>
+  // CHECK-DAG: vector.transfer_read{{.*}} tensor<4x4xf32>, vector<4x4xf32>
+  // CHECK: vector.contract{{.*}}vector<4x8xf32>, vector<8x4xf32> into vector<4x4xf32>
```
nirvedhmeshram, line 18: shouldnt this be `vector.contract{{.*}}vector<4x8xbf16>, vector<8x4xbf16> into vector<4x4xf32>` ?

build_tools/ci/run_matmul_test.sh
```diff
@@ -341,6 +341,18 @@ run_matmul_test \
     --vitis_path  "${VITIS}" \
     --pipeline "simple-pack"
 
+run_matmul_test \
+    --name "matmul_bf16_bf16_large_amd-aie_xrt_simple-pack" \
+    --lhs_rhs_type "bf16" \
+    --acc_type "f32" \
+    --shapes "large" \
```
nirvedhmeshram, line 348: So for the large shape, currently we have one "accumulate=false" type matmul. This is the zero-fill matmul. But from the lit test and your pass you are currently only supporting accumulate=true matmul due to the air bug. Can you please clarify?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/vectorization.mlir
newling: There is some casting I've omitted from CHECK here (but will add in for clarity), looks something like:

```
%58 = arith.extf %49 : vector<1x1x1x1x4x8xbf16> to vector<1x1x1x1x4x8xf32>
%59 = arith.extf %53 : vector<1x1x1x1x8x4xbf16> to vector<1x1x1x1x8x4xf32>
```

In the pass `LowerVectorToAIEVec` (run in aie2xclbin) this contract becomes

```
%71 = aievec.matmul %69, %70, %68 : vector<4x8xbf16>, vector<8x4xbf16> into vector<4x4xf32>
```  

It is a bit confusing and not the most direct path.

build_tools/ci/run_matmul_test.sh
newling: Fill is still supported, just not vectorized. It's when we run `linalg::vectorize(rewriter, fillOp)` that the bug is hit.  accumulate=true is supported.

nirvedhmeshram: Thanks, makes sense.

MaheshRavishankar (approved):

PR #175, opened 2024-02-20
[AMD-AIE] Add a pass to lower to Ukernels

Abhishek-Varma:
-- This commit adds a pass to lower to Ukernels.
-- Currently this targets the Pad based linalg.matmul or
   a linalg.generic which performs a matmul that of i32/bf16 variant,
   since we have a lead on the corresponding microkernels.
-- It also modifies AIETarget to enable adding "ukernels" config.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+  // These patterns are unconditionally enabled, because we have strong evidence
+  // that it is difficult for codegen to consistently approach microkernels
+  // performance, and that consideration overrides the benefit of fusions for
+  // these ops.
+  auto allTargets = [](auto target) { return true; };
+  // TODO(avarma): There is no need to have microkernels for anything but the
+  // target-optimal tiled layout. So, while building the microkernel pipeline, I
+  // need to have a linalg.matmul -> linalg.mmt4d.
+  patterns.insert<LowerToUKernelPattern<linalg::Mmt4DOp>>(context, allTargets);
```
Abhishek-Varma, line 250: I started off with `linalg.matmul`, but reverted to `linalg.mmt4d` based on my understanding that -> it makes sense to have microkernel only for tiled performant ISA specific implementations.

@MaheshRavishankar please let me know if we want to target `linalg.matmul` itself in this pass - which means we would be invoking this way earlier in the pipeline before the "tile+fuse" enchilada (since `linalg.matmul` gets lowered to a tiled `linalg.generic` in the pipeline).

Based on the discussion, I can reconfigure it to target `linalg.matmul` OR follow along with the `linalg.matmul -> linalg.mmt4d` comment I added as a `TODO`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
```diff
@@ -55,6 +58,14 @@ struct AMDAIEOptions {
     binder.opt<bool>("iree-amd-aie-enable-chess", useChess,
                      llvm::cl::cat(category),
                      llvm::cl::desc("Use the legacy chess compiler"));
+
+    binder.opt<std::string>(
+        "iree-amdaie-enable-ukernels", enableUkernels, llvm::cl::cat(category),
+        llvm::cl::desc(
+            "Enables microkernels in the llvmcpu backend. May be "
```
nirvedhmeshram, line 65: nit: amd-aie backend?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
newling: Just a heads up that it might be a vector.contract in the near future

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
```diff
@@ -0,0 +1,263 @@
...
+    flags = IREE_UK_FLAG_MMT4D_TYPE_BF16BF16F32;
+  } else if (lhsElemType.isBF16() && rhsElemType.isBF16() &&
+             outElemType.isBF16()) {
+    flags = IREE_UK_FLAG_MMT4D_TYPE_BF16BF16BF16;
+  } else {
+    return rewriter.notifyMatchFailure(
+        op, "unsupported combination of element types");
+  }
+
+  // // Check if the accumulator is zero-filled.
```
newling, line 162:
```suggestion
  // Check if the accumulator is zero-filled.
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
```diff
@@ -0,0 +1,263 @@
...
+  Value lhs = getInputForUKernel(op.getDpsInputOperand(0)->get());
+  Value rhs = getInputForUKernel(op.getDpsInputOperand(1)->get());
+  Value out = op.getDpsInitOperand(0)->get();
+  auto outType = llvm::cast<ShapedType>(out.getType());
+  Type lhsElemType = getElementTypeForUKernel(op.getDpsInputOperand(0)->get());
+  Type rhsElemType = getElementTypeForUKernel(op.getDpsInputOperand(1)->get());
+  Type outElemType = outType.getElementType();
+  uint32_t flags = 0;
+  if (lhsElemType.isBF16() && rhsElemType.isBF16() && outElemType.isF32()) {
+    flags = IREE_UK_FLAG_MMT4D_TYPE_BF16BF16F32;
```
newling, line 153: Where are these flags defined?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -95,6 +95,14 @@ def AMDAIELoweringStrategy :
   ];
 }
 
+def AMDAIELowerToUKernels :
+    Pass<"iree-amdaie-lower-to-ukernels", ""> {
+  let summary =
+      "Separate out parts of the IR that lower to a micro-kernel";
```
newling, line 101: It would be good to have a bit more in the summary about what level the micro-kernels are matching at. Is it a whole operation, or just the inner-most loop of the operation? Maybe it's not decided yet, based on the matmul vs mmt4d question earlier.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  // // Check if the accumulator is zero-filled.
+  if (isInitializedToZero(out)) {
+    // Not setting flags |= IREE_UK_FLAG_MMT4D_ACCUMULATE, so the mmt4d op won't
+    // read the existing accumulator, so its defining op can be discarded.
+    if (auto fillOp = out.getDefiningOp<linalg::FillOp>()) {
+      out = fillOp.getDpsInitOperand(0)->get();
+    }
+  } else {
```
newling, line 169: It could be initialized to 1, or something else? I guess that would be fine though, it'd just keep that initialization in the IR (`out` unchanged)

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_ukernel.mlir
```diff
@@ -0,0 +1,81 @@
...
+  %fill = linalg.fill ins(%cst : bf16) outs(%arg2 : tensor<?x?x?x?xbf16>) -> tensor<?x?x?x?xbf16>
+  %0 = linalg.mmt4d ins(%arg0, %arg1 : tensor<?x?x?x?xbf16>, tensor<?x?x?x?xbf16>)
+      outs(%fill : tensor<?x?x?x?xbf16>) -> tensor<?x?x?x?xbf16>
+  return %0 : tensor<?x?x?x?xbf16>
+}
+//      CHECK: func @mmt4d_fill(
+// CHECK-SAME:     %[[ARG0:[a-zA-Z0-9]+]]: tensor<?x?x?x?xbf16>
+// CHECK-SAME:     %[[ARG1:[a-zA-Z0-9]+]]: tensor<?x?x?x?xbf16>
+// CHECK-SAME:     %[[ARG2:[a-zA-Z0-9]+]]: tensor<?x?x?x?xbf16>
+//  CHECK-DAG:   %[[FLAGS:.+]] = arith.constant 6 : i32
```
newling, line 66: Would it possible to encode that the `accumulate` vs `initialize to 0` cases in a human readable form (i.e. something more obvious than the integers 6 and 262) ? Probably not without modifying IREE, I suppose.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
Abhishek-Varma: I've currently let go of the flag and added a TODO.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
Abhishek-Varma: I don't think so. I might be wrong but I believe this just checks if the `fill` op is present in the function body (to consume it directly while lowering and not having to read the zero filled accumulator).

Soon all these aspects of the code will get clearer as I make incremental progress - currently that's my honest understanding of the state (and demonstrating the same via the lit tests).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_ukernel.mlir
Abhishek-Varma: As mentioned in an earlier thread - I've, for now, gotten rid of the "flag" (value `6` and `262` is no more to be dealt with, for now though).

Once this pipeline gets clearer to me, I'll make incremental changes and modify.

But answering the initial ask - I'll have to check if some beautification (if needed) can be done on the encoded value later. Since the current revision doesn't have any of these encoding value (flag), I'd table off this for now.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
Abhishek-Varma: I've had a discussion with @MaheshRavishankar and have addressed the PR accordingly.

The decision was to target a `linalg.generic` since that's what we get after the last `pack-and-transpose` pass via simple-pack pipeline.

Therefore, I've added a utility that matches a tiled `linalg.generic` op which implements a matmul and have added TODOs/comments to explain the rationale in the pass too.

One of the cases that I bumped into while implementing was the presence of more than one `linalg.generic`s in the dispatch - I've added an implementation that deals with it, have demonstrated it via a lit test too and have added a TODO that highlights a possible "better" way to deal with the same.

NOTE: This pass seems like would go right after the last `pack-and-transpose + canonicalize + cse`. This is because `memref` related replacements are too ugly to deal with and the pass would crash too since a bufferized DPS linalg doesn't really have a "result". By "result", I mean : `%result = linalg.generic ....`. And due to this the `replaceOp` would cry.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
```diff
@@ -0,0 +1,296 @@
...
+    if (targetPredicate &&
+        !targetPredicate(IREE::HAL::ExecutableTargetAttr::lookup(op))) {
+      return failure();
+    }
+
+    FailureOr<IREE::Codegen::UKernelOpInterface> ukernelOp;
+    if (isMatmul(op)) {
+      ukernelOp = matchDAGForUKernel(rewriter, op, "matmul");
+    } else {
+      // TODO(avarma): This looks murky. There are dispatches with fused
```
MaheshRavishankar, line 239: This is OK actually. If we want to use a ukernel + fusion, we will need an intermediate stack buffer allocation in L1 to satisfy dependencies. So you can drop the TODO I think

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
```diff
@@ -0,0 +1,296 @@
...
+    // defining op can be discarded.
+    if (auto fillOp = out.getDefiningOp<linalg::FillOp>()) {
+      out = fillOp.getDpsInitOperand(0)->get();
+    }
+  }
+
+  Location loc = op.getLoc();
+  // Since this is invokved after the final pack-and-transpose we need to
+  // extract dimensions for M*K*k*m*m0*k0 x K*N*n*k*k0*n0 -> M*N*n*m*m0*n0.
+  Value M = rewriter.create<tensor::DimOp>(loc, lhs, 0);
```
MaheshRavishankar, line 187: We dont need to pass all this information. We need to pass only the tensor dimensions that are needed by the ukernel implementation. Shouldnt matter much, but just seems unnecessary.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
MaheshRavishankar: We will have to balance the use of microkernels and generic vectorization path, maybe add a flag, or prefer microkernels over direct vectorization. `vector.contract` is a harder op to lower to microkernel cause you dont have a pointer value. So we will have to figure out an ABI for vector types.

MaheshRavishankar (approved):
Thanks Abhishek! Looks good.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
```diff
@@ -0,0 +1,271 @@
...
+      outBlockArg.getOwner() != body || lhsBlockArg.getArgNumber() != 0 ||
+      rhsBlockArg.getArgNumber() != 1 || outBlockArg.getArgNumber() != 2) {
+    return false;
+  }
+  return true;
+}
+
+/// `isMatmul` is a utility function that aims to indentify whether a
+/// linalg op is a matmul op.
+static bool isMatmul(linalg::LinalgOp linalgOp) {
```
newling, line 99: An FYI: I have something similar in my vectorization work. If you land this first I might refactor it into a AMDAIELinalgUtils.h API.

newling (commented):
Looks good! There will be a few minor tweaks to my vectorization PR needed but I think vectorization can live alongside this harmoniously. 

Thanks for the clarifying comments in the code.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
```diff
@@ -0,0 +1,271 @@
...
+  // Step 0. Test if the op itself is a linalg.matmul op.
+  if (isa<linalg::MatmulOp>(linalgOp)) return true;
+  // Step 1. Test the body of the generic to indeed be what we expect for a
+  //         matmul.
```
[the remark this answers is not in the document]
Abhishek-Varma: Sure, I can add that as another `if` following right after the `isa<linalg::MatmulOp>(op)` check.

Although I think that wouldn't be necessary though since this part of the code is currently invoked by `linalg.generic/matmul` instances only - since we might be extending this for other "matmul" variants and anyway have to let go of this check.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
Abhishek-Varma: That'd be great!

Abhishek-Varma:
Merged it in to unblock adding an update to this pass - I'll raise a PR on Monday.

I can take up follow-up review comments' addressing as part of that PR itself.

PR #180, opened 2024-02-23
Add element type specific target inner-dimension pack sizes

newling:


newling:
FYI @jsetoain

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -29,6 +28,152 @@ static LogicalResult setRootConfigForPadPipeline(func::FuncOp entryPointFn,
...
+//   `vector<4x2xi16>`  | `vector<2x8xi16>`  | `vector<4x8xi32>`
+//   `vector<2x8xi16>`  | `vector<8x8xi8>`   | `vector<2x8xi64>`
+//   `vector<4x8xi16>`  | `vector<8x4xi8>`   | `vector<4x4xi64>`
+//   `vector<2x4xi16>`  | `vector<4x8xi16>`  | `vector<2x8xi64>`
+//   `vector<4x4xi16>`  | `vector<4x4xi16>`  | `vector<4x4xi64>`
+//   `vector<4x2xi32>`  | `vector<2x4xi16>`  | `vector<4x4xi64>`
+//   `vector<4x8xbf16>` | `vector<8x4xbf16>` | `vector<4x4xf32>`
+//
+// An instruction size (m, n, k) is returned for each of the above cases. Cases
+// which are not covered by the table return a failure.
```
yzhang93, line 48: If I understand correctly, the combination of lhs, rhs and accumulator data types should be in the above look up table , otherwise it will return a failure. Currently, the data types in the test of simple-pack pipeline are all i32, then it will return failure? Why there's no instruction for such case?

yzhang93 (commented):
Mostly looks good. Maybe it's better to move this functionality to a separate file, e.g. Utils.cpp?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -29,6 +28,152 @@ static LogicalResult setRootConfigForPadPipeline(func::FuncOp entryPointFn,
...
+  // pack the three bitwidths into a single 32-bit unsigned integer. This
+  // facilitates the use of a switch statement below.
+  auto getTriplet = [](uint32_t a, uint32_t b, uint32_t c) -> uint32_t {
+    return a + (b << 8) + (c << 16);
+  };
+
+  // Sanity check: getTriplet values are computed at compile time.
+  static_assert(getTriplet(1, 2, 3) == 1 + 2 * 256 + 3 * 65536);
+
+  switch (getTriplet(nBitsLhs, nBitsRhs, nBitsAcc)) {
```
MaheshRavishankar, line 93: Seems like a rather complicated way of just representing a table. I think you can just define a 
```
llvm::DenseMap<std::array<int, 3>, std::array<int, 3>> = {
 {8, 4, 32}, {4, 8, 16},
 ...
};
```

MaheshRavishankar (changes requested):
Nice! But maybe we can make the logic a bit simpler. Prioritizing readability a bit?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
newling: As far as I know AIE does not have a matmul instruction for i32 x i32 -> i32 or i32 x i32 -> i64. The table only has the datatypes which AIE natively supports. 

This function does indeed return failure if the types are not in the table. In this case a default size (currently n=m=4 and k=8) is selected in getPackedSize.  

BTW the test in samples/pack_pipeline_funcIR.mlir is i8 x i8 -> i32 for which the (pre-PR) value of (m=4 n=k=8) is correct according to this table.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
newling: This doesn't work, they key must have an llvm tombstone (more complex to extend with tombstone IMO than current design). I could use a std::map. Let me know if you'd like std::map instead of the change I've pushed now, happy to change again.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETargetUtils.cpp
```diff
@@ -0,0 +1,104 @@
...
+  return a + (b << 8) + (c << 16);
+};
+
+// Map from (LHS bitwidth, RHS bitwidth, Accumulator bitwidth) to the AIE
+// instruction size (m, n, k) for the integer types with those bitwidths.
+const auto& getIntegerMatmulInstructionSizeMap() {
+  // Sanity check.
+  static_assert(getElementTypeKey(1, 2, 3) == 1 + 2 * 256 + 3 * 65536);
+
+  static DenseMap<uint32_t, std::array<uint32_t, 3>> matmulIntSizes{
```
MaheshRavishankar, line 29: Nice! but I think the key can be an `std::array` as well and it effectively hashes the elements (I think). If it diesnt work, this is fine as well.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETargetUtils.h
```diff
@@ -0,0 +1,56 @@
...
+#ifndef IREE_AMD_AIE_TARGET_AIETARGETUTILS_H_
+#define IREE_AMD_AIE_TARGET_AIETARGETUTILS_H_
+
+#include <array>
+
+#include "mlir/IR/Types.h"
+
+namespace mlir::iree_compiler::AMDAIE {
+
+struct AMDAIEUtils {
```
MaheshRavishankar, line 16: I generally dislike such struct that are just carrying `static` methods. Its effectively just a `namespace`. Just drop the `struct` and have these as util functions. It doesnt need to like here though. This is information the compiler needs. The `Target` is more for managing the HAL. Just move this to `Transforms` folder or create a new `Utils` folder.

MaheshRavishankar (changes requested):
Just a few nits, but this is good.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETargetUtils.h
newling: Done. 

I like using structs like this because you get compile time errors instead of link time errors if the function in the .h and .cpp are slightly different. But users of APIs should come before developers of APIs, so I'm happy to make the change!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETargetUtils.cpp
newling: It can't, because the keys of DenseMaps need to define a tombstone key...

Issue #177, opened 2024-02-20
Example with Multiple dispatches has error in the runtime


nirvedhmeshram:
Please refer to this log where the CI captures this issue
https://gist.github.com/nirvedhmeshram/218ce409d5e26d390d053cfc74419fa4

To reproduce, you can run these CI actions locally,
https://github.com/nod-ai/iree-amd-aie/blob/main/.github/workflows/ci.yml#L102-L110

To cause the issue you will need the change done by https://github.com/nod-ai/iree-amd-aie/pull/176

nirvedhmeshram:
cc @MaheshRavishankar

daveliddell:
The problem is that during compilation of the two executables, which happens in parallel, some of the intermediate files have the same name, so the compilations step on each other.

Fixed in pending PR https://github.com/nod-ai/iree-amd-aie/pull/181

The fix makes separate subdirectories for each executable.

PR #181, opened 2024-02-24
added subdirs for executables, fixed use after free

daveliddell:
Fix for https://github.com/nod-ai/iree-amd-aie/issues/177

1. Added subdirs for each parallel executable, so that their commonly named files don't step on each other
2. Fixed a use-after-free bug that caused the aie2xclbin command line to have garbage in it


nirvedhmeshram:
looks like the CI script is failing, could you please see how it can be modified. Also cc @spaceotter to make sure the clean up is done correctly by the CI script after this change.

spaceotter (approved):

daveliddell:
Also got out-of-band approval from Nirvedh.  Merging now

Issue #147, opened 2024-02-10
Use scf.forall for lowering instead of air.launch and air.herd


powderluv:

powderluv:
The path for compilation we should use is what is implemented as the "Pack" pipeline. That path is more involved to support through AIR, and we are trying to match a specific "input" that AIR would work with. What we need instead is to have a more direct path for lowering the pack pipeline that has known to work to some extent due to evaluation with Elliots team. The more direct part would involve the following (and potentially more things).

MaheshRavishankar:
@erwei-xilinx , @vivek-amd and a few of us had a discussion on some related aspects. I think we need to have a bit more concrete on the scheduling we setup using tiling + packing transformations. Probably worth discussing in our sync tomorrow. @jtuyls would be good to have you chip in as well.

erwei-xilinx:
@MaheshRavishankar , @yzhang93 , @stephenneuendorffer In our discussion earlier today, we identified issue with the current proposed IR driving the MLIR-AIR's scheduling workflows. I am putting together an alternative scheduling strategy, where the new frontend IR for AIR is attached here: https://gist.gitenterprise.xilinx.com/erweiw/5bf859b389ee009223e5dfab21bbef07 Comments are greatly appreciated.

vivekkhandelwal1:
> @erwei-xilinx , @vivek-amd and a few of us had a discussion on some related aspects. I think we need to have a bit more concrete on the scheduling we setup using tiling + packing transformations. Probably worth discussing in our sync tomorrow. @jtuyls would be good to have you chip in as well.

Hi @MaheshRavishankar, I'm not sure if I'm tagged here by mistake or intentionally? Since, I'm not aware of any such discussion.

yzhang93:
> > @erwei-xilinx , @vivek-amd and a few of us had a discussion on some related aspects. I think we need to have a bit more concrete on the scheduling we setup using tiling + packing transformations. Probably worth discussing in our sync tomorrow. @jtuyls would be good to have you chip in as well.
> 
> Hi @MaheshRavishankar, I'm not sure if I'm tagged here by mistake or intentionally? Since, I'm not aware of any such discussion.

Hi @vivekkhandelwal1 sorry, it's a mistake. I should be the one to tag.

MaheshRavishankar:
Notes from meeting to get to the "streaming K" workflow.

1) Tile using scf.forall for first level tiling (for now a 2x2 logical view of the AIE cores)
[...]

PR #121, opened 2024-02-01
Add enum options to target lowering tablegen

jtuyls:
@Abhishek-Varma Could you help review?

PR #184, opened 2024-02-26
Add pad-pack pipeline to support large K dimension

yzhang93:
This pipeline is originally designed to solve the large K dimension of GEMM. It also works for other cases, such as small K dimension and/or large M, N dimensions.

I have tested a superset of OPT GEMM sizes, with M/N/K as [8, 16, 32, 64, 2048, 8192] combinations. Most test cases have passed, and the failed cases are those when two of M/N/K dimensions are large, e.g. `8192 x 8192 x k`, `m x 8191 x 2048`, `m x 8192 x 8192`. Note `8192 x 2048 x 8192` compiles without problem. For the next goal, we should look into those large failure cases.


erwei-xilinx:
I had a quick test with (8k, 8k, 2048), and encountered this error: 
```
error: 'aiex.ipu.dma_memcpy_nd' op Stride 2 exceeds the [1:1M] range.
note: see current operation: "aiex.ipu.dma_memcpy_nd"(%arg1) <{id = 4 : i64, metadata = @airMemcpyId10, operandSegmentSizes = array<i32: 1, 0, 0, 0>, static_offsets = array<i64: 0, 0, 0, 0>, static_sizes = array<i64: 4, 4, 512, 64>, static_strides = array<i64: 64, 4194304, 8192>, x = 0 : i64, y = 0 : i64}> : (memref<2048x8192xi32>) -> ()
```
The BD generator in `-air-to-aie` needs to be able to tile the 2nd stride dimension to avoid this issue. Copying @jtuyls into this discussion.

`xilinx::airrt::createAIRRtToIpuPass()` seems to be very slow.

erwei-xilinx (approved):
LGTM.

yzhang93:
> I had a quick test with (8k, 8k, 2048), and encountered this error:
> 
> ```
> error: 'aiex.ipu.dma_memcpy_nd' op Stride 2 exceeds the [1:1M] range.
> note: see current operation: "aiex.ipu.dma_memcpy_nd"(%arg1) <{id = 4 : i64, metadata = @airMemcpyId10, operandSegmentSizes = array<i32: 1, 0, 0, 0>, static_offsets = array<i64: 0, 0, 0, 0>, static_sizes = array<i64: 4, 4, 512, 64>, static_strides = array<i64: 64, 4194304, 8192>, x = 0 : i64, y = 0 : i64}> : (memref<2048x8192xi32>) -> ()
> ```
> 
> The BD generator in `-airrt-to-ipu` needs to be able to tile the 2nd stride dimension to avoid the stride value being bigger than 1M. Copying @jtuyls into this discussion.
> 
> `xilinx::airrt::createAIRRtToIpuPass()` also seems to be very slow...

I've opened a new [issue](https://github.com/nod-ai/iree-amd-aie/pull/184), and let's move the discussion there.

MaheshRavishankar:
@yzhang93  I was able start with the transform dialect script you were using and adapt it to make it work in a way that represents the final sequence of transformations we probably want  See https://github.com/nod-ai/iree-amd-aie/issues/147#issuecomment-1965665796 

I can take a look at this PR, but can we adapt this to use that approach? We need to do one more pass to fuse the `fill` with `scf.for`. We can discuss this tomorrow. If you want me to still review this PR, please let me know.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -142,7 +144,9 @@ def AMDAIETileAndFuse :
     Option<"useSCFFor", "use-scf-for", "bool", /*default=*/"false",
       "Use scf.forall by default for the corresponding tiling level">,
     Option<"tilingLevel", "tiling-level", "int64_t", /*default=*/"-1",
-      "Use default tiling level used to retrieve the configuration from lowering_config">  
+      "Use default tiling level used to retrieve the configuration from lowering_config">,
+    Option<"targetOp", "target-op", "int64_t", /*default=*/"1",
+      "Target consumer op for tiling, 1 for linalg.matmul, 2 for weight linalg.copy and 3 for input linalg.copy">
```
Abhishek-Varma, line 149: Better to create an Enum or an equivalent to make this verbose.

Then even the API `consumerToSkip` (which uses this) would be more "readable".

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/tile_copy_using_scf_for.mlir
```diff
@@ -0,0 +1,61 @@
...
+// RUN: iree-opt --pass-pipeline='builtin.module(func.func(iree-amdaie-tile-and-fuse{use-scf-for tiling-level=1 target-op=3}))' --split-input-file %s | FileCheck %s --check-prefix=CHECK-LHS-COPY
+
+#map = affine_map<(d0) -> (d0 * 64)>
+#config = #iree_codegen.lowering_config<tile_sizes = [[64, 64], [0, 256], [256, 0]]>
+func.func @matmul_example_dispatch_0_matmul_2048x2048x2048_i32() {
+  %c0_i32 = arith.constant 0 : i32
+  %c0 = arith.constant 0 : index
+  %0 = hal.interface.binding.subspan set(0) binding(0) type(storage_buffer) alignment(64) offset(%c0) flags(ReadOnly) : !flow.dispatch.tensor<readonly:tensor<2048x2048xi32>>
+  %1 = hal.interface.binding.subspan set(0) binding(1) type(storage_buffer) alignment(64) offset(%c0) flags(ReadOnly) : !flow.dispatch.tensor<readonly:tensor<2048x2048xi32>>
+  %2 = hal.interface.binding.subspan set(0) binding(2) type(storage_buffer) alignment(64) offset(%c0) : !flow.dispatch.tensor<writeonly:tensor<2048x2048xi32>>
```
Abhishek-Varma, line 11: The test case can be made concise - for instance, you can get rid of all the `hal` constructs.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
```diff
@@ -128,9 +154,14 @@ void AMDAIETileAndFusePass::runOnOperation() {
...
   // If `consumerOp` has its own lowering config, we prefer using it.
   // Otherwise, fallback to find a lowering_config from other operations.
+  // When the consumerOp is a linalg.copy, the lowering config is originally
+  // added to the matmul op. To tile the linalg.copy op, take lowering configs
+  // from the its user matmul.
   SmallVector<int64_t> tileSizesVal;
   if (auto loweringConfig = getLoweringConfig(consumerOp)) {
     tileSizesVal = loweringConfig.getTileSizeVals(tilingLevel);
+  } else if (isa<linalg::CopyOp>(consumerOp) && getLoweringConfig(matmulOp)) {
+    tileSizesVal = getLoweringConfig(matmulOp).getTileSizeVals(tilingLevel);
```
Abhishek-Varma, line 164: I guess the following should work :-

```
if (isa<linalg::CopyOp>(consumerOp) && ( auto loweringConfig = getLoweringConfig(matmulOp)))
```

Just to reuse the loweringConfig and not re-invoke the same function.

Abhishek-Varma (changes requested):
Just adding review comments irrespective of where this is headed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
```diff
@@ -110,10 +125,21 @@ void AMDAIETileAndFusePass::runOnOperation() {
...
+    // pipeline.
+    if (isa<linalg::CopyOp>(op)) copyCnt++;
+    if (isa<linalg::MatmulOp>(op)) matmulOp = op;
+
     // Find the next consumer op if it does not have loops OR it is from
-    // the skip ops list which currently contains linalg.copy and tensor.unpack.
-    if (op.getLoopIteratorTypes().empty() || consumerToSkip(op))
+    // the skip ops list.
+    if (op.getLoopIteratorTypes().empty() ||
+        consumerToSkip(op, targetOp, copyCnt))
```
jtuyls, line 142: I am a bit concerned about the generality of the logic we're adding here. The logic here and above in `consumerToSkip` feels very specific to the `copy + matmul` pattern we see here during tiling?  Is there a way to generalize this more?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -192,6 +202,60 @@ static LogicalResult setRootConfigForPackPipeline(func::FuncOp entryPointFn,
...
+  auto initType = matmulOp.getDpsInitOperand(0)->get().getType();
+  auto initShape = llvm::cast<ShapedType>(initType).getShape();
+  auto tileM0 = std::min((int)initShape[0], 64);
+  auto tileN0 = std::min((int)initShape[1], 64);
+  auto tileM1 = std::max((int)tileM0 / 2, 1);
+  auto tileN1 = std::max((int)tileN0 / 2, 1);
+  auto lhsType = matmulOp.getDpsInputOperand(0)->get().getType();
+  auto lhsShape = llvm::cast<ShapedType>(lhsType).getShape();
+  auto tileK0 = std::min((int)lhsShape[1], 256);
+  auto tileK1 = std::min((int)lhsShape[1] / 8, 4);
```
jtuyls, line 222: There are a bunch of hardcoded values here, which I think might have to do with avoiding for example L2 buffer allocation overflow (`tileM0`, `tileM1`?) and using a 2x2 herd size (`/2`)? Also, I am not sure for example where `tileK0 == 256` exactly comes from? I think it would be useful to document the hardcoded values and any baked-in assumptions in the comments above, for others to understand where these values come from. And long term, I think we should read these values from somewhere else as they might change depending on data type, device, a uKernel to be targeted etc.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -29,10 +29,17 @@ static LogicalResult setRootConfigForPadPipeline(func::FuncOp entryPointFn,
       IREE::Codegen::DispatchLoweringPassPipeline::None);
 }
 
-static SmallVector<int64_t> getPackedSize(Operation *op) {
+static SmallVector<int64_t> getPackedSize(Operation *op, const int packLevel) {
```
jtuyls, line 32: Maybe `packLevel` can be made an `enum`?

jtuyls (changes requested):
@yzhang93 I think we should add some `pad-pack` testcases with the functional OPT shapes to https://github.com/nod-ai/iree-amd-aie/blob/main/build_tools/ci/run_matmul_test.sh.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
yzhang93: It doesn't work in this way.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
yzhang93: I tried to use llvm;:cl::values(clEnumValN(1, ...), clEnumValN(2, ...), clEnumValN(3, ...)) for the options, but got build errors. Let me know if you have a better working solution for this.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
yzhang93: I agree this is customized to the current IR, but I didn't find a better way to generalize this.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: Here I'd like to use `int` to keep consistent with the other parts we used for packLevel, padLevel and bufferizeLevel.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: Tile sizes (64, 64, 256) are chosen empirically for large GEMM sizes. And (tileM1, tileN1) = (tileM0/2, tileN0/2) are chosen to satisfy the 2x2 herd. This is a preliminary setting for tile sizes, we should come up a more sophisticated tiling strategy and a way to choose parameters in the future.

MaheshRavishankar (changes requested):
Lets talk about generalizing the fusion logic a bit here. Maybe we can start with something simple and reasonably future proof. Have other comments as well.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
Abhishek-Varma: I'd have based it on @jtuyls 's PR : https://github.com/nod-ai/iree-amd-aie/pull/121 + defining the Enums too.

Perhaps @jtuyls can also help if we know what build errors you're facing.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -284,6 +287,108 @@ void addPackBasedPassPipeline(OpPassManager &pm, TilingConfig &tilingConfig) {
...
+  tileOptions.useSCFFor = false;
+  tileOptions.targetOp = 1;
+  modulePassManager.addNestedPass<func::FuncOp>(
+      createAMDAIETileAndFusePass(tileOptions));
+  modulePassManager.addNestedPass<func::FuncOp>(createAMDAIECleanupPass());
+  modulePassManager.addPass(createCanonicalizerPass());
+  modulePassManager.addPass(createCSEPass());
+
+  // Fuse fill into forall loop
+  modulePassManager.addNestedPass<func::FuncOp>(
```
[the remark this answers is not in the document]
yzhang93: You are right. In this case, fill can be fused greedily.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
yzhang93: > Confused by why it is a consumer op?

Normally the consumer is the tiling target e.g., matmulOp. I think it is set as "consumer" is because we want to fuse greedily to include its producers.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
yzhang93: I've figured this out. Now it's more readable.

Abhishek-Varma (approved):
Thanks for addressing the changes Vivian! LGTM - approving from my end.

yzhang93:
> Lets talk about generalizing the fusion logic a bit here. Maybe we can start with something simple and reasonably future proof. Have other comments as well.

@MaheshRavishankar As we discussed, I've refactored the codes to have a separate tiling pass. Could you please review it again?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -130,15 +133,15 @@ void addSimplePackBasedPassPipeline(OpPassManager &pm,
...
 
-  AMDAIETileAndFuseOptions tileOptions;
+  AMDAIETileAndFuseOptions tileFuseOptions;
   AMDAIEPackAndTransposeOptions packOptions;
   AMDAIEBufferizeToAllocationOptions bufferizeOptions;
 
   // First level tiling using scf.forall
-  tileOptions.tilingLevel = 0;
-  tileOptions.useSCFFor = false;
+  tileFuseOptions.tilingLevel = 0;
```
MaheshRavishankar, line 141: Can be a foot gun to reuse the `tileFuseOptions`, etc.

Can we move them closer to the pass. Do something like
```
{
    AMDAIETileAndFuseOptions tileFuseOptions;
    ...
    modulePassManager.addNestedPass...(createAMDAIETileAndFusePass(tileFuseOptions));
}
```

Should be a simple change on top of this PR. Wont block on this, but please add that here or as a follow up.


```

File: compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp [hunk 9]
```diff
@@ -284,6 +287,93 @@ void addPackBasedPassPipeline(OpPassManager &pm, TilingConfig &tilingConfig) {
...
 }
 
+void addPadPackBasedPassPipeline(OpPassManager &pm,
+                                 TilingConfig &tilingConfig) {
+  auto &modulePassManager = pm.nest<ModuleOp>();
+  modulePassManager.addNestedPass<func::FuncOp>(createAMDAIECleanupPass());
+  pm.addPass(createCanonicalizerPass());
+  pm.addPass(createCSEPass());
+
+  AMDAIETileOptions tileOptions;
```
MaheshRavishankar, line 297: Same here, Lets not reuse the same options structs across different pass invocations.

MaheshRavishankar (approved):
This looks good. Thanks for the change. I have one more comment about the options struct handling. Please do that as part of this pass or as a follow up.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
yzhang93: Done.

jtuyls (approved):
LGTM

PR #188, opened 2024-02-28
Workaround for xclbinutil kernel name size limitation

jtuyls:
When the kernel name provided to the `aie2xclbin` tool exceeds 64 characters, the underlying `xclbinutil` tool will complain with an error like below and no xclbin can be generated. This issue came up accidentally when running the tests with larger shapes as the kernel name would increase, but this is a general issue which can occur at any time as different parts of the flow change/increase the length of the resulting ExportOp SymName here: https://github.com/nod-ai/iree-amd-aie/blob/25a1f58415744a1dbe63b29ef16e87480856798f/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp#L175
```
****** Bootgen v2023.2
  **** Build date : Feb 20 2024-08:50:06
    ** Copyright 1986-2022 Xilinx, Inc. All Rights Reserved.
    ** Copyright 2022-2023 Advanced Micro Devices, Inc. All Rights Reserved.


[INFO]   : Bootimage generated successfully

XRT Build Version: 2.17.0 (HEAD)
       Build Date: 2024-02-20 06:19:26
          Hash ID: 22150929f8c3c19b514a515168fce1797db7f28c
Creating a default 'in-memory' xclbin image.

Section: 'MEM_TOPOLOGY'(6) was successfully added.
Size   : 88 bytes
Format : JSON
File   : '/proj/rdi/staff/jornt/versal/iree-amd-aie/tests/matmul/module_matmul_8192x2048_8192xi32__dispatch_0_amdaie_xclbin_fb/mem_topology.json'

Section: 'AIE_PARTITION'(32) was successfully added.
Size   : 14848 bytes
Format : JSON
File   : '/proj/rdi/staff/jornt/versal/iree-amd-aie/tests/matmul/module_matmul_8192x2048_8192xi32__dispatch_0_amdaie_xclbin_fb/aie_partition.json'
Info: Embedded Metadata section is missing project.platform.device.core element, adding it.
ERROR: The m_name entry length (73), exceeds the allocated space (64).  Name: 'matmul_8192x2048_8192xi32__dispatch_0_matmul_8192x2048x8192_i32:MLIRAIEV1'
loc("/proj/rdi/staff/jornt/versal/iree-amd-aie/tests/matmul/module_matmul_8192x2048_8192xi32__dispatch_0_amdaie_xclbin_fb/module_matmul_8192x2048_8192xi32__dispatch_0_amdaie_xclbin_fb.aiecc.mlir":1:1): error: 'builtin.module' op failed to execute xclbinutil
loc("/proj/rdi/staff/jornt/versal/iree-amd-aie/tests/matmul/module_matmul_8192x2048_8192xi32__dispatch_0_amdaie_xclbin_fb/module_matmul_8192x2048_8192xi32__dispatch_0_amdaie_xclbin_fb.aiecc.mlir":1:1): error: 'builtin.module' op Failed to generate XCLBin
```
cc @yzhang93 @erwei-xilinx This might help avoid issues when adding tests to CI with the larger OPT shapes.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -172,7 +172,11 @@ LogicalResult AIETargetBackend::serializeExecutable(
   // generated by `aie2xclbin`
   SmallVector<std::string> entryPointNames;
   for (auto exportOp : variantOp.getExportOps()) {
-    entryPointNames.emplace_back(exportOp.getSymName());
+    // The xclbin kernel name, appended with instance name suffix (`:MLIRAIEV1`,
+    // 10 chars) is required by the xclbinutil to have a length smaller or equal
+    // to 64 right now. To have some additional wiggle room for suffix changes,
+    // we use the 48 first characters for the kernel name.
+    entryPointNames.emplace_back(exportOp.getSymName().substr(0, 48));
```
nirvedhmeshram, line 179: I am not sure this will work for all cases, as the way the naming scheme works, we have kernels like matmul_M1_N1_K1
and then another code be matmul_M1_N1_k2 so both might end up being matmul_ML_N1 with this way.

jtuyls: Right, but different dispatches have a separate xclbin? This is the internal kernel name within the xclbin.

jtuyls: For me, we can also generate a uuid or something from the kernel name, but I thought that might be obscure as it would not contain any information about the type of operation. = maybe overkill, as personally, I think the xclbinutil should allow kernel names with dynamic size and we could check with the (I believe) XRT team on whether this could be changed.

jtuyls: Alternatively, we could work with kernel keys like `KERNEL_0`, `KERNEL_1`, etc and then possibly keep a map around somewhere to the SymName

nirvedhmeshram: so the entrypointName is put in our vmfb artifact, L268 of this file, and then it is used to find the xclbin of a given dispatch at runtime, 
https://github.com/nod-ai/iree-amd-aie/blob/main/runtime/src/iree-amd-aie/driver/xrt/native_executable.cc#L174
That being said, currently we are supporting only one entry point function per executable (per xclbin), so you are right, this shortening should be fine due to that

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -172,7 +172,11 @@ LogicalResult AIETargetBackend::serializeExecutable(
   // generated by `aie2xclbin`
   SmallVector<std::string> entryPointNames;
   for (auto exportOp : variantOp.getExportOps()) {
-    entryPointNames.emplace_back(exportOp.getSymName());
+    // The xclbin kernel name, appended with instance name suffix (`:MLIRAIEV1`,
+    // 10 chars) is required by the xclbinutil to have a length smaller or equal
+    // to 64 right now. To have some additional wiggle room for suffix changes,
+    // we use the 48 first characters for the kernel name.
```
nirvedhmeshram, line 178:
```suggestion
    // we use the 48 first characters for the kernel name.
    // This is okay to do for now because we are only supporting single entry point.
```

nirvedhmeshram (approved):
LGTM.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
nirvedhmeshram, line 179: do we need to check if the original size is greater than 44 first?

nirvedhmeshram (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
jtuyls: This should be safe: 
```
Returns a substring [pos, pos + count). If the requested substring extends past the end of the string, i.e. the count is greater than size() - pos (e.g. if count == npos), the returned substring is [pos, [size()](https://en.cppreference.com/w/cpp/string/basic_string/size)).
```
from https://en.cppreference.com/w/cpp/string/basic_string/substr

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
jtuyls: Done

nirvedhmeshram (approved):

nirvedhmeshram:
@jtuyls you should have access to merge this, but if its still being figured out, please let me know if you want me to.

jtuyls:
> @jtuyls you should have access to merge this, but if its still being figured out, please let me know if you want me to.

Thanks. I got access now to merge this.

PR #190, opened 2024-02-29
Add pad-pack and loop peel based transform dialect

yzhang93:


jtuyls (approved):
Thanks, this is very useful to experiment with the peeling.

MaheshRavishankar (changes requested):
Can we turn the transform dialect scripts here into tests? Basically run this using `iree-opt --iree-transform-dialect-interpreter` and make it a check test?

We should do that for all the scripts here, but let's start with this one.


Related: I have been playing around with this as well,
1) I have been looking at non divisible shapes. After working through it a bit , the pad bufferization is a mess. It generates a fill and copy and is unmanageable. So in my draft I moved back to pack at both levels.
2) I will send that out as a draft on top of this PR.

yzhang93:
> Can we turn the transform dialect scripts here into tests? Basically run this using `iree-opt --iree-transform-dialect-interpreter` and make it a check test?
> 
> We should do that for all the scripts here, but let's start with this one.
> 
> Related: I have been playing around with this as well,
> 
> 1. I have been looking at non divisible shapes. After working through it a bit , the pad bufferization is a mess. It generates a fill and copy and is unmanageable. So in my draft I moved back to pack at both levels.
> 2. I will send that out as a draft on top of this PR.

- I added lit tests for transform dialect scripts. 
- I still want to keep `matmul_fill_spec_pad_pack_peel.mlir` in case we want to revisit it in the future. 
- I haven't pushed your transform dialect script, because I think it will need your hack codes in llvm-project? Probably add it when you have the upstream codes merged?

MaheshRavishankar (approved):
Sounds good. Thanks!

PR #191, opened 2024-03-01
Flags to print IR in mlir-aie passes (i.e. in aie2xclbin)

newling:
Can now use 

`iree-compile --aie2xclbin-print-ir-after-all ...`

etc. to print mlir-aie pass information (pipes to stderr or stdout, whichever mlir pipes to by default). 

I should add a test for this. 

4 flags:  
print ir before/after all
print ir module scope
disable threading.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
@@ -115,34 +115,50 @@ LogicalResult AIETargetBackend::serializeExecutable(
     const SerializationOptions &serOptions,
     IREE::HAL::ExecutableVariantOp variantOp, OpBuilder &executableBuilder) {
   ModuleOp moduleOp = variantOp.getInnerModule();
+
   auto basename =
       llvm::join_items("_", serOptions.dumpBaseName, variantOp.getName());
 
-  // If an intermediates path has been specified, assume it is common for all
-  // executables compiling in parallel, so create an executable-specific
-  // subdir to keep this executable's intermediates separate.
-  SmallString<128> workDir;
-  if (!serOptions.dumpIntermediatesPath.empty()) {
-    workDir = serOptions.dumpIntermediatesPath;
-    llvm::sys::path::append(workDir, basename);
-    (void)llvm::sys::fs::create_directories(workDir);
-  }
 
-  // No path for intermediates: make a temporary directory for this executable
-  // that is certain to be distinct from the dir of any other executable.
-  else {
-    auto err =
-        llvm::sys::fs::createUniqueDirectory(variantOp.getName(), workDir);
-    if (err) {
-      return moduleOp.emitOpError() << "failed to create temporary working "
-                                       "directory for xclbin generation: "
-                                    << err.message();
+  auto maybeWorkDir = [&]() -> FailureOr<SmallString<128>> {
+    // If a path for intermediates has been specified, assume it is common for
+    // all executables compiling in parallel, and so create an
+    // executable-specific subdir to keep this executable's intermediates
+    // separate.
+    if (!serOptions.dumpIntermediatesPath.empty()) {
+      SmallString<128> workDir{serOptions.dumpIntermediatesPath};
+      llvm::sys::path::append(workDir, basename);
+      auto ecode = llvm::sys::fs::create_directories(workDir);
+      if (ecode) {
+        return moduleOp.emitError()
+               << "failed to create working directory " << workDir
+               << ". Error message : " << ecode.message();
+      }
+      return workDir;
     }
-  }
 
-  std::string errorMessage;
+    // No path for intermediates: make a temporary directory for this
+    // executable that is certain to be distinct from the dir of any other
+    // executable.
+    SmallString<128> workDirFromScratch;
+    auto err = llvm::sys::fs::createUniqueDirectory(
+        /* prefix = */ variantOp.getName(), workDirFromScratch);
+
+    if (err)
+      return moduleOp.emitOpError()
+             << "failed to create working directory for xclbin generation: "
+             << err.message();
+
+    return workDirFromScratch;
+  }();
+
+  if (failed(maybeWorkDir)) return failure();
+  auto workDir = maybeWorkDir.value();
+
   SmallString<128> inputMlirPath(workDir);
   llvm::sys::path::append(inputMlirPath, basename + ".aiecc.mlir");
+
+  std::string errorMessage;
   {
     auto inputMlirOut = openOutputFile(inputMlirPath, &errorMessage);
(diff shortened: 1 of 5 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -239,11 +257,75 @@ LogicalResult AIETargetBackend::serializeExecutable(
     for (auto s : cmdArgs) llvm::dbgs() << s << " ";
     llvm::dbgs() << "\n";
   }
-  int result = llvm::sys::ExecuteAndWait(cmdArgs[0], cmdArgs, cmdEnv);
-  if (result != 0) {
-    return moduleOp.emitOpError(
-        "Failed to produce an XCLBin with external tool");
+
+  auto getTempFileName =
```
nirvedhmeshram, line 261: This is great to handle and output/error messaging here will definitely help our development velocity. I have following questions 
1. why cant the tool do this on its side and we need to do it here?
2. If we must do it here, can we somehow make it less visible by putting it in a different function, as when reading the code and understanding the flow, a reader might not necessarily care about these low level IO)handling details.

newling: Thanks Nirvedh, 

1. Which tool, do you mean aie2xclbin? I added the ability for aie2xclbin to print in https://github.com/Xilinx/mlir-aie/pull/1058, this current PR just propagates those options to iree-compile. 

2. Agree, I'll try and refactor.

newling: (a test would definitely make point 1 clearer)

nirvedhmeshram: About 1. My question is about handling the stdout/stderr here, I was thinking we pass the flags from here and the tool should by default give the output/error streams ?

newling: Ok I think I understand your question now. 

We call aie2xclbin through llvm::sys::ExecuteAndWait, which doesn't seem to have the option to pipe the stdout/stderr from the called program. It only has the ability to write them to files which is what this new code does:

```
  std::array<std::optional<StringRef>, 3> redirects{
      /* std::cin = */
      std::optional<StringRef>{}, stdoutFile.str(), stderrFile.str()};

  int result =
      llvm::sys::ExecuteAndWait(cmdArgs[0], cmdArgs, cmdEnv, redirects);
 ```
 
 
 I looked around for an alternative which doesn't involve temporary files, but couldn't find any. The approach I've implemented here is similar to that in IREE here: https://github.com/openxla/iree/blob/e6397cb853de7f3b28dcdc05987c40f0a4d1e9fe/compiler/plugins/target/CUDA/CUDATarget.cpp#L129

Comment (nirvedhmeshram) [reply compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp:261]: Cool, thanks for explaining!

File: compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h [hunk 2]
```diff
@@ -23,12 +25,25 @@ struct AMDAIEOptions {
...
+  bool useChess{false};
+
+  // Print IR after all MLIR passes run in aie2xclbin (to stderr).
+  bool aie2xclbinPrintIrAfterAll{false};
+
+  // Print IR before all MLIR passes run in aie2xclbin (to stderr).
+  bool aie2xclbinPrintIrBeforeAll{false};
+
+  // Disable theading in MLIR passes in aie2xclbin.
+  bool aie2xclbinDisableTheading{false};
```
MaheshRavishankar, line 41: This should probably be on by default if you want to print this.

MaheshRavishankar (changes requested):
I am not sure we want this enabled through `iree-compile` this way. We can always dump out the intermediates and run the `aie2xclbin` command manually. Also I am not 100% sure but this seems to be creating a temp file always?

newling:
> I am not sure we want this enabled through `iree-compile` this way. We can always dump out the intermediates and run the `aie2xclbin` command manually. Also I am not 100% sure but this seems to be creating a temp file always?

True.. but it seems like more work than necessary -- my goal in this PR is to make development faster, and it seems like we will wanting to inspect IR in aie2xcblin quite often.

>  Also I am not 100% sure but this seems to be creating a temp file always?

Do you mean, even when `--aie2xclbin-print-ir-after-all` and `--aie2xclbin-print-ir-before-all` and `--aie2xclbin-print-ir-module-scope` are all false? If so -- yes that's correct. 2 temp files. I can make this path conditional on these flags being true, so that the path before this PR is unchanged.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
newling: I know it's necessary if you're printing at the module scope


```
iree-opt --mlir-print-ir-after-all --mlir-print-ir-module-scope  test.mlir
error: IR print for module scope can't be setup on a pass-manager without disabling multi-threading first.
```

Do you thin there is something particular to this case which means we should disable threading even in the non module scope case? I wanted to make it as close the the upstream behaviour as possible.

newling:
I have removed all temporary file generation now.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
MaheshRavishankar: In general its better to avoid threading. Makes the output more determininstic and easy to diff and such. If we are printing logs, there is no real need to be efficient. We are trying to debug anyway

MaheshRavishankar (approved):
Looks good to me. Thanks. Please address comments.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
newling: If I change this to 

```
bool aie2xclbinDisableTheading{false};
```

then mlir-aie compilation will be single threaded by default, irrespective of whether we are printing anything or not. 
For this reason I think it is better to leave it as 

```
bool aie2xclbinDisableTheading{false};
```

and let the user toggle it, exactly the same way as the user would toggle the MLIR --disable-threading, --print-ir-before-all, etc flags (in any MLIR project). 
 
I could set aie2xclbinDisableTheading to true only if 

 1) user has not specified its value
 2) user has enabled printing
 
 ? I guess that's possible, but starts to get complicated...

PR #193, opened 2024-03-04
[AMD-AIE] Update lower-to-ukernel pass

Abhishek-Varma:
-- This commit adds various updates to `--iree-amdaie-lower-to-ukernel`
   pass.
-- The updates made are as follows :-
   a. Target Simple-Pack pipeline.
   b. Add fn_def to the generated iree_codegen.ukernel op.
   c. Add a pass option `path-to-ukernels` to help add the same to the fn_def as mentioned in point b.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
@@ -8,6 +8,8 @@
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenDialect.h"
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/UKernelOps.h"
 #include "iree/compiler/Codegen/Utils/Utils.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/Path.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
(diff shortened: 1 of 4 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -219,9 +224,19 @@ void addSimplePackBasedPassPipeline(OpPassManager &pm,
...
+    options.pathToUkernels = clPathToUkernels;
+    modulePassManager.addNestedPass<func::FuncOp>(
+        createAMDAIELowerToUKernelsPass(options));
+  }
   // Comprehensive bufferization
   addAMDAIEBufferizePasses(modulePassManager);
+  modulePassManager.addPass(createLowerUKernelOpsToCallsPass());
   modulePassManager.addPass(memref::createFoldMemRefAliasOpsPass());
+
+  modulePassManager.addPass(createLowerUKernelOpsToCallsPass());
```
yzhang93, line 239: What is this pass for? Why do you call this pass twice?

Abhishek-Varma: The pass is used for this conversion: `iree_codegen.ukernel @myfunc` -> `func.call @myfunc`.

The second invocation was a mistake - it didn't affect the flow because the first invocation changed all `iree_codegen.ukernel` to `func.call` - thanks for catching - I've updated removed it.

yzhang93: Looks good to me. I'll leave for Mahesh to take a look also.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
```diff
@@ -142,12 +154,85 @@ static bool isMatmul(linalg::LinalgOp linalgOp) {
   return indexingMaps == maps;
 }
 
+static bool matchMatmulIteratorAndIndexingMapForSimplePack(
```
MaheshRavishankar, line 157: Nit: Rename to something that matches better with the op itself.

MaheshRavishankar (approved):

PR #194, opened 2024-03-04
Tile the outer dimensions of linalg.generics which describe matmuls 

newling:
In my vectorization pipeline, I run this pass before vectorization. Without this pass, the vector.contract ops generated are not in a form aievec can digest (they have multiple non-singleton reduction dimensions). 

Pass description:
  
```  
      This pass transforms all linalg.generic operations with matmul-like
      inner-dimension semantics. It replaces all outer-dimensions with scf.for
      loops. For example, it replaces a generic operation that describes a
      batched matmul with an scf.for loop containing a linalg.generic that
      describes a lower rank non-batched matmul. In other words, it replaces
      the batch dimension in the linalg.generic with an scf.for loop.

      All outer dimensions are replaced with scf.for loops. The three
      inner-dimensions must describe a matmul: 2 parallel dimensions and 1
      reduction dimension, at correct indices. The pass does not transform
      transposed matmuls.

      The motivation for this pass is to enable a subsequent vectorization pass
      to generate vector.contract operations which map easily to the AIEVec
      dialect.
```


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileForMatmulLikeInnerLoop.cpp
```diff
@@ -0,0 +1,168 @@
...
+      auto firstNonSingleton = [&]() -> int {
+        for (auto i : llvm::enumerate(shape)) {
+          if (i.value() > 1) {
+            return i.index();
+          }
+        }
+        return nDims;
+      }();
+      return nDims - firstNonSingleton;
+    };
```
yzhang93, line 63: Put this to a separate function?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileForMatmulLikeInnerLoop.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    // Don't transform to scf.for loops unless there is at least one
+    // non-singleton loop to construct (this ensures termination of the pass).
+    {
+      auto numDimsFromFirstNonSingleton0 =
+          getNumDimsFromFirstNonSingleton(genericOp->getOperand(0));
+      auto numDimsFromFirstNonSingleton1 =
+          getNumDimsFromFirstNonSingleton(genericOp->getOperand(1));
+      auto numDimsFromFirstNonSingleton2 =
+          getNumDimsFromFirstNonSingleton(genericOp->getOperand(2));
```
yzhang93, line 73: Refactor the codes to iterate over all operands using getOperands()?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileForMatmulLikeInnerLoop.cpp
```diff
@@ -0,0 +1,168 @@
...
+            genericOp,
+            llvm::formatv(
+                "The number of non-singleton trailing dimensions is less than "
+                "3 for all operands ({0}, {1}, {2}), so there are no outer "
+                "dimensions to tile.",
+                numDimsFromFirstNonSingleton0, numDimsFromFirstNonSingleton1,
+                numDimsFromFirstNonSingleton2));
+      }
+    }
+
```
yzhang93, line 88: If you can refactor the above `numDimsFromFirstNonSingleton` to a SmallVector, I think you can use llvm::any_of for the condition.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileForMatmulLikeInnerLoop.cpp
```diff
@@ -0,0 +1,168 @@
...
+    if (map1.getResult(nResults1 - 1) != map2.getResult(nResults2 - 1))
+      return rewriter.notifyMatchFailure(
+          genericOp,
+          "unable to locate 'N' in affine maps [(M,K) x (K,N) -> (M,N)]");
+    if (map0.getResult(nResults0 - 1) != map1.getResult(nResults1 - 2))
+      return rewriter.notifyMatchFailure(
+          genericOp,
+          "unable to locate 'K' in affine maps [(M,K) x (K,N) -> (M,N)]");
+
+    // Ran the gauntlet, now perform the tiling of the outer-dimensions.
```
yzhang93, line 137: The file name `AMDAIETileForMatmulLikeInnerLoop` is a little confusing and seems contradicting this comment?

yzhang93 (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileForMatmulLikeInnerLoop.cpp
newling: Do you mean the pass name `--iree-amdaie-tile-for-matmul-like-inner-loop` could be more descriptive? 


The pass tiles a linalg.generic so that it becomes an scf.for loop with a simple matmul inside. I've gone with matmul "like" in the pass name because I'm not checking that the reduction is actually a multiply accumulate. I could add this check, but it seems like unneccessary complexity. 

Some options:

--iree-amdaie-decompose-to-loop-with-simple-matmul 
--iree-amdaie-tile-outer-dims-of-matmul-like-generic
--iree-amdaie-tile-for-matmul-like-inner-loop
--iree-amdaie-insert-loops-for-vectorization

or combinations of the above :-)

newling: > The file name AMDAIETileForMatmulLikeInnerLoop is a little confusing and seems contradicting this comment?

The tiling is trivial: Use tile size of 1 for all but dimensions except the 3 inner-most dimensions (which don't get tiled, i.e. get tiles with '0'). So the outer dimensions get tiled.

yzhang93: I think `--iree-amdaie-insert-loops-for-vectorization` maybe a better choice.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ iree_lit_test_suite(
     "tile_and_fuse_using_scf_for.mlir"
     "tile_and_fuse_using_scf_forall.mlir"
     "tile_copy_using_scf_for.mlir"
+    "insert_loops_for_vectorization.mlir"
```
MaheshRavishankar, line 29: Please keep ordered.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -0,0 +1,172 @@
...
+    tileSizes[numIterators - 1] = 0;
+    auto outerOpts = linalg::LinalgTilingOptions().setTileSizes(tileSizes);
+    auto outerTiled = linalg::tileLinalgOp(rewriter, genericOp, outerOpts);
+    rewriter.replaceOp(genericOp, outerTiled.value().loops[0]->getResult(0));
+    return success();
+  }
+
+ private:
+  // Example: (1,1,4,1,2,3,4) has 5 dimensions from the first non-singleton.
+  static int getNumDimsFromFirstNonSingleton(Value v) {
```
MaheshRavishankar, line 132: Use of singleton is confusing. I am not sure what it is supposed to refer to.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -0,0 +1,172 @@
...
+    registry.insert<linalg::LinalgDialect, scf::SCFDialect>();
+  }
+  void runOnOperation() override;
+};
+
+void AMDAIEInsertLoopsForVectorizationPass::runOnOperation() {
+  MLIRContext *context = &getContext();
+  auto operation = getOperation();
+  IRRewriter rewriter(context);
+  RewritePatternSet patterns(operation.getContext());
```
MaheshRavishankar, line 160: For these it is better to not use rewrite patterns. They are just one-off transformations. Rewriter is useful for fixed point, which this isnt. Just walk the program and find the operation you want to tile, and tile it once. That is much easier.

MaheshRavishankar (changes requested):
THanks James! Mostly looks good! Few nits.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
newling: Changed to getNumInnerDims

newling:
I've changed from a Pattern to a simple Pass, although logic remains identical. Reduced line count, mostly because of the removal of the diagnostics with notifyMatchFailure.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -47,6 +47,33 @@ def AMDAIEFuseFillIntoForall :
   let constructor = "mlir::iree_compiler::AMDAIE::createAMDAIEFuseFillIntoForallPass()";
 }
 
+def AMDAIEInsertLoopsForVectorization :
+    Pass<"iree-amdaie-insert-loops-for-vectorization", "func::FuncOp"> {
+
+    let summary = "Transforms linalg.generic operations with matmul-like semantics.";
```
yzhang93, line 53: Add more descriptions for summary. "Transforms .... to replace outer-dimension with scf.for loops.

yzhang93 (approved):
Looks good to me. Just another minor comment.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -0,0 +1,134 @@
...
+    rewriter.replaceOp(genericOp, loops[0]->getResult(0));
+  }
+
+  void runOnOperation() final {
+    MLIRContext *context = &getContext();
+    func::FuncOp operation = getOperation();
+
+    IRRewriter rewriter(context);
+    operation->walk([&](linalg::GenericOp genericOp) -> WalkResult {
+      auto iteratorTypes = genericOp.getIteratorTypesArray();
```
MaheshRavishankar, line 69: Nit: Move the lambda implementation into a function for better readabilty. You can make that return a `LogicalResult`. There is an automatic conversion from `LogicalResult` -> `WalkResult`.

MaheshRavishankar (approved):

PR #198, opened 2024-03-05
mlir-aie: Bump wheel and commit hash

newling:


nirvedhmeshram:
Hi @newling are you planning on pushing on this, or do you want me to take a stab at the bump, btw when I was trying to update mlir-aie, I also had to bump mlir-air as they have dependencies

newling:
Hi @nirvedhmeshram I was meaning to get back to this. I am going to try again (with air too) and see what fails, then let's chat later.

nirvedhmeshram:
> Hi @nirvedhmeshram I was meaning to get back to this. I am going to try again (with air too) and see what fails, then let's chat later.

Actually looks like Abhishek already did a bump so its just the wheel that needs to be updated now, although since the CI is still running, looks like the old wheel is okay for now.

newling:
> > Hi @nirvedhmeshram I was meaning to get back to this. I am going to try again (with air too) and see what fails, then let's chat later.
> 
> Actually looks like Abhishek already did a bump so its just the mlir-aie wheel that needs to be updated now, although since the CI is still running, looks like the old wheel is okay for now.

Ah yes, I just noticed that too. I'll try and update mlir-aie again anyway.

PR #199, opened 2024-03-05
Disable conv2d test

newling:
This was just a POC transform dialect script being tested, so no concern here of a bug somewhere important being ignored.


nirvedhmeshram (approved):

yzhang93:
@newling Sorry for the late comment. You can keep the transform dialect test using iree-opt. The example is https://github.com/nod-ai/iree-amd-aie/blob/main/tests/transform_dialect/matmul_fill_spec_simple_pack.mlir#L1.

PR #202, opened 2024-03-07
Add air.dma_memcpy_nd canonicalization pass

nirvedhmeshram:
This pass adds simplifying canonicalizations that we expect to be legalized into aie dma operations using subsequent passes.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/canonicalize_dma.mlir
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/canonicalize_dma.mlir
@@ -1,0 +1,18 @@
+  // RUN: iree-opt --pass-pipeline="builtin.module(iree-amdaie-canonicalize-dma)" %s | FileCheck %s
+  // CHECK-LABEL: @canonicalize_dma
+  func.func @canonicalize_dma() {
+    %c16 = arith.constant 16 : index
+    %c32 = arith.constant 32 : index
+    %c64 = arith.constant 64 : index
+    %c128 = arith.constant 128 : index
+    %c8 = arith.constant 8 : index
+    %c4 = arith.constant 4 : index
+    %c2 = arith.constant 2 : index
+    %c1 = arith.constant 1 : index
+    %c0 = arith.constant 0 : index
+    %alloc = memref.alloc() : memref<1x1x8x16xi32, 1>
+    %alloc_0 = memref.alloc() : memref<1x1x2x2x4x8xi32, 2>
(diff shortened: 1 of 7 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/canonicalize_dma.mlir
```diff
@@ -0,0 +1,18 @@
...
+    %c64 = arith.constant 64 : index
+    %c128 = arith.constant 128 : index
+    %c8 = arith.constant 8 : index
+    %c4 = arith.constant 4 : index
+    %c2 = arith.constant 2 : index
+    %c1 = arith.constant 1 : index
+    %c0 = arith.constant 0 : index
+    %alloc = memref.alloc() : memref<1x1x8x16xi32, 1>
+    %alloc_0 = memref.alloc() : memref<1x1x2x2x4x8xi32, 2>
+    // CHECK: air.dma_memcpy_nd (%alloc_0[%c0] [%c128] [%c1], %alloc[%c0, %c0, %c0] [%c2, %c8, %c8] [%c8, %c16, %c1]) : (memref<1x1x2x2x4x8xi32, 2>, memref<1x1x8x16xi32, 1>)
```
erwei-xilinx, line 15: I wonder if there is a benefit to lower to `%alloc_0[][][]` rather than `%alloc_0[%c0] [%c128] [%c1]`? Because the latter version will try to lower to a BD with one wrap of 128, and subsequently fail if the memref has more elements than the wrap limit by AIE DMA hw. Instead, I think the former won't, because it will not try to use any wrap.

nirvedhmeshram: Thinking more about this, I could add a "dropTrivialSingleDim" pattern to get to `%alloc_0[][][]` but just want to understand what that implies happens lower down the stack and if this is the right way to represent it.

nirvedhmeshram: cool, makes sense! So since there is pattern to arrive at the empty lists in AIR, I think I dont need to add it here? This way at this level its more explicit which would be easier to read I think.

erwei-xilinx: Yeah that makes sense to me.

erwei-xilinx (approved):
LGTM!

PR #205, opened 2024-03-08
Make pad-pack as the default pipeline and decouple its lowering from other pipelines

nirvedhmeshram:
This is in preparation for changing the lowering flow without the maintenance burden of supporting it for non-default pipelines.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
@@ -37,7 +37,7 @@ static llvm::cl::opt<AIEPassPipeline> clUsePipeline(
         clEnumValN(AIEPassPipeline::PadPackPipeline, "pad-pack",
                    "Use the IREE lowering to AIR dialect through "
                    "pad and pack operations")),
-    llvm::cl::init(AIEPassPipeline::SimplePackPipeline));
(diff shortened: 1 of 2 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -549,6 +549,106 @@ void addMLIRAIRAIELoweringPasses(OpPassManager &passManager) {
   passManager.addPass(createCanonicalizerPass());
 }
 
+void addMLIRAIRAIELegacyLoweringPasses(OpPassManager &passManager) {
```
yzhang93, line 552: What is changed in `addMLIRAIRAIELegacyLoweringPasses` compared to `addMLIRAIRAIELoweringPasses`?

MaheshRavishankar (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
nirvedhmeshram: No changes yet, this just de-couples the default pipeline (which will be pad-pack after this commit) from the rest of them, so that we can start changing the lowering in `addMLIRAIRAIELoweringPasses` without worrying about if it will break various pipelines.

PR #206, opened 2024-03-08
Add transpose support for pad-pack pipeline

nirvedhmeshram:
Since pad-pack is going to be our default pipeline for a while. We should have the transpose support that we have for simple-pack in this pipeline. Currently this is failing here
https://github.com/Xilinx/mlir-air/blob/main/mlir/lib/Transform/AIRMiscPasses.cpp#L592
The reason is, that function is set up to deal with single symbol affine maps like `()[s0] -> (s0 * 4)` but with this lowering we have affine maps like `()[s0, s1] -> (s0 + s1 * 4)` . @erwei-xilinx could you please take a look at running the test added in this Pr and let us know, if this is an IREE issue that such affine maps are made or is this something we need to handle at the AIR level?


erwei-xilinx:
Thanks for the bug report. I can reproduce the issue. Let me take a look and get back to you.

erwei-xilinx:
The crash is due to AIR's broadcast detection incorrectly detecting this dma op as being broadcastable. I will put together a PR in AIR to fix that. 

Based on AIR's current broadcast pattern detection method, this is supposed to be considered as not statically broadcastable, because the operation `%6 = affine.apply #map5()[%arg16, %arg12]` takes as operands both air.herd's induction variable, which is static, and air.launch's induction variable, which is assumed to be dynamic. So if we were to have a broadcastable pattern in AIE, we may need to consider either removing the affine map's dependence to launch induction variables, or rethink on how we do broadcast. Would love to know how you think about this.

erwei-xilinx:
@nirvedhmeshram Can you check if this PR fixes the issue? https://github.com/Xilinx/mlir-air/pull/485

MaheshRavishankar:
> The crash is due to AIR's broadcast detection incorrectly detecting this dma op as being broadcastable. I will put together a PR in AIR to fix that.
> 
> Based on AIR's current broadcast pattern detection method, this is supposed to be considered as not statically broadcastable, because the operation `%6 = affine.apply #map5()[%arg16, %arg12]` takes as operands both air.herd's induction variable, which is static, and air.launch's induction variable, which is assumed to be dynamic. So if we were to have a broadcastable pattern in AIE, we may need to consider either removing the affine map's dependence to launch induction variables, or rethink on how we do broadcast detection. Would love to know how you think about this.

Yeah this like a symptom of the `scf.forall` -> `air.launch/air.herd` conversion. In a general setting, the constraints that you mentioned make sense. But here that abstraction breaks down cause we really have/should have one herd.

```
scf.forall (%iv0, %iv1) = (%lb0, %lb1) to (%ub0, %ub1) step (%step0, %step1) {
  ...
  scf.forall (%iv2, %iv3) = (%lb2, %lb3) to (%ub2, %ub3) step (%step2, %step3) {
    ... 
  } {#gpu.thready, #gpu.threadx}
  ...
} {#gpu.blocky, #gpu.blockx}

```

Now for sake of argument lets say we put this entire thing in one herd (I understand that `air.herd` is not meant to "handle" L2 objects, but that aside). Lets assume that we use a logical block of size (`%blocksize_x, %blocksize_y`)

```
air.herd (%core_id_y, %core_id_x) in (%num_cores_y, %num_cores_y) {
  %num_blocks_y = ceilDiv(%num_cores_y, %block_size_y)
  %num_blocks_x = ceilDiv(%num_cores_x, %block_size_x)
  
  %block_id_y = %core_id_y / %block_size_y
  %block_id_x = %core_id_x / %block_size_x

  %local_id_y = %block_id_y mod %block_size_y
  %local_id_x = %block_id_x mod %block_size_x

  scf.for (%iv0, %iv1) = (%lb0 + %block_id_y * %step_0, %lb1 + %block_id_x * %step_1) to (%ub0, %ub1) step(%num_blocks_y * %step0, %num_blocks_x * %step1) {
  ...
    scf.for (%iv2, %iv3) = (%lb2 + %local_id_y * %step_2, %lb3 + %local_id_x * %step_3) to (%ub2, %ub3) step(%block_sizes_y * %step2, %block_size_x * %step3) {
     ...
   }
   ...
}
```

To find broadcast patterns, you look at the indices accessed. If any of the accesses does not contain `%core_id_y` or `%core_id_x` then it is a broadcast.

nirvedhmeshram:
@erwei-xilinx It didnt solve the issue as in the e2e testing, it is crashing in the aie tool, even for large matmul with that change, the error is 
```
Compiling /proj/xcohdstaff6/nmeshram/module_generic_matmul_transpose_static_dispatch_0_amdaie_xclbin_fb/module_generic_matmul_transpose_static_dispatch_0_amdaie_xclbin_fb.aiecc.mlir
Created temporary directory /proj/xcohdstaff6/nmeshram/module_generic_matmul_transpose_static_dispatch_0_amdaie_xclbin_fb
loc("/proj/xcohdstaff6/nmeshram/module_generic_matmul_transpose_static_dispatch_0_amdaie_xclbin_fb/module_generic_matmul_transpose_static_dispatch_0_amdaie_xclbin_fb.aiecc.mlir":382:12): error: Dialect `hal' not found for custom op 'hal.interface.binding.subspan' 
/proj/xcohdstaff6/nmeshram/configured_module_generic_matmul_transpose_static_dispatch_0.mlir:8:4: error: 'builtin.module' op Failed to produce an XCLBin with external tool.
    builtin.module {
```

erwei-xilinx:
> To find broadcast patterns, you look at the indices accessed. If any of the accesses does not contain %core_id_y or %core_id_x then it is a broadcast.

@MaheshRavishankar Thanks for your comment. In MLIR-AIR we currently do two-stage broadcast lowering consisting of broadcast pattern detection and indices (offsets) specialization. The method you described is also how MLIR-AIR currently detects broadcast. The issue that we are currently running into is after the patten gets identified as broadcastable, then it fails in the indices specialization pass, where AIR is trying to convert the variable offsets into constants, and fails because the variable, i.e. result of the `affine.apply`, is not static.

I think your comment makes sense, that fundamentally this pattern should be broadcastable. Perhaps if we just try detecting the broadcast, but gracefully failing the indices specialization, then it might work... Let me try something and keep you both posted in this thread.

erwei-xilinx:
> It didnt solve the issue as in the e2e testing, it is crashing in the aie tool, for the transpose matmul with that change, the error is

I see. Let take a look at this downstream problem and keep you posted.

erwei-xilinx:
@nirvedhmeshram The IR generated from matmul transpose e2e tests seem to be missing `linalg(or memref).copy` ops between L3 and L2, and instead were generating the following `linalg.generic` ops:

```
            linalg.generic {indexing_maps = [#map, #map], iterator_types = ["parallel", "parallel"]} ins(%0 : memref<8x16xi32>) outs(%alloc : memref<8x16xi32, 1 : i32>) {
            ^bb0(%in: i32, %out: i32):
              linalg.yield %in : i32
            }
```
What seems to be happening now is that AIR doesn't recognize them as L3-L2 data movements, and instead ignores them, leading to designs with no Shim DMA activity.

Would it be possible to modify the two tests so that they generate copy ops? From `simple_pack_pipeline_e2e.mlir` I see this issue with `generic_matmul_transpose_static` and `matmul_transpose_b_static`, but not with `matmul_small` or `matmul_large`.

Here is an IR dump after `AMDAIEBridgeToAIRPass` for `generic_matmul_transpose_static`: https://gist.github.com/erwei-xilinx/fcfc6b43e2cf35b9499adf03456886e2

nirvedhmeshram:
Thanks @erwei-xilinx , that's a good catch. I will take a look at why we don't have the copies there.

nirvedhmeshram:
@erwei-xilinx can you please take a look again with all the updates
locally I am seeing
```
/proj/xcohdstaff6/nmeshram/configured_module_generic_matmul_transpose_static_dispatch_0.mlir:8:4: error: 'air.wait_all' op operand #0 must be variadic of async token type, but got '!airrt.event'
```

erwei-xilinx:
@nirvedhmeshram Sure let me look into this. Could you please provide an IR dump which triggers this error?

nirvedhmeshram:
@erwei-xilinx here it is https://gist.github.com/nirvedhmeshram/4de137094af4b9b28d66534a04c86c41

erwei-xilinx:
@nirvedhmeshram I think the AIR pass `air-specialize-channel-wrap-and-stride`, whose goal is to fold parent scf.for loops into extra `air.channel` put/get op wrap-and-stride dimensions, is failing to handle the case where scf.for induction variable goes to channel op offsets via a 2-operand affine.apply. Let me work on that.

erwei-xilinx:
@nirvedhmeshram Following up on this thread, I put together an RFC for a new feature in AIR to process the folding of scf.for loop into air.channel op's wraps and strides, where a multi-operand affine map is involved: https://gist.github.com/erwei-xilinx/84f3f4fd4ad79deb5081aae779f35c91

nirvedhmeshram:
@erwei-xilinx Thanks for all the help with this, looks like the CI is passing.
@MaheshRavishankar , @yzhang93 or @Abhishek-Varma could one of you please review this.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -99,10 +99,10 @@ void addPadBasedPassPipeline(OpPassManager &pm, TilingConfig &tilingConfig) {
     }
     if (i == 2) {
       modulePassManager.addNestedPass<func::FuncOp>(createAMDAIECleanupPass());
-      pm.addPass(createCanonicalizerPass());
-      pm.addPass(createCSEPass());
     }
     {
+      modulePassManager.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+      modulePassManager.addNestedPass<func::FuncOp>(createCSEPass());
```
Abhishek-Varma, line 105: If they aren't being guarded then they can just be pulled out of the `{....}` block.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -324,8 +324,8 @@ void addPadPackBasedPassPipeline(OpPassManager &pm,
                                  TilingConfig &tilingConfig) {
   auto &modulePassManager = pm.nest<ModuleOp>();
   modulePassManager.addNestedPass<func::FuncOp>(createAMDAIECleanupPass());
-  pm.addPass(createCanonicalizerPass());
-  pm.addPass(createCSEPass());
+  modulePassManager.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+  modulePassManager.addNestedPass<func::FuncOp>(createCSEPass());
```
Abhishek-Varma, line 328: This is for my understanding - why would we want to invoke some `--canonicalize --cse` like a `nestedPass` while the others can be as it is?

tests/samples/pad_pack_pipeline_e2e.mlir
```diff
@@ -43,3 +43,56 @@ func.func @matmul_large(%lhs: tensor<2048x2048xi32>, %rhs: tensor<2048x2048xi32>
                     outs(%fill: tensor<2048x2048xi32>) -> tensor<2048x2048xi32>
   return %res : tensor<2048x2048xi32>
 }
+
+// -----
+
+// This test demonstrates Pack pipeline based e2e lowering for a linalg.generic implementing
```
Abhishek-Varma, line 49: nit : `demonstrates Pack pipeline` -> `demonstrates Pad-Pack pipeline`.

tests/samples/pad_pack_pipeline_e2e.mlir
```diff
@@ -43,3 +43,56 @@ func.func @matmul_large(%lhs: tensor<2048x2048xi32>, %rhs: tensor<2048x2048xi32>
...
+    %19 = arith.muli %in, %in_0 : i32
+    %20 = arith.addi %out, %19 : i32
+    linalg.yield %20 : i32
+  } -> tensor<8x32xi32>
+  return %matmul_transpose : tensor<8x32xi32>
+}
+
+// -----
+
+// This test demonstrates Pack pipeline based e2e lowering for a linalg.matmul_transpose_b.
```
Abhishek-Varma, line 78: Likewise.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoForLoop.cpp
```diff
@@ -66,8 +66,8 @@ void AMDAIEFusePackIntoForLoopPass::runOnOperation() {
   }
 
   if (sliceOps.empty()) {
-    funcOp->emitOpError("There is no extract tensor slice.");
-    return signalPassFailure();
+    // Pack ops are already fused.
+    return;
```
Abhishek-Varma, line 70: Is this occurring only in case of matmul-transpose-like ops and we wouldn't want to "error out"?

I'm trying to understand the difference between the two and the "why" behind it.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPad.cpp
```diff
@@ -186,13 +186,13 @@ class AMDAIEPadPass : public impl::AMDAIEPadBase<AMDAIEPadPass> {
...
-  linalg::MatmulOp matmulOp;
-  funcOp->walk<WalkOrder::PostOrder, ReverseIterator>([&](TilingInterface op) {
-    // Find the next matmul op if it does not have loops.
-    if (op.getLoopIteratorTypes().empty() || !isa<linalg::MatmulOp>(op))
-      return WalkResult::advance();
-    matmulOp = cast<linalg::MatmulOp>(op);
-    return WalkResult::interrupt();
+  linalg::LinalgOp matmulOp;
+  funcOp->walk([&](linalg::LinalgOp op) {
+    if (linalg::isaContractionOpInterface(op)) {
```
Abhishek-Varma, line 191: Nice!

Abhishek-Varma (changes requested):
Thanks for adding this! A few comments.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoForLoop.cpp
yzhang93: This makes sense. We don't do fusion if there is no sliceOp or pack ops are already fused. So that it won't throw an error and break the passes. But could you add a LLVM_DEBUG message?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
yzhang93: Actually do we really need to run cleanup, canonicalization, cse at the beginning? I don't think these are useful, we'll need to check and clean up these passes in the following patch.

yzhang93 (approved):
Mostly LGTM. Approve now, please solve all the comments.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -18,7 +18,8 @@
 namespace mlir::iree_compiler::AMDAIE {
 
 static LogicalResult setRootConfigForPadPipeline(func::FuncOp entryPointFn,
-                                                 linalg::LinalgOp linalgOp) {
+                                                 linalg::LinalgOp linalgOp,
+                                                 AIEConfig cfg) {
```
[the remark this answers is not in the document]
nirvedhmeshram: There were some pipelines (e.g padpack) that have this config as an argument (even though nothing is done with it) while others dont. I had to make this consistent so that they can use the same functions. I choose to go in the direction of making all of them use them with the reasoning that - yes currently the pipelines are not doing anything smart with the config, but if that changes, we have the plumbing for it.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
nirvedhmeshram: So about the nesting - without the nesting these cleanup passes were not called along with the nested passes and were all called in the end. So yes techincally all of them should be nested, I will update that. 
Regarding why we need it - The tiling leaves behind some dead ops from the previous state. So I ended up accidently applying padding on the untiled dead matmul rather than the new tiled matmul. So its easier to have a claen state before starting a pass.

nirvedhmeshram: Just looking closely now, I think adding the nested pass is not necessary but we do have to apply the passes to modulePassManager and not just the pass manager. Also, looking at the IR yes some of these clenup passes are unnecessary so I will remove them.

yzhang93: Nice, thanks for cleaning up the passes!

Abhishek-Varma (approved):
LGTM!

PR #208, opened 2024-03-09
[Numerical issue reproducer] Run matmul test in loop. 

newling:
I get intermittent failures on main with the matmul test, for certain shapes and pipelines. 

In this PR I run 1000 times and count the number of failures. This is what I get locally (i.e. not in CI):

image

In the comment below you can see the failure rate in CI is much lower, but still fails sometimes.

### DIFF
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -42,7 +42,7 @@ if [ "$#" -lt 2 ] || [ "$#" -gt 6 ]; then
             "\n     6) <vitis-install-dir>        (optional)" \
             "\n Example, dependent on environment variables:" \
             "\n     ./run_matmul_test.sh  " \
-            "results_dir_tmp  \$MLIR_AIE_INSTALL_DIR  \$IREE_INSTALL_DIR  " \
+            "results_dir_tmp  \$IREE_BUILD_OR_INSTALL_DIR  \$MLIR_AIE_INSTALL_DIR  " \
             "\$PEANO_INSTALL_DIR  /opt/xilinx/xrt  \$VITIS_INSTALL_PATH"
     exit 1
 fi
@@ -287,13 +287,13 @@ function run_matmul_test() {
 
   echo "Running command: ${COMMAND}"
 
-  # Run the command n_runs times. Cound the number of times the return code is not 0. Do NOT fail automatocally when 0 is returned (see 'set -euox pipefail. 
-  # 1) initialize the count of number of fails to 0:
-  n_fails=0
+  # Run the command n_runs times.
+  # Count the number of times the return code is not 0.
+  # Don't fail automatocally when 0 is returned (see 'set -euox pipefail')
 
-  #2) Run the command n_runs times, incrementing n_fails when the return code is not 0. Ensure that we don't fail the script if eval returns a non-zero status.
+  n_fails=0
   for i in $(seq 1 $n_runs); do
-    # Disable exit on error:
+    # Disable exit on error temporarily:
     set +e
     eval "${COMMAND}"
     command_status=$?
@@ -358,7 +358,7 @@ run_matmul_test \
     --lhs_rhs_type "i32" \
     --acc_type "i32" \
     --shapes "large" \
-    --n_runs 1000 \
+    --n_runs 2000 \
     --target_backend "amd-aie" \
     --device "xrt" \
     --peano_install_path "${PEANO}" \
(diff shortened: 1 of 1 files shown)


newling:
So in CI I get 2 / 1000 fails:

image

The failure pattern is different to what I get locally. In CI it is: 

image

Whereas locally (where I got 536/1000 failures) it is more localized:

image

(although the failure pattern is different across runs).

nirvedhmeshram:
Just in the interest of ruling things out, can you do with `llvm-cpu` target instead of `amd-aie` and `local-sync` instead of `xrt` device and verify there are no failures? Interesting bug, thanks for catching it! I wonder if there is still some intermediate file dependency we havent caught yet. The reason I suspect that is, I think your setup is in a slow NFS file system while the CI is not which could explain the rareness on CI.

newling:
> Just in the interest of ruling things out, can you do with llvm-cpu target instead of amd-aie and local-sync instead of xrt device and verify there are no failures?

`--target-backend llvm=cpu --device xrt`
Command failed 0 times out of 1000 runs.

`--target-backend llvm=cpu --device local-sync`
Command failed 0 times out of 1000 runs.


> I wonder if there is still some intermediate file dependency we havent caught yet. The reason I suspect that is, I think your setup is in a slow NFS file system while the CI is not which could explain the rareness on CI.

When I run with with `--n_runs 1` I still get a high failure rate (locally). I guess that means if it a file system thing, it's maybe not related to a file name issue. I remove results_dir_tmp between runs where I'm running 

```
  ./run_matmul_test.sh   results_dir_tmp ../../../iree-build-with-peano/   $MLIR_AIE_INSTALL_DIR    $PEANO_INSTALL_DIR  /opt/xilinx/xrt  $VITIS_INSTALL_PATH 
```

Thanks @nirvedhmeshram, let me know if you have any more test ideas to run (at this point I'm a bit in the dark as to what to try!).

newling:
Additional data points. I run the following sequence of experiments locally. I modified the TestShape returned in tests/matmul/generate_e2e_matmul_tests.py and the -n_runs flag in  build_tools/ci/run_matmul_test.sh. 

The wall clock time is included -- notice that for some shape it is 10x slower per iteration of the run -- I can't explain this. 

# real 0m9.861s                                                                                                                                                                                                                                                                                                             # 10 / 10 failures                                                                                                                                                                                                                                                                                                          # TestShape(m=64, k=8, n=64, accumulate=False)
                                                                                                                                                                                                                                                                                                                            # real 0m1.594s                                                                                                                                                                                                                                                                                                             # 7 / 10 failures                                                                                                                                                                                                                                                                                                           #TestShape(m=64, k=16, n=64, accumulate=False)    
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  # real 0m1.896s                                                                                                                                                                                                                                                                                                             # 6 / 10 failures                                                                                                                                                                                                                                                                                                           #TestShape(m=64, k=32, n=64, accumulate=False)
                                                                                                                                                                                                                                                                                                                            # Fails to compile.                                                                                                                                                                                                                                                                                                         # TestShape(m=64, k=48, n=64, accumulate=False),
                                                                                                                                                                                                                                                                                                                            # real 0m9.971s                                                                                                                                                                                                                                                                                                             # 0 / 10 failures                                                                                                                                                                                                                                                                                                           #TestShape(m=64, k=64, n=64, accumulate=False)
                                                                                                                                                                                                                                                                                                                            # real 0m10.104s                                                                                                                                                                                                                                                                                                            # 0 / 10 failures                                                                                                                                                                                                                                                                                                           # TestShape(m=64, k=96, n=64, accumulate=False)  
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   #real 0m9.871s                                                                                                                                                                                                                                                                                                              # 0 / 10 failures                                                                                                                                                                                                                                                                                                           # TestShape(m=32, k=16, n=64, accumulate=False)
                                                                                                                                                                                                                                                                                                                            # real 0m29.174s                                                                                                                                                                                                                                                                                                            # 4 / 1000 failures                                                                                                                                                                                                                                                                                                         # TestShape(m=32, k=16, n=32, accumulate=False)                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                            # Then I ran m=8 k=16 n=32 and it froze (20 seconds, not a single run). Ctrl+C to stop.                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                            # Then running TestShape(m=32, k=16, n=32, accumulate=False) again resulted in 10/10 failures. So the state of the machine had changed.

Issue #210, opened 2024-03-09
Map scf.forall to a single herd with block_id/core_id


yzhang93:
Similar to `MapForallToBlocks` and `MapNestedForallToThreads` transforms used in GPU, we should also consider to map the scf.forall to blocks and cores with block_id/core_id.

```
scf.forall (%iv0, %iv1) = (%lb0, %lb1) to (%ub0, %ub1) step (%step0, %step1) {
  ...
  scf.forall (%iv2, %iv3) = (%lb2, %lb3) to (%ub2, %ub3) step (%step2, %step3) {
    ... 
  } {#gpu.thready, #gpu.threadx}
  ...
} {#gpu.blocky, #gpu.blockx}
```

Lets assume that we use a logical block of size (%blocksize_x, %blocksize_y), then the IR after the transformation should look like

```
air.herd (%core_id_y, %core_id_x) in (%num_cores_y, %num_cores_x) {
  %num_blocks_y = ceilDiv(%num_cores_y, %block_size_y)
  %num_blocks_x = ceilDiv(%num_cores_x, %block_size_x)
  
  %block_id_y = %core_id_y / %block_size_y
  %block_id_x = %core_id_x / %block_size_x

  %local_id_y = %core_id_y mod %block_size_y
  %local_id_x = %core_id_x mod %block_size_x

  scf.for (%iv0, %iv1) = (%lb0 + %block_id_y * %step_0, %lb1 + %block_id_x * %step_1) to (%ub0, %ub1) step(%num_blocks_y * %step0, %num_blocks_x * %step1) {
  ...
    scf.for (%iv2, %iv3) = (%lb2 + %local_id_y * %step_2, %lb3 + %local_id_x * %step_3) to (%ub2, %ub3) step(%block_sizes_y * %step2, %block_size_x * %step3) {
     ...
   }
   ...
}
```

PR #217, opened 2024-03-13
Add pass to map nested scf.forall ops to cores

yzhang93:
This pass is implemented following the logic in https://github.com/nod-ai/iree-amd-aie/issues/210. Currently the pass only works after bufferization. The full outputs for the tests can be found [here](https://gist.github.com/yzhang93/26f597a3211b022a6206c67630cdf983), and an example of peeled IR after mapping can be found [here](https://gist.github.com/yzhang93/7ab35c4da0f958d4ffaca31977b91c7f).


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/map_forall_to_cores.mlir
```diff
@@ -0,0 +1,85 @@
...
+    %subview_1 = memref.subview %2[%3, %4] [128, 32] [1, 1] : memref<128x128xi32> to memref<128x32xi32, strided<[128, 1], offset: ?>>
+    scf.forall (%arg2, %arg3) in (4, 1) {
+      %5 = affine.apply #map1(%arg2)
+      %6 = affine.apply #map1(%arg3)
+      %subview_2 = memref.subview %subview[%5, 0] [32, 512] [1, 1] : memref<128x512xi32, strided<[512, 1], offset: ?>> to memref<32x512xi32, strided<[512, 1], offset: ?>>
+      %subview_3 = memref.subview %subview_1[%5, %6] [32, 32] [1, 1] : memref<128x32xi32, strided<[128, 1], offset: ?>> to memref<32x32xi32, strided<[128, 1], offset: ?>>
+      linalg.fill ins(%c0_i32 : i32) outs(%subview_3 : memref<32x32xi32, strided<[128, 1], offset: ?>>)
+      linalg.matmul ins(%subview_2, %subview_0 : memref<32x512xi32, strided<[512, 1], offset: ?>>, memref<512x32xi32, strided<[128, 1], offset: ?>>) outs(%subview_3 : memref<32x32xi32, strided<[128, 1], offset: ?>>)
+    } {mapping = [#gpu.thread<y>, #gpu.thread<x>]}
+  } {mapping = [#gpu.block<y>, #gpu.block<x>]}
```
nirvedhmeshram, line 72: It was my understanding that we are to use the block and thread id's in these mappings rather than deciding the number of cores/ number of tiles explicitly as done in the pass?

yzhang93: At the beginning I was thinking to use GPU blocks/threads as shown in this IR https://gist.github.com/yzhang93/9702c546462d16c70d451ed0e5f02cdb. Then we found out we cannot use the GPU launch, also if we are using gpu.block_id  x/y, we'll still need to find a way to lower from GPU dialect to scf.for loops. I'll leave it for @jtuyls to review and see if it can be applied to his work. We can have further discussions next week.

jtuyls (approved):

nirvedhmeshram (approved):

Issue #209, opened 2024-03-09
XRT sync related numerical error 


newling:
This describes a numerical issue, and the current fix in: https://github.com/nod-ai/iree-amd-aie/pull/218

A brief history, and summary of findings:

- matmul of size m=n=64 k=16 with int32 fails about 50% of the time for me, locally. This failure is when running on the iree-amd-aie stack (i.e. using the HAL runtime) You should be able to reproduce this by cloning the following PR (assuming you have peano, phoenix AIE, etc) : https://github.com/nod-ai/iree-amd-aie/pull/208  which just runs the test in a loop. 

On CI there are also failures, but much rarer (1 in 1000 failures). 

- To narrow down the issue, I wanted to reproduce (or not) the error on a different stack. I built mlir-air and managed to run https://github.com/Xilinx/mlir-air/tree/main/test/xrt/08_gemm_extern_vec successfully (thanks @erwei-xilinx for helping me)

- I modified the above mlir-air test to use the xclbin and lx6 instructions generated in https://github.com/nod-ai/iree-amd-aie/pull/208 . Running 2,000 times,  I saw no numerical errors. This made me think this is NOT a compiler error (the xclbin seems to be totally fine when run through mlir-air). But rather something specific to our setup in iree-amd-aie. I started digging into the runtime, adding print statements to see when we sync: see branch https://github.com/newling/iree-amd-aie/tree/intermittent_numerics_debugging for the print additions. 

- I noticed that the mlir-air matmul test has a sync for the 'C' (C = matmul(A, B)) tensor *before* running the kernel, but that iree-amd-aie does not have such a sync (iree-amd-aie only syncs C after running the kernel). @nirvedhmeshram and I figured out that adding the sync here (see this current PR) makes C sync before the matmul, and resolves the numerical intermittent error. 

Consider this test case:

```
func.func @matmul_64x64_32xi32_(%lhs : tensor<64x16xi32>, %rhs : tensor<16x64xi32>) -> tensor<64x64xi32> {
  %init_acc = tensor.empty() : tensor<64x64xi32>
  %c0_acc_type = arith.constant 7 : i32
  %acc = linalg.fill ins(%c0_acc_type : i32) outs(%init_acc : tensor<64x64xi32>) -> tensor<64x64xi32>
[...]

PR #218, opened 2024-03-13
Fix for intermittent numerical error 

newling:
This is now described in https://github.com/nod-ai/iree-amd-aie/issues/209

This bug is sufficiently annoying (especially on machines with old drivers @daveliddell) that I'd like to land this now and get back to finding a better fix later.

### DIFF
--- a/runtime/src/iree-amd-aie/driver/xrt/direct_command_buffer.cc
+++ b/runtime/src/iree-amd-aie/driver/xrt/direct_command_buffer.cc
@@ -348,18 +348,16 @@ static iree_status_t iree_hal_xrt_direct_command_buffer_dispatch(
     iree_host_size_t base_index =
         iree_hal_xrt_pipeline_layout_base_binding_index(kernel_params.layout,
                                                         i);
-
     for (iree_host_size_t j = 0; j < binding_count; ++j) {
       xrt::bo arg_buffer =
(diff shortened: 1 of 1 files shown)


nirvedhmeshram:
This is great investigation and debugging James! I was curious to see one thing, in the air example if you remove the memcpy for the C matrix,
https://github.com/Xilinx/mlir-air/blob/main/test/xrt/08_gemm_extern_vec/test.cpp#L136
and then remove the sync on the C buffer does it then work? This is just a way to confirm that as you mentioned
> This makes us think that IREE is writing 0's to C after the AIE has written the result (39) back from device.

newling:
> This is great investigation and debugging James! I was curious to see one thing, in the air example if you remove the memcpy for the C matrix, https://github.com/Xilinx/mlir-air/blob/main/test/xrt/08_gemm_extern_vec/test.cpp#L136 and then remove the sync on the C buffer does it then work? This is just a way to confirm that as you mentioned
> 
> > This makes us think that IREE is writing 0's to C after the AIE has written the result (39) back from device.

It works. So the failure case is only when you initialize C and then don't sync. So I'm quite confident 

 > This makes us think that IREE is writing 0's to C after the AIE has written the result (39) back from device.

newling:
We decided this is not needed after the driver upgrade.

PR #219, opened 2024-03-14
Fix windows build

spaceotter:
Removed the code to amend the library search path, since libcdo_driver is now bundled with mlir-aie. Vitis path is only configured when using chess. Fixed some CMake issues finding the boost headers on Windows. Propogate C++ exceptions as IREE status when loading the xclbin file.


spaceotter:
This is WIP since it depends on three other commits and a change to the CI scripts that I haven't written yet.

nirvedhmeshram (approved):
LGTM!

PR #221, opened 2024-03-14
Add more lit tests for transform dialect scripts

yzhang93:


tests/transform_dialect/CMakeLists.txt
```diff
@@ -8,8 +8,10 @@ iree_lit_test_suite(
   NAME
     lit
   SRCS
+    "conv_fill_spec_pad.mlir"
```
newling, line 11: This transform dialect conv 'test' should be disabled because of the Windows issue, is it still disabled?

tests/transform_dialect/matmul_fill_spec_pad_pack.mlir
```diff
@@ -1,7 +1,23 @@
-// RUN: iree-compile --iree-hal-target-backends=amd-aie --compile-to=executable-sources %S/../samples/pad_pack_pipeline_e2e.mlir | iree-opt --pass-pipeline="builtin.module(hal.executable(hal.executable.variant(iree-hal-translate-target-executable-variants{target=amd-aie})))" --iree-codegen-transform-dialect-library=%s
+// RUN: iree-opt --iree-transform-dialect-interpreter %s | FileCheck %s
 
 // This script shows an example lowering matmul through pad-pack pipeline for AIE device.
 
+func.func @matmul_i32() {
```
newling, line 5: Can the hal boilerplate be removed? I.e. can %arg0 be an tensor.empty() etc

tests/transform_dialect/matmul_fill_spec_pad_pack.mlir
```diff
@@ -135,3 +149,39 @@ module attributes { transform.with_named_sequence } {
...
+//       CHECK:       linalg.generic
+//       CHECK:       memref.dealloc %{{.*}} : memref<4x8x4x8xi32, 2>
+//       CHECK:       memref.dealloc %{{.*}} : memref<4x4x8x8xi32, 2>
+//       CHECK:     }
+//       CHECK:     iree_linalg_ext.unpack %{{.*}} outer_dims_perm = [1, 0] inner_dims_pos = [0, 1] inner_tiles = [4, 8] into %{{.*}} : (memref<4x8x4x8xi32, 2> memref<32x32xi32, strided<[64, 1], offset: ?>, 1>)
+//       CHECK:     memref.dealloc %{{.*}} : memref<4x8x4x8xi32, 2>
+//       CHECK:   }
+//       CHECK:   linalg.copy ins(%{{.*}} : memref<64x64xi32, 1>) outs(%{{.*}} : memref<64x64xi32, strided<[512, 1], offset: ?>, #hal.descriptor_type<storage_buffer>>)
+//       CHECK:   memref.dealloc %{{.*}} : memref<64x2048xi32, 1>
+//       CHECK:   memref.dealloc %{{.*}} : memref<2048x64xi32, 1>
```
newling, line 185: Maybe remove checks for dealloc, I'm assuming they're not important to the transform being tested

newling (approved):
Looks fine to me, just the 3 minor comments

tests/transform_dialect/CMakeLists.txt
yzhang93: Now it uses iree-opt instead of iree-compile. I don't think it will cause the issue? @newling @erieaton-amd could you give it a try and see if breaks Windows?

tests/transform_dialect/matmul_fill_spec_pad_pack.mlir
yzhang93: These hal ops are needed if we use `iree-opt --iree-transform-dialect-interpreter`.

tests/transform_dialect/matmul_fill_spec_pad_pack.mlir
yzhang93: I would keep the check for `memref.dealloc`. The location of these alloc and dealloc ops are important in AIR, so want to make it clear when someone look at this test IR.

tests/transform_dialect/matmul_fill_spec_pad_pack.mlir
MaheshRavishankar: It's better to keep the hal ops. That is a better representation of how these transformations get lowered

PR #224, opened 2024-03-15
[WIP] An example of using XRT directly

newling:
[will not land, as it does not test any functionality] This is based on an mlir-air matmul test. It is a minimal reproducer illustrating why we need to sync (or not write zeros unnecessarily) result tensors in https://github.com/nod-ai/iree-amd-aie/pull/218

### DIFF
--- a/build_tools/ci/use_xrt_directly/run_test.sh
+++ b/build_tools/ci/use_xrt_directly/run_test.sh
@@ -48,25 +48,16 @@ if [ ! -d "${IREE_INSTALL_DIR}" ]; then
   exit 1
 fi
 
-# Search for iree-compile and iree-e2e-matmul-test in the user provided directory.
+# Search for iree-compile in the user provide
 IREE_COMPILE_EXE=""
-TEST_RUNNER=""
 for dir in "${IREE_INSTALL_DIR}" "${IREE_INSTALL_DIR}/bin" "${IREE_INSTALL_DIR}/tools"; do
   if [ -f "${dir}/iree-compile" ]; then
     IREE_COMPILE_EXE="${dir}/iree-compile"
-  fi
-  if [ -f "${dir}/iree-e2e-matmul-test" ]; then
-    TEST_RUNNER="${dir}/iree-e2e-matmul-test"
   fi
 done
 
 if [ -z "${IREE_COMPILE_EXE}" ]; then
   echo "No 'iree-compile' found in any of the following directories: " \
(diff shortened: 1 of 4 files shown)


newling:
This does 

C = 7 + A @ B 

where A and B are tensors of 1's, and M=N=64 and K=16. 

So we expect C to be all 23's. 

This is what I observe when I run this test locally. 

When I run it in CI, C is populated with 16's though. 

@nirvedhmeshram another mystery to solve tomorrow!

build_tools/ci/use_xrt_directly/run_test.sh
```diff
@@ -0,0 +1,193 @@
...
+  XCLBIN_FILE="${file}"
+done
+if [ -z "${XCLBIN_FILE}" ]; then
+  echo "No .xclbin file found in the output directory or any of its subdirectories."
+  exit 1
+else
+  echo "Found .xclbin file: ${XCLBIN_FILE}"
+fi
+
+# Maybe sign the XCLBIN file (CI moonshot).
```
newling, line 170: @erieaton-amd here I sign the XCLBIN directly, without first copying it and then signing the copy. 

Question about the clean-up: how does copying it with a new name ensure all xclbins are removed in CI? Based on a comment on discord.  Is what I'm doing here bad? How I can see what is actually happening on the CI machine, how can I check what junk is being created and left there? I guess that turned into 3 questions... 

FYI @daveliddell

erieaton-amd: This is cluttering the CI machine. The renaming scheme is expected by the cleanup phase to make sure that we delete every kernel that got copied to the firmware directory by this github job and not any other github jobs.

newling: I understand now. 

Repeating what you already know to help me remember... $SIGNER moves the xclbin it "signs" to a directory outside of github's CI temporary directory. It doesn't sign in-place. The place it moves to is always the same, so it's a problem if multiple CI jobs run in parallel -- the signer is moving xcblins from different locations (each CI starts in a different temporary directory) to the same location -- race conditions. 

To prevent this, each xclbin is given a unique name, based on it's job id. This uniquely named xclbin is what is then copied to the "special place" in /lib/firmware for signing. 

So that resolves the race condition, but not the problem of cleaning up the garbage left in /lib/firmware. This is done in another CI script.

PR #225, opened 2024-03-15
Expand the logic to find tiling and packing config

nirvedhmeshram:
This PR adds a largest factor based logic to select tile sizes and packing config for the pad-pack pipeline. With this arbitrary shapes seem to be working in local testing. Some CI e2e tests with arbitrary shapes are added here. Since the other legacy pipelines are not expected to support such shapes we have to have a new grouping for the shapes.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -215,15 +233,14 @@ static LogicalResult setRootConfigForPadPackPipeline(func::FuncOp entryPointFn,
...
+  auto tileM0 = findLargestFactor((int)initShape[0], 64);
+  auto tileN0 = findLargestFactor((int)initShape[1], 64);
+  auto tileM1 = findLargestFactor((int)tileM0, 32);
+  auto tileN1 = findLargestFactor((int)tileN0, 32);
   auto lhsType = linalgOp.getDpsInputOperand(0)->get().getType();
   auto lhsShape = llvm::cast<ShapedType>(lhsType).getShape();
-  auto tileK0 = std::min((int)lhsShape[1], 256);
-  auto tileK1 = std::min((int)lhsShape[1] / 8, 4);
-
+  auto tileK0 = findLargestFactor((int)lhsShape[1], 256);
```
newling, line 242: This call might loop 256 times. I asked ChatGPT4 for a more efficient algo and it came up with an O(sqrt(max)) approach which isn't significantly more compilated. Inner loop looks something like

```
 for (int i = 2; i <= std::sqrt(num); ++i) {
        if (num % i == 0) { // If i is a factor
           auto correspondingFactor = num / i;
           // update to max of i and correspondingFactor. 
        }
 }
 ```

File: tests/matmul/generate_e2e_matmul_tests.py [hunk 2]
```diff
@@ -34,8 +34,8 @@ class MatrixElemTypeId(enum.Enum):
 # The values are the accepted values for the --shapes= flag.
 @enum.unique
 class ShapesId(enum.Enum):
-    SMALL = "small"
```
newling, line 37: My preference: I'd like to remove this layer of indirection and specify shapes in tests directly. 

(would be a separate PR, and I'm happy to do it after this PR is merged)

tests/matmul/generate_e2e_matmul_tests.py
```diff
@@ -102,14 +102,31 @@ def get_test_shapes(shapes_id: ShapesId):
...
     #    latency.
+
     if shapes_id == ShapesId.SMALL:
+        return [
+            # some "nice" multiple of 8 shapes
+            TestShape(m=8, k=16, n=32, accumulate=False),
+            TestShape(m=16, k=8, n=16, accumulate=False),
+            # some arbitory shapes
+            TestShape(m=9, k=15, n=7, accumulate=False),
+            TestShape(m=52, k=63, n=52, accumulate=False),
```
newling, line 113: Do these shape really work? Very nice if so! Maybe not for bfloat16 yet though?

tests/matmul/generate_e2e_matmul_tests.py
```diff
@@ -102,14 +102,31 @@ def get_test_shapes(shapes_id: ShapesId):
     # 2. Some shapes are commented out: they used to be tested but have been
     #    disabled to improve the trade-off between test coverage and build
     #    latency.
+
     if shapes_id == ShapesId.SMALL:
+        return [
+            # some "nice" multiple of 8 shapes
+            TestShape(m=8, k=16, n=32, accumulate=False),
+            TestShape(m=16, k=8, n=16, accumulate=False),
+            # some arbitory shapes
```
newling, line 111:
```suggestion
            # some arbitrary shapes
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -30,29 +30,47 @@ static LogicalResult setRootConfigForPadPipeline(func::FuncOp entryPointFn,
...
+  }
+  return 1;
+}
+
+static SmallVector<int64_t> getPackedSize(linalg::LinalgOp linalgOp,
+                                          const int packLevel, int m = 0,
+                                          int n = 0, int k = 0) {
   // TODO: consider emiting an error/warning if the default sizes are used as a
   // fallback.
   SmallVector<int64_t> defaultSizes;
```
newling, line 52: If the '4,4,8' is related to vector sizes for matmul inner loop, the optimal sizes (element type dependent) can be obtained from getAIEMatmulInstructionSize

(not a request for change here as the 4,4,8, are here already, just some info).

tests/matmul/generate_e2e_matmul_tests.py
nirvedhmeshram: Yes, I would expect them to work for scalar path, which doesnt exist for bf16.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -240,7 +257,7 @@ static LogicalResult setRootConfigForPadPackPipeline(func::FuncOp entryPointFn,
   // ------------------------------------------------------
   MLIRContext *context = entryPointFn.getContext();
   const int packLevel = 1;
-  auto packedSizes = getPackedSize(linalgOp, packLevel);
+  auto packedSizes = getPackedSize(linalgOp, packLevel, tileM1, tileN1, tileK1);
```
nirvedhmeshram, line 260: Here `tilek1` is not the right thing to pass, for the 2kx2kx2k example, tilkek1 is 4 but the innerloop pack op has a source of k=16, I am trying to figure out how to statically figure that out.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
nirvedhmeshram: not sure I follow, what does the outer loop looks like?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: For 2kx2kx2k example, as mentioned above the inner pack size is selected fixed as (4, 4, 8). While the inner tile size for k should be set as 32 according to Erwei's experimental value. Since we have a fusion pass that fuse the pack ops into the scf.for loop, the tile size for k should be adjusted to 32/8=4, which is how 4 shows in the code.

yzhang93 (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -30,29 +30,47 @@ static LogicalResult setRootConfigForPadPipeline(func::FuncOp entryPointFn,
...
+                                          int n = 0, int k = 0) {
   // TODO: consider emiting an error/warning if the default sizes are used as a
   // fallback.
   SmallVector<int64_t> defaultSizes;
+  int minM = m ? findLargestFactor(m, 4) : 4;
+  int minN = n ? findLargestFactor(n, 4) : 4;
+  int minK = k ? findLargestFactor(k, 8) : 8;
   if (packLevel == 1) {
-    defaultSizes = {4, 4, 8};
+    defaultSizes = {{minM, minN, minK}};
```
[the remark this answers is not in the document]
nirvedhmeshram: we do not yet support padding though, the code just errors out if the pack sizes are not a multiple of the inner allocations.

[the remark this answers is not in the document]
nirvedhmeshram: The way I was thinking to go about this is, that these are just default configs, and the vectorization path can over-ride them with something that works for that path.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -215,15 +233,14 @@ static LogicalResult setRootConfigForPadPackPipeline(func::FuncOp entryPointFn,
...
   auto initType = linalgOp.getDpsInitOperand(0)->get().getType();
   auto initShape = llvm::cast<ShapedType>(initType).getShape();
-  auto tileM0 = std::min((int)initShape[0], 64);
-  auto tileN0 = std::min((int)initShape[1], 64);
-  auto tileM1 = std::max((int)tileM0 / 2, 1);
-  auto tileN1 = std::max((int)tileN0 / 2, 1);
+  auto tileM0 = findLargestFactor((int)initShape[0], 64);
+  auto tileN0 = findLargestFactor((int)initShape[1], 64);
+  auto tileM1 = findLargestFactor((int)tileM0, 32);
+  auto tileN1 = findLargestFactor((int)tileN0, 32);
```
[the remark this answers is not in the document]
nirvedhmeshram: so for the matmul shapes thar are multiples of 64 (the initial chosen level 0 tile, we can do this, I made sure we have the same config for such cases, e.g for 2kx2kx2k case we will still get
` #iree_codegen.lowering_config<tile_sizes = [[64, 64], [0, 0, 256], [32, 32], [0, 0, 4]]>, packing_config = #amdaie.packing_config<packing_config = [{packedSizes = [4, 4, 8],`
Howver for random shapes, we have to disticute differently and the cores can run multiple tiles. This seems to be working in my experimentation. 
We should still be able to adjust these factors for the agreeable shapes with your upcoming patch.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
nirvedhmeshram: Currently getAIEMatmulInstructionSize this is not returning a config for i32 case, perhaps because there is no vector intrinsic for it, so we need to pick some packing size for such cases separately I think.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
nirvedhmeshram: I am not clear how you are getting the 32 here though in the first place. As mentioned in the other comment the configuration is not changing for cases like 2kx2kx2k.
What I noticed is currently happening is. With the outer k tile of 256. The inner k tile divides that and we have a k of 256/4 = 64. So from a code-gen point of view any k packing value that divides 64  is fine, but from hardware point of view since we prefer 8, we choose that since it does divide 64. However lets say the outer size was something random like 50 needed tom support a random shape. then we will pick inner k as 2 because that's closest to 4 we can get and still divide 50, now we have a inner k of 50/2=25 and then we will choose a packing of 5, which is closest we can get to desired packing of 8 and still divide 25.

nirvedhmeshram:
I am neutral on if we actually want to land this, my main purpose was to put some pressure on the lower level stack to see if there are any bugs, as such I have found 2 issues. 
1. 300x300x300 shape seems to have a correctness error. 
@erwei-xilinx here is the gist for it https://gist.github.com/nirvedhmeshram/37fc0e276d624b5bca61e1b54720ac9a
Please let me know if it looks reasonable to you or you see some obvious bug. 
2. 9x15x7 (m,k,n) is failing in llvm with error `LLVM ERROR: unable to legalize instruction: %149:_(s32) = G_EXTRACT_VECTOR_ELT %148:_(<4 x s32>), %108:_(s32) (in function: core_0_2)`
I am going to update my llvm to latest and see if the bug persist and if it does I will provide the llvm IR dump so that we can get a fix for that.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
nirvedhmeshram: as discussed offline, we are going to stick with the unoptimized version for now.

erwei-xilinx:
> 300x300x300 shape seems to have a correctness error.

I had a quick scan through the IR dump and one issue I can see is that since the air.launch is (5, 5), it seems to have created an error with IPU instructions generation, which is trying to tile it by a factor of (4, 4) to accommodate for the SHIM DMA BD count limit (<=16). Since 5 isn't divisible by 4, it seems to have messed up there...

nirvedhmeshram:
> > 300x300x300 shape seems to have a correctness error.
> 
> I had a quick scan through the IR dump and one issue I can see is that since the air.launch is (5, 5), it seems to have created an error with IPU instructions generation, which is trying to tile it by a factor of (4, 4) to accommodate for the SHIM DMA BD count limit (<=16). Since 5 isn't divisible by 4, it seems to have messed up there...

@erwei-xilinx could you please share the pass in which this incorrect lowering may be happening? I can try to  investigate why it happens in this case and not some other random shapes that are passing the CI.

erwei-xilinx:
> could you please share the pass in which this incorrect lowering may be happening? I can try to investigate why it happens in this case and not some other random shapes that are passing the CI.

Sure, thanks. This is the [link](https://github.com/nod-ai/iree-amd-aie/blob/e00788c26fb454597eef89e1a685a9d4558b9fa6/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp#L552) to the pass.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: While the inner k tile is not 64 in the previous setting. If you like we can have a chat to discuss about this.

yzhang93 (commented):
Forgot to submit the additional review comments last week... I think there are some misunderstanding, if you like we can have some discussion.

nirvedhmeshram:
@yzhang93 making an example without this changes, for the 300x300x300 size, in a bit..

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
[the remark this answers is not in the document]
nirvedhmeshram: resolved offline.

nirvedhmeshram:
@yzhang93 PTAL

PR #227, opened 2024-03-18
Vectorization on pad-pack pipeline 

newling:
- add required passes 
- add e2e test

### DIFF
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -18,7 +18,7 @@
 #      `iree-e2e-matmul-test` to include support for the runtime HAL
 #      driver/device you wish to test.
 #   2. Update the paths in this script or specify them via environment variables
-#   3. Run: `./run_matmul_tests.sh <output_dir_path> <iree_install_path> [<mlir_aie_install_path>] [<peano_install_path>] [<xrt_path>] [<vitis_path>]`
+#   3. Run: `./run_matmul_tests.sh <output_dir_path> <iree_install_path> [<mlir_aie_install_path>] [<peano_install_path>] [<xrt_path>] [<vitis_path>] [<do_signing>]`
 #      The directories above in square brackets are optional, the first 2 directories are required.
 
 set -euox pipefail
@@ -33,7 +33,7 @@ if [ "$#" -lt 2 ] || [ "$#" -gt 7 ]; then
    #    5) <xrt-dir>               (optional)
    #    6) <vitis-install-dir>     (optional)
    #    7) <do-signing>            (optional)
-    echo -e "Illegal number of parameters: $#, expected 2,3,4,5, or 6 parameters." \
+    echo -e "Illegal number of parameters: $#, expected 2-7 parameters." \
             "\n The parameters are as follows:" \
             "\n     1) <output-dir>               (required)" \
             "\n     2) <iree-install-dir>         (required)" \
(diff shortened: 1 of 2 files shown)


newling:
@nirvedh if https://github.com/nod-ai/iree-amd-aie/pull/225 lands first I'll change to large-legacy in this PR.

nirvedhmeshram (approved):
LGTM!

yzhang93:
Do we want to add a flag to switch on/off the vectorization? I assume vectorization doesn't work if it doesn't tile and pack to the intrinsic sizes as we discussed today.

nirvedhmeshram:
> @Nirvedh if #225 lands first I'll change to large-legacy in this PR.

I suspect, you will have to change the shape of legacy large, the reason is the k is 16, without #225 , the inner k is choosen as 16/8=2 and a packing k of 8, with the new logic, it will choose inner k of 4 and a packing k of 4 which the vectorization wont support if I understand correctly?

nirvedhmeshram:
> Do we want to add a flag to switch on/off the vectorization? I assume vectorization doesn't work if it doesn't tile and pack to the intrinsic sizes as we discussed today.

I think vectorization will automatically not happen if the size doesnt match so no need of flag @newling ?

nirvedhmeshram:
> > @Nirvedh if #225 lands first I'll change to large-legacy in this PR.
> 
> I suspect, you will have to change the shape of legacy large, the reason is the k is 16, without #225 , the inner k is choosen as 16/8=2 and a packing k of 8, with the new logic, it will choose inner k of 4 and a packing k of 4 which the vectorization wont support if I understand correctly?

https://github.com/nod-ai/iree-amd-aie/pull/229 should address this, now we can even vectorize for k=8

newling:
@yzhang93 -- @nirvedhmeshram is kind of right: vectorization does not happen if the inner matmul dimensions don't match the AIE shapes. However technically it's the pass from the vector dialect to the AIEVec dialect where nothing happens (that pass in in mlir-aie). The vectorization pass here in iree-amd-aie still massages the outer dimensions to unit dims, irrespective of the inner dim shapes (that is done in the pass https://github.com/nod-ai/iree-amd-aie/pull/194) and it still converts to the vector dialect (that is done in this pass https://github.com/nod-ai/iree-amd-aie/pull/172). So even if the inner dimensions are m=3 n=5 k=7 (say) it still ends up in a vector.contract. This might not be the best behavior, and it is definitely possible to make the 2 passes mentioned do nothing unless the inner dimensions are good for vector->aievec to convert. But I think it's fine for now.

nirvedhmeshram:
> @yzhang93 -- @nirvedhmeshram is kind of right: vectorization does not happen if the inner matmul dimensions don't match the AIE shapes. However technically it's the pass from the vector dialect to the AIEVec dialect where nothing happens (that pass in in mlir-aie). The vectorization pass here in iree-amd-aie still massages the outer dimensions to unit dims, irrespective of the inner dim shapes (that is done in the pass #194) and it still converts to the vector dialect (that is done in this pass #172). So even if the inner dimensions are m=3 n=5 k=7 (say) it still ends up in a vector.contract. This might not be the best behavior, and it is definitely possible to make the 2 passes mentioned do nothing unless the inner dimensions are good for vector->aievec to convert. But I think it's fine for now.

Ah thanks for explaining, I think this is why other backends have different pipelines for when they think they cant match to that specific vector.contract. We dont have to go that far yet but I agree with @yzhang93 that we should guard with a flag, to have more flexibility, you can keep it default on will just give us flexibility to have a scalar path.

newling:
> > @yzhang93 -- @nirvedhmeshram is kind of right: vectorization does not happen if the inner matmul dimensions don't match the AIE shapes. However technically it's the pass from the vector dialect to the AIEVec dialect where nothing happens (that pass in in mlir-aie). The vectorization pass here in iree-amd-aie still massages the outer dimensions to unit dims, irrespective of the inner dim shapes (that is done in the pass #194) and it still converts to the vector dialect (that is done in this pass #172). So even if the inner dimensions are m=3 n=5 k=7 (say) it still ends up in a vector.contract. This might not be the best behavior, and it is definitely possible to make the 2 passes mentioned do nothing unless the inner dimensions are good for vector->aievec to convert. But I think it's fine for now.
> 
> Ah thanks for explaining, I think this is why other backends have different pipelines for when they think they cant match to that specific vector.contract. We dont have to go that far yet but I agree with @yzhang93 that we should guard with a flag, to have more flexibility, you can keep it default on.

Sounds good, I see no harm in having a flag to disable vectorization (or any other pass!) Will add one (future PR).

PR #228, opened 2024-03-18
gcc-13 compiler error fixes

newling:
Why am I building with g++? Because I wanted to use stacktrace (C++23 feature) with gcc has better support for than clang++. 

Side note: stack traces are not useful when running iree-run-module and trying to figure out what the call stack looks like. I guess because HAL virtual machine has it's own stack.


nirvedhmeshram (approved):

stellaraccident:
PSA: updating to newer c++ standard versions is a topic that typically requires careful multi project cooperation and shouldn't just be done locally.

keryell:
[runtime/src/iree-amd-aie/driver/xrt/.native_executable.cc.swp](https://github.com/nod-ai/iree-amd-aie/pull/228/files#diff-ab24fd2b0f5795b2dc34da6a9b766351e275ecb995bdf955e4cecd6f18630229) ?
Otherwise this is not specially updating the compiler, just avoiding to fail when using a modern one, which I find useful to improve the code quality by finding some bugs and cleaning the code.

newling:
> [runtime/src/iree-amd-aie/driver/xrt/.native_executable.cc.swp](https://github.com/nod-ai/iree-amd-aie/pull/228/files#diff-ab24fd2b0f5795b2dc34da6a9b766351e275ecb995bdf955e4cecd6f18630229) ? 

Oof that was bad, thanks for noticing that file. Fix PR https://github.com/nod-ai/iree-amd-aie/pull/235

> Otherwise this is not specially updating the compiler, just avoiding to fail when using a modern one, which I find useful to improve the code quality by finding some bugs and cleaning the code.

Yes that's correct I have no intention of adding a cmake flag requiring C++23 any time soon! I was just experimenting with a new g++ for debugging purposes (I wanted to see if it could produce stack traces, so I could avoid boost)

PR #229, opened 2024-03-18
Change tiling/packing order to be more vectorization friendly

nirvedhmeshram:
This PR tries to first pack to improve the chances of getting the desired pack and then chooses the inner k tile based on the resulting untiled size.


yzhang93:
This commit LGTM. But I would suggest to further refactor the selection of tile sizes including the ones used by ukernel path that @Abhishek-Varma have currently.

nirvedhmeshram:
> This commit LGTM. But I would suggest to further refactor the selection of tile sizes including the ones used by ukernel path that @Abhishek-Varma is working currently.
@Abhishek-Varma is going to work on that tomorrow.

yzhang93 (approved):

PR #231, opened 2024-03-19
Provide flag to make signing optional when running e2e matmul script

nirvedhmeshram:
Currently there is a guard on if we have the script to do the signing, however there are cases where you dont want to sign it even if you have the script. So this PR adds an option to do that.

### DIFF
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -23,7 +23,7 @@
 
 set -euox pipefail
 
-if [ "$#" -lt 2 ] || [ "$#" -gt 6 ]; then
+if [ "$#" -lt 2 ] || [ "$#" -gt 7 ]; then
 
    # The expected parameters are
    #    1) <output-dir>            (required)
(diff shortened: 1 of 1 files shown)


build_tools/ci/run_matmul_test.sh
```diff
@@ -255,7 +264,8 @@ function run_matmul_test() {
   # Make a guess as to whether we need to sign the XCLBIN:
   SIGNER=${XRT_DIR}/amdxdna/setup_xclbin_firmware.sh
   # 1) check if $XRT_DIR/amdxdna/setup_xclbin_firmware.sh exists:
-  if [ ! -f "$SIGNER" ]; then
+  echo $DO_SIGNING
+  if [ ! -f "$SIGNER" ] || [ $DO_SIGNING -eq 0 ]; then
```
newling, line 268: Please change this to

 if [ $DO_SIGNING -eq 0 ]
 echo "**** Skipping XCLBIN signing: DO_SIGNING set to 0 ****"
 elif [ ! -f "$SIGNER" ]
    echo "**** Skipping XCLBIN signing: $SIGNER not found ****"
 else
   ...
   
so that the message remains accurate.

newling (approved):
Makes sense to me, just the one minor comment about the message

PR #232, opened 2024-03-19
Tweak logic related to signing script

newling:
In the case do_signing=1 (the default) and the signing script is not found, fail immediately.


nirvedhmeshram (approved):
LGTM!

PR #234, opened 2024-03-20
Option to disable vectorization 

newling:
While it will be unusual to not use the default here (bfloats and small integer types **must** currently be vectorized in all situations because AIE does not support scalar operations on these types. And types of size >= 32 bits **must not** be vectorized because AIE does not have vector operations for these), it might be useful for debugging purposes to disable vectorization. If/when peano can emulate scalar bfloat operations, this might be especially useful to switch between vectorized and unvectorized lowering.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -55,6 +48,23 @@ static llvm::cl::opt<std::string> clPathToUkernels(
     "iree-amdaie-path-to-ukernels",
     llvm::cl::desc("Path to microkernels' directory"));
 
+static llvm::cl::opt<bool> clEnableVectorizationInPassPipeline(
+    "iree-amdaie-enable-vectorization-in-pass-pipeline",
```
yzhang93, line 52: I think flag name `iree-amdaie-enable-vectorization` is enough?

newling: Ok. I actually initially had that, but then thought it was weird that something like 

```
iree-opt --iree-amdaie-enable-vectorization=0 --iree-amdaie-vectorization
```

does vectorize. But on seconds thoughts it's a weird thing to do. I'll change it tomorrow

yzhang93: Okay, I see. Then maybe `iree-amdaie-enable-vectorization-passes`?

newling: A good compromise, I'll run with that.

yzhang93 (approved):
LGTM. Another minor comment... Maybe move the test `disable_vectorization.mlir` to the tests under compiler/.

PR #235, opened 2024-03-20
remote vim temp file

newling:
I accidentally included this in https://github.com/nod-ai/iree-amd-aie/pull/228 (Thanks for noticing this @keryell)


nirvedhmeshram (approved):
This is a fairly common thing to happen and you can blame it on our lacking .gitignore file, maybe add these lines to the .gitignore in this PR itself?
https://github.com/openxla/iree/blob/main/.gitignore#L46-L47

nirvedhmeshram:
@newling could you please add back the three project specific things in the .gitignore, I think they were there because users had found them useful in the past.

newling:
> @newling could you please add back the three project specific things in the .gitignore, I think they were there because users/ CI scripts had found them useful in the past.

Apologies, I assumed they were left over from someone's local setup/experiment. Added back here: https://github.com/nod-ai/iree-amd-aie/pull/238

PR #236, opened 2024-03-20
[AMD-AIE] Generalize tiling/packing based on a given element type

Abhishek-Varma:
-- This commit generalizes tiling/packing config based on a given
   element type.
-- The rationale behind this is that element type with lesser memory
   footprint can exploit higher tiling/packing sizes and vice versa.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAddLoweringStrategy.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAddLoweringStrategy.cpp
@@ -4,8 +4,10 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+#include "iree-amd-aie/IR/AMDAIEDialect.h"
 #include "iree-amd-aie/Transforms/KernelDispatch.h"
 #include "iree-amd-aie/Transforms/Passes.h"
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenDialect.h"
 #include "iree/compiler/Dialect/HAL/IR/HALDialect.h"
 #include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtDialect.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
@@ -23,6 +25,7 @@ class AMDAIELoweringStrategyPass
  public:
   void getDependentDialects(DialectRegistry &registry) const override {
     registry.insert<
+        AMDAIE::AMDAIEDialect, IREE::Codegen::IREECodegenDialect,
         IREE::HAL::HALDialect, IREE::LinalgExt::IREELinalgExtDialect,
         bufferization::BufferizationDialect, linalg::LinalgDialect,
         LLVM::LLVMDialect, pdl::PDLDialect, pdl_interp::PDLInterpDialect,
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
@@ -230,19 +230,28 @@ static LogicalResult setRootConfigForPadPackPipeline(func::FuncOp entryPointFn,
   // -------------- Set lowering config -------------------
   // ------------------------------------------------------
   // Assume working on a 2x2 AIE array. Currently, the tile sizes are chosen
-  // empirically for large GEMM sizes, which are [64, 64, 256] for the first
-  // level and [32, 32, 32] for the second level. Basic min/max constraints are
-  // added to avoid failure for small GEMM sizes.
+  // empirically for large GEMM sizes, which are [32*s, 32*s, 256] for the first
+  // level and [16*s, 16*s, 16*s] for the second level, where 's' is the scaling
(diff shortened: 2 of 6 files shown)


nirvedhmeshram (commented):
I am wondering why we end up with the same packing config for all types and why is that not scaling-aware?, do we always prefer (4,4,8) for that? or it is more flexible and the cases where we have vector intrinsics, we prefer 4,4,8 and for cases where we dont, it doesn't really matter?

Abhishek-Varma:
> I am wondering why we end up with the same packing config for all types and why is that not scaling-aware?, do we always prefer (4,4,8) for that? or it is more flexible and the cases where we have vector intrinsics, we prefer 4,4,8 and for cases where we dont, it doesn't really matter?

The same packing config works for i32 (which is exactly same as what we get without the PR's diff) and for bf16 (ukernel) vmfb - so it doesn't really matter perhaps?

As a context, I was working on a stale `main` branch and observed the packing config to have changed ("scaling aware") while changing tile sizes.
But on pulling the latest changes https://github.com/nod-ai/iree-amd-aie/commit/f8fbee9446bf1771d34ea43357df92e1b258e893 the packing config remained the same - I verified it too by running the bf16 ukernel.

Let me know your thoughts.

nirvedhmeshram (approved):
Makes sense, yeah I changed the order of packing/tiling in a recent commit that would explain the change you observed from stale branch to latest. LGTM but lets wait to land until you come on tomorrow morning to give some time for other reviewers if they have comments/concerns.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAddLoweringStrategy.cpp
```diff
@@ -23,6 +25,7 @@ class AMDAIELoweringStrategyPass
  public:
   void getDependentDialects(DialectRegistry &registry) const override {
     registry.insert<
+        AMDAIE::AMDAIEDialect, IREE::Codegen::IREECodegenDialect,
```
yzhang93, line 28: Why do you have to add these? I don't see what you added has to use these?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -223,26 +223,51 @@ static LogicalResult setRootConfigForPackPipeline(func::FuncOp entryPointFn,
...
+///     c. 16 bit width -> scaling factor will be 4.
+/// This can easily be scaled to other element data types by just changing the
+/// check in the `assert`.
+/// The rationale/need for this is - higher bit width element types can be
+/// accomodated using small tile sizes, whereas lower bit width element types
+/// can make use of larger tile sizes as they can be easily accomodated within
+/// the same. Therefore this tiling scale factor helps dynamically
+/// increase/decrease the tiling window depending on the element type's bit
+/// width.
+static unsigned getTilingScaleFactor(Type elemType) {
```
yzhang93, line 240: It's better to move this function to Utils.cpp.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -223,26 +223,51 @@ static LogicalResult setRootConfigForPackPipeline(func::FuncOp entryPointFn,
   return success();
 }
 
+/// Utility function that, given an element type, computes tiling scaling factor
+/// as : 64/bitWidth.
+/// Therefore, for example, if the element type has :-
+///     a. 64 bit width -> scaling factor will be 1.
+///     b. 32 bit width -> scaling factor will be 2.
+///     c. 16 bit width -> scaling factor will be 4.
+/// This can easily be scaled to other element data types by just changing the
```
yzhang93, line 232: Don't you also need to set scaling factor according to data types, i.e., int or float?

yzhang93 (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAddLoweringStrategy.cpp
Abhishek-Varma: I added a new lit test file which necessitated their addition.

We never had any lit test that was verifying the tile/pack config - and since the lit test invokes the lowering strategy pass as a standalone pass, the MLIRContext doesn't have the required Dialects' info to form the attributes which KernelDispatch forms.

This wasn't required previously because we were using some pass pipeline (bigger picture/context) to test it - it gets loaded in MLIRContext by at least one of the passes being used in that pass pipeline we use.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
Abhishek-Varma: Nope. It should only be based on the memory footprint which in turn is based on the bitwidth.

Eg: `fp16` uses 16 bit to represent a data. `int16` uses 16 bit to represent a data. Therefore their memory footprint is the same.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
Abhishek-Varma: Done.

yzhang93 (approved):
Thanks!

PR #238, opened 2024-03-20
Put previously removed .gitignore lines back

newling:
See https://github.com/nod-ai/iree-amd-aie/pull/235#issuecomment-2009953221


nirvedhmeshram (approved):
Thanks!

PR #239, opened 2024-03-20
Deprecate pad and simple-pack pipelines

yzhang93:
We want to deprecate the pad and simple-pack pipelines to lower the maintenance requirement. From now on, we will be focusing on two pipelines

- Pad-pack pipeline: is the current default pipeline, which can be lowered through AIR and run e2e on hardware.
- Pack pipeline: using pack operations for data movement from L3->L2, L2->L1, and peeling operation to enable initialization in L1. Currently this pipeline is not fully optimized to run through AIR, and is used for funcIR/future development.


newling:
Looks good to me, I can't think of any reason why any of these pipelines/tests should not be removed. Nice simplification!

MaheshRavishankar:
+1! Good for me as well. Let's wait for others to chime in

nirvedhmeshram (approved):
Nice cleanup, LGTM!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
```diff
@@ -266,9 +265,7 @@ static bool isMatmul(linalg::LinalgOp linalgOp, AIEPassPipeline passPipeline) {
     return false;
   }
 
-  if (passPipeline == AIEPassPipeline::PadPipeline) {
-    return match2DLinalgGenericMatmul(linalgOp);
```
Abhishek-Varma, line 270: This function in the file will also get eliminated - you may remove it.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_ukernel.mlir
```diff
@@ -1,5 +1,3 @@
-// RUN: iree-opt --split-input-file --pass-pipeline="builtin.module(func.func(iree-amdaie-lower-to-ukernels,cse,canonicalize))" %s | FileCheck %s
```
Abhishek-Varma, line 1: I still see `CHECK` statements - you may convert them to `PAD_PACK`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_ukernel.mlir
```diff
@@ -1,5 +1,3 @@
...
-// RUN: iree-opt --split-input-file --pass-pipeline="builtin.module(func.func(iree-amdaie-lower-to-ukernels{pass-pipeline=simple-pack path-to-ukernels="/custom/path/to/ukernels"},cse,canonicalize))" %s | FileCheck %s --check-prefix=SIMPLE_PACK_AND_UK_PATH
 // RUN: iree-opt --split-input-file --pass-pipeline="builtin.module(func.func(iree-amdaie-lower-to-ukernels{pass-pipeline=pad-pack},cse,canonicalize))" %s | FileCheck %s --check-prefix=PAD_PACK
```
Abhishek-Varma, line 1: Since others are removed, please add a `path-to-ukernels="/custom/path/to/ukernels"` like what was there previously.
And replace all uses of `mm.o` with `/custom/path/to/ukernels/mm.o`.
It should just work.

Abhishek-Varma (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
yzhang93: I think this function is useful. We can leave it here in case you'll need this in the future?

Abhishek-Varma: In fact, we won't need both `match2D*` (Pad) and `match6D*` (Simple-Pack) - you may remove both of them - no issues.

They'll not find use anywhere else in the pipeline.

Abhishek-Varma (approved):
Unblocking the PR which can be merged after the above changes are addressed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
yzhang93: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_ukernel.mlir
yzhang93: All use CHECK statements now

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_ukernel.mlir
yzhang93: Done.

PR #240, opened 2024-03-21
Example of unit testing C++ code 

newling:
I guess there are 2 different things up for review here

1) unit tests for C++ code. I know that we should use lit with filecheck for testing whenever possible, because it is fast and doesn't require compiling and linking (projects where linking C++ tests are slow are a pain to work on...). There is 1 example of gtest in IREE. 

2) new algo for findLargestFactor (less important than 1).

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
@@ -1,4 +1,4 @@
-// Copyright 2020 The IREE Authors
+// Copyright 2024 The IREE Authors
 //
 // Licensed under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
@@ -8,6 +8,7 @@
 
 #include "iree-amd-aie/IR/AMDAIEAttrs.h"
 #include "iree-amd-aie/Transforms/AMDAIEUtils.h"
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h"
 #include "iree/compiler/Codegen/Utils/CPUUtils.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/MemRef/Transforms/Transforms.h"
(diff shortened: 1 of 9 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -17,6 +18,36 @@
...
+
+  int largestLowFactor = 1;
+  for (int lowFactor = 2; lowFactor <= max; ++lowFactor) {
+    const int highFactor = num / lowFactor;
+
+    // This early exit is what makes this O(sqrt(num)) instead of O(num).
+    if (highFactor < lowFactor) return largestLowFactor;
+
+    const bool areActuallyFactors = num % lowFactor == 0;
+    if (areActuallyFactors) {
```
yzhang93, line 38: I know @nirvedhmeshram is the original author of this function, but I think it's better to move this utility to AMDAIEUtils.cpp.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/AMDAIEUtilsTest.cpp
```diff
@@ -0,0 +1,52 @@
...
+  EXPECT_EQ(detail::findLargestFactor(/* num = */ 11, /* max = */ 6), 1);
+  EXPECT_EQ(detail::findLargestFactor(/* num = */ 12, /* max = */ 6), 6);
+  EXPECT_EQ(detail::findLargestFactor(/* num = */ 13, /* max = */ 6), 1);
+
+  EXPECT_EQ(detail::findLargestFactor(/* num = */ 12, /* max = */ 5), 4);
+  EXPECT_EQ(detail::findLargestFactor(/* num = */ 12, /* max = */ 4), 4);
+  EXPECT_EQ(detail::findLargestFactor(/* num = */ 12, /* max = */ 3), 3);
+  EXPECT_EQ(detail::findLargestFactor(/* num = */ 12, /* max = */ 2), 2);
+
+  // Example of example which is faster because of sqrt algo:
```
nirvedhmeshram, line 41: Nit: Example of example -> Example

nirvedhmeshram (commented):
looks like gtest is not on the CI machine and other users will not have it either. So we would have to change the CI script to install it and also the readme so that users know to install it. And then we should also have instructions on running the gtests in README and run them in CI.

newling:
@nirvedhmeshram  -- IREE has googletest in third_party. I checked the build dir files and can confirm we're using the third_party googletest here. So I've remove the cmake configuration requirement.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
newling: I've moved it. FYI the only reason I've put this in a header at all is so that it can be tested. Because it's not really intended for end user use, I've put it in the 'detail' namespace. This seems to a be a common pattern in LLVM from what I can tell

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/AMDAIEUtilsTest.cpp
```diff
@@ -0,0 +1,52 @@
...
+  EXPECT_EQ(detail::findLargestFactor(/* num = */ 10000, /* max = */ 20001),
```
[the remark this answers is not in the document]
newling: Thanks for spotting, I must have been confused when writing this example... better tests added now

nirvedhmeshram (approved):
LGTM!

PR #243, opened 2024-03-22
Add support to peel the last iteration of the for loop

yzhang93:
This PR adds support to peel the last iteration out of the scf.for loop. This functionality is a special case of [peelForLoop](https://github.com/llvm/llvm-project/blob/556fe5f290ea88dcbb7ced16b0f057dcebce1fd0/mlir/lib/Dialect/SCF/Transforms/LoopSpecialization.cpp#L119) in llvm-project. However, there is a hard constraint in the upstream function that peeling only works if the step size doesn't divide the upper bound evenly. In our application, we'll need a function to peel the last iteration off in all situations. To not break the upstream patterns for loop peeling, a special function is added in this repository.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPeelForLoop.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPeelForLoop.cpp
@@ -14,6 +14,45 @@ namespace mlir::iree_compiler::AMDAIE {
 
 namespace {
 
+LogicalResult peelForLoopLastIteration(RewriterBase &b, scf::ForOp forOp,
+                                       scf::ForOp &lastIteration) {
+  RewriterBase::InsertionGuard guard(b);
+  auto lbInt = getConstantIntValue(forOp.getLowerBound());
+  auto ubInt = getConstantIntValue(forOp.getUpperBound());
+  auto stepInt = getConstantIntValue(forOp.getStep());
+
+  // Check again when the first iteration is already peeled off.
+  // Peeling is not needed if there is one or less iteration.
+  if (lbInt && ubInt && stepInt && ceil(float(*ubInt - *lbInt) / *stepInt) <= 1)
+    return failure();
+
+  AffineExpr lbSymbol, ubSymbol, stepSymbol;
+  bindSymbols(b.getContext(), lbSymbol, ubSymbol, stepSymbol);
+
+  // New upper bound: %ub - (%ub - %lb) mod %step
+  auto modMap =
+      AffineMap::get(0, 3, {ubSymbol - ((ubSymbol - lbSymbol) % stepSymbol)});
+  b.setInsertionPoint(forOp);
+  auto loc = forOp.getLoc();
+  Value splitBound = b.createOrFold<affine::AffineApplyOp>(
+      loc, modMap,
+      ValueRange{forOp.getLowerBound(), forOp.getUpperBound(),
+                 forOp.getStep()});
+
+  // Create ForOp for partial iteration.
+  b.setInsertionPointAfter(forOp);
+  lastIteration = cast<scf::ForOp>(b.clone(*forOp.getOperation()));
+  lastIteration.getLowerBoundMutable().assign(splitBound);
+  b.replaceAllUsesWith(forOp.getResults(), lastIteration->getResults());
+  lastIteration.getInitArgsMutable().assign(forOp->getResults());
+
+  // Set new upper loop bound.
+  b.modifyOpInPlace(forOp,
+                    [&]() { forOp.getUpperBoundMutable().assign(splitBound); });
+
+  return success();
+}
+
 class AMDAIEPeelForLoopPass
     : public impl::AMDAIEPeelForLoopBase<AMDAIEPeelForLoopPass> {
  public:
@@ -23,6 +62,8 @@ class AMDAIEPeelForLoopPass
 
   AMDAIEPeelForLoopPass() = default;
   AMDAIEPeelForLoopPass(const AMDAIEPeelForLoopPass &pass){};
+  AMDAIEPeelForLoopPass(const AMDAIEPeelForLoopOptions &options)
+      : AMDAIEPeelForLoopBase(options) {}
   void runOnOperation() override;
 };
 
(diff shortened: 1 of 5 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.h
```diff
@@ -23,7 +23,7 @@ enum class AIEPassPipeline : int32_t {
 };
 
 /// Enum for target tiling op.
-enum class TilingOp : int32_t { Matmul = 0, RhsCopy = 1, LhsCopy = 2 };
+enum class PeelLoopType : int32_t { First = 0, Last = 1, FirstLast = 2 };
```
Abhishek-Varma, line 26: I see there were no uses of `TilingOp` enum anyway.

With regards to `PeelLoopType` change the preceding comment.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -208,9 +208,23 @@ def AMDAIEPad :
 
 def AMDAIEPeelForLoop :
     Pass<"iree-amdaie-peel-for-loop", "func::FuncOp"> {
-  let summary = "Pass to peel the first iteration out of the scf.for loop.";
+  let summary = "Pass to peel the first or/and the last iteration out of the scf.for loop.";
   let constructor =
       "mlir::iree_compiler::AMDAIE::createAMDAIEPeelForLoopPass()";
+  let options = [
+    Option<"peelLoopType", "peel-loop-type",
```
Abhishek-Varma, line 215: Nit: `peel-loop-type` naming-wise suggests "type of the loop", to me.
How about `peeling-type` ?

This wouldn't matter in the greater scheme of work but I thought I should add a little nudge on this.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -208,9 +208,23 @@ def AMDAIEPad :
...
+      "mlir::iree_compiler::AMDAIE::PeelLoopType",
+      /*default=*/"mlir::iree_compiler::AMDAIE::PeelLoopType::First",
+      "Choose which type of loop peeling to perform",
+      [{::llvm::cl::values(
+        clEnumValN(mlir::iree_compiler::AMDAIE::PeelLoopType::First, "first",
+                   "Peel the first iteration."),
+        clEnumValN(mlir::iree_compiler::AMDAIE::PeelLoopType::Last, "last",
+                   "Peel the last iteration."),
+        clEnumValN(mlir::iree_compiler::AMDAIE::PeelLoopType::FirstLast, "firstLast",
```
Abhishek-Varma, line 224: Nit: I'm not sure we'd want camel-casing in pass invoking/flag options.

How about `first-last` or `first-and-last` ?

Abhishek-Varma (approved):
LGTM! Thanks! Only a few nit comments from my side.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
newling: Or 'first-then-last' ?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/peel_for_loop.mlir
```diff
@@ -46,17 +80,48 @@ func.func @no_peeling_example() -> i32 {
...
+//      PEEL-FIRST:     scf.yield %[[ADD]]
+//      PEEL-FIRST:   }
+//      PEEL-FIRST:   return %[[RESULT]]
+
+// -----
+
+#map = affine_map<(d0, d1)[s0] -> (s0, d0 - d1)>
+func.func @two_iteration_example() -> i32 {
+  %c0_i32 = arith.constant 0 : i32
+  %lb = arith.constant 0 : index
```
newling, line 103: Maybe to have better coverage, make a test (or modify this one) which doesn't start at a value for which lb % step != 0 ?

newling (commented):
Logic all looks good to me! Just a few design/doc nits.

MaheshRavishankar:
Not sure why the upstream peeling doesnt work out-of-the-box.

yzhang93:
> Not sure why the upstream peeling doesnt work out-of-the-box.

It is blocked by this condition https://github.com/llvm/llvm-project/blob/556fe5f290ea88dcbb7ced16b0f057dcebce1fd0/mlir/lib/Dialect/SCF/Transforms/LoopSpecialization.cpp#L133. It returns failure if the step size divides upper bound perfectly. In our application, we should do peeling irrespective of this condition.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPeelForLoop.cpp
```diff
@@ -42,17 +83,39 @@ void AMDAIEPeelForLoopPass::runOnOperation() {
...
+          return signalPassFailure();
+        }
+        break;
+      case PeelLoopType::FirstLast:
+        if (failed(scf::peelForLoopFirstIteration(rewriter, forOp, result))) {
+          forOp->emitOpError("failed to peel the first iteration.");
+          return signalPassFailure();
+        }
+        if (failed(peelForLoopLastIteration(rewriter, forOp, result))) {
+          LLVM_DEBUG(llvm::dbgs() << "Skip peeling for the last iteration.\n");
```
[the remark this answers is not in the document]
yzhang93: Here it doesn't send out signalPassFailure() is because we don't want to fail the pass if there are exactly two iterations of the loop. In this case, we don't actually run `peelForLoopLastIteration`,  since the remaining is the last iteration.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/peel_for_loop.mlir
yzhang93: Done.

newling (approved):

PR #245, opened 2024-03-22
Changes to get matmul with i8 operands (and i32 result) e2e working

newling:
Will also require this small change in aie: https://github.com/Xilinx/mlir-aie/pull/1149

I've verified that with these changes, I can add a test to our matmul script. 

Is there anywhere else we assume element types are >= 16 bits? Please let me know if so.


nirvedhmeshram (approved):
LGTM, I dont know of any other places with type bitwidth limits.

yzhang93:
Would love to see an e2e test. Maybe good to add to ci as well?

newling:
> Would love to see an e2e test. Maybe good to add to ci as well? https://github.com/nod-ai/iree-amd-aie/blob/main/build_tools/ci/run_matmul_test.sh

Will definitely add an e2e test, but I first need to get https://github.com/Xilinx/mlir-aie/pull/1149 landed.

Abhishek-Varma (approved):
> Is there anywhere else we assume element types are >= 16 bits? Please let me know if so.

Hi @newling  - you'd also need to reflect this change in the error reporting comment [here](https://github.com/nod-ai/iree-amd-aie/blob/bfe28758cce61fb59d0ea6b77694123467368ad2/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp#L163). LGTM otherwise.

PR #247, opened 2024-03-25
Fix affine map for peeling the last iteration

yzhang93:
This PR fixes the bug in the previous PR https://github.com/nod-ai/iree-amd-aie/pull/243, so that it works when (%ub - %lb) mod (%step) == 0.


nirvedhmeshram (approved):
This makes sense to me. Also we need this for peeling the first iteration too which would mean we cant use upstream utility for it?

yzhang93:
> This makes sense to me. Also we need this for peeling the first iteration too which would mean we cant use upstream utility for it?

I fixed a bug in upstream for peeling the first iteration, so we can use the upstream function for it.

PR #251, opened 2024-03-27
Bump mlir-air

erwei-xilinx:
New features in MLIR-AIR:
- Improved code coverage for small or asymmetrical GEMM sizes.
- Added a new pass `air-split-l2-memref` which optimizes for memtile buffer allocation, enabling for scaling the design to map to multiple columns.
- Bugfixes.


newling (approved):
We normally try to keep mlir-air and mlir-aie in sync, but it seems we've already fast-forwarded mlir-aie:

This updated mlir-air uses mlir-aie at `export HASH=2903588040333ea31c7a21251ae83bc0293c879a` (see utils/clone-mlir-aie.sh) which is dated `Mon Feb 26 09:25:51 2024 -0800`

But iree-amd-aie is already using a newer version: 57a29a806f43494191850c2e6339fd21fd54fd5c (see https://github.com/nod-ai/iree-amd-aie/tree/main/third_party) which is a March commit. 

So I don't have any objection to landing this, as mlir-air and mlir-aie are already out of sync and this bump brings them closer. 

Hoping we can bump again soon, if/when https://github.com/Xilinx/mlir-air/pull/512 lands.

erwei-xilinx:
> Hoping we can bump again soon, if/when https://github.com/Xilinx/mlir-air/pull/512 lands.

Thanks for the info, @newling . I think given that iree-amd-aie already bumped mlir-aie, then perhaps we wait for the MLIR-AIR PR to land before we bump MLIR-AIR from iree-amd-aie, so that we do not have to revert the mlir-aie version in iree-amd-aie.

newling:
Sounds good if you're ok with that. When Xilinx/mlir-air#512 lands, we bump both mlir-air and mlir-aie (to the hash in [utils/clone-mlir-aie.sh](https://github.com/Xilinx/mlir-air/pull/512/files#diff-0c7c73174cea88a203ef4ba84ee00beedeac13aeebe5fc9e5e66e7c4adfe2c1b))

PR #254, opened 2024-03-28
Make matmul tests take m,n,k directly 

newling:
@nirvedhmeshram I've been wanting to try this for a while, and your comment in the standup this morning made me just do it. 

If I understood correctly (sorry I was having audio issues) when you compile 2+ matmuls together, you hit runtime issues? 

This PR does:
1) removes the groups 'large' 'small' 'legacy_*' and now a test must be specified directly with m,n,k values. 
2) some minor superficial changes.

### DIFF
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -175,24 +175,42 @@ cd ${OUTPUT_DIR}
 function run_matmul_test() {
 
 
-  # Options without defaults:
+  # Options without defaults
+  # ========================
   local lhs_rhs_type=""
   local acc_type=""
   local m=""
   local n=""
   local k=""
 
-  # Options with defaults:
+  # Options with defaults
+  # =====================
+  # name_prefix: A prefix for the name of the test. The full test name will be
+  # extended with m,n,k if they are unique.
   local name_prefix="noprefix"
+
   local target_backend="amd-aie"
+
   local device="xrt"
+
   local peano_install_path="${PEANO}"
+
   local mlir_aie_install_path="${MLIR_AIE_INSTALL}"
+
   local vitis_path="${VITIS}"
+
   local pipeline="pad-pack"
+
+  # By default, the m,n,k provided are used, and there are no dynamic tensor
+  # dimensions.
   local dynamicity="static"
+
   local accumulate="false"
+
+  # By default we do not expect a compilation failure.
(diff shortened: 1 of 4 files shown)


runtime/src/iree-amd-aie/driver/xrt/direct_command_buffer.cc
```diff
@@ -353,6 +353,12 @@ static iree_status_t iree_hal_xrt_direct_command_buffer_dispatch(
           xrt::bo(*command_buffer->descriptor_sets[i].bindings[j],
                   command_buffer->descriptor_sets[i].lengths[j],
                   command_buffer->descriptor_sets[i].offsets[j]);
+
+      // TODO(jn) remove from PR. This is a temporary fix to avoid a
```
nirvedhmeshram, line 357: remove?

nirvedhmeshram (approved):
There is a trade-off here that we are taking the pressure away from runtime to be able to actually handle multiple dispatches at once. At the same time this allows us more flexibility in running different dtypes and shapes which is more important for the development phase we are in so lets do it this way. Thanks for the change!

newling:
> There is a trade-off here that we are taking the pressure away from runtime to be able to actually handle multiple dispatches at once.

That was my thinking too. Hmm. I guess I could keep the option to run a group of shapes too, and have a test commented out with a "TODO: resolve runtime issues". Because another advantage of having multiple matmuls in the vmfb might be test time? I'll play around with this option, and if this PR changes dramatically I'll ask for another review before landing.

nirvedhmeshram:
> > There is a trade-off here that we are taking the pressure away from runtime to be able to actually handle multiple dispatches at once.
> 
> That was my thinking too. Hmm. I guess I could keep the option to run a group of shapes too, and have a test commented out with a "TODO: resolve runtime issues". Because another advantage of having multiple matmuls in the vmfb might be test time? I'll play around with this option, and if this PR changes dramatically I'll ask for another review before landing.

Thanks, thats a good plan!

newling:
Hey @nirvedhmeshram I've updated this to accept multiple matmul shapes in a single test. I've also added options for the case where we (1) just want to compile and (2) expect a compilation failure, to better capture the cases commented out currently. I'm going to be away for 8 days: please feel free to land or change this if you want. Thanks!

nirvedhmeshram (approved):
LGTM

PR #255, opened 2024-03-29
[DELEGATE] Add a PDL matcher to match a linalg.matmul

nirvedhmeshram:
PDL matching has compile time performance benefit over transform dialect matching for large models. Here we are doing the matching a linalg level, mostly following what happened at the torch level in the upstream IREE matcher test. I was able to run the resulting IR through iree-compile and was able to delegate the matmul to AIE device with iree-run-module.

### DIFF
--- a/experimental/delegate/CMakeLists.txt
+++ b/experimental/delegate/CMakeLists.txt
@@ -53,9 +53,9 @@ iree_lit_test_suite(
   NAME
     lit
   SRCS
-   "mlp.mlir"
+    "mlp.mlir"
   DATA
-    "lianlg.pdl.mlir"
+    "linalg.pdl.mlir"
   TOOLS
     FileCheck
     iree-compile
--- a/experimental/delegate/CMakeLists.txt
+++ b/experimental/delegate/CMakeLists.txt
@@ -49,4 +49,17 @@ add_dependencies(iree-sample-deps
   mlp_bf16_aie_delegate
 )
 
+iree_lit_test_suite(
+  NAME
+    lit
+  SRCS
+   "mlp.mlir"
+  DATA
+    "lianlg.pdl.mlir"
+  TOOLS
+    FileCheck
+    iree-compile
+    iree-opt
+)
+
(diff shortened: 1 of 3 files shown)


experimental/delegate/CMakeLists.txt
```diff
@@ -49,4 +49,17 @@ add_dependencies(iree-sample-deps
   mlp_bf16_aie_delegate
 )
 
+iree_lit_test_suite(
+  NAME
+    lit
+  SRCS
+   "mlp.mlir"
+  DATA
+    "lianlg.pdl.mlir"
```
MaheshRavishankar, line 58: nit: typo `s/lianlg/linalg`

experimental/delegate/linalg.pdl.mlir
```diff
@@ -0,0 +1,122 @@
...
+
+  %m_op = pdl.operation "tensor.dim"(%lhs, %zero : !pdl.value, !pdl.value) -> (%index_type : !pdl.type)
+  %m = pdl.result 0 of %m_op
+  %n_op = pdl.operation "tensor.dim"(%rhs, %one : !pdl.value, !pdl.value) -> (%index_type : !pdl.type)
+  %n = pdl.result 0 of %n_op
+
+  %empty_op = pdl.operation "tensor.empty" (%m, %n : !pdl.value, !pdl.value) -> (%matmul_type : !pdl.type)
+  %empty = pdl.result 0 of %empty_op
+  
+  %fill_op = pdl.operation "linalg.fill" (%zero_f32, %empty : !pdl.value, !pdl.value) -> (%matmul_type : !pdl.type)
```
MaheshRavishankar, line 74: I think you can avoid all the `tensor.dim` matching and just match the `linalg.fill` and `arith.constant 0.0 : f32` , capturing the `tensor.empty` as an operand.

MaheshRavishankar (changes requested):

MaheshRavishankar (approved):

PR #259, opened 2024-04-01
Numerical test in loop 

newling:
Running this on the old driver, these are results from a few runs:

**With band-aid:**
fails after 6 iterations 
fails after 1906 iterations
frozen machine
fails after 1934 iterations
fails after 836 iterations
frozen machine

**If I remove the band-aid:**
fails after 1 iterations
fails after 3 iterations
fails after 2 iterations
... 

Without the band-aid, I see the classic row of zeros in the incorrect result. 

With the band-aid, the pattern is not zeros: looks like random bits, and over a larger area of C.

### DIFF
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -190,6 +190,10 @@ function run_matmul_test() {
         name="$2"
         shift 2
         ;;
+      --n_runs)
+        n_runs="$2"
+        shift 2
+        ;;
       --lhs_rhs_type)
         lhs_rhs_type="$2"
         shift 2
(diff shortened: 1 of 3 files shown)


nirvedhmeshram:
@newling looks like CI passes with the temp fix (band-aid) can we try without it?

newling:
> @newling looks like CI passes with the temp fix (band-aid) can we try without it?

Yeah, I'll set one off now.  

> @newling looks like CI passes with the temp fix (band-aid) can we try without it?

I modified this PR to remove the temp fix, and have just relaunched.

newling:
Yes I think this band-aid was only useful for the bug which we saw with the old driver. There might still be a bug somewhere in stack, but it has a a much lower frequency that the one we saw with the old driver, and I think it is orthogonal to this band-aid, so I'm going to close this PR.

PR #262, opened 2024-04-02
Fix FusePackIntoLoop pass

yzhang93:
This PR fixes the problem that the user of loop block argument is not linalg.generic. This PR is needed for pack-peel based pipeline.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
@@ -83,31 +83,36 @@ void AMDAIEFusePackIntoLoopPass::runOnOperation() {
     return signalPassFailure();
   }
 
-  // Based on the `fusePackDepth`, we would continue to greedily fuse the
-  // producer tensor.pack ops.
   LoopLikeOpInterface loops = cast<LoopLikeOpInterface>(scfLoopOp);
+  auto loopBody = useSCFFor ? cast<scf::ForOp>(loops).getBody()
+                            : cast<scf::ForallOp>(loops).getBody();
+
+  // Based on the `fusePackDepth`, we would greedily fuse the producer
+  // tensor.pack ops.
   for (unsigned depth = 1; depth <= fusePackDepth; depth++) {
     // Search the compute op and its producer slices.
-    BlockArgument bbArg = loops.getRegionIterArgs()[0];
-    SmallVector<tensor::ExtractSliceOp> sliceOps;
-    SmallVector<Operation *> allUsers(bbArg.getUsers().begin(),
-                                      bbArg.getUsers().end());
-    while (!allUsers.empty()) {
-      auto user = allUsers.pop_back_val();
-      if (isa<tensor::ExtractSliceOp>(user)) {
-        allUsers.insert(allUsers.begin(), user->getUsers().begin(),
-                        user->getUsers().end());
-        continue;
-      }
-
-      if (auto genericOp = dyn_cast<linalg::GenericOp>(user)) {
-        for (auto [index, operand] : llvm::enumerate(genericOp.getOperands())) {
-          FailureOr<tensor::ExtractSliceOp> sliceOp =
-              getTensorExtractSliceDefiningOp(operand);
-          if (!failed(sliceOp)) {
-            sliceOps.push_back(sliceOp.value());
+    linalg::GenericOp genericOp;
+    loopBody->walk<WalkOrder::PostOrder, ReverseIterator>(
+        [&](linalg::LinalgOp op) {
+          if (isa<linalg::GenericOp>(op) &&
+              linalg::isaContractionOpInterface(op)) {
+            genericOp = cast<linalg::GenericOp>(op);
+            return WalkResult::interrupt();
           }
-        }
+          return WalkResult::advance();
+        });
+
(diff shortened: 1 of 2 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
```diff
@@ -78,14 +78,28 @@ void AMDAIEFusePackIntoLoopPass::runOnOperation() {
...
+    SmallVector<Operation *> allUsers(bbArg.getUsers().begin(),
+                                      bbArg.getUsers().end());
+    while (!allUsers.empty()) {
+      auto user = allUsers.pop_back_val();
+      if (isa<tensor::ExtractSliceOp>(user)) {
+        allUsers.insert(allUsers.begin(), user->getUsers().begin(),
+                        user->getUsers().end());
+        continue;
+      }
+
```
Abhishek-Varma, line 102: Please add a comment about the logic.

From what it seems : it's taking up initial users -> adding more users to the beginning of the search party if it bumps into tensor.extract_slice

If we expect only one `linalg.generic` in the body of the for loop - why don't we encash on that?

I'm trying to understand and see if we can get rid of the expensive `allUsers.insert` operation.

Abhishek-Varma (commented):
I'll take another look at this - especially the test case since they seemed modified.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
yzhang93: Thanks @Abhishek-Varma for providing the suggestion! I've simplified the code logic.

yzhang93:
> I'll take another look at this - especially the test case since they seemed modified.

The new added test was failed with the previous codes because it couldn't find the generic op by looking for the user of loop bbArg. This PR changed the logic of the search codes so it will work for such test cases.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
```diff
@@ -78,22 +78,41 @@ void AMDAIEFusePackIntoLoopPass::runOnOperation() {
...
+  auto loopBody = useSCFFor ? cast<scf::ForOp>(loops).getBody()
+                            : cast<scf::ForallOp>(loops).getBody();
+
+  // Based on the `fusePackDepth`, we would greedily fuse the producer
+  // tensor.pack ops.
   for (unsigned depth = 1; depth <= fusePackDepth; depth++) {
-    // Search the compute op and its producer slices within the For loop.
-    BlockArgument bbArg = loops.getRegionIterArgs()[0];
+    // Search the compute op and its producer slices.
     SmallVector<tensor::ExtractSliceOp> sliceOps;
```
Abhishek-Varma, line 94: It's a good practice to declare a variable closest to its first use.

This way if the code had to exit early it wouldn't end up declaring variables which had no use till that time.

You can add this at [this line](https://github.com/nod-ai/iree-amd-aie/blob/caec355eecddae4e57f68d63b5f11b7f91ada949/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp#L110).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
```diff
@@ -78,22 +78,41 @@ void AMDAIEFusePackIntoLoopPass::runOnOperation() {
...
-        for (auto [index, operand] : llvm::enumerate(genericOp.getOperands())) {
-          FailureOr<tensor::ExtractSliceOp> sliceOp =
-              getTensorExtractSliceDefiningOp(operand);
-          if (!failed(sliceOp)) {
-            sliceOps.push_back(sliceOp.value());
+    linalg::GenericOp genericOp;
+    loopBody->walk<WalkOrder::PostOrder, ReverseIterator>(
+        [&](linalg::LinalgOp op) {
+          if (isa<linalg::GenericOp>(op) &&
+              linalg::isaContractionOpInterface(op)) {
```
Abhishek-Varma, line 99: Why are we placing a check on `ContractionOpInterface` ?

If there was some other GenericOp with defining ops from [tensor.extract_slice, tensor.pack] wouldn't we want to fuse the tensor.pack ops in that case as well?

Abhishek-Varma (commented):
Thanks for adding the new logic @yzhang93 !

A few comments.

MaheshRavishankar (approved):

Abhishek-Varma (approved):
Approving this since it looks good to me overall and I'd not want to block the merging!
Just few comments to address from the earlier thread.

Thanks @yzhang93 !

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
yzhang93: This is in case we have something like linalg.generic (matmul) + linalg.generic (elementwise), and we don't want to take the elementwise op.

PR #266, opened 2024-04-09
Bump mlir-air; fixup nondeterministic AIE routing error

erwei-xilinx:
Summary of the issue:
- With GEMM E2E tests big enough to use 4x4 AIE tiles and 4 AIE memtiles, nondeterministic failures happen indicating failure to generate a legal routing.

Summary of the fix:
- Identified an MLIR-AIR pass which buffers `memref.alloc` ops for code motion using `std::map<Operation *, Operation *>`.
- `std::map` sorts the keys, leading to nondeterministic sequence of `memref.alloc` ops being generated after the code motion.
- Those output IRs, with some `memref.alloc` ops randomly permutated, are all legal MLIR-AIR designs. But when AIE memtile buffer allocator maps them to physical memtiles as round robin, there are scenarios where the routings become unnecessarily complicated, and encountered routing congestion.
- Current solution is to replace the use of `std::map<Operation *, Operation *>` with `std::vector<std::pair<Operation *, Operation *>>`, which comes with a deterministic ordering of elements.


erwei-xilinx:
@yzhang93 , @nirvedhmeshram , the CI was repeated for 10 times and did not trigger the error anymore.

yzhang93 (approved):
Thanks for fixing the issue!

PR #267, opened 2024-04-09
Add logical objectfifo DMA ops and conversion from AIR

jtuyls:
This is a first PR in a sequence to bring in compilation through logical objectfifos, using a peeled matmul as an E2E example. See following branch for all new functionality: https://github.com/nod-ai/iree-amd-aie/tree/jornt_peeled_matmul_e2e. This PR adds basic infrastructure for AMDAIE ops with two core ones:
- `AMDAIE::LogicalObjectFifoFromMemrefOp`: a new op which creates a logical objectfifo from a memref. The logical objectfifo encapsulates the memref and adds asynchronous access to it through produce/consume or acquire/release ops, which will be added later on. 
- `AMDAIE::DmaCpyNdOp`: Similar to the AIR DMA op this represents a strided ND copy operation, but operates on logical objectfifos and with an index result. 

To demonstrate usage of these new ops, this PR adds a pass that converts `AIR::DmaMemcpyNdOp` into `AMDAIE::DmaCpyNdOp`. This is the first lowering pass run after `pack-to-dma` and `copy-to-dma` in this new pipeline demonstrated in the branch above.

@nirvedhmeshram @yzhang93 @Abhishek-Varma , could you help with reviewing?

cc @MaheshRavishankar


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAIRDmaToAMDAIEDma.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include "iree-amd-aie/IR/AMDAIEDialect.h"
+#include "iree-amd-aie/IR/AMDAIEOps.h"
+#include "iree-amd-aie/Transforms/Passes.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Support/LogicalResult.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+#define DEBUG_TYPE "iree-amdaie-air-dma-to-amdaie-dma"
+
+using namespace xilinx;
```
nirvedhmeshram, line 17: I dont think you need this as we should explicitly call out the xilinx namespace ops so its clear they are coming from mlir-air (which as far as I see you already are).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAIRDmaToAMDAIEDma.cpp
```diff
@@ -0,0 +1,78 @@
...
+    auto dst = rewriter.create<AMDAIE::LogicalObjectFifoFromMemrefOp>(
+        rewriter.getUnknownLoc(), AMDAIELogicalObjectFifoType::get(dstType),
+        op.getDst());
+
+    rewriter.setInsertionPoint(op);
+    rewriter.create<AMDAIE::DmaCpyNdOp>(
+        op.getLoc(), rewriter.getIndexType(), dst, op.getDstOffsets(),
+        op.getDstSizes(), op.getDstStrides(), src, op.getSrcOffsets(),
+        op.getSrcSizes(), op.getSrcStrides());
+    rewriter.eraseOp(op);
```
nirvedhmeshram, line 46: you can use `replaceOpWithNewOp` here so that the setInsertionPoint and eraseOp is not needed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDialect.cpp
```diff
@@ -22,9 +26,126 @@ struct AMDAIEDialectOpAsmInterface : public OpAsmDialectInterface {
...
+}
+
+mlir::MemRefType AMDAIELogicalObjectFifoType::getElementType() {
+  // 'getImpl' returns a pointer to the internal storage instance.
+  return getImpl()->elementType;
+}
+
+size_t AMDAIELogicalObjectFifoType::getStaticSize() {
+  auto shape = getElementType().getShape();
+  return std::accumulate(shape.begin(), shape.begin(), 1,
```
nirvedhmeshram, line 147: This seems wrong, shouldnt the second argument be shape.end()

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDialect.h
```diff
@@ -19,4 +19,37 @@
...
+
+  /// This method is used to verify the construction invariants.
+  static mlir::LogicalResult verify(
+      llvm::function_ref<mlir::InFlightDiagnostic()> emitError,
+      mlir::MemRefType elementType);
+
+  static constexpr llvm::StringLiteral name = "amdaielogicalobjectfifo";
+  /// Returns the element type of this LogicalObjectFifoType.
+  mlir::MemRefType getElementType();
+  ///
```
nirvedhmeshram, line 49: missing comment on why this function is needed?

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDialect.cpp
```diff
@@ -22,9 +26,126 @@ struct AMDAIEDialectOpAsmInterface : public OpAsmDialectInterface {
...
+    MemRefType elementType) {
+  // Call into a helper 'get' method in 'TypeBase' to get an uniqued instance
+  // of this type.
+  MLIRContext *ctx = elementType.getContext();
+  return Base::get(ctx, elementType);
+}
+
+LogicalResult AMDAIELogicalObjectFifoType::verify(
+    function_ref<InFlightDiagnostic()> emitError, MemRefType elementType) {
+  return success();
```
nirvedhmeshram, line 137: do we want to do some actual verification here, e.g fail on dynamic shaped memrefs?

nirvedhmeshram (changes requested):
Looking good overall. Thanks.

nirvedhmeshram (commented):
One more thing, we need to add round-trip tests to make sure the parser code is correctly tested. Also if there are are certain cases that are expected to fail verification then we also add invalid ir tests. Here is an example from IREE
https://github.com/openxla/iree/blob/main/compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/roundtrip.mlir
https://github.com/openxla/iree/blob/main/compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/invalid.mlir

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAIRDmaToAMDAIEDma.cpp
jtuyls: Didn't work for me as `AMDAIE::DmaCpyNdOp` has a result and the air version doesn't. I think `replaceOpWithNewOp` doesn't work in that case, but let me know if it should or if another method applies in that case.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDialect.cpp
jtuyls: Yes, dynamic shapes might be a good one to check. Will add it.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDialect.h
jtuyls: I will remove the method as it's not needed yet. Will introduce it when needed.

jtuyls:
> One more thing, we need to add round-trip tests to make sure the parser code is correctly tested. Also if there are are certain cases that are expected to fail verification then we also add invalid ir tests. Here is an example from IREE https://github.com/openxla/iree/blob/main/compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/roundtrip.mlir https://github.com/openxla/iree/blob/main/compiler/src/iree/compiler/Dialect/LinalgExt/IR/test/invalid.mlir

Thanks for the pointers. I will add additional tests.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/test/invalid.mlir
```diff
@@ -0,0 +1,101 @@
+// RUN: iree-opt --split-input-file --verify-diagnostics %s
+
```
jtuyls, line 2: I tried to add the following test to verify dynamic shape failure. The error seems to be returned, but somehow the test still fails. Couldn't figure out what I am doing wrong immediately, so if anyone has an idea?
```
// expected-error @+1 {{should encapsulate static memref}}
func.func @test(!amdaie.logicalobjectfifo<memref<?x8x16xi32, 1>>)
```

jtuyls:
@nirvedhmeshram Could you have another look?

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/test/invalid.mlir
yzhang93: Have you figured this out? I didn't see build failure.

yzhang93 (approved):
LGTM, thanks!

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/test/invalid.mlir
yzhang93: Have you figured this out? I didn't see

jtuyls: No, I didn't put it into the file as it doesn't pass for me locally. However, it reports the error that is expected..

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAIRDmaToAMDAIEDma.cpp
nirvedhmeshram: ok, it is fine the way it is.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/test/invalid.mlir
nirvedhmeshram: can you add the FileCheck, we can help if we see the failure in CI.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDmaOpInterface.td
```diff
@@ -0,0 +1,94 @@
...
+        Common verifier for doubly-strided operations. Checks whether source and target
+        access descriptors (`offsets`, `sizes`, `strides`) have the same number of
+        dimensions.
+      }],
+      /*retTy=*/"::mlir::LogicalResult",
+      /*methodName=*/"verifyCommon",
+      /*args=*/(ins),
+      /*methodBody=*/"",
+      /*defaultImplementation=*/[{
+        if ($_op.getTargetSizes().size() != $_op.getTargetOffsets().size()) {
```
nirvedhmeshram, line 72: I think air.dma op allows offsets to be missing and then they are assumed zero? should we pass for those cases?

nirvedhmeshram: you could handle it in the lowering and gaurantee that you always have the zero offsets in which case this verifier is correct.

jtuyls: We could relax the checks or handle it when we encounter this case? For now, I haven't seen this yet.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/test/invalid.mlir
jtuyls: Added it, waiting for a runner.. Could it be offline as I don't see anything else running?

jtuyls: I believe I figured it out :) see https://github.com/nod-ai/iree-amd-aie/pull/267/commits/5aaea5c0a0cdfd1f36e21b3bae8067eefaeaadc4

jtuyls: Yeah, CI passed

nirvedhmeshram: there is no FileCheck here?

nirvedhmeshram: oh I see, thats how the invalid tests work look likes.

nirvedhmeshram (approved):
LGTM, thanks!

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDialect.cpp
```diff
@@ -8,6 +8,10 @@
 
 #include "iree-amd-aie/IR/AMDAIEAttrs.h"
 #include "iree-amd-aie/IR/AMDAIEDialect.cpp.inc"
+#include "iree-amd-aie/IR/AMDAIETypes.cpp.inc"
```
MaheshRavishankar, line 11: Typically these are kept separately from the actual .h includes cause these are tablegen generated . See examples here https://github.com/openxla/iree/blob/954cb36d8201638823a08d0c65eb35325758c730/compiler/src/iree/compiler/Dialect/HAL/IR/HALTypes.h#L30

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDialect.cpp
```diff
@@ -22,9 +26,120 @@ struct AMDAIEDialectOpAsmInterface : public OpAsmDialectInterface {
   }
 };
 
+/// Parse an instance of a type registered to the AMDAIE dialect.
+/// Parse an AMDAIE type in the following forms:
+///   AMDAIE-type
+///         ::= `logicalobjectfifo` `<` type `>`
+static OptionalParseResult amdaieTypeParser(DialectAsmParser &parser,
```
MaheshRavishankar, line 33: Is there a reason why you used a custom type parser and printer instead of just doing `assemblyFormat` (https://mlir.llvm.org/docs/DefiningDialects/AttributesAndTypes/#adding-a-new-attribute-or-type-definition) . I think its great you went and looked at how to do it. It is useful to have this know how, but its easier to maintain with  just writing the assembly format.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDmaOpInterface.td
```diff
@@ -0,0 +1,94 @@
...
+        stride.
+      - `strides`: defines the strides or number of elements to be skipped to go to the
+        next access in every dimension.
+  }];
+  let cppNamespace = "mlir::iree_compiler::AMDAIE";
+
+  let methods = [
+    InterfaceMethod<
+      /*desc=*/"Returns ...",
+      /*retTy=*/"::mlir::ValueRange",
```
MaheshRavishankar, line 34: While we are adding these maybe just do these the same way `subview` does these things (https://github.com/llvm/llvm-project/blob/2996cb7970fb536d7db1088d571dc8d2f9ad35fa/mlir/include/mlir/Dialect/MemRef/IR/MemRefOps.td#L1833) . This makes it much easier to handle constant propagation etc. especially if you want to have static sizes.

MaheshRavishankar (commented):
Nice! I left a few comments as well. Sorry I wasnt able to review before it landed. Some of these would be good changes to incorporate. Its easier to do it now rather than later where things get a bit more entrenched.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDialect.cpp
jtuyls: No, initially, I had an unsuccessful try with tablegen, probably due to unfamiliarity, and then  I based the implementation on the mlir-aie objectfifo type that uses this more verbose style. I created a new PR to use tablegen instead: https://github.com/nod-ai/iree-amd-aie/pull/270

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDialect.cpp
jtuyls: I think I got what you mean and I addressed this here: https://github.com/nod-ai/iree-amd-aie/pull/270

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDmaOpInterface.td
jtuyls: This should be addressed in this new PR: https://github.com/nod-ai/iree-amd-aie/pull/271.

PR #269, opened 2024-04-11
Changes to pack-peel pipeline for matmul-elementwise ops fusion

yzhang93:
- Add data layout propagation in the pack-peel pipeline, so the pack ops can be propagated for elementwise ops.
- Modify the tileAndFusePass to include an option to not tile the elementwise ops.
- After these changes the generated IR before fusion is as https://gist.github.com/yzhang93/2d83d40f4cec20f58a135af73209759e
- TODOs:
1) Modify bufferizeToAllocation pass to include memory promotion for elementwise ops.
2) Add a method to automatically detect matmul-elementwise vs matmul-only dispatches, so that we can set different peel options. For matmul-only dispatch just the first iteration needs to be peeled, while for matmul-elementwise dispatch both the first and the last iterations need to be peeled.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
@@ -9,6 +9,7 @@
 #include "mlir/Dialect/GPU/IR/GPUDialect.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Linalg/Transforms/TilingInterfaceImpl.h"
+#include "mlir/Dialect/Linalg/Utils/Utils.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/Dialect/SCF/Transforms/TileUsingInterface.h"
@@ -57,7 +58,9 @@ static bool isTilingReductionDimension(TilingInterface consumerOp,
 }
 
 static bool consumerToSkip(TilingInterface op) {
-  if (isa<linalg::CopyOp>(op) || isa<tensor::UnPackOp>(op)) return true;
+  if (isa<linalg::CopyOp>(op) || isa<tensor::PackOp>(op) ||
+      isa<tensor::UnPackOp>(op))
+    return true;
   return false;
 }
 
@@ -107,6 +110,9 @@ void AMDAIETileAndFusePass::runOnOperation() {
     // Find the next consumer op if it does not have loops OR it is from
     // the skip ops list which currently contains linalg.copy and tensor.unpack.
     if (op.getLoopIteratorTypes().empty() || consumerToSkip(op))
+      return WalkResult::advance();
+    if (!tileElementwise && isa<linalg::GenericOp>(op) &&
+        isElementwise(cast<linalg::LinalgOp>(op.getOperation())))
       return WalkResult::advance();
     consumerOp = op;
     return WalkResult::interrupt();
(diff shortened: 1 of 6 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
```diff
@@ -108,6 +111,9 @@ void AMDAIETileAndFusePass::runOnOperation() {
     // the skip ops list which currently contains linalg.copy and tensor.unpack.
     if (op.getLoopIteratorTypes().empty() || consumerToSkip(op))
       return WalkResult::advance();
+    if (!tileElementwise && isa<linalg::GenericOp>(op) &&
```
MaheshRavishankar, line 114: Can you add some comments as to what this is doing?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/propagate_data_layout.mlir
```diff
@@ -5,36 +5,86 @@
...
+      linalg.yield %9 : i32
+    } -> tensor<64x64xi32>
+    scf.forall.in_parallel {
+      tensor.parallel_insert_slice %8 into %arg5[%arg3, %arg4] [64, 64] [1, 1] : tensor<64x64xi32> into tensor<1024x1024xi32>
+    }
+  } {mapping = [#gpu.block<y>, #gpu.block<x>]}
+  return %1 : tensor<1024x1024xi32>
+}
+
+// CHECK-LABEL: matmul_elementwise_1024x1024x512_i8xi8xi32
```
MaheshRavishankar, line 84: I know the existing tests dont do that but it would be better to have better test here... Basically capture the results of the first generic and make sure it is used in the second generic op.

MaheshRavishankar (changes requested):
Thanks. A  few minor comments.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
yzhang93: Refactored this part of codes and added some comments.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/propagate_data_layout.mlir
yzhang93: Done.

Abhishek-Varma (approved):
LGTM.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
```diff
@@ -108,6 +111,15 @@ void AMDAIETileAndFusePass::runOnOperation() {
     // the skip ops list which currently contains linalg.copy and tensor.unpack.
     if (op.getLoopIteratorTypes().empty() || consumerToSkip(op))
       return WalkResult::advance();
+
+    // For matmul + elementwise dispatch, we use flag `tileElementwise` to
+    // indicate whether we want to tile the elementwise op. If flag
+    // `tileElementwise == false`, and the linalg op is an elementwise op, it
+    // will advance to find the next target op for tiling.
+    auto linalgOp = dyn_cast_or_null<linalg::LinalgOp>(op.getOperation());
```
MaheshRavishankar, line 119: Isnt tile elementwise always false. In which case we are just avoiding elementwise op tiling always? If so, I'd just make that the case by default and avoid all the complexity with the flag. I see why you have it. Its OK to leave it in there too.

MaheshRavishankar (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
yzhang93: In the current setting, for the first level we'd like to tile both matmul and elementwise ops, in which case `tileElementwise == true`, and when it finds the elementwise op the walk function will be interrupted.

PR #271, opened 2024-04-15
Update DoublyStridedOpInterface to include both static and dynamic access attributes

jtuyls:
Addresses: https://github.com/nod-ai/iree-amd-aie/pull/267#r1563241598

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDmaOpInterface.td
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDmaOpInterface.td
@@ -106,7 +106,7 @@ def DoublyStridedOpInterface : OpInterface<"DoublyStridedOpInterface"> {
       /*desc=*/[{
         Return a vector of all the static or dynamic target offsets of the op.
       }],
-      /*retTy=*/"::llvm::SmallVector<::mlir::OpFoldResult, 4>",
+      /*retTy=*/"::llvm::SmallVector<::mlir::OpFoldResult>",
       /*methodName=*/"getTargetMixedOffsets",
       /*args=*/(ins),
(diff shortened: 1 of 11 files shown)


jtuyls:
@MaheshRavishankar Do you want to review this PR?

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDmaOpInterface.td
```diff
@@ -25,70 +25,233 @@ def DoublyStridedOpInterface : OpInterface<"DoublyStridedInterface"> {
...
+      /*methodBody=*/"",
+      /*defaultImplementation=*/[{
+        return $_op.getTargetStaticStrides();
+      }]
+    >,
+    InterfaceMethod<
+      /*desc=*/[{
+        Return a vector of all the static or dynamic target offsets of the op.
+      }],
+      /*retTy=*/"::llvm::SmallVector<::mlir::OpFoldResult, 4>",
```
MaheshRavishankar, line 109: Nit: drop the `, 4` . What ends up happening is if you do something like

```
SmallVector<OpFoldResult, 5> ... = getTargetMixedOffsets();
```

then the move methods dont kick in and it creates a copy. So easiest is to just drop the size template parameter.

MaheshRavishankar (approved):
Oh nice idea having a new interface for this! Cool!

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDmaOpInterface.td
jtuyls: Cool, thanks for pointing this out! Updated the PR.

Issue #273, opened 2024-04-15
[DRAFT] i8 matmul m=n=32 k=64: numerical failure 


newling:
This issue can be closed when the corresponding test passes

newling:
Drifted too far for this to still be relevant

PR #274, opened 2024-04-15
Numerical failure in i8 matmul: m=n=32 k=64 

newling:
This PR adds this failing test, and labels it as an expected numerical failure. By adding it like this we might be more likely to track and fix it. Issue created: https://github.com/nod-ai/iree-amd-aie/issues/273


newling:
This passes for me locally. i.e. the numerical test fails for me locally. So there's something between CI and my local setup.

PR #276, opened 2024-04-16
[Delegate] Changed to use a matmul shape for OPT

daveliddell:
Enhanced delegate to include the following new features:

- Now using a 4x4 vector matmul of shape 8x768x768, stored in Azure
- dtypes of model and AIE kernel inputs and outputs now defined in one place
- All tensor conversion and copying code templatized to automate those functions and use minimal implementations when possible
- New PDL transform to convert 8x768x768 matmul in linalg MLIR to external func call

See updated README.md for details.


experimental/delegate/kernels/CMakeLists.txt
```diff
@@ -15,6 +15,8 @@ message(STATUS "Processing AIE Delegate Kernels")
 # List of all kernels stored in Azure (without file extension)
 set(AIE_DELEGATE_KERNELS
     "matmul/matmul-bf16-256x256x256-v1"
+    "matmul/matmul-bf16-8x768x768-v1"  # OPT ref vec matmul (doesn't work)
```
nirvedhmeshram, line 18: we can remove this if its not functional?

nirvedhmeshram (commented):
I am wondering if its worth waiting for the bump and then landing this, with the new commands, e.g (iree-compile instead of iree-opt) and maybe a bash script that runs on CI and tests the output as well? That being said if this not getting merged is blocking you from making progress on next steps we can check this in. Let me know.

daveliddell:
> I am wondering if its worth waiting for the bump and then landing this, with the new commands, e.g (iree-compile instead of iree-opt) and maybe a bash script that runs on CI and tests the output as well? That being said if this not getting merged is blocking you from making progress on next steps we can check this in. Let me know.

Entirely up to you!  If this PR will get in your way, I can hold it.  Not checking it in is not stopping me from making progress in other areas.  Actually, not having the bump is more of an impediment, as it looks like I'll have to replicate more of the work that went into IREE main to get a full model working.

Anyway, this PR should be independent of the bump if you don't care either way.  I believe it works with a single matmul with the pinned IREE.  I don't have a CI test yet for a couple of reasons.  First, I'm not sure about the readiness of the CI test environment to accept a second e2e test, though I think James might have been working in this area.  Second, the way the delegate .so finds its AIE artifacts (xclbin and insts.txt) is incompatible with the way the CI manages xclbins, so I'm going to have to come up with a new solution.

experimental/delegate/mlp_aie_bf16_plugin.cpp
```diff
@@ -256,33 +421,18 @@ int setupNPUAccelerator() {
...
+    A_DATATYPE *bufA = xrtState->boA.map<A_DATATYPE *>();
+    TensorCopier<ModelLhsDType, A_DATATYPE>::copy(bufA, params->lhs.get(), aVolume);
     B_DATATYPE *bufB = xrtState->boB.map<B_DATATYPE *>();
-    for (int i = 0; i < K; i++)
-        for (int j = 0; j < N; j++)
-            *(bufB + i * N + j) = toBfloat16(params->rhs.getElement(i, j, N));
+    TensorCopier<ModelRhsDType, B_DATATYPE>::copy(bufB, params->rhs.get(), bVolume);
 
-    // copy output to XRT BO
+    // copy output to XRT BO (is this really necessary?)
```
nirvedhmeshram, line 432: maybe try without, although really depends on  the kernel implementation.

experimental/delegate/mlp_aie_bf16_plugin.cpp
```diff
@@ -256,33 +421,18 @@ int setupNPUAccelerator() {
...
-        }
-        // std::cout << "]," << std::endl;
-    }
 
-    cnt = 0;
-
-    // quantize and copy input to XRT BO
+    // Copy inputs to kernel input BOs
+    A_DATATYPE *bufA = xrtState->boA.map<A_DATATYPE *>();
+    TensorCopier<ModelLhsDType, A_DATATYPE>::copy(bufA, params->lhs.get(), aVolume);
```
nirvedhmeshram, line 428: I believe none of these copies should be necessary with this API call, youi can just mapo your host buffer to xrt buffer?
https://xilinx.github.io/XRT/master/html/xrt_native.main.html#_CPPv4N3xrt2bo2boERKN3xrt6deviceEPv6size_tN2bo5flagsE12memory_group

nirvedhmeshram (changes requested):
Took a closer look. Fine to land after addressing the comments, no need to wait for the bump.

daveliddell:
> I am wondering if its worth waiting for the bump and then landing this, with the new commands, e.g (iree-compile instead of iree-opt) and maybe a bash script that runs on CI and tests the output as well? That being said if this not getting merged is blocking you from making progress on next steps we can check this in. Let me know.

Wanted to address the CI test comment for the record: I don't have a CI test yet for a couple of reasons.  First, I'm not sure about the readiness of the CI test environment to accept a second e2e test, though I think James might have been working in this area.  Second, the way the delegate .so finds its AIE artifacts (xclbin and insts.txt) is incompatible with the way the CI manages xclbins, so I'm going to have to come up with a new solution.  We should discuss, but I'm thinking maybe environment variable since I'm not aware of other options for passing location info from IREE to the delegate .so.

daveliddell:
Also for the record: AIE delegate post-bump is now dependent on an IREE fix that hasn't landed yet.  See https://github.com/iree-org/iree/issues/17177.

daveliddell:
Started this PR over again with a cleaner history

PR #279, opened 2024-04-17
[WIP] Does CI pass when multiple xclbins in vmfb?

newling:


newling:
I guess the answer is no. I think the plan is to use a single xclbin for multiple dispatches, so this isn't something I'm going to dig into deeper...

PR #280, opened 2024-04-18
Add pass which inserts AIE workgroup and cores

jtuyls:
Adds a pass which inserts AIE workgroup ops around `scf::forall` with thread mapping (can be extended to blocks as well in the future). The workgroup consists of two parts:

1.  A control code region which will contain the uController control code instructions to drive the AIE array.
2.  An AIE configuration region (I call this an AIE state machine), which will be converted into AIE cores' elfs + a set of fixed dma instructions.

The idea of creating this workgroup is that subsequent passes will:
- Lower the AIE configuration region to AIE core and fixed DMA instructions
- Add required control and synchronization ops into the control code region 
- Move external temporal loops into the control code region
- Recognize and fuse sibling workgroups

This pass comes with some new ops:

- `WorkgroupOp`: designates a region around the parallel loops selected for parallelization across the AIE array dimensions (columns, rows). For now, the loops with thread mapping are selected, but this can be extended in the future to blocks as well.
- `ControlCodeOp`: each workgroup will contain one control code region op which will contain the uController control code instructions to drive the AIE array. This operation in inserted in every `WorkgroupOp` through the build methods to avoid workgroups without corresponding control code. The `ControlCodeOp` is the terminator as well for easy retrievability + deterministic location.
- `CoreOp`: Represents an AIE core on a certain location (column, row) within the AIE array and the operations which should be executed on this core. Note that this operation differs from the `CoreOp` defined in the AIE dialect in that this op accepts the column and row as indices instead of integers (via a TileOp in the AIE dialect). Eventually this `AMDAIE_CoreOp` will be converted in the `AIE_CoreOp` when we have static coordinates.
- `EndOp`: A generic terminator for AMDAIE ops' regions, used inside the `CoreOp`.
- `LogicalObjectFifoConsume`: Operation consuming the result of a DMA operation. This is a blocking operation,
    waiting for the DMA to produce data. Typically, this operation will reside inside a `CoreOp` to synchronize with external DMA operations producing data into the respective core's local memory.
- `LogicalObjectFifoProduce`: Operation producing the input of a DMA operation. This is a release-type operation,
    where the DMA will be waiting for the data to be produced. Typically, this operation will reside inside a `CoreOp` to synchronize with external DMA operations waiting for data from the respective core's local memory to be released.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_aie_workgroup.mlir
```diff
@@ -0,0 +1,98 @@
...
+    scf.forall (%arg0, %arg1) in (1, 1) {
+      scf.forall (%arg2, %arg3) in (1, 2) {
+        %9 = affine.apply #map(%arg2)
+        %10 = affine.apply #map(%arg3)
+        %11 = amdaie.dma_cpy_nd(%3[] [] [], %1[%9, %c0] [%c32, %c64] [%c1024, %c1]) : (!amdaie.logicalobjectfifo<memref<32x64xi32, 1>>, !amdaie.logicalobjectfifo<memref<32x1024xi32>>)
+        %12 = amdaie.dma_cpy_nd(%2[] [] [], %0[%c0, %10] [%c64, %c32] [%c64, %c1]) : (!amdaie.logicalobjectfifo<memref<64x32xi32, 1>>, !amdaie.logicalobjectfifo<memref<1024x64xi32>>)
+        linalg.fill ins(%c0_i32 : i32) outs(%alloc_1 : memref<4x8x4x8xi32, 2>)
+        %13 = amdaie.dma_cpy_nd(%5[%c0, %c0, %c0, %c0] [%c8, %c8, %c4, %c8] [%c256, %c32, %c8, %c1], %3[%c0, %c0, %c0, %c0] [%c8, %c8, %c4, %c8] [%c8, %c256, %c64, %c1]) : (!amdaie.logicalobjectfifo<memref<8x8x4x8xi32, 2>>, !amdaie.logicalobjectfifo<memref<32x64xi32, 1>>)
+        %14 = amdaie.dma_cpy_nd(%4[%c0, %c0, %c0, %c0] [%c4, %c8, %c8, %c8] [%c512, %c64, %c8, %c1], %2[%c0, %c0, %c0, %c0] [%c4, %c8, %c8, %c8] [%c8, %c256, %c32, %c1]) : (!amdaie.logicalobjectfifo<memref<4x8x8x8xi32, 2>>, !amdaie.logicalobjectfifo<memref<64x32xi32, 1>>)
```
yzhang93, line 76: Does it require all the `amdaie.dma_cpy_nd` within the inner scf.forall to be able to include in the workgroup? In real application, how do you make all these `amdaie.dma_cpy_nd` all under the inner scf.forall?  I think the first two should be outside the inner scf.forall.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEOpUtils.h
```diff
@@ -0,0 +1,33 @@
...
+SmallVector<OpTy> getParentsOfTypeIncluding(Operation *op) {
+  SmallVector<OpTy> res;
+  auto *current = op;
+  do
+    if (auto typedParent = dyn_cast<OpTy>(current)) res.push_back(typedParent);
+  while ((current = current->getParentOp()));
+  return res;
+}
+
+/// Return a vector with the most nested forall loops with the provided op's
```
yzhang93, line 27: Nit: most -> innermost?

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/test/roundtrip.mlir
```diff
@@ -53,3 +65,40 @@ func.func @dma_cpy_nd_mixed(%arg0: !amdaie.logicalobjectfifo<memref<1x1x8x16xi32
...
+// CHECK: amdaie.controlcode
+// CHECK: amdaie.end
+func.func @workgroup() {
+  amdaie.workgroup {
+    amdaie.controlcode {
+      amdaie.end
+    }
+  }
+  return
+}
```
yzhang93, line 104: Why not include `amdaie.core`?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertAIEWorkgroup.cpp
```diff
@@ -0,0 +1,169 @@
...
+        getInnermostForallLoops(workgroupOp);
+    if (innermostForallLoops.empty()) {
+      workgroupOp.emitOpError() << "contains no nested forall op";
+      return WalkResult::interrupt();
+    }
+    if (innermostForallLoops.size() != 1) {
+      workgroupOp.emitOpError()
+          << "multiple innermost foralls not supported for now";
+      return WalkResult::interrupt();
+    }
```
yzhang93, line 80: For peeled codes, there will be multiple innermost foralls. Do you have plans to add support in the following PRs?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEOpUtils.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/IR/Block.h"
+#include "mlir/IR/Operation.h"
+
+namespace mlir::iree_compiler::AMDAIE {
+
+/// Return a vector of the parent operations that are of type 'OpTy'
+template <typename OpTy>
+SmallVector<OpTy> getParentsOfTypeIncluding(Operation *op) {
```
yzhang93, line 18: This function name is not clear to me. Maybe rename it to `getOpTyParentOp` or something?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertAIEWorkgroup.cpp
```diff
@@ -0,0 +1,169 @@
...
+      if (op->getParentOfType<AMDAIE::CoreOp>()) return WalkResult::advance();
+      if (auto dmaOp = dyn_cast<AMDAIE::DmaCpyNdOp>(op)) {
+        auto sourceMemspace = dmaOp.getSourceObjectFifo().getMemorySpace();
+        auto targetMemspace = dmaOp.getTargetObjectFifo().getMemorySpace();
+        if (sourceMemspace &&
+            dyn_cast<IntegerAttr>(sourceMemspace).getInt() == 2 &&
+            targetMemspace &&
+            dyn_cast<IntegerAttr>(targetMemspace).getInt() == 2) {
+          dmaOp->emitOpError()
+              << "dma op with both source and target on L2 is not supported";
```
yzhang93, line 115: Why do you just check both source and target on L2 but not L1?

yzhang93 (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_aie_workgroup.mlir
jtuyls: > I think the first two should be outside the inner scf.forall.

Yes, they mostly should be and if they end up outside the workgroup, they can be fused into the workgroup later, using:
https://github.com/nod-ai/iree-amd-aie/compare/main...jornt_peeled_matmul_e2e#diff-61d9a09130faccae2d9f693e42252c7fb1b1780b26b88f9502ff59e29e0a07e7R20.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
```diff
@@ -24,10 +24,145 @@ class AMDAIE_Op<string mnemonic, list<Trait> traits = []> :
...
+      2. `DmaCpyNdOps` representing data movement between different memories in
+         the memory hierarchy.
+      3. Computational operations (linalg), which will be moved inside the AIE
+         core operations.
+      4. A newly inserted `ControlCodeOp`, representing the sequence of host
+         control code instructions.
+
+    The invariants of this op are:
+      1. The workgroup contains a single block, of which the terminator should
+         be a `ControlCodeOp`.
```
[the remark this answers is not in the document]
jtuyls: An invariant is a property, condition or relation that is always true. See for example https://stackoverflow.com/questions/112064/what-is-an-invariant

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertAIEWorkgroup.cpp
jtuyls: Peeling is already supported as the workgroup will end up around every `forall`, so in case of peeling there will be multiple workgroups.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/test/roundtrip.mlir
jtuyls: I had a separate test for that, but added it here as well.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEOpUtils.h
jtuyls: This function is named after `Operation::getParentOfType`: https://mlir.llvm.org/doxygen/classmlir_1_1Operation.html#a9968b118b25a099be92400e88239153c, and I think it should stay consistent with that one. Is the `including` part unclear?

jtuyls: Renamed it to `getInclusiveParentsOfType` and explained it in the comment.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertAIEWorkgroup.cpp
jtuyls: I think I made a type here, should be `L1`. But then I guess you would have the same question for `L2`? It's not checked for because both source and target on L2 will not enter any `if/else` block and will just be skipped. The logic here is just concerned about DMA operations taking data to/from L1 to insert the necessary synchronization ops.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEOpUtils.h
jtuyls: Done

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
```diff
@@ -24,10 +24,145 @@ class AMDAIE_Op<string mnemonic, list<Trait> traits = []> :
...
+// IREE AMDAIE Workgroup Ops
 //===----------------------------------------------------------------------===//
 
-def LogicalObjectFifoFromMemrefOp : AMDAIE_Op<"logicalobjectfifo.from_memref", []> {
+def AMDAIE_ControlCodeOp : AMDAIE_Op<"controlcode", [HasParent<"WorkgroupOp">,
+    SingleBlock, Terminator]> {
+  let summary = "Operation containing the host control code instructions.";
+  let description = [{
+    The control code op contains the host code instructions for a workgroup
+    in a single block. Therefore, this operation is expeced to always have a
```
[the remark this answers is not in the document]
jtuyls: Done

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertAIEWorkgroup.cpp
jtuyls: I updated the test to multiple innermost `forall` ops so you can see what the output looks like.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEOpUtils.h
yzhang93: I think this is better. Thanks!

yzhang93 (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEOpUtils.h
```diff
@@ -0,0 +1,34 @@
...
+
+/// Return a vector of the parent operations that are of type 'OpTy', including
+/// this op if it has type 'OpTy'
+template <typename OpTy>
+SmallVector<OpTy> getInclusiveParentsOfType(Operation *op) {
+  SmallVector<OpTy> res;
+  auto *current = op;
+  do
+    if (auto typedParent = dyn_cast<OpTy>(current)) res.push_back(typedParent);
+  while ((current = current->getParentOp()));
```
Abhishek-Varma, line 24: Opening/closing braces for `if` as well as `do/while` should be used even if we have one statement in the body.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_aie_workgroup.mlir
```diff
@@ -0,0 +1,138 @@
+// RUN: iree-opt --split-input-file --pass-pipeline="builtin.module(iree-amdaie-insert-aie-workgroup)" %s | FileCheck %s
+
+// CHECK-LABEL: @insert_aie_workgroup
+// CHECK: scf.forall (%arg0, %arg1) in (1, 1) {
```
Abhishek-Varma, line 4: Here and in other parts of the CHECK statements - we shouldn't use the SSA values directly as it is in the lit test.

`%[[VARIABLE:.*]] = ` for defining and `%[[VARIABLE]]` for using it is what we should do.

Abhishek-Varma (changes requested):
Looks good overall - a few comments to address.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_aie_workgroup.mlir
jtuyls: Done

Abhishek-Varma (approved):
Thanks @jtuyls ! LGTM!

PR #281, opened 2024-04-18
Bump IREE to latest

nirvedhmeshram:
Transform dialect scripts are failing with this bump which was some what expected. We will be xfailing these for now and adding follow up work to bring them back up.
This PR also bumps mlir-air.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/Translation/AIESerializer.cpp
+++ b/compiler/plugins/target/AMD-AIE/Translation/AIESerializer.cpp
@@ -516,7 +516,7 @@ LogicalResult AccelSerializer::processOperation(Operation *opInst,
 
   auto status =
       TypeSwitch<Operation *, LogicalResult>(opInst)
-          .Case<func::FuncOp, scf::ForallOp, scf::ForOp,
+          .Case<FunctionOpInterface, scf::ForallOp, scf::ForOp,
                 IREE::LinalgExt::PackOp, IREE::LinalgExt::UnPackOp,
                 memref::AllocOp, linalg::FillOp, linalg::GenericOp>(
               [&](auto typedOp) { return processOperation(typedOp, scope); })
@@ -528,7 +528,7 @@ LogicalResult AccelSerializer::processOperation(Operation *opInst,
   return status;
 }
 
-LogicalResult AccelSerializer::processOperation(func::FuncOp funcOp,
+LogicalResult AccelSerializer::processOperation(FunctionOpInterface funcOp,
                                                 ScopeInfo &scope) {
   ScopeInfo subScope = scope.getSubScope();
   std::string fnName = "primfn";
@@ -609,7 +609,7 @@ LogicalResult AccelSerializer::processOperation(func::FuncOp funcOp,
       "\"from_legacy_te_schedule\": True, \"param_device_types\": [], "
       "\"global_symbol\": \"main\"}";
 
-  Region &body = funcOp.getBody();
+  Region &body = funcOp.getFunctionBody();
   if (!llvm::hasSingleElement(body)) {
     return funcOp.emitOpError("unhandled operation with multiple blocks");
   }
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAddLoweringStrategy.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAddLoweringStrategy.cpp
@@ -10,11 +10,15 @@
 #include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenDialect.h"
 #include "iree/compiler/Dialect/HAL/IR/HALDialect.h"
 #include "iree/compiler/Dialect/LinalgExt/IR/LinalgExtDialect.h"
+#include "mlir/Dialect/Bufferization/IR/Bufferization.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
-#include "mlir/Dialect/Linalg/TransformOps/LinalgTransformOps.h"
 #include "mlir/Dialect/PDL/IR/PDL.h"
 #include "mlir/Dialect/PDLInterp/IR/PDLInterp.h"
+#include "mlir/Dialect/Transform/IR/TransformDialect.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Pass/PassRegistry.h"
 
 namespace mlir::iree_compiler::AMDAIE {
 
@@ -43,22 +47,20 @@ class AMDAIELoweringStrategyPass
 }  // namespace
 
 void AMDAIELoweringStrategyPass::runOnOperation() {
-  IREE::HAL::ExecutableVariantOp variantOp = getOperation();
-  ModuleOp moduleOp = variantOp.getInnerModule();
-  if (!moduleOp) {
-    getOperation()->emitError(
-        "Expected a variantOp root with an inner ModuleOp");
-    return signalPassFailure();
-  }
+  ModuleOp moduleOp = getOperation();
   // To simplify development, the number of cores can be passed as a flag during
   // compilation. In the future these parameters could be read from file.
   struct AIEConfig cfg = {numCores};
-  if (failed(initAIELaunchConfig(moduleOp, usePassPipeline, cfg))) {
-    return signalPassFailure();
+  for (auto funcOp : moduleOp.getOps<FunctionOpInterface>()) {
+    // Set the strategy with default heuristics.
+    if (failed(initAIELaunchConfig(funcOp, usePassPipeline, cfg))) {
+      funcOp.emitOpError("failed to set lowering configuration");
+      return signalPassFailure();
+    }
   }
 }
 
-std::unique_ptr<Pass> createAMDAIELoweringStrategyPass(
+std::unique_ptr<OperationPass<ModuleOp>> createAMDAIELoweringStrategyPass(
     AMDAIELoweringStrategyOptions options) {
   return std::make_unique<AMDAIELoweringStrategyPass>(options);
 }
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
@@ -9,8 +9,12 @@
 #include "air/Conversion/Passes.h"
 #include "air/Transform/Passes.h"
 #include "iree-dialects/Dialect/LinalgTransform/Passes.h"
+#include "iree/compiler/Codegen/Common/PassUtils.h"
 #include "iree/compiler/Codegen/Common/Passes.h"
+#include "iree/compiler/Dialect/LinalgExt/Transforms/Passes.h"
+#include "iree/compiler/Utils/PassUtils.h"
 #include "mlir/Dialect/Affine/Passes.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/Linalg/Passes.h"
 #include "mlir/Dialect/MemRef/Transforms/Passes.h"
 #include "mlir/Pass/PassManager.h"
@@ -50,13 +54,13 @@ static llvm::cl::opt<bool> clEnableVectorizationPasses(
                    "purposes only."),
     llvm::cl::init(true));
 
-void appendVectorizationToPipeline(OpPassManager &pm) {
+void appendVectorizationToPipeline(OpPassManager &funcPassManager) {
   if (!clEnableVectorizationPasses) return;
-  pm.addNestedPass<func::FuncOp>(createAMDAIECleanupPass());
-  pm.addNestedPass<func::FuncOp>(createAMDAIEInsertLoopsForVectorizationPass());
-  pm.addNestedPass<func::FuncOp>(createAMDAIEVectorizationPass());
-  pm.addPass(createCanonicalizerPass());
-  pm.addPass(createCSEPass());
+  funcPassManager.addPass(createAMDAIECleanupPass());
+  funcPassManager.addPass(createAMDAIEInsertLoopsForVectorizationPass());
+  funcPassManager.addPass(createAMDAIEVectorizationPass());
+  funcPassManager.addPass(createCanonicalizerPass());
+  funcPassManager.addPass(createCSEPass());
 }
 
 //===---------------------------------------------------------------------===//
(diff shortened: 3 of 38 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.h
```diff
@@ -90,7 +91,7 @@ std::unique_ptr<Pass> createAMDAIELowerToUKernelsPass(
     AMDAIELowerToUKernelsOptions options = {});
 
 /// Create a pass to lower workgroup count region of entry point operations.
-std::unique_ptr<Pass> createAMDAIELowerWorkgroupCountPass();
+std::unique_ptr<InterfacePass<FunctionOpInterface>> createAMDAIELowerWorkgroupCountPass();
```
MaheshRavishankar, line 94: This needs to be a variant pass. Can be a module pass also if needed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -181,13 +181,13 @@ def AMDAIELowerToUKernels :
 }
 
 def AMDAIELowerWorkgroupCount :
-    Pass<"iree-amdaie-lower-workgroup-count", "mlir::iree_compiler::IREE::HAL::ExecutableVariantOp"> {
+    InterfacePass<"iree-amdaie-lower-workgroup-count", "mlir::FunctionOpInterface"> {
```
MaheshRavishankar, line 184: Duplicate comment: THis needs to be a variant pass or module pass.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -247,30 +251,30 @@ void addPackPeelBasedPassPipeline(OpPassManager &pm,
   modulePassManager.addPass(createCSEPass());
 
   // Comprehensive bufferization
-  addAMDAIEBufferizePasses(modulePassManager);
+  addAMDAIEBufferizePasses(pm);
 }
 
-void addPadPackBasedPassPipeline(OpPassManager &pm,
+void addPadPackBasedPassPipeline(OpPassManager &modulePassManager,
```
MaheshRavishankar, line 257: Nit: rename to `funcPassManager`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -347,52 +351,61 @@ void addPadPackBasedPassPipeline(OpPassManager &pm,
...
   }
-
+  /*
   // Lower to UKernels
   {
     AMDAIELowerToUKernelsOptions options;
     options.passPipeline = AIEPassPipeline::PadPackPipeline;
     options.pathToUkernels = clPathToUkernels;
-    modulePassManager.addNestedPass<func::FuncOp>(
+    modulePassManager.addPass(
```
MaheshRavishankar, line 363: Make this pass a `FunctionOpInterface` pass, it can actually be an unspecified pass.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -347,52 +351,61 @@ void addPadPackBasedPassPipeline(OpPassManager &pm,
...
   }
-
+  */
   // Vectorization passes
   appendVectorizationToPipeline(modulePassManager);
 
   // Comprehensive bufferization
   addAMDAIEBufferizePasses(modulePassManager);
-  modulePassManager.addPass(createLowerUKernelOpsToCallsPass());
+  //modulePassManager.addPass(createLowerUKernelOpsToCallsPass());
```
MaheshRavishankar, line 372: Move this out of here just before invoking the conversion-to-air-to-aie pass pipelines.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -347,52 +351,61 @@ void addPadPackBasedPassPipeline(OpPassManager &pm,
...
 }
 
-void buildAMDAIETransformPassPipeline(OpPassManager &pm) {
-  addCommonTargetExecutablePreprocessingPasses(pm);
+void buildAMDAIETransformPassPipeline(OpPassManager &variantPassManager) {
+
+  OpPassManager &modulePassManager = variantPassManager.nest<ModuleOp>();
+  modulePassManager.addPass(createLowerExecutableUsingTransformDialectPass());
```
MaheshRavishankar, line 378: Needs to be done after `createAMDAIELoweringStrategyPass`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -347,52 +351,61 @@ void addPadPackBasedPassPipeline(OpPassManager &pm,
...
+  {
+    FunctionLikeNest funcPassManager(modulePassManager);
+    addCommonTargetExecutablePreprocessingPasses(funcPassManager);
+  }
```
MaheshRavishankar, line 382: After these passes we need to run the `MaterializeUserConfig` pass from IREE core

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -347,52 +351,61 @@ void addPadPackBasedPassPipeline(OpPassManager &pm,
...
   {
     AMDAIELoweringStrategyOptions options;
     options.usePassPipeline = clUsePipeline;
     options.numCores = clNumCores;
-    pm.addPass(createAMDAIELoweringStrategyPass(options));
+    modulePassManager.addPass(createAMDAIELoweringStrategyPass(options));
   }
   {
+    FunctionLikeNest funcPassManager(modulePassManager);
```
MaheshRavishankar, line 390: Before this (i.e. after LoweringStrategyPass) the `createLowerExecutableUsingTransformDialectPass` should run.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -347,52 +351,61 @@ void addPadPackBasedPassPipeline(OpPassManager &pm,
...
     AMDAIELowerExecutableTargetOptions options;
     options.usePassPipeline = clUsePipeline;
-    pm.addPass(createAMDAIELowerExecutableTargetPass(options));
+    //FunctionLikeNest(modulePassManager).addPass(createAMDAIELowerExecutableTargetPass(options));
+    funcPassManager.addPass([&](){ return createAMDAIELowerExecutableTargetPass(options);})
+                   .addPass([&](){ return createAMDAIELowerWorkgroupCountPass();});
```
MaheshRavishankar, line 395: The lower workgroup count should be a module pass, so what is commented out two lines below is how it should be.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -347,52 +351,61 @@ void addPadPackBasedPassPipeline(OpPassManager &pm,
...
   }
-  pm.addPass(createAMDAIELowerWorkgroupCountPass());
+  //modulePassManager.addPass(createAMDAIELowerWorkgroupCountPass());
```
MaheshRavishankar, line 397: Right after this you can run the lowerukerneltofuncop pass.

nirvedhmeshram:
@erwei-xilinx I am seeing a new failure, in the loop fusion pass with the air-bump in this PR that is required to match the llvm of the iree bump happening here. Here is a gist with the IR and the failure, could you please confirm this can be reproduced in latest AIR for you?
https://gist.github.com/nirvedhmeshram/b45d47c444e0dd7597d5f1f5ddb20038

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAddLoweringStrategy.cpp
```diff
@@ -43,22 +47,20 @@ class AMDAIELoweringStrategyPass
...
 
 void AMDAIELoweringStrategyPass::runOnOperation() {
-  IREE::HAL::ExecutableVariantOp variantOp = getOperation();
-  ModuleOp moduleOp = variantOp.getInnerModule();
-  if (!moduleOp) {
-    getOperation()->emitError(
-        "Expected a variantOp root with an inner ModuleOp");
-    return signalPassFailure();
-  }
+  auto moduleOp = getOperation();
```
Abhishek-Varma, line 50: It'd be nice to use `ModuleOp` here instead of `auto`.

It's a good practice to use `auto` only in those cases where the in-line context mentions it explicitly.
For instance the line below `auto funcOp = moduleOp.getOps<FunctionOpInterface>()` is a correct use since. Another example is `cast<SomeType>(...)`.

compiler/plugins/target/AMD-AIE/Translation/AIESerializer.cpp
```diff
@@ -516,7 +516,7 @@ LogicalResult AccelSerializer::processOperation(Operation *opInst,
 
   auto status =
       TypeSwitch<Operation *, LogicalResult>(opInst)
-          .Case<func::FuncOp, scf::ForallOp, scf::ForOp,
+          .Case<mlir::FunctionOpInterface, scf::ForallOp, scf::ForOp,
```
Abhishek-Varma, line 519: Any way we could skip having to specify the `mlir::` namespace here for `FunctionOpInterface` ?
Just how it is done in the same PR in [AMDAIELowerExecutableTarget.cpp](https://github.com/nod-ai/iree-amd-aie/blob/e9fc58296903ae570474d3890d4395d5733d878b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerExecutableTarget.cpp#L70).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -22,7 +22,7 @@ def AMDAIEBridgeToAIR : Pass<"iree-amdaie-bridge-to-air", ""> {
 }
 
 def AMDAIEBufferizeToAllocation :
-    Pass<"iree-amdaie-bufferize-to-allocation", "func::FuncOp"> {
+    InterfacePass<"iree-amdaie-bufferize-to-allocation", "mlir::FunctionOpInterface"> {
```
Abhishek-Varma, line 25: In `.td` the use of namespace `mlir::` is correct. Thank you for this!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_loops_for_vectorization.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: iree-opt --iree-amdaie-insert-loops-for-vectorization %s | FileCheck %s
+// RUN: iree-opt --pass-pipeline="builtin.module(func.func(iree-amdaie-insert-loops-for-vectorization))" %s | FileCheck %s
```
Abhishek-Varma, line 1: Wow. The recent update wouldn't let us use a standalone pass the way it was? :/

tests/transform_dialect/matmul_fill_spec_pad.mlir
```diff
@@ -1,5 +1,5 @@
 // RUN: iree-opt --iree-transform-dialect-interpreter %s | FileCheck %s
-
+// XFAIL: *
```
Abhishek-Varma, line 2: This is for my understanding - why don't we get rid of these transform dialect? And why are we xfailing them now?

Abhishek-Varma (approved):
Wow. These are huge updates! Thank you @nirvedhmeshram !
A few nit comments and questions from my end.

To unblock you, I'm hitting an approve. Again, thank you for this!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_loops_for_vectorization.mlir
nirvedhmeshram: Ya I think the stand alone way just looks concise but under the hood it is with the builtin.module pass manager. But now we need to be more explicit in mentioning that this pass is meant for a function pass manager.

tests/transform_dialect/matmul_fill_spec_pad.mlir
nirvedhmeshram: I think the transform dialect can still be useful for experimentation. Why we are xfailing is that with the upstream changes a lot has changed that needs updating of the script and the plan is to do that in a seperate PR and not block the bump on it.

nirvedhmeshram:
The new failure is interesting! The reason is previously in lowering strategy pass we used to attach a `{translation_info = #iree_codegen.translation_info<None>}` attribute to the hal.executable.export op. But after this PR this is attached to the func.func op that is passed to the aie2xclbin tool which fails with the error 
```
loc("module_matmul_small_1_dispatch_0_amdaie_xclbin_fb.aiecc.mlir":635:185): error: #"iree_codegen"<"translation_info<Custom>"> : 'none' attribute created with unregistered dialect. If this is intended, please call allowUnregisteredDialects() on the MLIRContext, or use -allow-unregistered-dialect with the MLIR opt tool used
```
There can be several solutions such as deleting the attribute after its purpose is served, trying with the `-allow-unregistered-dialect` flag but wanted check what the right thing to do is? @MaheshRavishankar

MaheshRavishankar:
You can add the `reconcileTranslationInfo` pass as is done here https://github.com/iree-org/iree/blob/e8f4948d5c62beea27a109161b697d9088d5b3bd/compiler/src/iree/compiler/Codegen/LLVMGPU/Passes.cpp#L843

I notice I missed adding this on the CPU side, but this essentially deletes the translation info and sets the values appropriately

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -52,9 +56,9 @@ static llvm::cl::opt<bool> clEnableVectorizationPasses(
 
 void appendVectorizationToPipeline(OpPassManager &pm) {
```
MaheshRavishankar, line 57: Nit: upstream I tried to capture the nesting in the name.there is no other way to know what the nesting is though unfortunately.

MaheshRavishankar (approved):
Awesome!!! Did you want me to look at the TD tests.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
nirvedhmeshram: makes sense, changed the passManager's name to funcPassManager so the nesting is clear.

nirvedhmeshram:
> Awesome!!! Did you want me to look at the TD tests.

I can take a stab at it on Monday unless I am needed on something higher priority.

PR #275, opened 2024-04-16
Add support to bufferize the operands from elementwise ops

yzhang93:
- We will need to run this pass for elementwise ops after the elementwise op is fused into the last iteration of the matmul computation.

PR #283, opened 2024-04-19
Overwrite option to force peeling first and last iterations for matmul-elementwise fusion

yzhang93:
- In pack-peel pipeline, for matmul dispatch only the first iteration is peeled from the scf.for, while for matmul-elementwise dispatch, both the first and the last iterations need to be peeled. To avoid bringing extra flags to switch between such cases, the  default option `peelingType` is overwritten as `FirstLast` when it detects the matmul-elementwise pattern. 
- Clean up the pass option names as suggested in the previous review comment https://github.com/nod-ai/iree-amd-aie/pull/275#discussion_r1569154836.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
@@ -8,6 +8,7 @@
 
 #include "llvm/ADT/DenseMap.h"
 #include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/Iterators.h"
 
 namespace mlir::iree_compiler::AMDAIE {
 
@@ -301,6 +302,7 @@ bool isMatmul(linalg::LinalgOp linalgOp) {
 /// its producer is a matmul-like op.
 bool isMatmulProducerOfElementwise(linalg::LinalgOp linalgOp) {
   if (!isElementwise(linalgOp)) return false;
+  if (isa<linalg::FillOp>(linalgOp)) return false;
   // Check if any of the defining op is a matmul-like op. To simplify the
   // problem, currently only check if it is a contraction op.
   for (auto operand : linalgOp->getOperands()) {
@@ -310,6 +312,18 @@ bool isMatmulProducerOfElementwise(linalg::LinalgOp linalgOp) {
           return true;
         }
         break;
+      }
+      if (auto forOp = dyn_cast_or_null<scf::ForOp>(defOp)) {
+        bool findMatmul = false;
+        forOp.getBody()->walk<WalkOrder::PostOrder, ReverseIterator>(
+            [&](linalg::LinalgOp op) {
+              if (isMatmul(op)) {
+                findMatmul = true;
+                return WalkResult::interrupt();
+              }
+              return WalkResult::advance();
+            });
+        if (findMatmul) return true;
       }
       operand = defOp->getOperand(0);
     }
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
@@ -298,13 +298,12 @@ bool isMatmul(linalg::LinalgOp linalgOp) {
          match6DLinalgGenericMatmul(linalgOp);
 }
 
-/// Utility to indentify whether a generic op is an elementwise op and whether
-/// its producer is a matmul-like op.
+/// Utility to identify if `linalgOp` is an elementwise operation with a
+/// matmul-like op upstream in its computation tree.
 bool isMatmulProducerOfElementwise(linalg::LinalgOp linalgOp) {
   if (!isElementwise(linalgOp)) return false;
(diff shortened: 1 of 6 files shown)


newling:
> To avoid bringing extra flags to switch between such cases, the default option peelingType is overwritten as FirstLast when it detects the matmul-elementwise pattern.

I haven't looked at the PR yet, but this this sounds more like an 'auto' behavior than a 'default' behavior..

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/peel_for_loop.mlir
```diff
@@ -1,6 +1,7 @@
 // RUN: iree-opt --pass-pipeline="builtin.module(func.func(iree-amdaie-peel-for-loop{peeling-type=first}))"  --split-input-file %s | FileCheck %s --check-prefix=PEEL-FIRST
 // RUN: iree-opt --pass-pipeline="builtin.module(func.func(iree-amdaie-peel-for-loop{peeling-type=last}))"  --split-input-file %s | FileCheck %s --check-prefix=PEEL-LAST
 // RUN: iree-opt --pass-pipeline="builtin.module(func.func(iree-amdaie-peel-for-loop{peeling-type=first-last}))"  --split-input-file %s | FileCheck %s --check-prefix=FIRST-LAST
+// RUN: iree-opt --pass-pipeline="builtin.module(func.func(iree-amdaie-peel-for-loop{peeling-type=first}))"  --split-input-file %s | FileCheck %s --check-prefix=OVERWRITE-FLAG
```
newling, line 4: I think this could be confusing allowing a user specified peeling-type to be overwritten like this. If there was peeling-type=auto, or peeling-type=unset, changing those to another peeling-type would make sense .

newling, line 4: I think this could be confusing allowing a user specified peeling-type to be overwritten like this. If there was peeling-type=auto, or peeling-type=unset, changing those to another peeling-type would make sense .

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
```diff
@@ -301,6 +302,7 @@ bool isMatmul(linalg::LinalgOp linalgOp) {
 /// its producer is a matmul-like op.
 bool isMatmulProducerOfElementwise(linalg::LinalgOp linalgOp) {
```
newling, line 303: This function doesn't seem to have a very tight definition (in code or documentation) and it's not tested in a unit test, so I'm wondering why it is is AMDAIEUtils.hpp and not just local to the translation unit it is used in. 

My thinking is that AMDAIEUtils.hpp is for code which might be used in multiple places (or must be in a header for unit testing).

newling (commented):
I'm a bit out of the loop on the peeling work, when you refer to matmul-elementwise fusion is this specifically unary elementwise ops? 

i.e. `relu(matmul(A,B))` but not `add(matmul(A,B),C)` (add being a binary elementwise operation).

yzhang93:
> I'm a bit out of the loop on the peeling work, when you refer to matmul-elementwise fusion is this specifically unary elementwise ops?
> 
> i.e. `relu(matmul(A,B))` but not `add(matmul(A,B),C)` (add being a binary elementwise operation).

The target is more like the latter one. But I think it shouldn't be really matter?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/peel_for_loop.mlir
yzhang93: So the default peelingType is set as `PeelingType::First` at https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td#L243 because it is the peeling function we want to use for GEMM. And this flag is equivalent to `--pass-pipeline="builtin.module(func.func(iree-amdaie-peel-for-loop))"`, and it is overwritten when it detects the matmul-elementwise pattern.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
yzhang93: This function is already used in multiple places. For another use you could refer to this https://github.com/nod-ai/iree-amd-aie/blob/e296b578f33f04873a94cee000d1d4409d051d9b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp#L50 for example.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/peel_for_loop.mlir
yzhang93: Thanks for your suggestion. Now I change the default setting in td file as "None", and in implementation call `peelForLoopFirstIteration` by default. In this way, for normal GEMM it will peel the first iteration automatically, and for matmul-elementwise case it will first assign the PeelingType as "FirstLast".

MaheshRavishankar (approved):

PR #284, opened 2024-04-19
Remove legacy pipeline

newling:
This PR removes 2 things:

1)  The method `addMLIRAIRAIELegacyLoweringPasses` which is not needed it seems (pad-pack uses 
 the very similar `addMLIRAIRAIELoweringPasses` and pack-peel uses neither (is this an oversight?)). 
 
 2) ` static PassPipelineRegistration<> AIELoweringPipeline("iree-amdaie-aie-lowering-pipeline", ... ` which doesn't seem to be used anywhere. 
 
 Is it ok to just go ahead with these removals, or is a plan to use them in the future?

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
@@ -354,6 +354,8 @@ void addPadPackBasedPassPipeline(OpPassManager &funcPassManager,
   // Vectorization passes
   appendVectorizationToPipeline(funcPassManager);
 
+  modulePassManager.addPass(createCanonicalizerPass());
+
   // Comprehensive bufferization
   addAMDAIEBufferizePasses(funcPassManager);
 }
@@ -384,9 +386,7 @@ void buildAMDAIETransformPassPipeline(OpPassManager &variantPassManager) {
     addMLIRAIRAIELoweringPasses(modulePassManager, false);
   } else if (clUsePipeline == AIEPassPipeline::PackPeelPipeline) {
     addMLIRAIRAIELoweringPasses(modulePassManager, true);
-  } else {
-    addMLIRAIRAIELegacyLoweringPasses(modulePassManager);
-  }
+  } 
   variantPassManager.addPass(createReconcileTranslationInfoPass());
   variantPassManager.addPass(createAMDAIELowerWorkgroupCountPass());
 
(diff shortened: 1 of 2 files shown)


nirvedhmeshram:
Not sure if the registration function had some purpose, @MaheshRavishankar ?
Other then that the PR looks good to me. The pad-pack is not using any lowering on purpose as its not plumbed through e2e.

MaheshRavishankar:
I think this was added for some other out-of-tree users... @yzhang93 has more context. Please wait for her to take a look

yzhang93 (changes requested):
I think it's fine to remove `MLIRAIRAIELegacyLoweringPasses`, don't remember why it was created in the first place. We should always try to keep a single set of AIRLoweringPasses. Minor change request, otherwise LGTM.

yzhang93:
@erwei-xilinx Could you please make sure you are using passes within `MLIRAIRAIELoweringPasses` instead of `MLIRAIRAIELegacyLoweringPasses` when working for pack-peel pipeline? @nirvedhmeshram mentioned there was some problem running `MLIRAIRAIELoweringPasses` on simple-pack pipeline previously.

erwei-xilinx:
> @erwei-xilinx Could you please make sure you are using passes within `MLIRAIRAIELoweringPasses` instead of `MLIRAIRAIELegacyLoweringPasses` when working for pack-peel pipeline? @nirvedhmeshram mentioned there was some problem running `MLIRAIRAIELoweringPasses` on simple-pack pipeline previously.

@yzhang93 Sure, in my local experiments with pack-peel pipeline, I haven't attempted to change the IREE pipeline, and I think that the default pack-peel pipeline was using `MLIRAIRAIELoweringPasses`.

yzhang93:
> > @erwei-xilinx Could you please make sure you are using passes within `MLIRAIRAIELoweringPasses` instead of `MLIRAIRAIELegacyLoweringPasses` when working for pack-peel pipeline? @nirvedhmeshram mentioned there was some problem running `MLIRAIRAIELoweringPasses` on simple-pack pipeline previously.
> 
> @yzhang93 Sure, in my local experiments with pack-peel pipeline, I will be sure to stick to the default `MLIRAIRAIELoweringPasses` pipeline.

Thanks! @newling I would suggest to hold this PR for a while and wait for @erwei-xilinx to make sure `AMDAIEPackToDmaPass()` work for pack-peel pipeline instead of using `AMDAIEDecomposeLinalgExtPackUnPackToAIRPass()`. Then we can safely delete the legacy lowering passes.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -403,9 +405,7 @@ void buildAMDAIETransformPassPipeline(OpPassManager &pm) {
   if (clUsePipeline == AIEPassPipeline::PadPackPipeline) {
     auto &modulePassManager = pm.nest<ModuleOp>();
     addMLIRAIRAIELoweringPasses(modulePassManager);
-  } else if (clUsePipeline != AIEPassPipeline::PackPeelPipeline) {
-    assert(false && "Expected either PadPackPipeline or PackPeelPipeline");
-  }
+  }
```
yzhang93, line 408: You can remove the if condition as well. In the future we should use `addMLIRAIRAIELoweringPasses` for all pipelines.

yzhang93: If you can make this change and rebase on main, then we can merge this PR.

yzhang93 (approved):
Thanks!

Issue #285, opened 2024-04-22
Compile all matmul shapes is SD3 model for AIE


newling:
Tracker task for the shapes we should support with direct codegen

2 compiler errors need resolving. The errors can be reproduced with square matmuls. 

### Error 1: m=n=k=2304 (=2048+256) bf16 operands f32 output. 

The failure is in `irrt-to-ipu`

The error message is 

```
// iree-compile: iree-amd-aie/third_party/mlir-air/mlir/lib/Conversion/AIRRtToIpuPass.cpp:551: void 
(anonymous namespace)::tileIllegalWrapDim(airrt::DmaMemcpyNdOp): Assertion 
`!(const_wrap % (AIE2_WRAP_UPPER_BOUND / 2)) && "Currently do not support remainder tiles"' failed.
```

@erwei-xilinx is aware of this:
https://teams.microsoft.com/l/message/19:meeting_Zjc5ZmZhM2EtZDcxZS00NzYxLTliYmQtNGFlNzY1MDJhNjMy@thread.v2/1713812572221?context=%7B%22contextType%22%3A%22chat%22%7D

### Error 2: m=n=k=2432 (=2048+256+128) bf16 operands f32 output. 
Failure in `air-label-scf-for-to-ping-pong`

Error message is quite generic, needs further understanding:
```
error: block with no terminator, has %40 = "air.wait_all"(%arg8) : (!air.async.token) -> !air.async.token
note: see current operation: %40 = "air.wait_all"(%arg8) : (!air.async.token) -> !air.async.token
```

Example input IR:
```
!lhs = tensor<2432x2432xbf16>
!rhs = tensor<2432x2432xbf16>
!out = tensor<2432x2432xf32>
func.func @matmul_32x32_32xf32_(%lhs : !lhs, %rhs : !rhs) -> !out {
  %init_acc = tensor.empty() : !out
  %c0_acc_type = arith.constant 0.0 : f32
  %acc = linalg.fill ins(%c0_acc_type : f32) outs(%init_acc : !out) -> !out
  %result = linalg.matmul ins(%lhs, %rhs: !lhs, !rhs) outs(%acc: !out) -> !out
  return %result: !out
}
```

newling:
### Investigation of Error 2:

m=n=k=2304 does not fail compilation as early as m=n=k=2432 so I compare the IR generated by passes between these 2, to see where the 2432 case goes hors piste. 

[...]

PR #286, opened 2024-04-22
Add all 9 sd3 matmuls to CI testing 

newling:
Currently 0/9 of the matmuls for sd3 compile. They are added to CI with the  
--expect-compile-failure="1" flag. 

There are 2 different compilation failures seen in the matmuls, these are documented in the tests it this PR.  

The tracker task for get these matmuls compiling and then running is https://github.com/nod-ai/iree-amd-aie/issues/285

### DIFF
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -209,15 +209,8 @@ function run_matmul_test() {
   # The default is to compile and run the numerical test.
   local compile_only="0"
 
-  # Experimenting with ways to make the host side baseline computation faster.
-  local matrix_rank="0"
-
   while [ "$#" -gt 0 ]; do
     case "$1" in
-      --matrix-rank)
-        matrix_rank="$2"
-        shift 2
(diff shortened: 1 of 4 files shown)


MaheshRavishankar:
```
error: operand #2 does not dominate this use
```

seems like a easily fixable issue. I am assuming you are going to post more info here or on the issue

newling:
```
error: operand #2 does not dominate this use
```
is no longer the error after the IREE and mlir-air bumps (https://github.com/nod-ai/iree-amd-aie/pull/281) it is now the equally generic

```
error: block with no terminator
```
more comments in  https://github.com/nod-ai/iree-amd-aie/issues/285 but needs further analysis

newling:
(Currently fails CI because it needs #300)

newling:
All matmuls pass! Except for the one with N=2 as expected. 

CI caveat: 
1) running the tests is really slow because of the host side matmul to create the reference
2) CI currently fails for large tensors, something to do with github runner

PR #291, opened 2024-04-23
Fail informatively when pack/unpack not converted in pack-to-dma

newling:
Fail if the pack-to-dma pass does not convert all pack and unpack operations to dma copy operations. 

To ensure that the failure message contains the reason that the pack/unpack was not converted, I have replaced the pattern-based approach with a direct walk-the-graph-once approach.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/pack_to_dma.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: iree-opt --pass-pipeline="builtin.module(iree-amdaie-pack-to-dma)" %s | FileCheck %s
+// RUN: iree-opt --iree-amdaie-pack-to-dma --cse --split-input-file %s | FileCheck %s
```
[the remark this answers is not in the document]
newling: It was necessary -- the approach with patterns seems to run it automatically. 

Without cse, duplicate constants are used --  `%c0` and  `%c0_1` both appear as operands of the resulting dma copy. 


I know it's not great having cse/canonicalization in unit tests like this...

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPackToDma.cpp
```diff
@@ -164,105 +170,114 @@ LogicalResult getDmaInputs(RewriterBase &rewriter, Operation *&operandOp,
...
-
-  LogicalResult matchAndRewrite(OpType op,
-                                PatternRewriter &rewriter) const override {
-    // 1. Filter out NYI cases.
-    llvm::ArrayRef<int64_t> innerTiles = op.getStaticInnerTiles();
-    if (llvm::any_of(innerTiles, [](int64_t size) {
-          return ShapedType::isDynamic(size);
-        })) {
-      return rewriter.notifyMatchFailure(op, "non-static shape NYI");
+LogicalResult processInputs(Operation *op, SmallVector<OpFoldResult> &offsets,
```
nirvedhmeshram, line 209: please add a comment explaining the function.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPackToDma.cpp
```diff
@@ -164,105 +170,114 @@ LogicalResult getDmaInputs(RewriterBase &rewriter, Operation *&operandOp,
...
+    if (failed(unPackDmaInputs(unPackOp, offsets, sizes, strides))) {
+      return failure();
+    }
+  }
+  return success();
+}
 
-    Value input = op.getInput();
-    Value output = op.getOutput();
+LogicalResult rewriteAsDma(IRRewriter &rewriter, Operation *op, Value input,
```
nirvedhmeshram, line 224: here as well. sorry the original pattern should have had a comment too.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/pack_to_dma.mlir
nirvedhmeshram, line 1: Lets add canonicalize and cse passes in the pass pipeline where this pass is used so its nice to read when we look at IR.

nirvedhmeshram (changes requested):
Looks good, overall, just requesting some minor changes.

newling:
Hi @nirvedhmeshram thanks for the review. I've added 2 comments, and added canonicalization to the pipeline after this pass (which is now needed as before with the pattern-based approach it happened automatically).

nirvedhmeshram:
curious what the \p in the two comments added is about? Since we are not generating any docs if its a formatting thing, I am not sure if the value add is enough to justify the loss in readability for people directly reading the code?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -409,6 +409,8 @@ void addMLIRAIRAIELoweringPasses(OpPassManager &passManager, bool packPeel) {
   else
     passManager.addPass(createAMDAIEPackToDmaPass());
 
+  passManager.addPass(createCanonicalizerPass());
```
nirvedhmeshram, line 412: I would only add these to the else part of the above if/else to not unintentionally mess with the peel-pack flow.

newling:
> curious what the \p in the two comments added is about? Since we are not generating any docs if its a formatting thing, I am not sure if the value add is enough to justify the loss in readability for people directly reading the code?
> 
> Also wondering what exactly is causing the tests to fail, perhaps some of the dma ops are getting removed by canonicalization as they appear like dead code? This may be why the placement of these passes is so sensitive (not saying it should be of course).

https://www.doxygen.nl/manual/commands.html#cmdp
The use of /// suggests doxygen-style comment so I was being consistent, but can change. 
Hmm yes, probably adding canonicalization to the pass pipeline tripped a lit test. I didn't run the tests locally. Will investigate asap.

newling:
Hmm with my setup in vim "\p myParam" is nicely colored and formatted. Not is VSCode?

newling:
> Also wondering what exactly is causing the tests to fail, perhaps some of the dma ops are getting removed by canonicalization as they appear like dead code? This may be why the placement of these passes is so sensitive (not saying it should be of course).

It was running canonicalization in the pipeline (CSE is fine, canonicalization results in a compilation error). I've left a TODO in the code -- we should be able to run canonicalization here AFAICT. 

I've removed the doxygen style comments

nirvedhmeshram:
> Hmm with my setup in vim "\p myParam" is nicely colored and formatted. Not is VSCode?

Nope, but frankly no strong opinion on that.

nirvedhmeshram (approved):
Thanks!

PR #293, opened 2024-04-23
Do packing before tiling in pad-pack pipeline

nirvedhmeshram:
This makes sure we make a reasonable tiling/packing config for any shape that is a multiple of the minimum packing config.
The order is 
1. Set the packing config first.
2. Then the inner tile which should be a multiple of the packing.
3. Then the outer tile that should be a multiple of the inner tile.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
@@ -176,22 +176,15 @@ static LogicalResult setRootConfigForPadPackPipeline(
     return linalgOp.emitOpError("expected bitwidth 64/32/16/8");
   }
   unsigned tilingScaleFactor = maybeTilingScaleFactor.value();
-  // TODO (nmeshram) : We should be able to use fixed tiling config after we
-  // have padding support.
-  auto tileM0 = findLargestFactor((int)initShape[0], 64 * tilingScaleFactor);
-  auto tileN0 = findLargestFactor((int)initShape[1], 64 * tilingScaleFactor);
-  auto tileM1 = findLargestFactor((int)tileM0, 16 * tilingScaleFactor);
-  auto tileN1 = findLargestFactor((int)tileN0, 16 * tilingScaleFactor);
-  auto tileK0 = findLargestFactor((int)lhsShape[1], 256);
 
-  // Do packing first to allow larger k packing
+  // Do packing first to allow better packing configs
   // ------------------------------------------------------
   // --------------- Set packing config -------------------
   // ------------------------------------------------------
   MLIRContext *context = entryPointFn.getContext();
   const int packLevel = 1;
-  auto packedSizes =
-      getPackedSize(linalgOp, packLevel, tileM1, tileN1, (int)tileK0);
+  auto packedSizes = getPackedSize(linalgOp, packLevel, (int)initShape[0],
+                                   (int)initShape[1], (int)lhsShape[1]);
   SmallVector<int64_t> transposePackIndices = {0, 1, 2};
   SmallVector<bool> unpackEmpty = {false, false, true};
   SmallVector<SmallVector<int64_t>> innerPerm = {{0, 1}, {1, 0}, {0, 1}};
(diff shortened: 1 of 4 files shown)


newling:
I like this. I will try it out on a problem size in https://github.com/nod-ai/iree-amd-aie/pull/286

newling:
> I think the packing/tiling config we arrived for this shape which is
tile_sizes = [[128, 128], [0, 0, 152], [32, 32], [0, 0, 1]], packedSizes = [4, 4, 8] is correct?

On this branch it is

```
#config = #iree_codegen.lowering_config<tile_sizes = [[152, 152], [0, 0, 152], [8, 8], [0, 0, 1]]>
 ```
 
 On main it is 
 
```
#config = #iree_codegen.lowering_config<tile_sizes = [[152, 152], [0, 0, 152], [38, 38], [0, 0, 1]]>
```

On https://github.com/newling/iree-amd-aie/tree/power_of_two (where is passes) it is

```
#config = #iree_codegen.lowering_config<tile_sizes = [[128, 128], [0, 0, 128], [64, 64], [0, 0, 8]]>
```

**Comment by newling** [comment]
@erwei-xilinx here is the IR: https://gist.github.com/newling/ba8ab6d035c60077525e2398e02c3f51

But note that this is not blocking -- I've discussed with @nirvedhmeshram and he's going to try a different algorithm (which prioritized a larger tile size on L1) which we think will result in a different tiling which we know does compile through mlir-air:

#config = #iree_codegen.lowering_config<tile_sizes = [[128, 128], [0, 0, 128], [64, 64], [0, 0, 8]]>

**Comment by nirvedhmeshram** [comment]
> > @erwei-xilinx here is the IR: https://gist.github.com/newling/ba8ab6d035c60077525e2398e02c3f51
> > But note that this is not blocking -- I've discussed with @nirvedhmeshram and he's going to try a different algorithm (which prioritized a larger tile size on L1) which we think will result in a different tiling which we know does compile through mlir-air:
> > #config = #iree_codegen.lowering_config<tile_sizes = [[128, 128], [0, 0, 128], [64, 64], [0, 0, 8]]>
> 
> Sounds great, thanks, @newling !

with the updated logic I ended up getting
`#iree_codegen.lowering_config<tile_sizes = [[128, 128], [0, 0, 152], [64, 64], [0, 0, 8]]>`
which is still crashing, is that the case for you as well @newling ?

**Comment by nirvedhmeshram** [comment]
> > > @erwei-xilinx here is the IR: https://gist.github.com/newling/ba8ab6d035c60077525e2398e02c3f51
> > > But note that this is not blocking -- I've discussed with @nirvedhmeshram and he's going to try a different algorithm (which prioritized a larger tile size on L1) which we think will result in a different tiling which we know does compile through mlir-air:
> > > #config = #iree_codegen.lowering_config<tile_sizes = [[128, 128], [0, 0, 128], [64, 64], [0, 0, 8]]>
> > 
> > 
> > Sounds great, thanks, @newling !
> 
> with the updated logic I ended up getting `#iree_codegen.lowering_config<tile_sizes = [[128, 128], [0, 0, 152], [64, 64], [0, 0, 8]]>` which is still crashing, is that the case for you as well @newling ?

Nevermind, added another constrain and now its working as expected.

**Comment by nirvedhmeshram** [comment]
> @nirvedhmeshram is this ready for review?

Yes it is. Also, I am getting 128 for k too now. There the outer tile should be a multiple of (inner k tile* packing) as the inner k logic is a bit different.

**Review by newling — APPROVED** [review]
Looks good!

File: compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp [hunk 1]
```diff
@@ -207,12 +200,26 @@ static LogicalResult setRootConfigForPadPackPipeline(
...
+  auto tileN1 = findLargestFactor((int)initShape[1], 16 * tilingScaleFactor,
+                                  (int)packedSizes[1]);
+  auto tileK1 = findLargestFactor((int)lhsShape[1] / (int)packedSizes[2],
                                   2 * tilingScaleFactor);
+
+  auto tileM0 =
+      findLargestFactor((int)initShape[0], 64 * tilingScaleFactor, (int)tileM1);
+  auto tileN0 =
+      findLargestFactor((int)initShape[1], 64 * tilingScaleFactor, (int)tileN1);
+  auto tileK0 = findLargestFactor((int)lhsShape[1], 256,
```
Abhishek-Varma, line 220: I understand this is coming off late but any changes to KernelDispatch's logic should be supplemented with a lit test.

[compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lowering_strategy.mlir](https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lowering_strategy.mlir) is where the lit test should go - helps in seeing the immediate effect of new additions/updates to the existing tests.

Can we please address this in a follow up PR? @nirvedhmeshram

nirvedhmeshram: done https://github.com/nod-ai/iree-amd-aie/pull/299

PR #296, opened 2024-04-24
Option to require tiles to be powers of 2

newling:
The good news is that this PR allows matmul of size m=n=k=2432 to compile to the end (before it was hitting a failure where pack ops failed to lower due to imperfect tiling). 

It needs tidyinng up, but feel free to comment now...  I know @nirvedhmeshram has a similar solution proposed: https://github.com/nod-ai/iree-amd-aie/pull/293

Alternative (and possibly more robust) solution is padding all inputs to be of a "sufficient" size (for this problem size, I have found that m=n=2560 works). There are still things to work out to make this solution feasible in my mind: (1) algo to find the smallest round up (2) do the actual padding in the producer of the tensors as mentioned in today's standup with @MaheshRavishankar

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
@@ -330,32 +330,56 @@ bool isMatmulProducerOfElementwise(linalg::LinalgOp linalgOp) {
   return false;
 }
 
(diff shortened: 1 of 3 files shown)


newling:
Sorry @yzhang93 this is a WIP, there is still a lot of clean up to do. Also I've thought a bit more about it and actually think a solution along the lines #293 is better.

newling:
I am closing because on second thoughts the solution in #293 makes more sense.

PR #297, opened 2024-04-24
[AMD-AIE] Add a pass to fuse consumer to the last innermost scf loop

Abhishek-Varma:
-- This commit adds a pass `--iree-amdaie-fuse-consumer-into-loop` to fuse in the consumer on the last innermost scf loop.

```
NOTE:
a. I have added a demarcation on the part of code that will get removed once the upstream changes land.
    So you may review anything outside the demarcated zone.
b. Have tried keeping the lit test as concise as I could while highlighting the use case
    that necessitates the need for this pass.
```


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseConsumerIntoLoop.cpp
```diff
@@ -0,0 +1,738 @@
...
+      [&](linalg::LinalgOp op) {
+        if (linalg::isaContractionOpInterface(op)) {
+          linalgOp = op;
+          return WalkResult::interrupt();
+        }
+        return WalkResult::advance();
+      });
+
+  if (!linalgOp) {
+    llvm::dbgs() << "----- Could not find any linalg op implementing "
```
nirvedhmeshram, line 696: please guard this with  LLVM_DEBUG.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseConsumerIntoLoop.cpp
```diff
@@ -0,0 +1,738 @@
...
+                    "ContractionOpInterface \n";
+    return;
+  }
+
+  Value::user_range users = linalgOp->getResult(0).getUsers();
+  if (!llvm::hasSingleElement(users)) {
+    llvm::dbgs() << "----- Expected only one user of the compute op : "
```
nirvedhmeshram, line 703: here as well.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseConsumerIntoLoop.cpp
```diff
@@ -0,0 +1,738 @@
...
+                 << linalgOp << "\n";
+    return;
+  }
+
+  Operation *terminatorStoreOp = *(users.begin());
+  if (!(isa<tensor::InsertSliceOp, tensor::ParallelInsertSliceOp>(
+          terminatorStoreOp))) {
+    llvm::dbgs() << "----- Expected either tensor.insert_slice OR "
```
nirvedhmeshram, line 711: here as well.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseConsumerIntoLoop.cpp
```diff
@@ -0,0 +1,738 @@
...
+                    "tensor.parallel_insert_slice to be the only user of the "
+                    "compute op : "
+                 << linalgOp << "\n";
+    return;
+  }
+
+  if (!terminatorStoreOp) {
+    llvm::dbgs() << "----- Consumer ops are already fused it seems \n";
```
nirvedhmeshram, line 719: here as well.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/fuse_consumer_into_loop.mlir
```diff
@@ -0,0 +1,98 @@
...
+            //      FORALL:   }
+            %extracted_slice = tensor.extract_slice %arg8[%arg6, %arg7] [32, 32] [1, 1] : tensor<64x64xf32> to tensor<32x32xf32>
+            %6 = linalg.matmul ins(%arg0, %arg1 : tensor<32x32xf32>, tensor<32x32xf32>) outs(%extracted_slice : tensor<32x32xf32>) -> tensor<32x32xf32>
+            scf.forall.in_parallel {
+                tensor.parallel_insert_slice %6 into %arg8[%arg6, %arg7] [32, 32] [1, 1] : tensor<32x32xf32> into tensor<64x64xf32>
+            }
+        }
+        %3 = tensor.empty() : tensor<64x64xf32>
+        %4 = tensor.empty() : tensor<64x64xf32>
+        %5 = linalg.elemwise_binary {fun = #linalg.binary_fn<add>} ins(%2, %3 : tensor<64x64xf32>, tensor<64x64xf32>) outs(%4 : tensor<64x64xf32>) -> tensor<64x64xf32>
```
nirvedhmeshram, line 43: maybe I am not caught up but will this pass wotk if this was a linalg.generic doing elementwise operations in its region? If yes, can we add that as a test? If not then how do we reach this input in our codegen especially if there are two or more elementwise in a single generic op region.

nirvedhmeshram (changes requested):
Looking good overall, just a question and few nits.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/fuse_consumer_into_loop.mlir
Abhishek-Varma: TLDR: Yes, I'll modify one of the test to use `linalg.generic` instead to demonstrate the same, sure.

More context:
Yes, it'd work on ANY consumer which implements `TilingInterface` and `DestinationPassingStyleInterace` - the caller of the upstream API only need to worry about the sending the right producer (`tensor.parallel_insert_slice` or `tensor.insert_slice`) and then the API will capture ANY consumer that satisfies the aforementioned constraint.

`linalg.generic`,`linalg.elementwise_*`, `linalg.matmul`, `linalg.softmax`, etc are few examples of the same. So, yes, it'd work.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/fuse_consumer_into_loop.mlir
```diff
@@ -0,0 +1,106 @@
+// RUN: iree-opt --split-input-file --pass-pipeline='builtin.module(func.func(iree-amdaie-fuse-consumer-into-loop))' %s | FileCheck %s --check-prefix=FORALL
+// RUN: iree-opt --split-input-file --pass-pipeline='builtin.module(func.func(iree-amdaie-fuse-consumer-into-loop{use-scf-for=true}))' %s | FileCheck %s --check-prefix=FOR
+
+// FORALL-LABEL: @fuse_consumer_into_scfforall
```
yzhang93, line 4: These checks are hard to read in the middle of the tests. Could you please add all the checks after the test?

yzhang93 (changes requested):
Mostly LGTM. Thanks for adding the pass!

nirvedhmeshram (approved):
Looks from my side.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseConsumerIntoLoop.cpp
```diff
@@ -0,0 +1,742 @@
...
+  loopBody->walk<WalkOrder::PostOrder, ReverseIterator>(
+      [&](linalg::LinalgOp op) {
+        if (linalg::isaContractionOpInterface(op)) {
+          linalgOp = op;
+          return WalkResult::interrupt();
+        }
+        return WalkResult::advance();
+      });
+
+  if (!linalgOp) {
```
yzhang93, line 695: Here you should add another check to see if there's a following elementwise op after this contractionOp. If there's no such following op, then directly return. Otherwise it will run this pass and fail when there's matmul only.

Abhishek-Varma: Very good point!

Well, the fuse-consumer API will take care of that check and return failure.

So on the caller site, instead of re-performing the same check, I have added an `emitWarning` to bail out cleanly.

Abhishek-Varma: I've also added a test case to demonstrate the same.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseConsumerIntoLoop.cpp
```diff
@@ -0,0 +1,743 @@
...
+  if (!terminatorStoreOp) {
+    LLVM_DEBUG(llvm::dbgs()
+               << "----- Consumer ops are already fused it seems \n");
+    return;
+  }
+
+  std::optional<SCFFuseConsumerOfSliceResult> fusedConsumer =
+      tileAndFuseConsumerOfSlice(rewriter, terminatorStoreOp);
+  if (!fusedConsumer) {
+    scfLoopOp->emitWarning(
```
nirvedhmeshram, line 730: wait we dont want the compiler emiting warning everytime it sees a gemm with no elemtwise in the end, so we should just have a debug print here. I am in favor of the test you added to show things work for this case though.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseConsumerIntoLoop.cpp
yzhang93: If there's an unpack op after forall loop, it will try to fuse unpack inside the loop. This needs to be avoided.

Abhishek-Varma:
@yzhang93 and @nirvedhmeshram - I've addressed your comments.

Please re-review/approve.

nirvedhmeshram (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseConsumerIntoLoop.cpp
yzhang93: @Abhishek-Varma Since you did force-push, it's hard for me to find the changes in your pass. Can you summarize how you solve the above problem?

Abhishek-Varma: To unblock us - I've restricted the fusion to only those `Linalg` ops which are implementing both `TilingInterface` and `DestinationPassingStyle`.

yzhang93 (approved):
LGTM. Thanks!

PR #302, opened 2024-04-25
Add workgroup unroll and logical objectfifo distribution pass

jtuyls:
Adds a pass which unrolls the threaded `scf.forall` within a workgroup, but looks for broadcast capability of dma ops and tries to hoist them out of the loops if possible. When dma ops are unrolled, new logical objectfifos are inserted to allow them to be distributed on different physical tiles. 

This pass consists of following steps:

1. Transform threaded `scf.forall` into `scf.for` (this transformation exists in upstream and will be pulled in from there once changes propagate)
2. After the `scf.forall` to `scf.for` transformations, typically all `affine.apply` are located within the inner `scf.for` and this transformation will try to hoist them to accommodate dma op hoisting.
3. Unroll the `scf.for`'s within a workgroup. Try to hoist dma ops that don't depend on the loops' induction variables to avoid duplicated dma ops. Unroll and distribute the remaining logical objectfifos in the `scf.for` loops.
4. Assign tiles to the logical objectfifos with local memory space (L1/core).
5. Assign physical AIE tiles to the remaining logical objectfifos. This rewrite takes an iterative approach by matching logical objectfifos and only assigning tiles when linked through dma ops with other logical objectfifos which already have tiles assigned. If the linked logical objectfifos don't have tiles assigned yet, we will return a failure and give the linked logical objectfifos a chance to assign tiles before returning to this one.


compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
```diff
@@ -161,7 +198,7 @@ def AMDAIE_LogicalObjectFifoConsume: AMDAIE_Op<"logicalobjectfifo.consume", []>
   }];
 }
 
-def AMDAIE_LogicalObjectFifoFromMemrefOp : AMDAIE_Op<"logicalobjectfifo.from_memref", []> {
+def AMDAIE_LogicalObjectFifoFromMemrefOp : AMDAIE_Op<"logicalobjectfifo.from_memref", [Pure]> {
```
yzhang93, line 201: What does "Pure" mean?

yzhang93 (commented):
I didn't quite follow the broadcast logic. It would be helpful if you can walk me through the codes. I'll do another review after I understand more.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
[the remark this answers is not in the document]
jtuyls: It add this to allow cse to eliminate this op if possible. 

> // Always speculatable operation that does not touch memory.  These operations
// are always legal to hoist or sink.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.cpp
```diff
@@ -164,6 +181,97 @@ LogicalObjectFifoFromMemrefOp DmaCpyNdOp::getTargetObjectFifo() {
...
+LogicalResult LogicalObjectFifoFromMemrefOp::canonicalize(
+    LogicalObjectFifoFromMemrefOp logicalObjectFifo,
+    PatternRewriter &rewriter) {
+  // We only canonicalize the tiles for now, so return if empty
+  if (logicalObjectFifo.getTiles().empty()) {
+    return success();
+  }
+
+  // Check whether we are dealing with AMDAIE::TileOp or xilinx::AIE::TileOp
+  if (isa<xilinx::AIE::TileOp>(
```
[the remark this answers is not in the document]
jtuyls: At some point after this transformation, the `LogicalObjectFifoFromMemref` operation's tiles can become `xilinx::AIE::TileOp`. But good point, this should probably not be necessary and  I can probably get rid of the need for by postponing `AMDAIE::TileOp` to `xilinx::AIE::TileOp` conversion. For now, I will just remove this logic from this PR.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Transforms.h
```diff
@@ -0,0 +1,25 @@
...
+
+namespace mlir::iree_compiler::AMDAIE {
+
+/// Convert a scf.forall into a set of nested scf.for loops.
+/// TODO(jornt): Replace with upstream `forallToFor` function once merged.
+LogicalResult forallToFor(RewriterBase &rewriter, scf::ForallOp forOp);
+
+/// Hoist an affine apply op on a scf.for op's induction variable
+/// TODO(jornt): Can we generalize this to go into upstream?
+LogicalResult hoistForAffineApplyOp(RewriterBase &rewriter, affine::AffineApplyOp applyOp);
```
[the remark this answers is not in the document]
jtuyls: Added it: https://github.com/nod-ai/iree-amd-aie/pull/302/files#diff-b0c567f0c0c8869a4b8458dd7f898561ef30760df9c6e5b48fe1fa3fa038677d

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.cpp
```diff
@@ -43,6 +43,18 @@ LogicalResult ControlCodeOp::verify() {
 // AMDAIE_CoreOp
 //===----------------------------------------------------------------------===//
 
+// Hardcoded row_offset == 2 -> AIE core rows start from 2
+// TODO(jornt): avoid hardcoding here. Add a device model/identifier to loop up
+// core offset. This will be handled in a follow-up.
```
Abhishek-Varma, line 48: Any doc comment should be in triple quotes. `//` -> `///`

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
```diff
@@ -55,20 +57,24 @@ def AMDAIE_CoreOp: AMDAIE_Op<"core", [SingleBlock]>, Results<(outs Index)> {
   }];
 
   let arguments = (
-    ins Index:$col,
-        Index:$row
+    ins Index:$tile
```
Abhishek-Varma, line 60: The `description` of this op should also change. I see instances of `col` and `row` in it.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEHoistForAffineApply.cpp
```diff
@@ -0,0 +1,56 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree-amd-aie/Transforms/Passes.h"
+#include "iree-amd-aie/Transforms/Transforms.h"
+#include "mlir/Dialect/Affine/IR/AffineOps.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+
+namespace mlir::iree_compiler::AMDAIE {
+
+LogicalResult hoistForAffineApplyOp(RewriterBase &rewriter,
```
Abhishek-Varma, line 14: Add a doc comment.

Abhishek-Varma, line 14: Whenever we implement a pass it's a good practice to add `DEBUG_TYPE` - similar to [this](https://github.com/nod-ai/iree-amd-aie/blob/71834c279050348faf747472747efa04afea0df2/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp#L21).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
```diff
@@ -0,0 +1,503 @@
...
+
+#include "aie/Dialect/AIE/IR/AIEDialect.h"
+#include "iree-amd-aie/IR/AMDAIEOps.h"
+#include "iree-amd-aie/Transforms/Passes.h"
+#include "iree-amd-aie/Transforms/Transforms.h"
+#include "mlir/Dialect/SCF/Transforms/Transforms.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+#define DEBUG_TYPE "iree-amdaie-unroll-and-distribute-workgroup"
```
Abhishek-Varma, line 15: Like this one.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
```diff
@@ -0,0 +1,503 @@
...
+LogicalResult hoistAffineApplyDependingOnFor(mlir::ModuleOp moduleOp) {
+  IRRewriter rewriter(moduleOp.getContext());
+  moduleOp->walk([&](affine::AffineApplyOp applyOp) {
+    (void)hoistForAffineApplyOp(rewriter, applyOp);
+  });
+  return success();
+}
+
+/// Unroll the scf.forall loops selected for parallel execution on the AIE
+/// array. Try to hoist dma ops that don't depend on the loops\ induction
```
Abhishek-Varma, line 52: I might be wrong but did you mean `loops'` instead of `loops\`?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
```diff
@@ -0,0 +1,498 @@
...
+
+      // Checks to determine internal logical objectfifos:
+      //   1) If source is on global memory, target is an interal logical
+      //   objectfifo.
+      //   2) If target is on global memory, source is an interal logical
+      //   objectfifo.
+      //   3) If target memory space is more local than source memory space,
+      //   target is an interal logical objectfifo.
+      //   4) If target memory space is less local than source memory space,
+      //   source is an interal logical objectfifo.
```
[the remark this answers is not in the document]
Abhishek-Varma: If I understand correctly this essentially means the following :-
```
internal_logical_objectfifo = (source.memory < target.memory) ? source : target
```
where `memory -> [L1, L2, Global]` and `L1 < L2 < Global`

@jtuyls please correct me if I'm wrong.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
```diff
@@ -0,0 +1,503 @@
...
+  LogicalResult hoistDmaOps(PatternRewriter &rewriter, scf::ForOp forOp) const {
+    bool hoistHappened{false};
+
+    // Create utility function to check whether an operand depends on a scf.for
+    // induction variable or a value within the scf.for's scope
+    auto dependsOnLoop = [&](OpOperand &operand) -> bool {
+      Operation *op = operand.get().getDefiningOp();
+      if (!op) return operand.get() == forOp.getInductionVar();
+
+      // Check fo induction var and whether the parent scf.for is the same as
```
Abhishek-Varma, line 115: `Check fo induction` -> `Check for induction` ?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
```diff
@@ -0,0 +1,503 @@
...
+    return failure();
+  }
+
+  /// Unroll the for loop and while doing so, create new clones of
+  /// AMDAIE::LogicalObjectFifoFromMemrefOp so they can be distributed
+  /// onto multiple physical locations later.
+  LogicalResult loopUnrollAndDistributeLogicalObjectFifos(
+      RewriterBase &rewriter, scf::ForOp forOp) const {
+    Block *loopBodyBlock = forOp.getBody();
+    auto builder = OpBuilder::atBlockTerminator(loopBodyBlock);
```
Abhishek-Varma, line 152: Expand `auto` to explicit types here and elsewhere in the code.

We should use `auto` only when the type can be seen from the in-line context. Cases where `auto` is favourable are `cast<SOME_TYPE>`, `dyn_cast<SOME_TYPE>`, `rewriter.create<SOME_OP>`, etc.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
```diff
@@ -0,0 +1,503 @@
...
+    if (stepInt == (ubInt - lbInt)) return failure();
+    Value forOpIV = forOp.getInductionVar();
+    forOp.setLowerBound(
+        rewriter.create<arith::ConstantIndexOp>(rewriter.getUnknownLoc(), 0));
+    forOp.setUpperBound(
+        rewriter.create<arith::ConstantIndexOp>(rewriter.getUnknownLoc(), 1));
+    forOp.setStep(
+        rewriter.create<arith::ConstantIndexOp>(rewriter.getUnknownLoc(), 1));
+
+    // Iterate through the loop and create body
```
Abhishek-Varma, line 172: Maybe add here that we're skipping the first iteration of the loop ?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
```diff
@@ -0,0 +1,503 @@
...
+
+  LogicalResult matchAndRewrite(
+      AMDAIE::LogicalObjectFifoFromMemrefOp logicalObjectFifo,
+      PatternRewriter &rewriter) const override {
+    if (!logicalObjectFifo.getTiles().empty()) {
+      return failure();
+    }
+
+    auto memSpace = logicalObjectFifo.getMemorySpace();
+    // Skip logical objectfifos with within local memory as they should
```
Abhishek-Varma, line 367: `Skip logical objectfifos within` ?

Abhishek-Varma (changes requested):
Thanks for this @jtuyls ! It involves quite a few heavy-lifting. This is great!

First round of comments from my end - I've only seen the aesthetics and few nuances.

I'd review this again in detail.

A comment: Since this PR involves two passes - it's preferred if we segregate them into two separate PRs - we can let this one be and perhaps make a note for future PRs. It helps in reviewing by minimising the window of focus.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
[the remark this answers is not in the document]
jtuyls: It was the other way around: `internal_logical_objectfifo = (source.memory < target.memory) ? target : source`, but there were some more complexities to handle internal producer-consumer dependencies. I think it was unclear code and probably not robust, so I rewrote it. I am still not fully happy with it as I am making an assumption that data will move more locally to the core on the input side and away on the output, but it's a reasonable assumption for now. Could you have a look at the new hoisting function? 

I am still looking at writing this more generally if possible. The structural issue is that these dmas implicitly depend on each other through the memory/objectfifo they write into/read from.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.cpp
jtuyls: Thanks, done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEHoistForAffineApply.cpp
jtuyls: Done.

jtuyls: Thanks, done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
jtuyls: Yes, I did. Thanks for the catch.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
jtuyls: Thanks, done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
jtuyls: Done.

jtuyls: Although, I don't think this is practiced very carefully throughout iree. I am myself more on the 'use more auto side', especially for simple/common types, but I am ok with being more strict in this codebase as long as it's consistent.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
jtuyls: Thanks, fixed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
jtuyls: Added it.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
Abhishek-Varma: You're correct. Just that I came across this coding style practice during one of the upstream PRs in `llvm-project`  recently and thus trying to adopt it here too. It definitely helps.

Thanks for following up!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
```diff
@@ -0,0 +1,511 @@
...
+    // Keep a pointer to the last non-terminator operation in the original block
+    // so that we know what to clone (since we are doing this in-place).
+    Block::iterator srcBlockEnd = std::prev(loopBodyBlock->end(), 2);
+
+    // Update loop bounds
+    int64_t lbInt = getConstantIntValue(forOp.getLowerBound()).value();
+    int64_t ubInt = getConstantIntValue(forOp.getUpperBound()).value();
+    int64_t stepInt = getConstantIntValue(forOp.getStep()).value();
+    if (lbInt != 0 || stepInt != 1) return failure();
+    if (stepInt == (ubInt - lbInt)) return failure();
```
Abhishek-Varma, line 170: It's a good practice to put braces around trivial bodied `if`, `for` and `while`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
```diff
@@ -0,0 +1,511 @@
...
+    Value forOpIV = forOp.getInductionVar();
+    forOp.setLowerBound(
+        rewriter.create<arith::ConstantIndexOp>(rewriter.getUnknownLoc(), 0));
+    forOp.setUpperBound(
+        rewriter.create<arith::ConstantIndexOp>(rewriter.getUnknownLoc(), 1));
+    forOp.setStep(
+        rewriter.create<arith::ConstantIndexOp>(rewriter.getUnknownLoc(), 1));
```
Abhishek-Varma, line 177: Guess I'm failing to understand part of `forOp.set<>Bound(...)`.

If we have made sure that `lb = 0` and `step = 1` then why do we need to set those bounds again? Shouldn't we just do that for the `ub` ?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
```diff
@@ -0,0 +1,511 @@
...
+      return failure();
+    }
+    return success();
+  }
+};
+
+/// Assign tiles to the logical objectfifos with local memory space (L1).
+/// The tiles are derived from the usage of the logical objectfifos within
+/// core operations, which are already assigned a tile location.
+LogicalResult assignLocalAieTiles(mlir::ModuleOp moduleOp) {
```
Abhishek-Varma, line 249: I guess you don't need `mlir::` namespace here.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
```diff
@@ -0,0 +1,511 @@
...
+  IRRewriter rewriter(moduleOp.getContext());
+
+  // Map from local objectfifos found to the tiles where they are used
+  DenseMap<AMDAIE::LogicalObjectFifoFromMemrefOp,
+           llvm::SmallSetVector<Value, 16>>
```
Abhishek-Varma, line 254: No need for `llvm::` here elsewhere in the code that's making use of any LLVM ADT.

And while we're at it can you instead try the following :-
1. `SmallSetVector<Value>` ? If this doesn't work then try the one below.
2. `SmallSetVector<Value, 16>`

I'm saying this because I got to know upstream that we don't require the second argument for `SmallVector` anymore.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
```diff
@@ -0,0 +1,511 @@
...
+      logicalObjectFifosToTiles;
+
+  // Utility function insert a local objectfifo - tile pair into the local
+  // objectfifo to tile map
+  auto insertTile = [&](AMDAIE::LogicalObjectFifoFromMemrefOp logicalObjectFifo,
+                        Value tileResult) -> void {
+    if (!logicalObjectFifosToTiles.contains(logicalObjectFifo)) {
+      llvm::SmallSetVector<Value, 16> tiles;
+      logicalObjectFifosToTiles[logicalObjectFifo] = tiles;
```
Abhishek-Varma, line 263: Wouldn't `logicalObjectFifosToTiles[logicalObjectFifo] = {};` just work?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+/// Return the tiles of the targets of the users of this logical
+/// objectfifo.
+LogicalResult getUserTargetTiles(
+    AMDAIE::LogicalObjectFifoFromMemrefOp logicalObjectFifo,
+    SmallVectorImpl<AMDAIE::TileOp> &tiles) {
+  llvm::SmallSetVector<AMDAIE::TileOp, 16> tileSet;
+  for (Operation *user : logicalObjectFifo->getUsers()) {
+    if (auto dmaOp = dyn_cast<AMDAIE::DmaCpyNdOp>(user);
+        dmaOp && dmaOp.getSourceObjectFifo() == logicalObjectFifo) {
```
Abhishek-Varma, line 338: I guess both of these functions can be combined into one since the only differing thing between the two is `(target, source)` and `(source, target)`.

So maybe you can pass in another variable to the unified function `fetchTarget` ?

And this way the `if` can become :-
```
if (auto dmaOp = dyn_cast<AMDAIE::DmaCpyNdOp>(user)) {
   // Based on `fetchTarget` perform the check and extraction.
   ...
   // Rest of the code remains the same.
}
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/test/canonicalize.mlir
```diff
@@ -0,0 +1,26 @@
...
+//===----------------------------------------------------------------------===//
+
+// CHECK-LABEL: func.func @logicalobjectfifo_from_memref
+// CHECK:       %[[C0:.+]] = arith.constant 0 : index
+// CHECK:       %[[C1:.+]] = arith.constant 1 : index
+// CHECK:       %[[TILE_0_0:.*]] = amdaie.tile(%[[C0]], %[[C0]])
+// CHECK:       %[[TILE_0_1:.*]] = amdaie.tile(%[[C0]], %[[C1]])
+// CHECK:       %[[TILE_1_0:.*]] = amdaie.tile(%[[C1]], %[[C0]])
+// CHECK:       %[[TILE_1_1:.*]] = amdaie.tile(%[[C1]], %[[C1]])
+// CHECK:       %[[I0:.*]] = amdaie.logicalobjectfifo.from_memref %[[ARG0:.*]], {%[[TILE_0_0]], %[[TILE_0_1]], %[[TILE_1_0]], %[[TILE_1_1]]}
```
Abhishek-Varma, line 14: No need to match `%[[I0:.*]]`.

Just do `amdaie.logicalobjectfifo.from_memref %[[ARG0:.*]], {%[[TILE_0_0]], %[[TILE_0_1]], %[[TILE_1_0]], %[[TILE_1_1]]}`

Abhishek-Varma (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEHoistForAffineApply.cpp
```diff
@@ -0,0 +1,60 @@
...
+  }
+  return failure();
+}
+
+namespace {
+struct AMDAIEHoistForLoopAffineApply
+    : public impl::AMDAIEHoistForLoopAffineApplyBase<
+          AMDAIEHoistForLoopAffineApply> {
+  void runOnOperation() override {
+    Operation *parentOp = getOperation();
```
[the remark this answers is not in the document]
jtuyls: I am writing an op-agnostic pass on purpose here as this transformation doesn't make any assumption on the operation it's executed on: https://mlir.llvm.org/docs/PassManagement/#op-agnostic-operation-passes. A lot of the core mlir passes are written this way, e.g. https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/SCF/Transforms/Passes.td. Does this have any disadvantage?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
jtuyls: I don't see this rule in the mlir/llvm codebase? I see `if/else` without braces everywhere.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/hoist_for_affine_apply.mlir
```diff
@@ -0,0 +1,43 @@
...
+      %0 = affine.apply #map(%i)
+      %1 = affine.apply #map1(%i)
+    }
+  }
+  // CHECK:     scf.for %[[IV1:.+]] = %[[LB1]] to %[[UB1]] step %[[STEP1]]
+  // CHECK:       %[[MAP0:.+]] = affine.apply #map(%[[IV1]])
+  // CHECK:       %[[MAP1:.+]] = affine.apply #map1(%[[IV1]])
+  // CHECK:       scf.for %[[IV2:.+]] = %[[LB2]] to %[[UB2]] step %[[STEP2]]
+  // CHECK-NOT:     %[[MAP2:.+]] = affine.apply #map(%[[IV1]])
+  // CHECK-NOT:     %[[MAP3:.+]] = affine.apply #map1(%[[IV1]])
```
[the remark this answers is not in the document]
Abhishek-Varma: I guess I didn't explain properly - the above snippet I suggested was for the part where the `affine.apply` is hoisted (i.e. between the loops).

For sake of completeness, this is what you should be doing instead :-
```
  //      CHECK:     scf.for %[[IV1:.+]] = 
  // CHECK-NEXT:       affine.apply #map(%[[IV1]])
  // CHECK-NEXT:       affine.apply #map1(%[[IV1]])
  // CHECK:          scf.for
  // CHECK-NOT:        affine.apply #map(%[[IV1]])
  // CHECK-NOT:        affine.apply #map(%[[IV1]])
```
(The above assumes that NO other op is being created by the pass between the two loops, hence `CHECK-NEXT`)

To reiterate what I was trying to suggest here is - always check the following while creating lit test to make it as concise as possible :- 
1. Are the SSA values being matched used anywhere? If not, we might as well move on to the next important/relevant match.
2. What other extra matching can I let go of which isn't communicating the main intention of the pass?

Let me know if that makes sense.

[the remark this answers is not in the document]
nirvedhmeshram: I think CHECK-NEXT would be wrong here because it implies that the particular check is on the next immediate line, however the pass does not guarantee that and it doesn't need to, that is either order is correct sc.for -> map -> map1  or sc.for-> map1 -> map, so just using CHECK is correct here as we dont want the test to fail in case that order changes in the future.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
jtuyls: Yes, this is indeed not needed for now because of those checks.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
jtuyls: Thanks, yes, this works.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/hoist_for_affine_apply.mlir
```diff
@@ -0,0 +1,43 @@
+// RUN: iree-opt --split-input-file --pass-pipeline="builtin.module(func.func(iree-amdaie-hoist-for-affine-apply))" %s | FileCheck %s
+
+// CHECK-LABEL: @hoist_affine_map
+// CHECK-SAME:  %[[LB1:.+]]: index, %[[UB1:.+]]: index, %[[STEP1:.+]]: index, %[[LB2:.+]]: index, %[[UB2:.+]]: index, %[[STEP2:.+]]: index
```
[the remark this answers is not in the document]
jtuyls: Done

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEHoistForAffineApply.cpp
```diff
@@ -0,0 +1,60 @@
...
+  // Check whether the parent op is a for op and the map operand is already the
+  // induction variable of the parent for op.
+  Value operand = applyOp.getMapOperands()[0];
+  if (auto forOp = dyn_cast<scf::ForOp>(applyOp->getParentOp());
+      forOp == scf::getForInductionVarOwner(operand)) {
+    return failure();
+  }
+
+  // Rewrite if operand is an induction variable.
+  if (auto forOp = scf::getForInductionVarOwner(operand)) {
```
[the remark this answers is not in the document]
jtuyls: Done

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/unroll_and_distribute_workgroup.mlir
```diff
@@ -0,0 +1,520 @@
...
+
+// CHECK-LABEL: @unroll_and_distribute_workgroup_1x4_cores
+// CHECK-DAG:   %[[C0:.*]] = arith.constant 0 : index
+// CHECK-DAG:   %[[C1:.*]] = arith.constant 1 : index
+// CHECK-DAG:   %[[C2:.*]] = arith.constant 2 : index
+// CHECK-DAG:   %[[C3:.*]] = arith.constant 3 : index
+// CHECK:       scf.forall (%[[ARG0:.*]], %[[ARG1:.*]]) in (1, 1) {
+// CHECK:         amdaie.workgroup {
+// CHECK:           %[[TILE_0:.*]] = amdaie.tile(%[[C0]], %[[C2]])
+// CHECK:           %[[CORE_0:.*]] = amdaie.core(%[[TILE_0]])
```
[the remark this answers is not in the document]
jtuyls: Done

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+  AMDAIEUnrollAndDistributeWorkgroupPass() = default;
+  AMDAIEUnrollAndDistributeWorkgroupPass(
+      const AMDAIEUnrollAndDistributeWorkgroupPass &pass){};
+  void runOnOperation() override;
+};
+
+void AMDAIEUnrollAndDistributeWorkgroupPass::runOnOperation() {
+  MLIRContext *context = &getContext();
+  if (failed(workgroupForallToFor(getOperation()))) {
```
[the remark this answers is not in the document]
jtuyls: Done

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
jtuyls: Yes, removed it

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/test/canonicalize.mlir
jtuyls: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEHoistForAffineApply.cpp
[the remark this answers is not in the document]
Abhishek-Varma: I'm unable to open the doc link - guess the site is down.
But if that's the case, you may definitely leave it just like that.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/hoist_for_affine_apply.mlir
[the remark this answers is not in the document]
nirvedhmeshram: Ya i think its fine this way too.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
nirvedhmeshram: we follow google coding standard which iree and hence iree-amd-aie follows. If this was in core then we follow llvm standards which say no trivial braces all the time I think.
I think its a bit nuanced then just "always trivial braces". I think the way it is now is correct in this case becuase it fits on single line and there is no else part.
https://google.github.io/styleguide/cppguide.html#Formatting_Looping_Branching

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/hoist_for_affine_apply.mlir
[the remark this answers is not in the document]
jtuyls: @nirvedhmeshram  I wrote my answer before I saw yours as I didn't refresh the page. I think you make a good point on `CHECK-NEXT` 

> that is either order is correct sc.for -> map -> map1 or sc.for-> map1 -> map

I think `CHECK` will also check for the order, so `scf.for-> map1 -> map` will fail.

Based on the discussion, I updated to `CHECK-DAG` to allow for a different order as well.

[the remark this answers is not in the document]
nirvedhmeshram: oh you are right, CHECK-DAG is most suitable here.

yzhang93 (approved):
Thanks for the changes!

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.cpp
```diff
@@ -164,6 +180,65 @@ LogicalObjectFifoFromMemrefOp DmaCpyNdOp::getTargetObjectFifo() {
...
+
+LogicalResult LogicalObjectFifoFromMemrefOp::verify() {
+  // Check whether the tile arguments are all of type AMDAIE::TileOp
+  bool allTileOps = llvm::all_of(getTiles(), [](Value result) {
+    return isa<TileOp>(result.getDefiningOp());
+  });
+  if (allTileOps) {
+    return success();
+  }
+  return failure();
```
Abhishek-Varma, line 227: You can just do the following :-
```
if(llvm::all_of(...)) {
    return success();
}
return failure();
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
```diff
@@ -0,0 +1,507 @@
...
+    // induction variable or a value within the scf.for's scope
+    auto dependsOnLoop = [&](OpOperand &operand) -> bool {
+      Operation *op = operand.get().getDefiningOp();
+      if (!op) return operand.get() == forOp.getInductionVar();
+
+      // Check for an induction var and whether the parent scf.for is the same
+      // as the one we're using for the hoist
+      auto parentForOp = op->getParentOfType<scf::ForOp>();
+      return (operand.get() == forOp.getInductionVar()) ||
+             (parentForOp && parentForOp == forOp);
```
Abhishek-Varma, line 88: Wouldn't just `parentForOp == forOp` work ?

Abhishek-Varma (approved):
Thanks for addressing the review comments @jtuyls ! LGTM!
Approving this - please address the few other changes requested and we can get this in.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
jtuyls: Yes, updated it

PR #304, opened 2024-04-26
MLIR --timing flag to aie2xclbin passes

newling:
- This is useful to locate slow passes in mlir-aie (I used it to find the slow pass for https://github.com/Xilinx/mlir-aie/pull/1431)
- Required bump of mlir-aie wheel used


nirvedhmeshram (approved):
LGTM!

PR #307, opened 2024-04-26
Integrate FuseConsumerIntoLoop to pack-peel pipeline and fix bufferization

yzhang93:
- Integrate FuseConsumerIntoLoopPass to pack-peel packline. Adjust the order of certain passes to avoid bufferization problem.
- Modified BufferizeToAllocationPass and FusePackIntoLoopPass to generate correct IR without extra alloc/copies after bufferization.
- Add an e2e matmul-elementwise example to demonstrate the lowering through AIR and generate NPU instructions.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
@@ -45,19 +45,31 @@ static LogicalResult applyBufferizeToAllocation(RewriterBase &rewriter,
 /// elementwise op's input is the output of the matmul op and has already been
 /// promoted, there is no need to promote such operand again.
 static SmallVector<Value> getInputOutputOperands(linalg::LinalgOp &linalgOp) {
-  SmallVector<Value> operands;
-  for (auto operand : linalgOp->getOperands()) {
-    if (isMatmulProducerOfElementwise(linalgOp)) {
-      auto defOp = operand.getDefiningOp<linalg::LinalgOp>();
-      // Continue the loop without including the operand which is the output of
-      // matmul.
-      if (defOp && isMatmul(defOp)) {
-        continue;
-      }
+  if (isMatmulProducerOfElementwise(linalgOp)) {
+    SmallVector<Value> operands;
+    for (auto operand : linalgOp->getOperands()) {
+      if (isMatmulInDefChain(operand)) continue;
+      operands.push_back(operand);
     }
-    operands.push_back(operand);
+    return operands;
   }
-  return operands;
+  return linalgOp->getOperands();
+}
+
+/// Utility to fetch input operands from the LinalgOp (matmul or elementwise
+/// op). For matmul-elementwise special case, since one of the elementwise op's
+/// input is the output of the matmul op and has already been promoted, there is
+/// no need to promote such operand again.
+static SmallVector<Value> getInputOperands(linalg::LinalgOp &linalgOp) {
+  if (isMatmulProducerOfElementwise(linalgOp)) {
+    SmallVector<Value> operands;
+    for (auto operand : linalgOp.getDpsInputs()) {
+      if (isMatmulInDefChain(operand)) continue;
+      operands.push_back(operand);
+    }
+    return operands;
+  }
+  return linalgOp.getDpsInputs();
 }
 
 /// Utility to fetch operands from the defining ops of LinalgOp's input
@@ -71,28 +83,13 @@ static SmallVector<Value> getInputOutputOperands(linalg::LinalgOp &linalgOp) {
 static FailureOr<SmallVector<Value>> getOperandsFromDefOp(
     linalg::LinalgOp &linalgOp) {
   SmallVector<Value> operands;
-  // For matmul only dispatch, we only want to fetch the input operand of the
-  // pack ops.
-  auto candidateOperands = isMatmulProducerOfElementwise(linalgOp)
-                               ? linalgOp->getOperands()
-                               : linalgOp.getDpsInputs();
-  for (auto operand : candidateOperands) {
-    auto defOp = operand.getDefiningOp();
-    if (!defOp) {
+  for (auto input : linalgOp.getDpsInputs()) {
+    auto defOp = input.getDefiningOp();
+    // The defining op has to be a pack op, fail otherwise.
+    if (!defOp || !isa<tensor::PackOp>(defOp)) {
       return failure();
     }
-    // For matmul-elementwise special case, continue the loop without including
-    // the operand which is the output of matmul.
-    if (isMatmulProducerOfElementwise(linalgOp)) {
-      auto defLinalgOp = dyn_cast<linalg::LinalgOp>(defOp);
-      if (defLinalgOp && isMatmul(defLinalgOp)) {
-        continue;
-      }
-    } else {
-      if (!isa<tensor::PackOp>(defOp)) {
-        return failure();
-      }
-    }
+    // We only want to fetch the input operand of the pack op.
     operands.push_back(defOp->getOperand(0));
   }
   return operands;
@@ -109,7 +106,7 @@ static FailureOr<SmallVector<Value>> getOperandsToBufferize(
       return getInputOutputOperands(linalgOp);
     /// Create new allocation only for Lhs, Rhs.
     case BufferizeOperand::Input:
-      return SmallVector<Value>(linalgOp.getDpsInputs());
+      return getInputOperands(linalgOp);
     /// Create new allocations only for Out.
     case BufferizeOperand::Output:
       return SmallVector<Value>(linalgOp.getDpsInits());
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
@@ -47,7 +47,7 @@ static LogicalResult applyBufferizeToAllocation(RewriterBase &rewriter,
 static SmallVector<Value> getInputOutputOperands(linalg::LinalgOp &linalgOp) {
   if (isMatmulProducerOfElementwise(linalgOp)) {
     SmallVector<Value> operands;
-    for (auto operand : linalgOp->getOperands()) {
+    for (Value operand : linalgOp->getOperands()) {
       if (isMatmulInDefChain(operand)) continue;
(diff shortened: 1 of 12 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
```diff
@@ -45,19 +45,31 @@ static LogicalResult applyBufferizeToAllocation(RewriterBase &rewriter,
...
+}
+
+/// Utility to fetch input operands from the LinalgOp (matmul or elementwise
+/// op). For matmul-elementwise special case, since one of the elementwise op's
+/// input is the output of the matmul op and has already been promoted, there is
+/// no need to promote such operand again.
+static SmallVector<Value> getInputOperands(linalg::LinalgOp &linalgOp) {
+  if (isMatmulProducerOfElementwise(linalgOp)) {
+    SmallVector<Value> operands;
+    for (auto operand : linalgOp.getDpsInputs()) {
```
Abhishek-Varma, line 66: `auto` -> `Value`.

Expand `auto` here and elsewhere.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
```diff
@@ -298,34 +287,44 @@ bool isMatmul(linalg::LinalgOp linalgOp) {
          match6DLinalgGenericMatmul(linalgOp);
 }
 
+/// Utility to identify if the input operand has matmul-like op in its
+/// def-chain.
+bool isMatmulInDefChain(Value operand) {
+  if (isa<BlockArgument>(operand)) return false;
```
Abhishek-Varma, line 293: Add braces around `return false`. It's a good practice to guard even if the body is a single statement. Do this here and elsewhere.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -135,7 +98,9 @@ def AMDAIEFusePackIntoLoop :
     Option<"fusePackDepth", "fuse-pack-depth", "int64_t", /*default=*/"1",
       "Set the depth until which we would keep fusing producer tensor.pack chain">,
     Option<"useSCFFor", "use-scf-for", "bool", /*default=*/"true",
-      "Set the innermost scf loop type to fuse tensor.pack ops into">
+      "Set the innermost scf loop type to fuse tensor.pack ops into">,
+    Option<"targetElementwise", "target-elementwise", "bool", /*default=*/"false",
+      "Set if the compute op within the loop is an elementwise op">
```
Abhishek-Varma, line 103: `Set if the compute op within the loop is an elementwise op` -> `Set if the target compute op within the loop is an elementwise op`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
```diff
@@ -298,34 +287,44 @@ bool isMatmul(linalg::LinalgOp linalgOp) {
...
+  while (Operation *defOp = operand.getDefiningOp()) {
```
Abhishek-Varma, line 294: I guess you can completely eliminate the `if` check here. The `while` check that follows would anyway help in that.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
```diff
@@ -298,34 +287,44 @@ bool isMatmul(linalg::LinalgOp linalgOp) {
...
+            if (isMatmul(op)) {
+              findMatmul = true;
+              return WalkResult::interrupt();
+            }
+            return WalkResult::advance();
+          });
+      if (findMatmul) return true;
+    }
+    operand = defOp->getOperand(0);
+    if (isa<BlockArgument>(operand)) return false;
```
Abhishek-Varma, line 315: No need for this. The control is flowing to the beginning of the `while` and the check there will anyway take care of it.

Abhishek-Varma (changes requested):
Requesting few changes. Looks good overall.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
yzhang93: Just wondering is there a particular reason to do so?

Abhishek-Varma: Thanks for addressing the comment.

Regarding the reason - yes, it increases the verbosity of the code.

`auto` is to be used only when the data type of the LHS is known within the one-line context. Example of cases where we're to use `auto` is `dyn_cast<....>`, `cast<....>`, `rewrite.create<...>`, etc.

In the example of cases like above the reader of the code doesn't know whether `getDpsInputs()` is returning a `Value`, `OpOperand` or some other data type. So, it helps in stating the LHS' data type in cases like these and use `auto` in cases where the RHS explicitly mentions what the data type is.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
```diff
@@ -71,28 +83,13 @@ static SmallVector<Value> getInputOutputOperands(linalg::LinalgOp &linalgOp) {
...
-  auto candidateOperands = isMatmulProducerOfElementwise(linalgOp)
-                               ? linalgOp->getOperands()
-                               : linalgOp.getDpsInputs();
-  for (auto operand : candidateOperands) {
-    auto defOp = operand.getDefiningOp();
-    if (!defOp) {
+  for (Value input : linalgOp.getDpsInputs()) {
+    auto defOp = input.getDefiningOp();
+    // The defining op has to be a pack op, fail otherwise.
+    if (!defOp || !isa<tensor::PackOp>(defOp)) {
```
Abhishek-Varma, line 89: Just use `getDefiningOp<tensor::PackOp>()` - and in this case `auto` would be correct else we should've used `Operation*` if we only use `getDefiningOp()`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
```diff
@@ -28,6 +29,15 @@ static FailureOr<tensor::ExtractSliceOp> getTensorExtractSliceDefiningOp(
               sliceOp.getSource().getDefiningOp())) {
         return sliceOp;
       }
+      if (isa<BlockArgument>(sliceOp.getSource())) {
+        auto blkArg = dyn_cast<BlockArgument>(sliceOp.getSource());
+        for (auto blkOperand :
```
Abhishek-Varma, line 34: Again, please expand `auto`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
```diff
@@ -298,33 +287,46 @@ bool isMatmul(linalg::LinalgOp linalgOp) {
...
+              findMatmul = true;
+              return WalkResult::interrupt();
+            }
+            return WalkResult::advance();
+          });
+      if (findMatmul) {
+        return true;
+      }
+    }
+    operand = defOp->getOperand(0);
```
Abhishek-Varma, line 315: Why do we only traverse the def-chain of the first operand here?

I understand that any LinalgOp which is not matmul-like will result in `false`, but what other operations besides `scf::ForOp` can we get in the def-chain whose first operand holds the relevance? I can only think of `tensor.pack/unpack`.

Abhishek-Varma (changes requested):
Few comments. Looks good overall! Thanks @yzhang93 .

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
yzhang93: Normally the defining op should be tensor.pack or tensor.extract_slice, and we only take the source operand from it.

MaheshRavishankar: I have the same question. Maybe some more context here would help. Ideally we dont want such a walk here. Its seems a bit fragile.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
```diff
@@ -298,33 +269,54 @@ bool isMatmul(linalg::LinalgOp linalgOp) {
...
+            return WalkResult::interrupt();
+          }
+          return WalkResult::advance();
+        });
+    if (findMatmul) {
+      return true;
+    }
+  }
+
+  for (auto operand : defOp->getOperands()) {
```
Abhishek-Varma, line 302: `auto` -> `Value`.

Abhishek-Varma (approved):
The def-chain logic now looks okay to me. A nit change. Looks good to me. Thanks @yzhang93 !

@MaheshRavishankar please take a look at the def-chain logic and stamp if looks you to as well.

yzhang93:
@erwei-xilinx I think the latest bump of MLIR-AIR broke the new added test for matmul_elementwise in `pack_peel_pipeline.mlir`. I see below errors:
```error: 'air.channel.put' op operand #1 must be memref of any type values, but got '!air.async.token'```.

MaheshRavishankar (approved):
Spoke offline about the trying to find "matmul producer" path. That seems more specific than what we need. We shouldnt be indexing on specific code path here. If something is being produced from a for loop, should just assume it is some fused computation and not really look at its body to look for matmuls. 

Just a  note for now, but would be good to unwind soon.

yzhang93:
@erwei-xilinx As discussed, I enabled/disabled special passes for matmul-elementwise fusion, and currently the new tests passed compilation. Please take a look at the changes here https://github.com/nod-ai/iree-amd-aie/pull/307/files#diff-42f0d0bb098689f25ee68e8f05ec6c2eaa89ce41a6394d7d99c1d1c912943b38R467-R495

erwei-xilinx (approved):
The changes on AIR pipeline look good to me. Thanks!

PR #308, opened 2024-05-01
Runtime changes to use DPU Sequence rather than run_pp mode

nirvedhmeshram:
Currently the kernels are compiled to run in an outdated run_pp (run as a post processing kernel) mode. This PR contains the runtime changes needed to run the kernels in the DPU Sequence mode which is soon to be put on maintenance mode but still better that run_pp. The goal is to run as a transactional binary with control packets so this can be a temporary stop gap till we move to that flow. The compiler side changes for this have to happen in mlir-aie.


nirvedhmeshram:
The E2E test wont pass since it uses the mlir-aie wheel that does not have the mlir-aie changes

PR #309, opened 2024-05-01
Bump IREE to latest

daveliddell:
Added a bunch of missing # includes.


yzhang93 (commented):
I wonder what change make these header files required now.

daveliddell:
> I wonder what change make these header files required now.

Yeah, I don't know!  Maybe some header files got cleaned up so that they didn't include as much unnecessary stuff.  In any case, it's a good idea in general for a .cpp to # include the stuff that it uses, so that it's less prone to such indirect breakage.  These files were using various dialects directly but not including their headers.

yzhang93 (approved):

daveliddell:
@yzhang93 Thanks, Vivian!

PR #299, opened 2024-04-25
[NFC] Add lowering_strategy test for pad-pack config changes

nirvedhmeshram:
Add a test to show the recent change in the packing/tiling config ordering for pad-pack pipeline

PR #310, opened 2024-05-01
For the pad-pack pipeline, avoid tile size ratio of 2

newling:
This is low priority as pad-pack isn't going to be our main pipeline, and this doesn't fix any hangs since https://github.com/Xilinx/mlir-air/pull/562/ landed,  but maybe a nice thing to have (we can use in peel-pack later).

The pad-pack pipeline currently targets a 4x4 AIE array, so for a fixed shared memory tiling in the M- and N- directions (M0 and N0) it is better to have 

```
M1=M0/4, N1 = N0/4
```
than 
```
M1=M0/2, N1 = N0/2
```

where M1 and N1 are the tiling sizes on the AIE core memory. 

This PR checks for M0/M1 = 2 and if possible halves M1 so that M0/M1 = 4 (ditto the N- dimension) 

Note that since https://github.com/Xilinx/mlir-air/pull/562/files has landed this change is purely for performance and not for correctness.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
@@ -151,8 +151,19 @@ FailureOr<ParameterSetting> ParameterSetting::create(linalg::LinalgOp linalgOp,
   const uint64_t N = initShape[1];
   const uint64_t K = lhsShape[1];
 
+  // If we are conservative with ensuring that tiles A, B, and C fit at the
+  // different memory levels, we should choose the scale factor based
+  // on the largest of the types of A, B, and C, which is the type of C, as
+  // accumulation always happens in an element type with at least as may bits as
+  // the operand types. We currently choose the scale factor based on the
+  // element type of the lhs operand, which works with the current workloads (we
+  // have not seen OOM errors).
+  //
+  // Long-term we should use a different approach, one which takes into account
+  // the element types of all tensors which need to be allocated in memory
+  // simultaneously.
   FailureOr<unsigned> maybeScaleFactor =
-      getTilingScaleFactor(initType.getElementType());
+      getTilingScaleFactor(lhsType.getElementType());
   if (failed(maybeScaleFactor)) {
     return linalgOp.emitOpError(
         "does not have the expected bitwidth (64, 32, 16, or 8), could not "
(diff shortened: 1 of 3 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -272,6 +281,20 @@ FailureOr<PadPackTiling> PadPackTiling::create(linalg::LinalgOp linalgOp) {
   uint32_t M0 = findLargestFactor(M, maxL0Size, M1);
   uint32_t N0 = findLargestFactor(N, maxL0Size, N1);
 
+  auto downScale = [](uint32_t pack, uint64_t memTile, uint64_t coreTile) {
+    if (memTile / coreTile != 2) return coreTile;
+    if (memTile % coreTile != 0) return coreTile;
+    if (coreTile % 2 != 0) return coreTile;
```
nirvedhmeshram, line 287: maybe a comment for reason behind each of the if statements?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -272,6 +281,20 @@ FailureOr<PadPackTiling> PadPackTiling::create(linalg::LinalgOp linalgOp) {
   uint32_t M0 = findLargestFactor(M, maxL0Size, M1);
   uint32_t N0 = findLargestFactor(N, maxL0Size, N1);
 
+  auto downScale = [](uint32_t pack, uint64_t memTile, uint64_t coreTile) {
```
nirvedhmeshram, line 284: maybe a comment saying this is doing step 3 or some summary of what step 3 above does.

nirvedhmeshram (approved):
LGTM, left some minor comments.

yzhang93 (changes requested):
Sorry, I'm adding a blocker here because I'm making a major change unifying parameter chosen for both pipelines. Let's evaluate afterwards.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lowering_strategy.mlir
```diff
@@ -132,7 +132,7 @@ builtin.module {
 
 // -----
 
-// CHECK-PAD-PACK{LITERAL}: #config = #iree_codegen.lowering_config<tile_sizes = [[128, 128], [0, 0, 128], [64, 64], [0, 0, 8]]>
+// CHECK-PAD-PACK{LITERAL}: #config = #iree_codegen.lowering_config<tile_sizes = [[128, 128], [0, 0, 128], [32, 32], [0, 0, 8]]>
```
yzhang93, line 135: Not sure why it generates 64 and 128 without this extra step. The original thought of using below codes (16 and 64) is trying to get the relation of 4 times between two tiling levels.
  ```
auto maxL1Size = 16 * scaleFactor;
  auto maxL0Size = 64 * scaleFactor;
```

newling: So here scaleFactor = 4. 

What happens is this:

Step 1: 64 is chosen at the AIE core tile size  (maxL1Size is 64)
Step 2: 128 is chosen as memtile size (maxL1Size is 256, but 2432 / 256  = 9.5)
Step 3 then shrinks 64 to 32. 

The logic for steps 1 and 2 was introduced in https://github.com/nod-ai/iree-amd-aie/pull/293 with tests added in https://github.com/nod-ai/iree-amd-aie/pull/299

The above PRs from @nirvedhmeshram  basically changed it from 
Step 1: Choose memtile size
Step 2: Choose AIE core size

To 
Step 1: Choose AIE core size
Step 2: Choose memtile size 

It was an ad hoc solution to getting an important matmul shape to compile. I think what was happening was that for 2432, the memtile size was being chosen as 152 (=19 * 8) which wasn't desirable. Sorry, I can't remember more details than that at the moment

newling: Anyway -- I'm happy to go back to the drawingboard for choosing tile sizes. We could consider for example that next gen AIE isn't a 4x4 grid of cores. We should also consider that its the _sum_ of A, B, and C tiles which are constrained simultaneously, so if A is forced to have small tiling due to constraints, the B can have larger tiling.

yzhang93: Thanks for your explanation, and I think it now makes sense to me. Also agreed that we should come up with better tiling strategy after recent demos. I'll remove the blocker and you can merge the PR if you like.

yzhang93 (approved):

PR #311, opened 2024-05-02
[Delegate] Enhancements in support of OPT matmul

daveliddell:
Added several enhancements to AIE delegates:

- Now using a 4x4 vector matmul of shape 8x768x768, stored in Azure
- dtypes of model and AIE kernel inputs and outputs now defined in one place
- All tensor conversion and copying code templatized to automate those functions and use minimal implementations when possible
- New PDL transform to convert 8x768x768 matmul in linalg MLIR to external func call
- prototype code for direct XRT buffer use (disabled until HAL buffers are allocated with XRT)
- copy from result buffer to kernel now switchable by macro (not needed for both kernels we're currently using)

### DIFF
--- a/experimental/delegate/.gitignore
+++ b/experimental/delegate/.gitignore
@@ -1,0 +1,2 @@
+*.vmfb
+*.sub.mlir
--- a/experimental/delegate/opt.mlir
+++ b/experimental/delegate/opt.mlir
@@ -1,0 +1,47 @@
+// RUN: iree-opt --pass-pipeline="builtin.module(iree-preprocessing-apply-pdl-patterns{patterns-file=%p/opt.pdl.mlir}, cse)" %s | FileCheck %s
+
+// TODO: Add filecheck checks
+
+#x86_64_target = #hal.executable.target<"llvm-cpu", "embedded-elf-x86_64", {
+  data_layout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128",
+  native_vector_size = 32 : index,
+  target_triple = "x86_64-none-elf"
+}>
+
+// The target devices that the program will run on. We can compile and run with
+// multiple targets, but this example is maintaining an implicit requirement
+// that the custom kernel being spliced in is supported by the target device,
+// hence we only support llvm-cpu here.
+#cpu_target = #hal.device.target<"llvm-cpu", [
+  #x86_64_target
+]>
+
+#map = affine_map<(d0, d1) -> (d0, d1)>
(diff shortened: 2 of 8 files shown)


daveliddell:
Readme is outdated.  I'll work on that next

experimental/delegate/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.vmfb
```
MaheshRavishankar, line 1: Is this intentional?

experimental/delegate/opt.mlir
```diff
@@ -0,0 +1,47 @@
+// RUN: iree-opt --pass-pipeline="builtin.module(iree-preprocessing-apply-pdl-patterns{patterns-file=%p/opt.pdl.mlir}, cse)" %s | FileCheck %s
+
+// TODO: Add filecheck checks
```
MaheshRavishankar, line 3: Maybe just add a simple FileCheck test... Just to show that there are two matmul calls.

experimental/delegate/.gitignore
daveliddell: Yes, the demos are designed to run from the delegate directory for now.  I intend for those build artifacts not to show up in git status, etc., for any potential users

experimental/delegate/opt.mlir
daveliddell: Good idea!  Will do

daveliddell: Added one for single OPT matmul as well

experimental/delegate/.gitignore
MaheshRavishankar: It would be good if the way this is done, it just either uses a tmp directory or it generates in the working directory. Polluting the codespace isnt great. But can be a follow up.

MaheshRavishankar (approved):

experimental/delegate/.gitignore
daveliddell: It generates in the cwd already, and no automation generates these files, so you get pollution only if you manually ask for it!  :-) .  You can output to a different dir, or run the demo from a different cwd if you'd rather.

Personally, for demos in a source tree, because the demo inputs are there, I find it easiest to generate my outputs there as well, as long as they don't show up in my git status.  I have also observed that most developers prefer to operate this way, probably because we're too lazy to make a copy of the demo tree or work with separate input and output dirs.

In any case, when I get around to making e2e tests for CI, I'll be sure not to pollute the source tree.  I agree that the build process should never create any artifacts mixed in with the source.

daveliddell:
Thanks, Mahesh!

PR #312, opened 2024-05-02
Add an e2e matmul transpose test 

newling:
Currently fails to compile (error message copied into test).

### DIFF
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -479,15 +479,24 @@ run_matmul_test \
     --compile-only "0" \
     --do_transpose_rhs "0"
 
-# TODO: fix this. 
-#: error: 'aie.dma_bd' op Cannot give more than 4 dimensions for step sizes and wraps in this  tile (got 5 dimensions).
 run_matmul_test \
-  --name_prefix "transpose" \
+  --name_prefix "transpose_int32" \
+  --lhs_rhs_type "i32" \
+  --acc_type "i32" \
+  --m "8" --n "16" --k "32" \
+  --do_transpose_rhs "1"
+
+# TODO: fix this. 
+#: error: 'aie.dma_bd' op Cannot give more than 4 dimensions for step sizes
+#  and wraps in this  tile (got 5 dimensions).
+run_matmul_test \
+  --name_prefix "transpose_bf16" \
   --lhs_rhs_type "bf16" \
   --acc_type "f32" \
   --m "256" --n "256" --k "256" \
   --do_transpose_rhs "1" \
   --expect-compile-failure "1"
+
 
 # The below matmul case passes with 
 # tile_sizes = [[1, 1], [0, 0, 250], [1, 1], [0, 0, 2]], packedSizes = [1, 1, 5]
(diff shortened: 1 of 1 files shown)


yzhang93 (changes requested):
It's useful if you also add another test for matmul_transpose_b with i32 type.

yzhang93:
Can you try to add vectorization passes and see if bf16 ci pass after that? @erwei-xilinx reported peano + bf16 can only work in ci with vectorization passes.

erwei-xilinx:
> Can you try to add vectorization passes and see if bf16 ci pass after that? @erwei-xilinx reported peano + bf16 can only work in ci with vectorization passes.

No I meant the pack-peel pipeline which is missing vectorization passes. The transpose pipeline is still using pad-pack which already has vectorization.

yzhang93:
> > Can you try to add vectorization passes and see if bf16 ci pass after that? @erwei-xilinx reported peano + bf16 can only work in ci with vectorization passes.
> 
> No I meant the pack-peel pipeline which is missing vectorization passes. The transpose pipeline is still using pad-pack which already has vectorization.

A good point! @newling We should also add matmul transpose test for pack-peel pipeline. If that works with vectorization, we don't have to waste time fixing failure for pad-pack.

newling:
Hey @yzhang93 and @erwei-xilinx PR https://github.com/nod-ai/iree-amd-aie/pull/319 adds vectorization to the pack peel pipeline, but the pipeline needs work to make vectorization work. So tests are not fully enabled in https://github.com/nod-ai/iree-amd-aie/pull/319

erwei-xilinx:
@newling the error comes from AIR failing to analyze a data access pattern with vectorized code. If `linalg.generic` can remain at AIR input then the test should pass. I wonder how difficult do you think it is to move the vectorization passes to after, or in the middle of, AIR passes?

newling:
> @newling the error comes from AIR failing to analyze a data access pattern with vectorized code. If `linalg.generic` can remain at AIR input then the test should pass. I wonder how difficult do you think it is to move the vectorization passes to after, or in the middle of, AIR passes?

I think the vectorization pass needs to happen before bufferization, but I'm not sure.

newling:
@erwei-xilinx @yzhang93 @MaheshRavishankar 

I tried permuting the order in which vectorization and bufferization happen from:


```
appendVectorizationToPipeline(funcPassManager);
addAMDAIEBufferizePasses(funcPassManager);                                                                                                                                                                 ```

to 

```
addAMDAIEBufferizePasses(funcPassManager);                                                                                                                                                                
appendVectorizationToPipeline(funcPassManager);
```

but the vectorization pass fails. I think it crashes in `linalg::tileLinalgOp(rewriter, genericOp, opts);` (assertion error about incorrect number of results). 

So hopefully you can find another solution, because running vectorization later doesn't seem like it will be a quick-fix. Hopefully the L1 memory can be created to be the correct size at construction..

erwei-xilinx:
> @erwei-xilinx @yzhang93 @MaheshRavishankar
> 
> I tried permuting the order in which vectorization and bufferization happen from:

Thanks for trying that, @newling . I put together this PR to AIR which attempts to analyze the data access pattern of post-vectorization kernel code: https://github.com/Xilinx/mlir-air/pull/565 This will actually get the CI to pass for pack-peel bf16 GEMMs, but currently causing pad-pack to fail... Will keep looking into that issue tomorrow.

newling:
Thanks Erwei. Yes I think ideally we'd have transpose working on both pipelines.

yzhang93 (approved):

PR #314, opened 2024-05-02
Add circular dma copy nd operation and pass to insert them

jtuyls:
This PR adds a new circular dma copy nd operation which represents an indefinite copy from the source to the target logical objectfifo. This operation waits for data to be produced on the source logical objectfifo and produces into the target logical objectfifo. This PR is an intermediary one to prepare for adding the `AMDAIEToAIEWorkgroup` pass and to simplify the review process: https://github.com/nod-ai/iree-amd-aie/compare/main...jornt_peeled_matmul_e2e#diff-45372414f96b9a5e4b0da5d7b089c11e8cf5cbe68f92eda0fe96d0c8c03bb010R290


compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.cpp
```diff
@@ -180,6 +180,113 @@ LogicalObjectFifoFromMemrefOp DmaCpyNdOp::getTargetObjectFifo() {
   return dyn_cast<LogicalObjectFifoFromMemrefOp>(getTarget().getDefiningOp());
 };
 
+// Build a CircularDmaCpyNdOp with mixed static and dynamic entries.
```
yzhang93, line 183: Does it have to be three build functions for static/dynamic entries?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCircularDmas.cpp
```diff
@@ -0,0 +1,62 @@
...
+#define DEBUG_TYPE "iree-amdaie-insert-circular-dma"
+
+namespace mlir::iree_compiler::AMDAIE {
+
+namespace {
+
+/// Insert circular dma copy operations from L2 to L1 and L1 to L2. This is
+/// hardcoded for now because of how MLIR-AIE is set up, but can be generalized
+/// in the future.
+LogicalResult insertCircularDmaCpyNdOps(Operation *op) {
```
yzhang93, line 20: This pass seems to just take the `AMDAIE::DmaCpyNdOp` between L1 and L2 memory spaces and replace it with this new op `AMDAIE::CircularDmaCpyNdOp`. First question, is it a necessary process? In your following passes, whenever you need it can you just directly detect the memory spaces like this pass does?

Second thing, if you have to keep this new op and pass, it sounds more reasonable to me if the name of the pass is `convertToCircularDmaCpyNdOps`. Because what it does is replacing the op instead of inserting a new op.

jtuyls: Earlier, I had it in a subsequent pass and I got it out to simplify that one. I am also expecting that this logic can become more complex in the future as:
- You could do L1 -> L2 and L2 -> L1 non-circular as well, but I don't think that works with current MLIR-AIE.
- We will probably have to insert new transformations after this pass and before the next one, for example to hoist these circular buffers  with no dependencies on the surrounding loops out of those loops.  This is a valid transformation for `CircularDmaCpyNdOp` in certain circumstances, but not for `DmaCpyNdOp`:

```
scf.for %i = ... {
  %0 = amdaie.circular_dma_cpy_nd(...)
  ...
}  
```

to 

```
%0 = amdaie.circular_dma_cpy_nd(...)
scf.for %i = ... {
  ...
}  
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.cpp
jtuyls: Yes, one for static, one for dynamic and one for static/dynamic mixed. I based myself on the `Subview` build functions for this: https://github.com/llvm/llvm-project/blob/2996cb7970fb536d7db1088d571dc8d2f9ad35fa/mlir/include/mlir/Dialect/MemRef/IR/MemRefOps.td#L1833 and https://github.com/llvm/llvm-project/blob/2996cb7970fb536d7db1088d571dc8d2f9ad35fa/mlir/lib/Dialect/MemRef/IR/MemRefOps.cpp#L2685

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCircularDmas.cpp
jtuyls: > Second thing, if you have to keep this new op and pass, it sounds more reasonable to me if the name of the pass is convertToCircularDmaCpyNdOps. Because what it does is replacing the op instead of inserting a new op.

Good point, thanks. I made this change.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaToCircularDma.cpp
```diff
@@ -0,0 +1,62 @@
...
+namespace {
+
+/// Convert dma copy operations to circular dma copy operations from L2 to L1
+/// and L1 to L2. This is hardcoded for now because of how MLIR-AIE is set up,
+/// but can be generalized in the future.
+LogicalResult convertDmaToCircularDma(Operation *op) {
+  IRRewriter rewriter(op->getContext());
+  op->walk([&](AMDAIE::DmaCpyNdOp dmaOp) {
+    auto sourceMemSpace = dmaOp.getSourceObjectFifo().getMemorySpace();
+    auto targetMemSpace = dmaOp.getTargetObjectFifo().getMemorySpace();
```
Abhishek-Varma, line 24: `auto` -> `Attribute` ?

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/test/roundtrip.mlir
```diff
@@ -26,6 +26,51 @@ func.func @logicalobjectfifo_from_memref(%arg0: memref<1x1x8x16xi32, 1>) {
...
+  return
+}
+
+// -----
+
+// CHECK-LABEL: func.func @circular_dma_cpy_nd_mixed
+// CHECK: amdaie.circular_dma_cpy_nd
+// CHECK-SAME: (%[[ARG0:.*]][%c0, %c0, %c0, %c0] [1, 1, %c8, %c16] [%c128, %c128, %c16, 1],
+// CHECK-SAME: %[[ARG1:.*]][%c0, %c0, %c0, %c0] [1, 1, %c8, %c16] [%c128, %c16, %c16, 1])
+// CHECK-SAME: (!amdaie.logicalobjectfifo<memref<1x1x8x16xi32, 1>>, !amdaie.logicalobjectfifo<memref<8x16xi32, 1>>)
```
Abhishek-Varma, line 62: nit: It'd be nice to indent these in a similar way you did with `dma_to_circular_dma.mlir`.

Abhishek-Varma (approved):
A very good PR! Thanks @jtuyls !
A few nits.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/test/roundtrip.mlir
jtuyls: Yes, done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaToCircularDma.cpp
jtuyls: Thanks for the catch, I copied some code from older code where I have more auto.. Updated it.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/test/roundtrip.mlir
jtuyls: I saw the rest of the file was not nicely indented either, so will create a separate small follow-up to clean up the file.

jtuyls:
@yzhang93 I merged this as you didn't block and I addressed your comments. As for the question whether this warrants a separate pass, we can always merge it back into a subsequent pass later if it becomes an issue, but I think it's better standalone. Let me know if there is anything else and I will take care of it in a follow-up.

PR #315, opened 2024-05-03
Lowering strategy code and diagnostics improvements 

newling:
This is a non-functional change tothe lowering strategy code. It just 

1) makes the error clearer when the tensor size does not divide the AIE instruction size
2) refactors the functions for obtaining packing sizes to be slightly clearer
3) avoids non-const ref vectors in APIs where vectors could be const 
4) small style changes 

I would like to follow it up with a small change to improve stategy to avoid using just 2 rows of 4x4 array.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
@@ -289,9 +289,9 @@ static LogicalResult setRootConfigForPadPackPipeline(
   // small GEMM sizes where possible. Assume for now that we are working on a
   // 4x4 AIE array.
 
-  auto maybe = PadPackTiling::create(linalgOp);
-  if (failed(maybe)) return failure();
-  auto tiling = maybe.value();
+  auto maybePadPackTiling = PadPackTiling::create(linalgOp);
+  if (failed(maybePadPackTiling)) return failure();
+  auto padPackTiling = maybePadPackTiling.value();
 
   // Do packing first to allow better packing configs
   // ------------------------------------------------------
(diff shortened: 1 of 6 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -156,76 +169,162 @@ static LogicalResult setRootConfigForPackPeelPipeline(
...
+    mlir::FunctionOpInterface entryPointFn, linalg::LinalgOp linalgOp,
+    AIEConfig cfg) {
+  // Currently, the tile sizes are chosen empirically for large GEMM sizes,
+  // which are [64*s, 64*s, 256] for the first level and [16*s, 16*s, 16*s] for
+  // the second level, where 's' is the scaling factor based on the element
+  // type's bit width. Basic min/max constraints are added to avoid failure for
+  // small GEMM sizes where possible. Assume for now that we are working on a
+  // 4x4 AIE array.
+
+  auto maybe = PadPackTiling::create(linalgOp);
```
nirvedhmeshram, line 292: nit -> maybePadPackTiling.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -156,76 +169,162 @@ static LogicalResult setRootConfigForPackPeelPipeline(
...
+  if (failed(maybe)) return failure();
+  auto tiling = maybe.value();
```
nirvedhmeshram, line 294: tiling-> padPackTiling.

nirvedhmeshram (approved):
LGTM, couple of variables could be renamed for better readability if you want.

PR #323, opened 2024-05-07
[Delegate] Updates to add large matmul kernel

daveliddell:
Several new features in support of the Tres Leches model:

- Added a new matmul kernel of shape 8192x9728x2432 with bf16 inputs and f32 outputs
- Added a PDL to match linalg.matmul and rewrite to a delegate call for the new kernel
- Enhanced error handling, including catching some configuration errors that used to result in a crash without error message
- Choice of kernel to use now selected with the `DELEGATE_KERNEL_TO_USE` macro


experimental/delegate/CMakeLists.txt
```diff
@@ -64,9 +64,11 @@ iree_lit_test_suite(
     "mlp.mlir"
     "matmul.mlir"
     "opt.mlir"
+    "tres-leches.mlir"
```
newling, line 67: @daveliddell  Does this file need to be checked in?

daveliddell: Absolutely it does!  (Eheheh)

nirvedhmeshram (changes requested):
Looks good, just one request, changes tresleches/ tres-leches to something like largematmul/ large-matmul as the former is not informative/sensible for a general open source observer of the repository.

daveliddell:
> Looks good, just one request, changes tresleches/ tres-leches to something like largematmul/ large-matmul as the former is not informative/sensible for a general open source observer of the repository.

Updated.  Sorry, I'll get the hang of addressing the audience properly eventually!

nirvedhmeshram (approved):
LGTM!

PR #325, opened 2024-05-07
[Delegate] Converted delegate to expect f32 inputs

daveliddell:
By default, the delegate now expects f32 inputs instead of bf16.

For the Large Matmul case, there are now two PDLs for rewriting to call the delegate, one that matches bf16 inputs (large-matmul.pdl.mlir) and one that matches f32 inputs (large-matmul-f32.pdl.mlir).

### DIFF
--- a/experimental/delegate/mlp_aie_bf16_plugin.cpp
+++ b/experimental/delegate/mlp_aie_bf16_plugin.cpp
@@ -109,8 +109,8 @@ using B_DATATYPE = bfloat16_t;
 using C_DATATYPE = float; // bfloat16_t;
 
 // Types of the matmul LHS, RHS, and result, as seen by the model
-using ModelLhsDType = bfloat16_t;
-using ModelRhsDType = bfloat16_t;
+using ModelLhsDType = float; // bfloat16_t;
+using ModelRhsDType = float; // bfloat16_t;
 using ModelReturnDType = float;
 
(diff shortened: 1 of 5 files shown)


experimental/delegate/mlp_aie_bf16_plugin.cpp
```diff
@@ -109,8 +109,8 @@ using B_DATATYPE = bfloat16_t;
 using C_DATATYPE = float; // bfloat16_t;
 
 // Types of the matmul LHS, RHS, and result, as seen by the model
-using ModelLhsDType = bfloat16_t;
-using ModelRhsDType = bfloat16_t;
+using ModelLhsDType = float; // bfloat16_t;
+using ModelRhsDType = float; // bfloat16_t;
```
nirvedhmeshram, line 113: hmmm, I was expecting you to convert the input buffers to bf16 manually some where for this kernel?

nirvedhmeshram: ah I see the kernel types are defined right above.

nirvedhmeshram (approved):
LGTM!

experimental/delegate/mlp_aie_bf16_plugin.cpp
daveliddell: The magic of templates!  :-D  The converter/copier classes do that automatically, and those decisions are made at compile time during template param binding.

PR #329, opened 2024-05-08
Enable vectorization passes in pack-peel pipeline

erwei-xilinx:
New feature in MLIR-AIR:
- Added logic in `MemrefShrinkage` pattern (part of `AIRSegmentLoopFusion` pass) to support vectorized kernel code.

### DIFF
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -584,6 +584,34 @@ run_matmul_test \
     --pipeline "pack-peel" \
     --lhs_rhs_type "i32" \
     --acc_type "i32" \
-    --m "64"  --n "64" --k "160"
+    --m "64"  --n "64" --k "128"
+
+run_matmul_test \
+    --name_prefix "packPeel" \
+    --pipeline "pack-peel" \
+    --lhs_rhs_type "bf16" \
+    --acc_type "f32" \
+    --m "64"  --n "64" --k "128"
+
+run_matmul_test \
+    --name_prefix "packPeelLarge" \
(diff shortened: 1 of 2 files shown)


yzhang93 (approved):

yzhang93:
matmul_elementwise test failed. You may want to delete this test for now https://github.com/nod-ai/iree-amd-aie/blob/main/tests/samples/pack_peel_pipeline_matmul_elementwise.mlir#L70

erwei-xilinx:
> matmul_elementwise test failed. You may want to delete this test for now if it caused the failure. https://github.com/nod-ai/iree-amd-aie/blob/main/tests/samples/pack_peel_pipeline_matmul_elementwise.mlir#L70

Let me take a look at this one...

yzhang93 (changes requested):
Need to fix the ci first.

build_tools/ci/run_matmul_test.sh
```diff
@@ -584,6 +584,34 @@ run_matmul_test \
     --pipeline "pack-peel" \
     --lhs_rhs_type "i32" \
     --acc_type "i32" \
-    --m "64"  --n "64" --k "160"
+    --m "64"  --n "64" --k "128"
```
newling, line 587: Is it necessary to change 128 to 160?

erwei-xilinx: No both 128 and 160 works. That was an arbitrary choice.

newling (approved):
Thanks!

yzhang93 (approved):
Thanks for fixing the issues!

PR #331, opened 2024-05-08
Make loops-for-vectorization recognize matmul_transpose_b

newling:
With the current pipeline, matmul_tranpose_b ops are converted to matmul-like generics here: https://github.com/nod-ai/iree-amd-aie/blob/784dedb18b9768ebfff7b344906ee9052e8674eb/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPackAndTranspose.cpp#L121 (this is for the pad-pack pipeline, not sure how peel-pack pipeline differs). As a result, by the time the vectorization pass runs, there are no matmul_transpose_b operations (or equivalent linalg.generics) in the IR. 

I would to change this (in a future PR). Specifically I would like to change inner permutation of packing (see https://github.com/nod-ai/iree-amd-aie/blob/784dedb18b9768ebfff7b344906ee9052e8674eb/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp#L134) to have no permutation. This will result in matmul_transpose_b not appearing as matmul in the vectorization passes, but remaining as a matmul_transpose_b semantically.  

The motivation for this (future) change is that performing a permutation of the inner dimensions doesn't work for data types <32b , because dma copies cannot have a stride of less than 32 bits. The solution that we are pursuing is to convert a vector.contract with a transpose map for the B operand into a vector.transpose operation followed by a vector.contract with normal matmul maps for operands. Then, vector.transpose will lower to aievec.shuffle operations (which need to be implemented, but will supported in hardware (see [here](https://www.xilinx.com/htmldocs/xilinx2023_2/aiengine_ml_intrinsics/intrinsics/group__intr__gpvectorop__interleave.html))) and an aievec.matmul operations.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
@@ -69,37 +69,31 @@ class AMDAIEInsertLoopsForVectorizationPass
     auto numIterators = iteratorTypes.size();
 
     // No outer dimensions to tile if fewer than 4 iterators.
-    if (numIterators < 4) {
-      return failure();
-    }
+    if (numIterators < 4) return failure();
 
     // Matmul-like ops have 3 operands.
-    if (genericOp->getNumOperands() != 3) {
-      return failure();
-    }
+    if (genericOp->getNumOperands() != 3) return failure();
 
     // Don't transform to scf.for loops unless there is at least one
     // non-singleton loop to construct. This isn't strictly necessary, but
     // avoids generating a bunch of loops of size 1.
     if (llvm::all_of(genericOp->getOperands(), [&](Value operand) {
           return getNumInnerDims(operand) < 3;
-        })) {
+        }))
       return failure();
-    }
 
     assert(iteratorTypes.size() >= 3 && "expected at least 3 iterators here");
 
     // Check that innermost 3 iterators are 'parallel, parallel, reduction'.
     for (auto i : {2, 3}) {
-      if (!linalg::isParallelIterator(iteratorTypes[numIterators - i])) {
+      if (!linalg::isParallelIterator(iteratorTypes[numIterators - i]))
         return failure();
-      }
     }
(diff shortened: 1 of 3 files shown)


newling:
FYI @jsetoain

jsetoain:
> FYI @jsetoain

Thanks! I'm on the case. Btw, notice the link to internal documentation in a public facing post. [Xilinx public AIE intrinsic documentation](https://www.xilinx.com/htmldocs/xilinx2023_2/aiengine_ml_intrinsics/intrinsics/group__intr__gpvectorop__interleave.html) is a better alternative.

yzhang93:
> (this is for the pad-pack pipeline, not sure how peel-pack pipeline differs). As a result, by the time the vectorization pass runs, there are no matmul_transpose_b operations (or equivalent linalg.generics) in the IR.
> I would to change this (in a future PR). Specifically I would like to change inner permutation of packing (see
> 
> https://github.com/nod-ai/iree-amd-aie/blob/784dedb18b9768ebfff7b344906ee9052e8674eb/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp#L134
> 
> ) to have no permutation. This will result in matmul_transpose_b not appearing as matmul in the vectorization passes, but remaining as a matmul_transpose_b semantically.

Yes, we have to change the pack-transpose to make it different for matmul and matmul-transpose. This is also on my todo list. I'm wondering if the intrinsic values would change for matmul-transpose-b?

yzhang93 (approved):

newling:
>  I'm wondering if the intrinsic values would change for matmul-transpose-b?

At the end of all the passes we'll still end up with an aievec.matmul with the same operand shape (m=n=4 k=8 for bf16, etc). We might need to tile the B operand in L1 and L2 differently, but actually I think the code is fine as it currently is. My next step is to experiment with this. 

@jsetoain and I have also discussed the possibililty that IREE targets a vector.contract with a larger size (m=n=8 k=8 for bf16, for example) which then gets lowered to 4 aievec.matmul operations (4 because 8*8 / 4*4 = 4). But we think this is an optimization which can come later. Why is this an optimization? Because we could amortize the cost of performing the transpose on the core.

PR #333, opened 2024-05-08
Workaround for CI host memory limit

newling:
Adds line

```
sudo prlimit -lunlimited --pid $$
```

to .github/workflows/ci.yml

Proof that this works:
Fails: https://github.com/nod-ai/iree-amd-aie/pull/332
Passes: https://github.com/nod-ai/iree-amd-aie/pull/330

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,8 +107,36 @@ jobs:
       - name: E2E correctness matmul test
         run: |
           source .venv/bin/activate
-          # See https://github.com/Xilinx/mlir-air/issues/566
+          # Without this additional line an error like
+          #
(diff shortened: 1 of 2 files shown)


nirvedhmeshram (approved):
Since this is on our local runner and not one of github's runner I dont think we need to worry about any pricing changes so I think this is good to go!

yzhang93:
With this workaround, can we add SD3 tests to ci now? https://github.com/nod-ai/iree-amd-aie/pull/286

newling:
> With this workaround, can we add SD3 tests to ci now? #286

In theory yes, although it'll be slow. Let's be selective. I'm also working on making the baseline computation faster.

PR #334, opened 2024-05-09
Add additional matmul tests 

newling:
@yzhang93 it doesn't take as long as I expected (in the case that the tests pass). It takes about 1 minute to run these for me locally.

### DIFF
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -211,10 +211,18 @@ function run_matmul_test() {
 
   local do_transpose_rhs="0"
 
+  # The maximum number of elements to check for correctness.
+  # See https://github.com/iree-org/iree/blob/tools/testing/e2e/test_utils.c#L40-L47
+  local max_elements_to_check="20000"
+
   while [ "$#" -gt 0 ]; do
     case "$1" in
       --compile-only)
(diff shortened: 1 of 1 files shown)


build_tools/ci/run_matmul_test.sh
```diff
@@ -614,4 +644,66 @@ run_matmul_test \
     --acc_type "f32" \
     --m "128"  --n "128" --k "2304"
 
+run_matmul_test \
+    --name_prefix "mm6" \
```
yzhang93, line 648: Why these `name_prefix` are not in order?

build_tools/ci/run_matmul_test.sh
```diff
@@ -614,4 +644,66 @@ run_matmul_test \
...
+    --lhs_rhs_type "bf16" \
+    --acc_type "f32" \
+    --m "308"  --k "9728" --n "2432"  \
+    --expect-compile-failure "0" \
+    --compile-only "0"
```
yzhang93, line 653: For this and other tests below, I don't think you need to put these two flags.

yzhang93 (approved):
Thanks for adding these tests! Can you monitor the time spend on github ci for these tests? We should merge the PR if it doesn't take long.

PR #335, opened 2024-05-09
Modify the config settings for pack-peel pipeline

yzhang93:
- This PR modified the tiling and packing parameter settings for pack-peel pipeline following the existing strategy for pad-pack pipeline. These changes are specially made to make challenging matmul input sizes compiling and running on the device. The performance is not the main concern at this time.
- Bump MLIR-AIR to include several fixes in AIR.
- With this PR 7/8 matmul tests added here https://github.com/nod-ai/iree-amd-aie/pull/334 have passed using pack-peel pipeline. The remaining one failed at AIE level. @erwei-xilinx will help take a further look.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
@@ -73,11 +73,71 @@ FailureOr<std::array<uint32_t, 3>> getPackedSize(linalg::LinalgOp linalgOp,
   }
   return instructionSize;
 }
-}  // namespace
 
-static LogicalResult setRootConfigForPackPeelPipeline(
-    mlir::FunctionOpInterface entryPointFn, linalg::LinalgOp linalgOp,
-    AIEConfig cfg) {
+// Container class for the tiling at level 0 (the AIE shared memory) and level 1
+// (the AIE core) in the M-, N-, and K-dimensions of a matmul operation, using
+// the pad-pack approach to tiling a matmul. Also contains the packing sizes for
+// the M, N, and K dimensions. The packing sizes correspond to matmul
+// vector-instruction sizes for vectorizable types.
+class ParameterSetting {
+ public:
+  SmallVector<int64_t> getPackSizeL0() const {
+    return {m0Pack, n0Pack, k0Pack};
+  }
+  SmallVector<int64_t> getPackSizeL1() const {
+    return {m1Pack, n1Pack, k1Pack};
+  }
+
+  static FailureOr<ParameterSetting> create(linalg::LinalgOp linalgOp,
+                                            bool isPackPeel);
+
+  uint32_t getM0() const { return M0; }
+  uint32_t getN0() const { return N0; }
+  uint32_t getK0() const { return K0; }
+  uint32_t getM1() const { return M1; }
+  uint32_t getN1() const { return N1; }
+  uint32_t getK1() const { return K1; }
+  uint32_t getM0Pack() const { return m0Pack; }
+  uint32_t getN0Pack() const { return n0Pack; }
+  uint32_t getK0Pack() const { return k0Pack; }
+  uint32_t getM1Pack() const { return m1Pack; }
+  uint32_t getN1Pack() const { return n1Pack; }
+  uint32_t getK1Pack() const { return k1Pack; }
+
+ private:
+  ParameterSetting(uint32_t M0, uint32_t N0, uint32_t K0, uint32_t M1,
+                   uint32_t N1, uint32_t K1, uint32_t m0Pack, uint32_t n0Pack,
+                   uint32_t k0Pack, uint32_t m1Pack, uint32_t n1Pack,
+                   uint32_t k1Pack)
+      : M0(M0),
+        N0(N0),
+        K0(K0),
+        M1(M1),
+        N1(N1),
+        K1(K1),
+        m0Pack(m0Pack),
+        n0Pack(n0Pack),
+        k0Pack(k0Pack),
+        m1Pack(m1Pack),
+        n1Pack(n1Pack),
+        k1Pack(k1Pack) {}
+
+  uint32_t M0;
+  uint32_t N0;
+  uint32_t K0;
+  uint32_t M1;
+  uint32_t N1;
+  uint32_t K1;
+  uint32_t m0Pack;
+  uint32_t n0Pack;
+  uint32_t k0Pack;
+  uint32_t m1Pack;
+  uint32_t n1Pack;
+  uint32_t k1Pack;
+};
+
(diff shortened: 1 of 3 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -239,7 +152,7 @@ FailureOr<PadPackTiling> PadPackTiling::create(linalg::LinalgOp linalgOp) {
   const uint64_t K = lhsShape[1];
 
   FailureOr<unsigned> maybeScaleFactor =
-      getTilingScaleFactor(lhsType.getElementType());
+      getTilingScaleFactor(initType.getElementType());
```
newling, line 155: I guess 'scale factor' isn't really the right abstraction for this. For an i8 -> i32 matmul, I don't think it makes sense to use the (small) scale factor for i32 to determine how many i8 elements can fit in AIE core memory or memtile memory. We should revisit this later..

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -264,32 +177,139 @@ FailureOr<PadPackTiling> PadPackTiling::create(linalg::LinalgOp linalgOp) {
...
+    uint32_t K1 = findLargestFactor(K / k1Pack, 2 * scaleFactor);
+    uint32_t K0 = findLargestFactor(K, 256, k1Pack * K1);
+
+    // In pad-pack pipeline, there is only one level of packing, set pack
+    // parameters for level 0 as 0.
+    uint32_t m0Pack = 0;
+    uint32_t n0Pack = 0;
+    uint32_t k0Pack = 0;
+
+    // Tiling at level 0 (shared memory) should be no smaller than tiling at
```
newling, line 226: I guess this assertion is not true on the pack-peel pipeline now, can you explain why?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -264,32 +177,139 @@ FailureOr<PadPackTiling> PadPackTiling::create(linalg::LinalgOp linalgOp) {
...
 
-  auto maxL1Size = 16 * scaleFactor;
-  uint32_t M1 = findLargestFactor(M, maxL1Size, mPack);
-  uint32_t N1 = findLargestFactor(N, maxL1Size, nPack);
-
-  auto maxL0Size = 64 * scaleFactor;
-  uint32_t M0 = findLargestFactor(M, maxL0Size, M1);
-  uint32_t N0 = findLargestFactor(N, maxL0Size, N1);
+  if (isPackPeel) {
+    // Assume working on a 2x2 AIE array, so the ideal level 1 tile sizes should
```
newling, line 181: Please add a comment in the code about why we assume we're currently assuming a 2x2 array or an issue link

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -73,158 +73,71 @@ FailureOr<std::array<uint32_t, 3>> getPackedSize(linalg::LinalgOp linalgOp,
...
-          entryPointFn, linalgOp, tileSizes,
-          IREE::Codegen::DispatchLoweringPassPipeline::Custom))) {
-    return failure();
-  }
-  return success();
-}
-
-namespace {
 
 // Container class for the tiling at level 0 (the AIE shared memory) and level 1
```
newling, line 77: This comment should be updated if it's no longer just for the pad-pack approach. Would it be possible to say, in terms of  the 15 parameters, how many elements of A, B, and C will be resident in the AIE core and memtile memories for the different pipelines? 

I think for pad-pack for A it is:

memtile: M0 * K0 elements
AIE core memory : M1 * K1 elements. 

is that correct? i.e. the packing parameters don't enter the equations. Actually in my mind that was the definition of M0, M1, K0, K1...

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -73,158 +73,71 @@ FailureOr<std::array<uint32_t, 3>> getPackedSize(linalg::LinalgOp linalgOp,
...
+  uint32_t m0Pack;
+  uint32_t n0Pack;
+  uint32_t k0Pack;
+  uint32_t m1Pack;
+  uint32_t n1Pack;
+  uint32_t k1Pack;
 };
 
-FailureOr<PadPackTiling> PadPackTiling::create(linalg::LinalgOp linalgOp) {
+FailureOr<ParameterSetting> ParameterSetting::create(linalg::LinalgOp linalgOp,
```
newling, line 139: Another design (maybe neater, maybe not...) would be to have 2 create methods

createForPadPack(...)
createForPackPeel(...)

newling (commented):
My comments are mostly requests for explanations about the peel-pack pipeline which I'm not too familiar with. It seems like it does packing from DDR->memtile (L3 -> L2) which the pack-pad pipeline does not?

yzhang93:
> My comments are mostly requests for explanations about the peel-pack pipeline which I'm not too familiar with. It seems like it does packing from DDR->memtile (L3 -> L2) which the pack-pad pipeline does not?

Yes, you are right. Pad-pack pipeline has `linalg.copy` to move data from L3->L2, and pack-peel pipeline has pack operations for that. So it has two levels of packing.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: I'm not sure which factor is correct to use. My thinking is we need to initialize and do computation on the output buffer in the AIE core, so we need to take into account the larger data type.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: Currently, for pack-peel pipeline AIR only supports 2x2 array. @erwei-xilinx will extend this to 4x4 array next.

newling (approved):
Looks good to me.

newling:
The numerical error is almost certainly not related to this PR, and probably intermittent. We haven't seen this in CI before though, so this is worrying.

yzhang93:
> The numerical error is almost certainly not related to this PR, and probably intermittent. We haven't seen this in CI before though, so this is worrying.

Yeah, let me bump mlir-air in a separate PR and see if some new commits in AIR breaks the ci.

PR #337, opened 2024-05-10
Numerical testing (aie vs llvm-cpu) of mlir funcs

newling:
Tests of arbitrary func.funs: perform e2e run with (1) AIE and (2) llvm-cpu backends. Compare results. 

Developer work-flow:

1) Add your .mlir file to build_tools/ci/cpu_comparison/test_files (see files there for expected format)
2) run ./run_test.sh (same workflow as ./run_matmul_script.sh which we currently have).

Only 1 is needed to get a test to run in CI. 

At some point we can try and deprecate this approach, or integrate into IREE.


MaheshRavishankar (commented):
Skimming this, this looks fine for me...

At @erieaton-amd  if this works, then I am happy to take this.

yzhang93:
I tried this locally for matmul-elementwise test, and it worked! Thanks for your effort. I haven't looked into the detail, but I can do so if @MaheshRavishankar and others think this is able to be used for all ci tests.

newling:
> I tried this locally for matmul-elementwise test, and it worked! Thanks for your effort. 

That's great :-) 

I haven't looked into the detail, but I can do so if @MaheshRavishankar and others think this is able to be used for all ci tests.

Not sure what you mean by _all_ ci tests.

yzhang93:
> > I haven't looked into the detail, but I can do so if @MaheshRavishankar and others think this is able to be used for all ci tests.
> 
> Not sure what you mean by _all_ ci tests.

Are we going to use this for matmul tests as well, i.e., use `run_test.sh` to replace `run_matmul_test.sh`?

newling:
> > > I haven't looked into the detail, but I can do so if @MaheshRavishankar and others think this is able to be used for all ci tests.
> > 
> > 
> > Not sure what you mean by _all_ ci tests.
> 
> Are we going to use this for matmul tests as well, i.e., use `run_test.sh` to replace `run_matmul_test.sh`?

I see. One nice thing about run_matmul_test.sh is that the reference calculation has the option to not compute (and compare) every single output value: https://github.com/nod-ai/iree-amd-aie/blob/d127b391a5346dbb85964a3dd6a272efd6843b13/build_tools/ci/run_matmul_test.sh#L216
this is used for very large matmuls, and significantly reduces CI time. Having such an option to work with this pipeline is not really possible. And so that's a big advantage to keeping run_matmul_test.sh

build_tools/ci/cpu_comparison/input_generator.py
```diff
@@ -0,0 +1,151 @@
...
+def float32_to_bfloat16(float32_value):
+
+    # Convert to a 32-bit integer representation
+    int32_repr = float32_value.view(np.int32)
+
+    # Truncate the lower 16 bits
+    bf16_int_repr = (int32_repr >> 16) & 0xFFFF
+
+    # Convert to numpy uint16 for consistency
+    return np.uint16(bf16_int_repr)
```
yzhang93, line 46: I don't know where to get the reference. Is this the standard way to convert f32 to bf16? Also it may be better to rename this function as `convert_f32_to_bf16`.

build_tools/ci/cpu_comparison/input_generator.py
```diff
@@ -0,0 +1,151 @@
...
+    else:
+        raise ValueError("Invalid or unsupported element type: " + element_type)
+
+
+def write_input(bin_filename, num_elements, element_type, input_number):
+
+    # Random integer values in range [lower_bound, upper_bound)
+    # will be generated for the input data.
+    lower_bound = 0
+    upper_bound = 5
```
yzhang93, line 86: This seems to be a small range for integer type.

build_tools/ci/cpu_comparison/input_generator.py
```diff
@@ -0,0 +1,151 @@
...
+
+    input_args = []
+
+    with open(filename, "r") as file:
+        input_number = 1
+
+        # Looking for lines of the form '# input 3x40xf32'
+        for line in file:
+            tokens = line.split()
+            if len(tokens) > 2 and tokens[0] == "//":
```
yzhang93, line 120: You should probably add some warning/error message if this condition doesn't hold.

build_tools/ci/cpu_comparison/output_comparer.py
```diff
@@ -0,0 +1,77 @@
...
+def compare(npy_cpu1, npy_aie1):
+
+    if npy_cpu1.shape != npy_aie1.shape:
+        raise ValueError(
+            "The two outputs have different shapes: {} and {}".format(
+                npy_cpu1.shape, npy_aie1.shape
+            )
+        )
+
+    are_close = np.allclose(npy_cpu1, npy_aie1, rtol=1e-3, atol=1e-3)
```
yzhang93, line 14: rtol and atol tolerance values seem to be large for float numbers? (I'm not sure what values are appropriate).

build_tools/ci/cpu_comparison/run_test.sh
```diff
@@ -0,0 +1,311 @@
...
+      --iree-amdaie-use-pipeline=${pipeline} \
+      --iree-amd-aie-peano-install-dir=${peano_install_path} \
+      --iree-amd-aie-mlir-aie-install-dir=${mlir_aie_install_path} \
+      --iree-amd-aie-vitis-install-dir=${vitis_path} \
+      --iree-hal-dump-executable-files-to=$PWD \
+      --iree-amd-aie-show-invoked-commands \
+      --iree-amd-aie-enable-chess \
+      --iree-amdaie-enable-ukernels=all \
+      --mlir-print-ir-after-all \
+      --mlir-print-ir-module-scope \
```
yzhang93, line 229: I don't think we need these flags.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
```diff
@@ -232,8 +232,8 @@ void AMDAIELowerToUKernelsPass::runOnOperation() {
                                                             pathToUkernels);
   patterns.insert<LowerToUKernelPattern<linalg::MatmulOp>>(context, allTargets,
                                                            pathToUkernels);
-  patterns.insert<LowerToUKernelPattern<linalg::FillOp>>(context, allTargets,
-                                                         pathToUkernels);
+//   patterns.insert<LowerToUKernelPattern<linalg::FillOp>>(context, allTargets,
+//                                                          pathToUkernels);
```
yzhang93, line 236: Is this something you don't want to check in?

yzhang93 (changes requested):

build_tools/ci/cpu_comparison/input_generator.py
newling: I've changed it to 10.

build_tools/ci/cpu_comparison/output_comparer.py
newling: It's context specific, there are cases where 1e-3 is probably too low. I'll try and make it a command line arg

yzhang93 (approved):
Thanks! I still see some parts need to be improved. However, I think we can check in the PR at this time, and optimize the codes progressively.

newling:
>  I still see some parts need to be improved.

Ok, let me know what they are (I think I addressed all your comments?) and I'm happy to address in a follow-up patch

yzhang93:
> > I still see some parts need to be improved.
> 
> Ok, let me know what they are (I think I addressed all your comments?) and I'm happy to address in a follow-up patch

My thinking is instead of adding individual tests with different shapes and data types to `build_tools/ci/cpu_comparison/test_files`, we should have something more general like a template mlir which takes input shapes and data types as input. I can also help with this improvement in a following patch.

PR #339, opened 2024-05-10
Revert "Modify the config settings for pack-peel pipeline"

yzhang93:
Reverts nod-ai/iree-amd-aie#335, since it seems to have caused intermittent numerical issue.


nirvedhmeshram:
looking at the post submit CI history I am not convinced the PR had anything to do with it, this is an intermittent issue happening in quite a few unrelated PR. I think we need to update our driver/kernel on the CI machine. Perhaps something Sai can help with once he is back from vacation.

yzhang93:
Yes, I think the issue isn't caused by this particular PR, but I can reproduce the same problem on a local system. I'm not going to revert this PR, but will keep looking into the cause.

PR #341, opened 2024-05-10
Add option to run matmul test multiple times (but compile only once)

newling:


yzhang93:
It is a good idea to be able to run matmul multiple times in local test, but I don't think it's necessary to add this to ci. The main reason is it's hard to determine the `n_repeats` value. Some intermittent error happens less frequently, and it would take more time repeating a larger number of times. I would suggest keep this for local test/debugging.

newling:
I have investigated intermittent numerical errors on three separate occassions, each time I need to add this flag back for my local testing. I would like it to be part of the script, so that next time I don't need to spend time adding it locally again. We can have 1 small test with n_repeats>1 just to show that it works, it won't effect CI time. Is your main concern CI time?

build_tools/ci/run_matmul_test.sh
```diff
@@ -206,19 +206,22 @@ function run_matmul_test() {
   # The default is to not expect a compilation failure.
   local expect_compile_failure="0"
 
-  # The default is to compile and run the numerical test.
-  local compile_only="0"
```
yzhang93, line 210: I think it's good to keep this flag

build_tools/ci/run_matmul_test.sh
```diff
@@ -206,19 +206,22 @@ function run_matmul_test() {
...
   local do_transpose_rhs="0"
 
   # The maximum number of elements to check for correctness.
   # See https://github.com/iree-org/iree/blob/tools/testing/e2e/test_utils.c#L40-L47
   local max_elements_to_check="20000"
 
+  # After compilation, the test with be run 'n_repeats' times. This option (when
+  # set greater than 1) is useful for shapes which might be 'flakey' and fail
+  # intermittently. It is also useful if a test is know to fail at runtime but
+  # should still be checked to compile (set n_repeats=0 in this case).
```
yzhang93, line 218: Most of the time the intermittent failure is caused by compilation problem not runtime. It's not helpful to just compile once and run multiple times.

yzhang93 (changes requested):

build_tools/ci/run_matmul_test.sh
newling: Ok, but the reason this flag is removed in this PR is that it is equivalent to n_repeats=0 (and so redundant)

yzhang93: Yeah, I figured that, but it seems to be more readable to use flag `compile_only`. Maybe to change `n_repeats` to `num_repeat_runs`?

yzhang93 (approved):

PR #342, opened 2024-05-10
[Delegate] Add transform dialect scripts for matmul+elementwise

nirvedhmeshram:
The matmul test shows parity with the PDL test and runs with correct numbers e2e in local testing.
The matmul+elemetwise test generates the expected dispatch and compiles. The numerics of these are not tested as there is no delegate (cpu or AIE) for this. 
The tests use iree-compile because there is an unrelated crash in using iree-opt.

### DIFF
--- a/experimental/delegate/CMakeLists.txt
+++ b/experimental/delegate/CMakeLists.txt
@@ -66,11 +66,14 @@ iree_lit_test_suite(
     "opt.mlir"
     "large-matmul.mlir"
     "large-matmul-f32.mlir"
+    "matmul-elementwise-f32.mlir"
   DATA
     "linalg.pdl.mlir"
     "opt.pdl.mlir"
     "large-matmul.pdl.mlir"
     "large-matmul-f32.pdl.mlir"
+    "mlp_spec_matmul.mlir"
+    "mlp_spec_matmul_elementwise.mlir"
   TOOLS
     FileCheck
     iree-compile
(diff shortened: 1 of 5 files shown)


daveliddell (approved):
LGTM!

MaheshRavishankar:
Whats the crash with `iree-opt`?

nirvedhmeshram:
> Whats the crash with `iree-opt`?

Just needs to add a bunch of dependent dialects. AFAICS custom dispatch is only demonstarted e2e in upstream and there is no example like below in upstream

```
iree-opt --pass-pipeline="builtin.module((iree-preprocessing-transform-interpreter{transform-spec-path=/experimental/delegate/mlp_spec_matmul_elementwise.mlir}))" /proj/xsjhdstaff4/nmeshram/iree-amd-aie/experimental/delegate/matmul-elementwise-f32.mlir
```
I just tried adding the dependent dialects and running the above command and it didnt crash but to my surprise no matching happened?
I dont see any failed matches in the debug print either.

nirvedhmeshram:
Here is a draft pull request with the dependent dialect stuff that should have solved the issue but didnt
https://github.com/iree-org/iree/pull/17354

nirvedhmeshram:
Landing this as the iree-opt issues are unrelated to the scripts being added here and we have the iree-compile way of testing them in CI for now.

PR #344, opened 2024-05-13
[AMD-AIE] Lower zero filling linalg.fill to ukernel

Abhishek-Varma:
-- This commit updates `--iree-amdaie-lower-to-ukernel` pass by
   adding zero filling ukernel matcher for a linalg.fill op.

Signed-off-by: Abhishek Varma <abhvarma@amd.com>

PR #345, opened 2024-05-13
Pipeline not used in ukernel matching pattern 

newling:
Not sure why the pipeline was being passed to the ukernel patterns, maybe some historical artifact?


yzhang93 (approved):

yzhang93:
> Not sure why the pipeline was being passed to the ukernel patterns, maybe some historical artifact?

Yes, it's the historical artifacts. I should have cleaned up this when I deprecated the pipelines. Thanks for cleaning up the codes.

newling:
Closing this as @Abhishek-Varma has these changes in https://github.com/nod-ai/iree-amd-aie/pull/344

PR #348, opened 2024-05-14
Add pass to convert logical objectfifo produce/consume to acquire/release

jtuyls:
Adds a pass to convert logical objectfifo consume/produce operations to acquire/release semaphore operations. These acquire/release operations can then later be converted into objectfifo acquire/release operations in the AIE dialect.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConsumeProduceToAcquireRelease.cpp
```diff
@@ -0,0 +1,139 @@
...
+  return getParentOpInBlock(block, parentOp);
+}
+
+/// Walk all consume/produce operations within the core operations and insert
+/// semaphore operations.
+template <typename OpTy>
+LogicalResult consumeProduceToAcquireRelease(Operation *parentOp) {
+  using IteratorType = std::conditional_t<
+      std::is_same<OpTy, AMDAIE::LogicalObjectFifoConsume>::value,
+      ForwardIterator, ReverseIterator>;
```
yzhang93, line 32: Why do you need both ForwardIterator and ReverseIterator?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConsumeProduceToAcquireRelease.cpp
```diff
@@ -0,0 +1,139 @@
...
+      // it was encountered before. Add it to the map and advance if not.
+      Operation *dmaOp = op.getDma().getDefiningOp();
+      if (!mapper.contains(dmaOp)) {
+        mapper.map(dmaOp, op.getOperation());
+        return WalkResult::advance();
+      }
+
+      // Find the new consume/produce operation's parent operation within the
+      // same block as the previous operation of the same type and operating on
+      // the same DMA. Use theisparent operation in the same block to set the
```
yzhang93, line 58: Typo: Use this parent

yzhang93 (approved):
Overall LGTM. Thanks!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConsumeProduceToAcquireRelease.cpp
jtuyls: For the produce operation, a `release` operation needs to be inserted at the place where the produce was as the data will be 'produced' and then the data object is released. And an acquire operation needs to be inserted earlier in the block to acquire the data object resource so we can produce data into it. To accomplish this, the `ReverseIterator` is more convenient. For the consume operator it's the other way around as an acquire operation needs to be inserted at the location of the consume operation so the data object can be read and the release needs to be inserted after the data is not needed anymore.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConsumeProduceToAcquireRelease.cpp
jtuyls: Done

PR #351, opened 2024-05-15
[AMD-AIE] Bump IREE 6/2/2024 + Update fuse-consumer pass to use upstream utility

Abhishek-Varma:
This PR consists of two commits :-

**Commit 1**
-- IREE bump to 6/2/2024.

**Commit 2**
-- This commit updates `--iree-amdaie-fuse-consumer-into-loop` pass to use the upstream utility instead.

Signed-off-by: Abhishek Varma <abhvarma@amd.com>

PR #352, opened 2024-05-15
Remove redundant data copy of elementwise op

yzhang93:
- Add [upstream support](https://github.com/shark-infra/llvm-project/blob/f2d824e381755993b6e8c32f4de89192189ca9a7/mlir/lib/Dialect/Linalg/Transforms/ElementwiseOpFusion.cpp#L1864) locally to remove output dependency, since the pattern is not exposed and cannot be called directly. This pattern is added in `AMDAIEPropagateDataLayoutPass`, so that whenever we do data layout propagation it will get the desirable output for matmul-elementwise fusion. @MaheshRavishankar please let me know if it is reasonable to add it here, otherwise I'm open to move it into a separate pass or add the pattern to the cleanup pass.
- Currently, tests without vectorization compile through, but crash at `AIRToAIESchedulingUtilsPass` with vectorization. @erwei-xilinx

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPropagateDataLayout.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPropagateDataLayout.cpp
@@ -6,6 +6,7 @@
 
 #include "iree-amd-aie/Transforms/Passes.h"
 #include "mlir/Dialect/Linalg/Transforms/Transforms.h"
+#include "mlir/Dialect/SparseTensor/IR/SparseTensor.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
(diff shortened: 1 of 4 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPropagateDataLayout.cpp
```diff
@@ -15,6 +16,52 @@ namespace mlir::iree_compiler::AMDAIE {
...
+    rewriter.startOpModification(op);
+    bool modifiedOutput = false;
+    Location loc = op.getLoc();
+    for (OpOperand &opOperand : op.getDpsInitsMutable()) {
+      if (!op.payloadUsesValueFromOperand(&opOperand)) {
+        Value operandVal = opOperand.get();
+        auto operandType = dyn_cast<RankedTensorType>(operandVal.getType());
+        if (!operandType) continue;
+
+        // If outs is sparse, leave it to the sparsifier.
```
MaheshRavishankar, line 37: Drop the sparse stuff.. unnecessary for now.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPropagateDataLayout.cpp
```diff
@@ -15,6 +16,52 @@ namespace mlir::iree_compiler::AMDAIE {
...
+    if (!modifiedOutput) {
+      rewriter.cancelOpModification(op);
+      return failure();
+    }
+    rewriter.finalizeOpModification(op);
+    return success();
+  }
+};
+
+void populateElementwiseOpsFusionPatterns(RewritePatternSet &patterns) {
```
MaheshRavishankar, line 61: Rename this since we are not doing fusion. You dont need the populate method. Just add the pattern directly below.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -76,9 +77,22 @@ void appendVectorizationToPipeline(OpPassManager &funcPassManager) {
 static FailureOr<Value> aieComprehensiveBufferizeAllocationFn(
     OpBuilder &builder, Location loc, MemRefType memRefType,
     ValueRange dynamicSizes, unsigned alignment) {
-  return builder
-      .create<memref::AllocaOp>(loc, memRefType, dynamicSizes,
-                                builder.getI64IntegerAttr(alignment))
+  int64_t numDims = memRefType.getShape().size();
+  AMDAIEMemSpace memSpace = AMDAIEMemSpace::Shared;
+  if (clUsePipeline == AIEPassPipeline::PadPackPipeline && numDims == 4) {
```
MaheshRavishankar, line 82: Why not use local always?

MaheshRavishankar (changes requested):

yzhang93:
> * Currently, tests without vectorization compile through, but crash at `AIRToAIESchedulingUtilsPass` with vectorization. @erwei-xilinx

This issue is fixed by fusing the unpack op into the last forall loop based on this https://github.com/nod-ai/iree-amd-aie/pull/351. Since that PR hasn't been merged, I disable the vectorization in the elementwise test to avoid failure in ci.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
MaheshRavishankar: If it doesnt its an error. For now just set local always.

MaheshRavishankar (approved):

PR #354, opened 2024-05-16
[Delegate] Support for 16k matmul

daveliddell:
Several enhancements in support of a 16384x16384x512 matmul with bf16 inputs and f32 outputs:

- New PDL for 16k matmul
- timers added for NPU setup and kernel execution
- cleanup of delegate tracing for better code readability
- re-indentation of some functions for consistency
- kernel (on Azure) generated via AIE direct codegen

### DIFF
--- a/experimental/delegate/mlp_aie_bf16_plugin.cpp
+++ b/experimental/delegate/mlp_aie_bf16_plugin.cpp
@@ -14,6 +14,7 @@
 #include <stdio.h>
 
 #include <algorithm>
+#include <chrono>
 #include <cstddef>
 #include <cstdint>
 #include <cstring>
@@ -38,6 +39,7 @@
 #define REF_MATMUL_DELEGATE_KERNEL 1
 #define OPT_DELEGATE_KERNEL 2
 #define LARGE_MATMUL_DELEGATE_KERNEL 3
+#define MATMUL_16K_DELEGATE_KERNEL 4
 
 //#############################################################################
 //
@@ -47,7 +49,8 @@
 // Uncomment the kernel to use
 // #define DELEGATE_KERNEL_TO_USE REF_MATMUL_DELEGATE_KERNEL
 // #define DELEGATE_KERNEL_TO_USE OPT_DELEGATE_KERNEL
-#define DELEGATE_KERNEL_TO_USE LARGE_MATMUL_DELEGATE_KERNEL
+// #define DELEGATE_KERNEL_TO_USE LARGE_MATMUL_DELEGATE_KERNEL
+#define DELEGATE_KERNEL_TO_USE MATMUL_16K_DELEGATE_KERNEL
 
 // Turn this on to use XRT buffers which are separate from HAL buffers.
 // There is a performance cost to copying between HAL and XRT buffer, but it
@@ -63,7 +66,7 @@
 // #define USE_BF16_CPU_ACCUMULATOR 1
 
 // Turn this on to print debug messages throughout the delegate run
-// #define TRACE_DELEGATE 1
+// #define ENABLE_TRACE_DELEGATE 1
 
 // Turn this on to dump matmul operand and result tensor values
 // #define DEBUG_VALUES 1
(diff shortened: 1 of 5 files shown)


daveliddell:
Looks like the pack_peel_matmul test failure was transient?

experimental/delegate/mlp_aie_bf16_plugin.cpp
```diff
@@ -47,7 +49,8 @@
 // Uncomment the kernel to use
 // #define DELEGATE_KERNEL_TO_USE REF_MATMUL_DELEGATE_KERNEL
 // #define DELEGATE_KERNEL_TO_USE OPT_DELEGATE_KERNEL
-#define DELEGATE_KERNEL_TO_USE LARGE_MATMUL_DELEGATE_KERNEL
+// #define DELEGATE_KERNEL_TO_USE LARGE_MATMUL_DELEGATE_KERNEL
```
newling, line 52: same as line above, probably forgot to remove?

experimental/delegate/mlp_aie_bf16_plugin.cpp
```diff
@@ -84,15 +87,65 @@
   #define CONVERSION_DEBUG(turnOn_) ;
 #endif
 
-// Fake bfloat16 type (assuming no C++ 23)
-using bfloat16_t = std::uint16_t;
+#ifdef ENABLE_TRACE_DELEGATE
+#define TRACE_DELEGATE(str_)                                      \
+  do {                                                            \
```
newling, line 92: What's the difference between `do {foo();} while (0)` and `foo();` ?

experimental/delegate/mlp_aie_bf16_plugin.cpp
```diff
@@ -47,7 +49,8 @@
 // Uncomment the kernel to use
 // #define DELEGATE_KERNEL_TO_USE REF_MATMUL_DELEGATE_KERNEL
 // #define DELEGATE_KERNEL_TO_USE OPT_DELEGATE_KERNEL
-#define DELEGATE_KERNEL_TO_USE LARGE_MATMUL_DELEGATE_KERNEL
+// #define DELEGATE_KERNEL_TO_USE LARGE_MATMUL_DELEGATE_KERNEL
+#define DELEGATE_KERNEL_TO_USE MATMUL_16K_DELEGATE_KERNEL
```
newling, line 53: What is the advantage of using macros for compile time control flow, instead of the usual c++ `if (matmul_16k) { ... }` control flow? 

The advantage of the usual c++ control flow is that CI checks everything compiles. 

I guess ideally we would classes for this, although it is in the 'experimental' directory so not necessary.

newling (approved):
LGTM

newling:
> Looks like the pack_peel_matmul test failure was transient?

I suspect we just need to bump mlir-air, but will check

newling:
> > Looks like the pack_peel_matmul test failure was transient?
> 
> I suspect we just need to bump mlir-air, but will check



> > Looks like the pack_peel_matmul test failure was transient?
> 
> I suspect we just need to bump mlir-air, but will check

It's a bit worrying actually because the commit which should have fixed the undefined behaviour (in mlir-air) was  67acba7614801508bcfde0aa19012a58853a537a which is behind the mlir that iree-amd-aie uses 85a99b3154f4c6559aa3657cdb72327ddf249c1f . So it seems like there is UB unaccounted for. 

(Just a heads up @erwei-xilinx)

experimental/delegate/mlp_aie_bf16_plugin.cpp
daveliddell: I know they all look the same, but one is OPT and the other is LARGE
:-D

experimental/delegate/mlp_aie_bf16_plugin.cpp
daveliddell: There is currently a lot of compile-time template-based processing going on to ensure type compatibility, automatic value conversion, and copy optimization, not to mention the types in the blob of data transferred from HAL.

When I first started the delegate, it wasn't clear which direction the design was going, or how much we really want to do before heterogeneous device support makes this all obsolete.  Now that a pattern is starting to emerge, maybe we want to enhance the delegate with the ability to select its behavior at run time, via a parameter passed from HAL to select from the catalog of kernel options built in to the delegate.  I could probably preserve the type safety via object factories.  However, we don't really go backward to older kernels, except rarely for debugging, so I'm not sure if it's worth the effort.

It would be even better if I didn't have to keep modifying the code every time we added a new kernel.  That would require reading metadata from some external source, which is doable also.  Maybe just a separate header would be good.

All effort on the delegate so far has led up to this support for the 16k matmul.  I wonder what we'll want to do with it next.  I would say that if there is a next whatever, that might be the time to make delegates easier to work with.

experimental/delegate/mlp_aie_bf16_plugin.cpp
daveliddell: That's an old standard trick to make a macro behave like a single statement.  Probably in this case, it doesn't matter since the body of the macro is a single statement already, but it was commonly used in general for cases like this:

```
#define BAD_MACRO(arg) \
    line1(arg); \
    line2

#define GOOD_MACRO(arg) \
    do { \
        line1(arg); \
        line2; \
    } while (0)

void func() {
    if (condition)
        BAD_MACRO(arg);
    else
        GOOD_MACRO(arg);
}
```

BAD_MACRO would cause problems, whereas GOOD_MACRO wouldn't.  Someone who wraps every block in curly braces might say that's what you get for not wrapping every block in curly braces!  :-D

I'm having second thoughts about it, though.  I think I'll go back and simplify it.

experimental/delegate/mlp_aie_bf16_plugin.cpp
newling: I mean...

#define DELEGATE_KERNEL_TO_USE LARGE_MATMUL_DELEGATE_KERNEL
// #define DELEGATE_KERNEL_TO_USE LARGE_MATMUL_DELEGATE_KERNEL

experimental/delegate/mlp_aie_bf16_plugin.cpp
newling: Yikes. Macros are... quite something!

experimental/delegate/mlp_aie_bf16_plugin.cpp
daveliddell: Actually, maybe I misinterpreted what you were saying.  I commented out the LARGE_MATMUL in favor of the next line for the 16k matmul.  The diff view is confusing, so I had to look at the original source to make sure I did it right!

experimental/delegate/mlp_aie_bf16_plugin.cpp
daveliddell: Simplified.

experimental/delegate/mlp_aie_bf16_plugin.cpp
newling: That makes sense to me. Heterogenous support can't come soon enough.

experimental/delegate/mlp_aie_bf16_plugin.cpp
daveliddell: Ah, sorry, our comments crossed paths (didn't see your latest).  So this is what is actually says, as seen from a freshly updated workspace:

```
// Uncomment the kernel to use
// #define DELEGATE_KERNEL_TO_USE REF_MATMUL_DELEGATE_KERNEL
// #define DELEGATE_KERNEL_TO_USE OPT_DELEGATE_KERNEL
// #define DELEGATE_KERNEL_TO_USE LARGE_MATMUL_DELEGATE_KERNEL
#define DELEGATE_KERNEL_TO_USE MATMUL_16K_DELEGATE_KERNEL
```

So it's just an GH diff optical illusion :-D

PR #358, opened 2024-05-20
Fix for undefined behavior

newling:
The string

```
 std::string newVitis = "VITIS=" + options.vitisInstallDir;
```
 
 was being created, then pushed back into a SmallVector<StringRef>, and then deallocated, which meant that the string "VITIS=/my/path/to/vitis" was deallocated before being used, which meant it was a sequence of random chars, which meant chess couldn't be found... 
 
 This PR changes SmallVector<StringRef> to SmallVector<std::string> so that this doesn't happen. Other strings adjusted accordingly


nirvedhmeshram (approved):
Thanks!

PR #359, opened 2024-05-20
CI: Use XRT that's built, not one in /opt

newling:


newling:
The build process installs to /opt/xilinx already

PR #364, opened 2024-05-21
Fix parallel loop mapping attributes

yzhang93:


Abhishek-Varma (approved):
Thanks for taking this up! LGTM.

CC: @jtuyls

PR #367, opened 2024-05-22
Add peeled matmul to objectfifo sample

jtuyls:
Example of peeled matmul lowering to AIE objectfifo. This can be temporary until the pipeline is set up and an e2e example can be created. In the meantime, this will ensure that the initial example stays functional.

### DIFF
--- a/tests/samples/matmul_peeled_objectfifo.mlir
+++ b/tests/samples/matmul_peeled_objectfifo.mlir
@@ -1,0 +1,116 @@
+// RUN: iree-opt --pass-pipeline="builtin.module(fold-memref-alias-ops,iree-amdaie-pack-to-dma,air-copy-to-dma,iree-amdaie-air-dma-to-amdaie-dma,iree-amdaie-insert-aie-workgroup,iree-amdaie-fuse-logicalobjectfifo-into-workgroup,cse,iree-amdaie-unroll-and-distribute-workgroup,cse,iree-amdaie-dma-to-circular-dma,func.func(iree-amdaie-create-aie-workgroup),cse,iree-amdaie-canonicalize-doubly-strided-op,iree-amdaie-consume-produce-to-acquire-release,cse,canonicalize,iree-amdaie-controlcode-loop-unroll,cse,canonicalize,canonicalize,iree-amdaie-lower-to-aie,canonicalize)" --split-input-file %s | FileCheck %s
+
+// CHECK:       aie.device(npu1)
+// CHECK-DAG:   %[[TILE_0_2:.+]] = aie.tile(0, 2)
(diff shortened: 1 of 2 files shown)


tests/samples/matmul_peeled_objectfifo.mlir
```diff
@@ -0,0 +1,116 @@
...
+    %1 = hal.interface.binding.subspan set(0) binding(0) type(storage_buffer) alignment(64) offset(%c0) flags(ReadOnly) : memref<32x1024xi32>
+    memref.assume_alignment %1, 64 : memref<32x1024xi32>
+    %2 = hal.interface.binding.subspan set(0) binding(2) type(storage_buffer) alignment(64) offset(%c0) : memref<32x64xi32>
+    memref.assume_alignment %2, 64 : memref<32x64xi32>
+    %alloc = memref.alloc() : memref<4x8x8x8xi32, 2>
+    %alloc_0 = memref.alloc() : memref<8x8x4x8xi32, 2>
+    %alloc_1 = memref.alloc() : memref<4x8x4x8xi32, 2>
+    %alloc_2 = memref.alloc() : memref<32x32xi32, 1>
+    %alloc_3 = memref.alloc() : memref<64x32xi32, 1>
+    %alloc_4 = memref.alloc() : memref<32x64xi32, 1>
```
yzhang93, line 41: These require first level to be pad instead of pack. Does objectfifo potentially work with more dimensions?

jtuyls: Yeah, that should be fine. I created this example when the main pipeline was `pad-pack`.

yzhang93 (commented):
It's okay to check in this for now. But we'd like to have a similar test with the IR from pack-peel pipeline. @Abhishek-Varma can you first try to use the same iree-opt flags on top of this IR https://gist.github.com/yzhang93/62f142bc71c7a9c247f98c71b2ccee4c?

yzhang93 (approved):

Abhishek-Varma:
> It's okay to check in this for now. But we'd like to have a similar test with the IR from pack-peel pipeline. @Abhishek-Varma can you first try to use the same iree-opt flags on top of this IR https://gist.github.com/yzhang93/62f142bc71c7a9c247f98c71b2ccee4c?

Hi @yzhang93 - This is an invalid IR - this doesn't take into account the thread ID mapping.

I'll continue looking at the IR journey via the current Pack-Peel pipeline (with the current known missing pieces) and see with the same flags (as well as the [C++ pipeline](https://github.com/nod-ai/iree-amd-aie/pull/368)).

yzhang93:
> Hi @yzhang93 - This is an invalid IR - this doesn't take into account the thread ID mapping.

Yes, this was generated before that fixing PR, you just need to manually modify the corresponding block ID to thread ID. The other parts should be the same.

PR #369, opened 2024-05-22
Fix deprecated cast warnings

jtuyls:
Fix deprecated `.cast<U>` and `.dyn_cast<U>` warnings. Use `mlir::cast<U>()` and `mlir::dyn_cast<U>()` instead.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/Translation/AIESerializer.cpp
+++ b/compiler/plugins/target/AMD-AIE/Translation/AIESerializer.cpp
@@ -55,10 +55,8 @@ static int64_t getLinearizedSize(SmallVector<int64_t> &shape) {
 
 /// Translate the memory space into shared / local.
 static FailureOr<std::string> translateMemorySpace(IntegerAttr memorySpace) {
-  if (isConstantIntValue(memorySpace, 1))
-    return std::string("shared");
-  if (isConstantIntValue(memorySpace, 2))
-    return std::string("local");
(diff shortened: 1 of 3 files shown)


compiler/plugins/target/AMD-AIE/Translation/AIESerializer.cpp
```diff
@@ -55,10 +55,8 @@ static int64_t getLinearizedSize(SmallVector<int64_t> &shape) {
 
 /// Translate the memory space into shared / local.
 static FailureOr<std::string> translateMemorySpace(IntegerAttr memorySpace) {
-  if (isConstantIntValue(memorySpace, 1))
-    return std::string("shared");
-  if (isConstantIntValue(memorySpace, 2))
-    return std::string("local");
+  if (isConstantIntValue(memorySpace, 1)) return std::string("shared");
+  if (isConstantIntValue(memorySpace, 2)) return std::string("local");
```
Abhishek-Varma, line 59: I wonder why the `clang-format` didn't catch this as an error while checking formatting.

Abhishek-Varma (approved):
Thanks for cleaning up!

compiler/plugins/target/AMD-AIE/Translation/AIESerializer.cpp
jtuyls: Actually, where does CI run clang-format?

PR #374, opened 2024-05-23
[AMD-AIE] Fix UnrollAndDistributeWorkgroup pass for non-normalised loop bound

Abhishek-Varma:
-- This commit adds a fix to `--iree-amdaie-unroll-and-distribute-workgroup`
   pass in order to handle non-normalised loop bounds.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
@@ -166,7 +166,13 @@ class AMDAIEUnrollWorkgroupLoops : public OpRewritePattern<scf::ForOp> {
     int64_t lbInt = getConstantIntValue(forOp.getLowerBound()).value();
     int64_t ubInt = getConstantIntValue(forOp.getUpperBound()).value();
     int64_t stepInt = getConstantIntValue(forOp.getStep()).value();
-    if (lbInt != 0 || stepInt != 1) return failure();
+    // TODO(avarma): Either :-
+    //               1. Enforce loop normalisation before this pass.
+    //                  OR
+    //               2. Adapt this better during PR review.
+    if (lbInt != 0) return failure();
+    ubInt = std::ceil(ubInt / stepInt);
+    stepInt = 1;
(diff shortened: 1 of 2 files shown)


Abhishek-Varma:
Relaying message from the WIP PR for adding C++ pipeline : https://github.com/nod-ai/iree-amd-aie/pull/368#issuecomment-2126755096

This helps fix a crash when trying to use the C++ pipeline -> the `amdaie.tile` op is hoisted out to the beginning of the `amdaie.workgroup`, but while trying to do so we're creating an invalid IR since the operand of the `amdaie.tile` is tied to the loop IV. This itself turned out to be a red-herring since the current implementation of the pass adds a strict check for normalised loop bounds, as a result of which it never unrolls the SCF loop and spuriously hoists the `amdaie.tile` op.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp
```diff
@@ -166,7 +166,13 @@ class AMDAIEUnrollWorkgroupLoops : public OpRewritePattern<scf::ForOp> {
     int64_t lbInt = getConstantIntValue(forOp.getLowerBound()).value();
     int64_t ubInt = getConstantIntValue(forOp.getUpperBound()).value();
     int64_t stepInt = getConstantIntValue(forOp.getStep()).value();
-    if (lbInt != 0 || stepInt != 1) return failure();
+    // TODO(avarma): Either :-
+    //               1. Enforce loop normalisation before this pass.
```
jtuyls, line 170: I was looking at the same :s I almost have it so will push shortly and you could have a look

jtuyls: Took a bit more time. Initially, I implemented it for `scf.for`, but actually it's needed on `scf.forall` before the `amdaie.core` operations are inserted. Here is the PR though: https://github.com/nod-ai/iree-amd-aie/pull/375

Abhishek-Varma: Aah okay. No worries. I'll close this PR. Taking a look at your PR now.

PR #375, opened 2024-05-23
Add loop normalization pass and add to workgroup insertion

jtuyls:


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Transforms.h
```diff
@@ -29,6 +29,15 @@ LogicalResult createLogicalObjectFifoLink(
...
                                     affine::AffineApplyOp applyOp);
 
+/// Normalize the loop bounds of the `scf.for` operation to lowerbound == 0 and
+/// step == 1.
+LogicalResult normalizeLoopBounds(RewriterBase &rewriter, scf::ForOp forOp);
+
+/// Normalize the loop bounds of the `scf.forall` operation to lowerbound == 0
+/// and step == 1.
+LogicalResult normalizeLoopBounds(RewriterBase &rewriter,
+                                  scf::ForallOp forallOp);
```
Abhishek-Varma, line 39: Instead maybe have just one API `normalizeLoopBounds(RewriterBase&, LoopLikeInterface)`. Within the API maybe multiplex to the respective static functions ?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertAIEWorkgroup.cpp
```diff
@@ -152,6 +167,10 @@ class AMDAIEInsertAIEWorkgroupPass
 };
 
 void AMDAIEInsertAIEWorkgroupPass::runOnOperation() {
+  // Normalize the loop bounds of `scf.forall` operations within the module.
+  if (failed(normalizeModuleLoopBounds(getOperation()))) {
+    return signalPassFailure();
+  }
```
Abhishek-Varma, line 173: Any reason why we're doing it this way ?

I guess this might not be the right way to do things.

I'd instead have the normalisation pass invoked before `AMDAIEInsertAIEWorkgroupPass`. This also helps during debugging. Else by clubbing such major changes within one pass, we are missing out on lot of ways which we can use during debugging an IR log.

I see a similar thing being done in [AMDAIEUnrollAndDistributeWorkgroupPass here](https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp#L467-L499). I faced the above issue today while debugging the pass hence thought it'd actually make sense to break down passes like these into smaller chunks that does one thing atomically. Just helps getting a good look of IR transformations.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIENormalizeLoopBounds.cpp
```diff
@@ -0,0 +1,191 @@
...
+#define DEBUG_TYPE "iree-amdaie-normalize-loop-bounds"
+
+namespace mlir::iree_compiler::AMDAIE {
+
+/// NOTE: Copy of existing utility struct in llvm-project:
+/// https://github.com/llvm/llvm-project/blob/172759492a162592da7ae9e03888661c108b1be4/mlir/lib/Dialect/SCF/Utils/Utils.cpp#L33C1-L40C15.
+/// Can be replaced with that utility struct once it is exposed.
+///
+// This structure is to pass and return sets of loop parameters without
+// confusing the order.
```
Abhishek-Varma, line 21: I was about to suggest `emitNormalizedLoopBounds` from upstream but unfortunately it's not exposed. Good catch.

In fact, I strongly feel this particular pass should be upstreamed! There's no reason for it to stay as an `AMDAIE` specific pass. While the upstream merge takes time, we can definitely unblock us by getting this pass in - but we should actually consider upstreaming this.

Abhishek-Varma (changes requested):
Hi @jtuyls - thanks for adding this!

Couple of comments/suggestions from my end.

Abhishek-Varma:
> @qedawkins was also looking for this. Maybe we can just move it into iree

Hi @MaheshRavishankar  - I feel this is more suited to be upstreamed into LLVM instead of IREE.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIENormalizeLoopBounds.cpp
```diff
@@ -0,0 +1,191 @@
...
+  // Compute the number of iterations the loop executes: ceildiv(ub - lb, step)
+  // assuming the step is strictly positive.  Update the bounds and the step
+  // of the loop to go from 0 to the number of iterations, if necessary.
+  if (isZeroBased && isStepOne) return {lb, ub, step};
+
+  Value diff =
+      isZeroBased ? ub : rewriter.createOrFold<arith::SubIOp>(loc, ub, lb);
+  Value newUpperBound =
+      isStepOne ? diff
+                : rewriter.createOrFold<arith::CeilDivSIOp>(loc, diff, step);
```
qedawkins, line 52: This can be done with an affine map + `makeComposedFoldedAffineApply` (after changing the helper operands to OpFoldResult)

Edit: I would just replace this helper with the version for `scf.forall` below.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIENormalizeLoopBounds.cpp
```diff
@@ -0,0 +1,191 @@
...
+/// Insert `affine.apply` operations to compute the denormalized index values.
+LogicalResult normalizeLoopBounds(RewriterBase &rewriter,
+                                  scf::ForallOp forallOp) {
+  OpBuilder::InsertionGuard g(rewriter);
+  if (forallOp.isNormalized()) return success();
+
+  SmallVector<OpFoldResult> newLbs;
+  SmallVector<OpFoldResult> newUbs;
+  SmallVector<OpFoldResult> newSteps;
+  for (auto &&[iv, lb, ub, step] : llvm::zip(
```
qedawkins, line 119: The logic here should be reusable for `scf.for` by passing an `ArrayRef<OpFoldResult>` for the loop ranges. There is a `Range` struct provided here https://github.com/llvm/llvm-project/blob/89245b6cf808d4ed27cdea8d1e85fd9f4c6c844b/mlir/include/mlir/Dialect/Utils/StaticValueUtils.h#L33 that you can replace `LoopParams` with.

MaheshRavishankar:
> > @qedawkins was also looking for this. Maybe we can just move it into iree
> 
> Hi @MaheshRavishankar - I feel this is more suited to be upstreamed into LLVM instead of IREE.

Upstream would work too.. Should be as simple as just exposing this method https://github.com/llvm/llvm-project/blob/b17d44558ba4c30a3005089b334f68593d6a9c7c/mlir/lib/Dialect/SCF/Utils/Utils.cpp#L485 

It unfortunately cant use the affine stuff, but does a reasonable job.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIENormalizeLoopBounds.cpp
jtuyls: Yeah, I also feel like it should be upstreamed, but I wanted to get this in to unblock first and have a discussion on how to upstream in the best way.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertAIEWorkgroup.cpp
jtuyls: Initially, I walked the workgroups here and only normalized the loops within the workgroups, but now, with normalizing the whole module, I agree we should just call the pass before. 

In general, is there a way to add dependency passes for a pass that are always run before/after the corresponding pass in a pipeline?

> I see a similar thing being done in [AMDAIEUnrollAndDistributeWorkgroupPass here](https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollAndDistributeWorkgroup.cpp#L467-L499). I faced the above issue today while debugging the pass hence thought it'd actually make sense to break down passes like these into smaller chunks that does one thing atomically. Just helps getting a good look of IR transformations.

Not sure about the `AMDAIEUnrollAndDistributeWorkgroupPass` as I think it's quite atomic, just with a couple of back-to-back steps. For example, splitting between `AMDAIEUnrollWorkgroupLoops` and `assignLocalAieTiles` actually doesn't give correct results if you would run `cse` in between. Maybe, you would get out the `workgroupForallToFor` and `hoistAffineApplyDependingOnFor` out of it, but I think it's quite specific to the unroll that follows?

As for debuggability, we can add debug statements in between the different steps that dump the IR in between?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Transforms.h
jtuyls: I addressed this in the upstream PR, we can discuss further there.

jtuyls:
I created an upstream PR for a longer term solution: https://github.com/llvm/llvm-project/pull/93781, so let's move those discussions to that PR. But I would like to still merge this one as a temporary solution. Could you check again @Abhishek-Varma ?

cc @MaheshRavishankar

PR #377, opened 2024-05-24
[WIP] Prevent hoisting of Affine.apply

Abhishek-Varma:
-- Affine.apply when hoisted out of the scf.for as part of `iree-amdaie-create-aie-workgroup`
   pass creates op dominating issue.

-- This commit/PR will be used to better collaborate/discuss on the issue.

Attaching the culprit IR log from the Matmul dispatch : https://gist.github.com/Abhishek-Varma/9e0f39fc3570521c1ee8d6e0032ef402

1. It contains the IR before `iree-amdaie-create-aie-workgroup`.

2. I tried to prefetch other possible fixes required, so after this stopgap fix the next issue is in `controlcode-loop-unroll` pass and the above Gist contains a short repro as well as the IR we get right before the pass.

   I tried triaging the above and found that the upstream API `forallToForLoop` is unable to replace the SSA values despite trying to `inlineBlockBefore`.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/create_aie_workgroup.mlir
```diff
@@ -381,3 +381,40 @@ func.func @complex_example(%arg0: memref<1x1x8x16xi32, 2>, %arg1: memref<8x16xi3
...
+// CHECK:             amdaie.npu.dma_wait(%[[IPU_DMA]], S2MM)
+func.func @do_not_hoist_affine(%arg0: memref<1x1x8x16xi32, 2>, %arg1: memref<8x16xi32, 1>) {
+  %c0 = arith.constant 0 : index
+  %c1 = arith.constant 1 : index
+  %c8 = arith.constant 8 : index
+  %0 = amdaie.logicalobjectfifo.from_memref %arg0, {} : memref<1x1x8x16xi32, 2> -> !amdaie.logicalobjectfifo<memref<1x1x8x16xi32, 2>>
+  %1 = amdaie.logicalobjectfifo.from_memref %arg1, {} : memref<8x16xi32, 1> -> !amdaie.logicalobjectfifo<memref<8x16xi32, 1>>
+  scf.for %arg2 = %c0 to %c8 step %c1  {  
+    %16 = affine.apply affine_map<(d0) -> (d0 * 32)>(%arg2)
+    %2 = amdaie.dma_cpy_nd(%0[] [] [], %1[0, 0, 0, 0] [1, 1, 8, 16] [128, 16, %16, 1]) : (!amdaie.logicalobjectfifo<memref<1x1x8x16xi32, 2>>, !amdaie.logicalobjectfifo<memref<8x16xi32, 1>>)
```
jtuyls, line 417: To handle a DMA operation depending on an `scf.for` from L2 to L1, I think the operation needs to be hoisted out of the `scf.for` block by moving the loop iteration inside the DMA strided access pattern instead:

```
%apply = affine.apply affine_map<(d0) -> (d0 * 32)>(%arg2)
%2 = amdaie.circular_dma_cpy_nd(%0[] [] [], %1[0, %apply, 0, 0] [1, 1, 8, 8] [128, 128, 16, 1])
```

would become:

```
%2 = amdaie.circular_dma_cpy_nd(%0[] [] [], %1[0, 0, 0, 0] [1, 8, 8, 8] [256, 32, 16, 1])
```

and now it can/should (`circular_dma_cpy_nd`: can, `dma_cpy_nd`: should) be hoisted out of the loop as it doesn't depend on it anymore. Note that for now, DMA operations from L2 to L1 should always be `amdaie.circular_dma_cpy_nd` operations for now because of the way objectfifos are set up within AIE.

I feel like this transformation requires a separate pass which should be run before.

jtuyls: After offline discussion, there are two separate issues that are exposed here:

1) When there is a dependency on a loop induction variable for a `amdaie.dma_cpy_nd` operation, this needs to be moved correctly to the control code as pointed out by @Abhishek-Varma above. Example input IR:

```
func.func @do_not_hoist_affine(%arg0: memref<1x1x8x16xi32, 2>, %arg1: memref<8x16xi32, 1>) {
  %c0 = arith.constant 0 : index
  %c1 = arith.constant 1 : index
  %c8 = arith.constant 8 : index
  %0 = amdaie.logicalobjectfifo.from_memref %arg0, {} : memref<1x1x8x16xi32, 2> -> !amdaie.logicalobjectfifo<memref<1x1x8x16xi32, 2>>
  %1 = amdaie.logicalobjectfifo.from_memref %arg1, {} : memref<8x16xi32, 1> -> !amdaie.logicalobjectfifo<memref<8x16xi32, 1>>
  scf.for %arg2 = %c0 to %c8 step %c1  {  
    %apply = affine.apply affine_map<(d0) -> (d0 * 32)>(%arg2)
    %2 = amdaie.dma_cpy_nd(%0[] [] [], %1[0, 0, 0, 0] [1, 1, 8, 16] [128, 16, %apply, 1]) : (!amdaie.logicalobjectfifo<memref<1x1x8x16xi32, 2>>, !amdaie.logicalobjectfifo<memref<8x16xi32, 1>>)
  }
  return
}
```

Expected output IR: 

```
func.func @do_not_hoist_affine(%arg0: memref<1x1x8x16xi32, 2>, %arg1: memref<8x16xi32, 1>) {
  amdaie.workgroup {
    %c0 = arith.constant 0 : index
    %c1 = arith.constant 1 : index
    %c8 = arith.constant 8 : index
    %0 = amdaie.logicalobjectfifo.from_memref %arg0, {} : memref<1x1x8x16xi32, 2> -> !amdaie.logicalobjectfifo<memref<1x1x8x16xi32, 2>>
    %1 = amdaie.logicalobjectfifo.from_memref %arg1, {} : memref<8x16xi32, 1> -> !amdaie.logicalobjectfifo<memref<8x16xi32, 1>>
    %2 = amdaie.circular_dma_cpy_nd(%0[] [] [], %1[] [] []) : (!amdaie.logicalobjectfifo<memref<1x1x8x16xi32, 2>>, !amdaie.logicalobjectfifo<memref<8x16xi32, 1>>)
    amdaie.controlcode {
      scf.for %arg2 = %c0 to %c8 step %c1  {
        %apply = affine.apply affine_map<(d0) -> (d0 * 32)>(%arg2)
        %3 = amdaie.npu.dma_cpy_nd %2([] [] [], [0, 0, 0, 0] [1, 1, 8, 16] [128, 16, %apply, 1])
        amdaie.npu.dma_wait(%3, MM2S)
      }
      amdaie.end
    }
  }
  return
}
```

2)  However, as the backend compiler currently doesn't support reprogramming of DMAs on memtiles and cores as described in the output snippet above, we need to avoid creating control code instructions targeting DMAs on memtiles and cores for moving data back-and-forth between L2 and L1. This can be done by integrating the addressing depending on the loop's induction variable into the DMA's access pattern and then hoisting the DMA operation out of the loop.  Example input IR for this transformation:

```
func.func @subsume_iteration_into_dma_and_hoist(%arg0: memref<1x1x8x16xi32, 2>, %arg1: memref<8x16xi32, 1>) {
  %c0 = arith.constant 0 : index
  %c1 = arith.constant 1 : index
  %c8 = arith.constant 8 : index
  %0 = amdaie.logicalobjectfifo.from_memref %arg0, {} : memref<1x1x8x16xi32, 2> -> !amdaie.logicalobjectfifo<memref<1x1x8x16xi32, 2>>
  %1 = amdaie.logicalobjectfifo.from_memref %arg1, {} : memref<8x16xi32, 1> -> !amdaie.logicalobjectfifo<memref<8x16xi32, 1>>
  scf.for %arg2 = %c0 to %c8 step %c1  {  
    %apply = affine.apply affine_map<(d0) -> (d0 * 32)>(%arg2)
    %2 = amdaie.circular_dma_cpy_nd(%0[] [] [], %1[0, %apply, 0, 0] [1, 1, 8, 8] [128, 128, 16, 1]) : (!amdaie.logicalobjectfifo<memref<1x1x8x16xi32, 2>>, !amdaie.logicalobjectfifo<memref<8x16xi32, 1>>)
  }
  return
}
```

Expected output IR: 

```
func.func @subsume_iteration_into_dma_and_hoist(%arg0: memref<1x1x8x16xi32, 2>, %arg1: memref<8x16xi32, 1>) {
  %c0 = arith.constant 0 : index
  %c1 = arith.constant 1 : index
  %c8 = arith.constant 8 : index
  %0 = amdaie.logicalobjectfifo.from_memref %arg0, {} : memref<1x1x8x16xi32, 2> -> !amdaie.logicalobjectfifo<memref<1x1x8x16xi32, 2>>
  %1 = amdaie.logicalobjectfifo.from_memref %arg1, {} : memref<8x16xi32, 1> -> !amdaie.logicalobjectfifo<memref<8x16xi32, 1>>
  %2 = amdaie.circular_dma_cpy_nd(%0[] [] [], %1[0, 0, 0, 0] [1, 8, 8, 8] [256, 32, 16, 1]) : (!amdaie.logicalobjectfifo<memref<1x1x8x16xi32, 2>>, !amdaie.logicalobjectfifo<memref<8x16xi32, 1>>)
  return
}
```

This results in a single programming of the DMAs needed to implement this `amdaie.circular_dma_cpy_nd` operation and no control code is needed on the uController side to update the DMAs.

Abhishek-Varma: Hi @jtuyls !

Thank you for this helpful write-up!

Just adding here points as per my understanding and for some more clarification :-

Case 1: Dependency on the loop induction variable.
This is basically a long-term goal which we can table off, for now.

Case 2: Subsuming the addressing depending on loop induction variable into DMA's stride pattern.
The INPUT IR expected in this case will have `amdaie.dma_cpy_nd` or `amdaie.circular_dma_cpy_nd` ?
             If it's the former, the we need to convert `scf.for(amdaie.dma_cpy_nd)` to `amdaie.circular_dma_cpy_nd(new stride access pattern)` ?
             If it's the latter, then basically the expectation is that the preceding passes should then instead generate a `amdaie.circular_dma_cpy_nd` ?

jtuyls: > The INPUT IR expected in this case will have amdaie.dma_cpy_nd or amdaie.circular_dma_cpy_nd ?
If it's the former, the we need to convert scf.for(amdaie.dma_cpy_nd) to amdaie.circular_dma_cpy_nd(new stride access pattern) ?
If it's the latter, then basically the expectation is that the preceding passes should then instead generate a amdaie.circular_dma_cpy_nd ?

The transformation could in principal be done on both (`DmaCpyNdBaseOp` could be used to accomplish that). However, in practice we will see `amdaie.circular_dma_cpy_nd` between L2/L1 as this currently is the only one that will get through correctly as the backend compiler currently doesn't support reprogramming of DMAs on memtiles and cores. The `AMDAIEDmaToCircularDmaPass` will take care of this: https://github.com/nod-ai/iree-amd-aie/blob/19c401f48eb32831ef1b3ba0368ea7499a763bff/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaToCircularDma.cpp#L38

jtuyls:
I think this has all been done and can be closed @Abhishek-Varma ?

PR #378, opened 2024-05-28
Modify the pack-transpose config for matmul_transpose_b

yzhang93:
This PR changes inner permutation of pack operation to keep the original form of matmul A(m,k) x B(k, n) and matmul_transpose_b A(m, k) x B(n, k).

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
@@ -283,9 +283,19 @@ FailureOr<ParameterSetting> ParameterSetting::create(linalg::LinalgOp linalgOp,
 }
 }  // namespace
 
+static SmallVector<int64_t> setInnerPermB(bool isMatmulTransposeB) {
+  SmallVector<int64_t> innerPermB;
+  if (isMatmulTransposeB) {
+    innerPermB = {0, 1};
+  } else {
+    innerPermB = {1, 0};
+  }
+  return innerPermB;
+}
+
 static LogicalResult setRootConfigForPackPeelPipeline(
     mlir::FunctionOpInterface entryPointFn, linalg::LinalgOp linalgOp,
-    AIEConfig cfg) {
+    AIEConfig cfg, bool isMatmulTransposeB) {
   auto maybePackPeelTiling = ParameterSetting::create(linalgOp, true);
   if (failed(maybePackPeelTiling)) return failure();
   auto packPeelTiling = maybePackPeelTiling.value();
@@ -295,12 +305,15 @@ static LogicalResult setRootConfigForPackPeelPipeline(
   // ------------------------------------------------------
   MLIRContext *context = entryPointFn.getContext();
 
-  // Transpose B matrix from [K N n k] to [K N k n]
+  // For matmul, transpose B matrix from [K N n k] to [K N k n]
+  // For matmul_transpose_b, we don't have to transpose the B matrix,
+  // since it is already [N K n k]
   SmallVector<int64_t> transposePackIndices = {1};
   // There is no corresponding unpack for the specified pack operation
   // 0 is used when unpack is empty
   SmallVector<bool> unpackEmpty = {false};
-  SmallVector<SmallVector<int64_t>> innerPerm = {{1, 0}};
+  SmallVector<int64_t> innerPermB = setInnerPermB(isMatmulTransposeB);
+  SmallVector<SmallVector<int64_t>> innerPerm = {innerPermB};
   SmallVector<SmallVector<int64_t>> outerPerm = {{0, 1}};
   auto packingConfigLevel1Attr = getPackingConfigPackingLevelAttr(
       context, packPeelTiling.getPackSizeL0(), transposePackIndices,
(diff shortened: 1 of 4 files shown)


newling:
Thanks, this is useful!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -295,12 +305,15 @@ static LogicalResult setRootConfigForPackPeelPipeline(
   // ------------------------------------------------------
   MLIRContext *context = entryPointFn.getContext();
 
-  // Transpose B matrix from [K N n k] to [K N k n]
+  // For matmul, transpose B matrix from [K N n k] to [K N k n]
```
newling, line 308: Is this comment correct? 

// For matmul, transpose B matrix from [K N n k] to [K N k n]

Why is B in form [K N n k] to start with, for matmul?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -295,12 +305,15 @@ static LogicalResult setRootConfigForPackPeelPipeline(
...
+  // For matmul_transpose_b, we don't have to transpose the B matrix,
+  // since it is already [N K n k]
   SmallVector<int64_t> transposePackIndices = {1};
   // There is no corresponding unpack for the specified pack operation
   // 0 is used when unpack is empty
   SmallVector<bool> unpackEmpty = {false};
-  SmallVector<SmallVector<int64_t>> innerPerm = {{1, 0}};
+  SmallVector<int64_t> innerPermB = setInnerPermB(isMatmulTranspose);
```
newling, line 315: nit: could do without function call:

SmallVector<int64_t> innerPermB  =  isMatmulTranspose ? {0,1} : {1,0};

newling (changes requested):
We'll probably want to support matmul(A.T, B) and matmul(A.T, B.T) in the future. In which case "isMatmulTranpose" will be ambiguous then, please make explicit the fact that this is matmul(A, B.T)

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: This is correct. Pack-transpose is performed after pack operation, and the packing is done by tiling iterator dimensions in their canonical order. Please refer to this https://github.com/llvm/llvm-project/blob/949ef57dd20f8d3f3257376b91af71ab8c380338/mlir/include/mlir/Dialect/Linalg/TransformOps/LinalgTransformOps.td#L761.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: It doesn't work to initialize the vector with constructors in this way. It has to be static during the compile time.

nirvedhmeshram (approved):
LGTM!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
newling: Ok. I think the function can be simplified, and maybe should be called 'get' rather than 'set'.

yzhang93:
> We'll probably want to support matmul(A.T, B) and matmul(A.T, B.T) in the future. In which case "isMatmulTranpose" will be ambiguous then, please make explicit the fact that this is matmul(A, B.T)

I'm not sure we have matmul(A.T, B) and matmul(A.T, B.T) in GEMM applications. But I agree it's more strict to use term "isMatmulTranposeB". This was following the name here https://github.com/nod-ai/iree-amd-aie/pull/378/files#diff-2d5f147d261b73e74cc0ea08e56b10bedf6cad9f61923d58064304a626e37c13R446.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
newling: Wow! Thanks

newling:
@yzhang93 

FYI with this PR we still get

 error: 'aie.dma_bd' op For <32b width datatypes, inner-most dim stride must be 1
 
from the tranpose matmul test :

https://github.com/nod-ai/iree-amd-aie/blob/78e473e5c9c6f2c5faad2f3afef775fa3495fefd/build_tools/ci/run_matmul_test.sh#L562

We must eventually get to a point where we don't ever do any transpose of the inner most dims in DMA copies (at least for any data types with <32 bits)

yzhang93:
> @yzhang93
> 
> FYI with this PR we still get
> 
> error: 'aie.dma_bd' op For <32b width datatypes, inner-most dim stride must be 1
> 
Yes, I've noticed that. For pack-peel pipeline it has a better situation, it can compile through but has numerical issue. There are still some issues within AIR. Erwei will look into it the next.

> We must eventually get to a point where we don't ever do any transpose of the inner most dims in DMA copies (at least for any data types with <32 bits)

I don't get this. I thought now there's no more transpose of innermost dims?

PR #385, opened 2024-05-31
Bump dependencies 5/30/2024

daveliddell:
Cherry-pick of #383 from main to tresleches branch to fix broken tresleches build.

Original PR description:

Some minor API changes.

### DIFF
--- a/sync_deps.py
+++ b/sync_deps.py
@@ -7,7 +7,7 @@
 ### Update with: shark-workspace pin
 
 PINNED_VERSIONS = {
-  "iree": "06e05b3f86870cd5620df2f57ea8af4cb136c120",
+  "iree": "ff8072818e5996fbc9fb937ce21c8433bba001d5",
 }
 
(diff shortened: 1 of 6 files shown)


sync_deps.py
```diff
@@ -7,7 +7,7 @@
 ### Update with: shark-workspace pin
 
 PINNED_VERSIONS = {
-  "iree": "06e05b3f86870cd5620df2f57ea8af4cb136c120",
+  "iree": "ff8072818e5996fbc9fb937ce21c8433bba001d5",
```
nirvedhmeshram, line 10: I think you might want to keep the old hash as now you are not using the IREE tresleches branch?

daveliddell: Yes, I suppose this is true.  And probably neither one is correct, as we want to pull in the latest optimizations.  I will rework the PR to pull in the HEAD of shared/tresleches-cpu.  Thanks for pointing this out!

Issue #380, opened 2024-05-29
Implement multi-kernel XCLBIN


nirvedhmeshram:
This issue is a amd-aie counterpart of upstream issue https://github.com/iree-org/iree/issues/7824

We will initially follow the SPIR-V commit sequence below
1. https://github.com/iree-org/iree/pull/15785 
2. https://github.com/iree-org/iree/pull/15788 (bug fix vibes will merge with 1) 
3. https://github.com/iree-org/iree/pull/15789 
4. https://github.com/iree-org/iree/pull/15802 

Once this is done the state will be that we will have a XCLBIN archive that combines multiple XCLBINs

Then in the last stage we will use the utility provided here, https://github.com/Xilinx/mlir-aie/pull/1508
to merge the XCLBIN's and achieve the desired multi-kernel XCLBIN

nirvedhmeshram:
I did some profiling on what we get by having a XCLBIN with multiple PDIs. My test example had three kernels and I measured the loading time in the following setups
1. Make a new xclbin for each kernel and load the xclbin and then the corresponding kernel from it.
2. Make one xclbin for all three kernels but load it three times, once for each kernel.
3. Make one xclbin, load it only onces and get the three kerenels from it.

Here are the times I saw for small GEMM shapes (M,N,K = 32,32,32) and large GEMM shapes (M,N.K = 1024,1024, 32)

| Setup| Small | Large |
|--------|-------|-------|
|1| 19.82 ms | 21.66 ms|
|2| 24.11 ms | 22.37 ms|
|3| 19.88 ms | 21.53 ms|

I think it can be concluded that the main bottleneck is getting the kernel from the xclbin itself and merging to one xcdlbin isnt changing the time taken to load the kernels much.

nirvedhmeshram:
@kumardeepakamd please see the findings above. Looks like we need to do some different pdi loads to actually benefit from xclbin merging?

PR #388, opened 2024-05-31
Add initial link executable pass

nirvedhmeshram:
This links the aie.device code from different modules into the same module

e.g 
```
hal.executable dispatch_0 {
  builtin.module {
    aie.device {
      func.func dispatch_0
    }
  }
}

hal.executable dispatch_1 {
  builtin.module {
    aie.device {
      func.func dispatch_1
    }
  }
}
```
Is changed to
```
hal.executable linked_dispatches {
  builtin.module {
    aie.device {
      func.func dispatch_0
    }
    aie.device {
      func.func dispatch_1
    }
  }
}
```
Progress towards: https://github.com/nod-ai/iree-amd-aie/issues/380


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinkExecutable.cpp
```diff
@@ -0,0 +1,79 @@
...
+                              DenseMap<StringRef, Operation *> &symbolMap) {
+        return mergeModuleInto(sourceInnerModule, linkedInnerModule, symbolMap);
+      };
+
+      // Try linking together all executables in moduleOp.
+      if (failed(linkExecutablesInto(moduleOp, sourceExecutableOps,
+                                     linkedExecutableOp, linkedTargetOp,
+                                     mergeModuleFn))) {
+        return signalPassFailure();
+      }
```
yzhang93, line 67: Why is this inside the for loop? What happens if there are multiple executables to link? Maybe it's good to add a test linking three executables as well?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinkExecutable.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+struct AMDAIELinkExecutablesPass
+    : public impl::AMDAIELinkExecutablesBase<AMDAIELinkExecutablesPass> {
+  AMDAIELinkExecutablesPass() = default;
+  void runOnOperation() override {
+    auto moduleOp = getOperation();
+    auto moduleBuilder = OpBuilder::atBlockBegin(moduleOp.getBody());
+
+    auto sourceExecutableOps =
+        llvm::to_vector<8>(moduleOp.getOps<IREE::HAL::ExecutableOp>());
```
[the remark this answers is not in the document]
nirvedhmeshram: Its not limited to 8, its a performance hint for the compiler to allocate space for a vector of 8, but the actual content can be more or less. I think using powers of 2 is common here and since typical program would have several executables we use 8 here. Btw I copied most of this pass from this upstream pass (which had the 8)
https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/LLVMCPU/LLVMCPULinkExecutables.cpp#L26

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinkExecutable.cpp
nirvedhmeshram: The loop is when there are several different executable targets. example from SPIR-V
https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/SPIRV/test/link_executables.mlir#L189-L192
So only the executables with a unique target get put in the same linked executables.. For AIE we only have one target so this feature is not strictly needed but its is good for future proofing and is easier to implement as we can just use the same upstream utility functions this way. 

Adding three executables will use the same exact code flow (as they will be still same target) so I think the two exec example is good enough. Let me know if you still want that.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinkExecutable.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include "iree-amd-aie/Transforms/PassDetail.h"
+#include "iree-amd-aie/Transforms/Passes.h"
+#include "iree/compiler/Codegen/Utils/LinkingUtils.h"
+#include "iree/compiler/Utils/ModuleUtils.h"
+#include "llvm/Support/FormatVariadic.h"
+#include "mlir/Pass/Pass.h"
+
+namespace mlir::iree_compiler::AMDAIE {
+
```
Abhishek-Varma, line 15: Add a `DEBUG_TYPE` for the pass.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinkExecutable.cpp
```diff
@@ -0,0 +1,79 @@
...
+namespace {
+
+struct AMDAIELinkExecutablesPass
+    : public impl::AMDAIELinkExecutablesBase<AMDAIELinkExecutablesPass> {
+  AMDAIELinkExecutablesPass() = default;
+  void runOnOperation() override {
+    auto moduleOp = getOperation();
```
Abhishek-Varma, line 22: `auto` -> `ModuleOp`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinkExecutable.cpp
```diff
@@ -0,0 +1,79 @@
...
+        llvm::formatv("{0}_linked_{1}", moduleName, "amd_aie");
+    auto linkedExecutableOp = moduleBuilder.create<IREE::HAL::ExecutableOp>(
+        moduleOp.getLoc(), linkedExecutableName);
+    linkedExecutableOp.setVisibility(
+        sourceExecutableOps.front().getVisibility());
+    auto executableBuilder =
+        OpBuilder::atBlockBegin(&linkedExecutableOp.getBlock());
+
+    // Gather all unique executable targets - we may have multiple.
+    auto executableTargetAttrs = gatherExecutableTargets(sourceExecutableOps);
```
Abhishek-Varma, line 43: Expand `auto`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/aie_link_executables.mlir
```diff
@@ -0,0 +1,71 @@
+// RUN: iree-opt --split-input-file --iree-amdaie-link-executables %s | FileCheck %s
+#executable_target_amdaie_xclbin_fb = #hal.executable.target<"amd-aie", "amdaie-xclbin-fb", {target_arch = "chip-tbd", ukernels = "none"}>
```
Abhishek-Varma, line 2: Add a new line space here after the `RUN` command.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinkExecutable.cpp
```diff
@@ -0,0 +1,79 @@
...
+  void runOnOperation() override {
+    auto moduleOp = getOperation();
+    auto moduleBuilder = OpBuilder::atBlockBegin(moduleOp.getBody());
+
+    auto sourceExecutableOps =
+        llvm::to_vector<8>(moduleOp.getOps<IREE::HAL::ExecutableOp>());
+    if (sourceExecutableOps.size() <= 1) return;
+
+    // Guess a module name, if needed, to make the output files readable.
+    auto moduleName = guessModuleName(moduleOp, "amdaie_module");
```
Abhishek-Varma, line 30: `std::string` ?

Abhishek-Varma (changes requested):
Hi @nirvedhmeshram - if the difference between this and the upstream utility is based on the `std::string` names, would it be better to create a common utility/pass for both wherein you just specify those names via a pass flag?

nirvedhmeshram:
> Hi @nirvedhmeshram - if the difference between this and the upstream utility is based on the `std::string` names, would it be better to create a common utility/pass for both wherein you just specify those names via a pass flag?

Good point, but I think we want our own pass here to do more custom things in the future (e.g xclbin specific requirements). The core logic of merging executables is already used from upstream.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinkExecutable.cpp
yzhang93: > For AIE we only have one target so this feature is not strictly needed but its is good for future proofing and is easier to implement as we can just use the same upstream utility functions this way.

Can you make a comment about this?

> Adding three executables will use the same exact code flow (as they will be still same target) so I think the two exec example is good enough. Let me know if you still want that.

It's ok if it works locally for you and not adding it to test.

yzhang93 (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinkExecutable.cpp
```diff
@@ -0,0 +1,82 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree-amd-aie/Transforms/PassDetail.h"
+#include "iree-amd-aie/Transforms/Passes.h"
+#include "iree/compiler/Codegen/Utils/LinkingUtils.h"
+#include "iree/compiler/Utils/ModuleUtils.h"
+#include "llvm/Support/FormatVariadic.h"
+#include "mlir/Pass/Pass.h"
+
+#define DEBUG_TYPE "iree-amdaie-linkexec"
```
Abhishek-Varma, line 14: The `DEBUG_TYPE` should be same as the pass name -> `iree-amdaie-link-executables`.

This helps when putting in debug statements as well as quickly finding the `.cpp` as well.

Abhishek-Varma (approved):
LGTM. Just one comment to address before merging this in.

PR #392, opened 2024-06-03
Update pack-peel pipeline and ci tests

yzhang93:
- With this work https://github.com/nod-ai/iree-amd-aie/pull/351, we change the pack-peel pipeline to always peel the first and the last iteration and fuse the unpack op into the inner forall loop for matmul only and matmul-elementwise dispatch.
- With this change, we have made progress on codegen for matmul-transpose-b and matmul-elementwise example dispatches to a reasonable state.
- A minor issue is with [upstream change](https://github.com/llvm/llvm-project/pull/90189), there is redundant data allocation and copy after bufferization ([example IR](https://gist.github.com/yzhang93/55f448368db32cccd2af31c730cc878a#file-gistfile1-txt-L342)). So currently I have to disable the canonicalization pass [here](https://github.com/nod-ai/iree-amd-aie/pull/392/files#diff-42f0d0bb098689f25ee68e8f05ec6c2eaa89ce41a6394d7d99c1d1c912943b38L256), but in the long term we may want to fix this issue.

### DIFF
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -690,7 +690,6 @@ run_matmul_test \
     --lhs_rhs_type "bf16" \
     --acc_type "f32" \
     --m "512"  --n "512" --k "512" \
-    --num_repeat_runs "0"
 
 run_matmul_test \
     --name_prefix "packPeel2304" \
@@ -698,7 +697,22 @@ run_matmul_test \
     --lhs_rhs_type "bf16" \
     --acc_type "f32" \
     --m "128"  --n "128" --k "2304" \
-    --num_repeat_runs "0"
+
+run_matmul_test \
+  --name_prefix "packPeel_t_i32" \
+  --pipeline "pack-peel" \
+  --lhs_rhs_type "i32" \
+  --acc_type "i32" \
+  --m "128" --n "256" --k "512" \
+  --do_transpose_rhs "1"
+
+run_matmul_test \
+  --name_prefix "packPeel_t_bf16" \
+  --pipeline "pack-peel" \
+  --lhs_rhs_type "bf16" \
+  --acc_type "f32" \
+  --m "128" --n "256" --k "512" \
+  --do_transpose_rhs "1" \
 
 ###################################################################
 
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
@@ -243,17 +243,13 @@ void addPackPeelBasedPassPipeline(OpPassManager &funcPassManager,
 
   // Peel the first and last iteration. Note: Do not run CSE pass afterwards,
   // because it may bring problem for bufferization.
-  {
-    AMDAIEPeelForLoopOptions peelOptions;
-    peelOptions.peelingType = PeelingType::FirstLast;
-    funcPassManager.addPass(createAMDAIEPeelForLoopPass(peelOptions));
-  }
+  funcPassManager.addPass(createAMDAIEPeelForLoopPass());
   funcPassManager.addPass(createCanonicalizerPass());
 
   // Fuse fill op into the first inner forall loop
   funcPassManager.addPass(createAMDAIEFuseFillIntoForallPass());
 
-  // Fuse unpack and elementwise op into the last inner forall loop
+  // Fuse unpack/elementwise consumer ops into the last inner forall loop
   funcPassManager.addPass(createAMDAIEFuseConsumerIntoLoopPass());
 
   // Fuse pack ops into the last inner forall loop
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
@@ -241,19 +241,20 @@ void addPackPeelBasedPassPipeline(OpPassManager &funcPassManager,
   funcPassManager.addPass(createCanonicalizerPass());
   funcPassManager.addPass(createCSEPass());
 
-  // Peel the for loop. For matmul dispatch, only peel the first iteration,
-  // while for matmul-elementwise dispatch, peel the first and last iteration.
-  // Note: Do not run CSE pass afterwards, because it may bring problem for
-  // bufferization.
-  funcPassManager.addPass(createAMDAIEPeelForLoopPass());
+  // Peel the first and last iteration. Note: Do not run CSE pass afterwards,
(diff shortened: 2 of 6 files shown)


Abhishek-Varma (commented):
Hi @yzhang93 - thanks for summarizing.

It'd be nice to rebase this on main once the other commits get merged and rename the current PR accordingly.

build_tools/ci/run_matmul_test.sh
```diff
@@ -690,15 +690,29 @@ run_matmul_test \
...
+
+run_matmul_test \
+  --name_prefix "packPeel_t_i32" \
+  --pipeline "pack-peel" \
+  --lhs_rhs_type "i32" \
+  --acc_type "i32" \
+  --m "128" --n "256" --k "512" \
+  --do_transpose_rhs "1"
+
+run_matmul_test \
```
newling, line 709: Nice one, first complete matmul transpose example I think? @jsetoain

yzhang93: This is adding matmul transpose example for pack-peel pipeline. There's already a test working for pad-pack pipeline https://github.com/nod-ai/iree-amd-aie/commit/16a2c08ca13f97ba903335adcbbd90091040ee4b#diff-dfa2b1719749b22586f350c17128c1ec6531c32e18bb17968b75fea850f44333R558

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -241,19 +241,20 @@ void addPackPeelBasedPassPipeline(OpPassManager &funcPassManager,
...
+    peelOptions.peelingType = PeelingType::FirstLast;
+    funcPassManager.addPass(createAMDAIEPeelForLoopPass(peelOptions));
+  }
   funcPassManager.addPass(createCanonicalizerPass());
 
   // Fuse fill op into the first inner forall loop
   funcPassManager.addPass(createAMDAIEFuseFillIntoForallPass());
 
-  // Fuse elementwise op into the last inner forall loop
+  // Fuse unpack and elementwise op into the last inner forall loop
```
Abhishek-Varma, line 256:
```suggestion
  // Fuse unpack/elementwise consumer ops into the last inner forall loop.
```

build_tools/ci/run_matmul_test.sh
newling: That's great!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -241,19 +241,20 @@ void addPackPeelBasedPassPipeline(OpPassManager &funcPassManager,
...
-  // Note: Do not run CSE pass afterwards, because it may bring problem for
-  // bufferization.
-  funcPassManager.addPass(createAMDAIEPeelForLoopPass());
+  // Peel the first and last iteration. Note: Do not run CSE pass afterwards,
+  // because it may bring problem for bufferization.
+  {
+    AMDAIEPeelForLoopOptions peelOptions;
+    peelOptions.peelingType = PeelingType::FirstLast;
+    funcPassManager.addPass(createAMDAIEPeelForLoopPass(peelOptions));
+  }
```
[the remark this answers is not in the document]
yzhang93: Done

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -263,7 +264,6 @@ void addPackPeelBasedPassPipeline(OpPassManager &funcPassManager,
     fusePackOptions.targetElementwise = true;
     funcPassManager.addPass(createAMDAIEFusePackIntoLoopPass(fusePackOptions));
   }
-  funcPassManager.addPass(createCanonicalizerPass());
```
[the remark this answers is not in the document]
yzhang93: 1. Yes, the redundant allocation and copy would cause issues in AIR. I talked with @MaheshRavishankar about the issue offline, and we think it might be a bug in bufferization. An issue https://github.com/iree-org/iree/issues/17571 is created to track this. On the hand, I've compared the IR after iree-comprehensive-bufferization passes (which includes two canonicalization passes after bufferization) with and without this pre canonicialization pass, and the IR dumps turn out to be identical. So before this bufferization issue is fixed we'd not use this canonicalization pass.
2. Another way if we want to keep the canonicalization pass, we have to disable the `ForallOpIterArgsFolder` pattern. I tried to call canonicalization in [this way](https://github.com/llvm/llvm-project/blob/e42864ecfb35466d38beeaf57c5da009c5bc77c1/mlir/unittests/Transforms/Canonicalizer.cpp#L72), but it requires adding `setDebugName("DisabledPattern")` to `ForallOpIterArgsFolder` rewritePattern's constructor. This change may not be desirable in upstream, but if the otherwise, I'm okay to change it in upstream and call canonalizer with disabled pattern.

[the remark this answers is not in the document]
Abhishek-Varma: Aah okay. I understand.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -304,7 +304,7 @@ def AMDAIEPeelForLoop :
   let options = [
     Option<"peelingType", "peeling-type",
       "mlir::iree_compiler::AMDAIE::PeelingType",
-      /*default=*/"",
+      /*default=*/"mlir::iree_compiler::AMDAIE::PeelingType::FirstLast",
```
Abhishek-Varma, line 307:
```suggestion
      /*default=*/"::mlir::iree_compiler::AMDAIE::PeelingType::FirstLast",
```

The preceding `::` ensures that the global `mlir` namespace is used.

Abhishek-Varma (approved):
LGTM! A small nit to address. Thanks for the changes!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
yzhang93: If this is the case, we'd need to update the namespace all over the place in this repository. I won't address this in this PR to keep consistent. If it's really needed, I'd do it in a follow-up PR.

Abhishek-Varma: Sure, it'd be great if you can address this in a follow-up PR. Not a blocker for this PR.
General code health of the repository and correct conventions go a long way in maintaining the same.

PR #395, opened 2024-06-06
Enable initial executable linking + mlir-air bump

nirvedhmeshram:
Adds the pass to link multiple HAL executables together to the pass pipeline. The resulting linke3d executable can have several entry points. At artifact creation time we create artifact for each entry point one at a time by passing the corresponding aie.device op to the `aie2xclbin` tool. In a later PR we will merge the xclbins when possible.
The executable schema is updated to account for flexibility needed to have shared xclbins / lx6 instruction streams.
We also add logic to use assigned ordinals of entry points rather than assuming ascending order as noted by https://github.com/iree-org/iree/pull/15905

An mlir-air bump is needed for this to work e2e.

Progress towards: https://github.com/nod-ai/iree-amd-aie/issues/380


compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -38,6 +38,43 @@ static llvm::cl::opt<std::string> clEnableAMDAIEUkernels(
                    "unprefixed microkernels to enable, e.g. `matmul`."),
     llvm::cl::init("none"));
 
+// Utility to find aie.device Op corresponding to the export Op, We have
```
yzhang93, line 41: Nit: Utility to find aie.device Op corresponding to the export Op. For example, we have

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -569,6 +569,17 @@ void addMLIRAIRAIELoweringPasses(OpPassManager &passManager, bool packPeel) {
...
+// NOTE: this runs on the top-level program module containing all hal.executable
+// ops.
+void buildAMDAIELinkingPassPipeline(OpPassManager &passManager) {
+  // Link together executables. This may produce some IR duplication.
+  passManager.addPass(createAMDAIELinkExecutablesPass());
+
+  // Cleanup IR duplication.
+  passManager.addNestedPass<IREE::HAL::ExecutableOp>(
+      mlir::createCanonicalizerPass());
+}
```
yzhang93, line 581: Why do you expose this in Passes.cpp? Can it be directly added to [AIETarget.cpp](https://github.com/nod-ai/iree-amd-aie/pull/395/files#diff-2a27c2baf66b75bcf426d927b55cbf5f3b5d7bfa69b9b7492ab0d3672805bac4R151)?

yzhang93 (commented):
I don't have much context, but the codes generally LGTM. Is this working currently? It's good to add a test.

newling:
Not sure what happened in CI, looks like an xclbin was created but then the `cp` failed because it couldn't see an xclbin

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -38,6 +38,43 @@ static llvm::cl::opt<std::string> clEnableAMDAIEUkernels(
...
+  moduleOp.walk([&](func::FuncOp funcOp) {
+    if (funcOp.getSymName() == exportOpName) {
+      deviceOp = dyn_cast_or_null<xilinx::AIE::DeviceOp>(funcOp->getParentOp());
+      return WalkResult::interrupt();
+    }
+    return WalkResult::advance();
+  });
+  if (!deviceOp) {
+    moduleOp.emitError()
+        << "failed to find aie.device Op containing func.func with symbol "
```
newling, line 72: nit: why `aie.device Op` but `func.func` -- they're both MLIR Ops. I'd personally drop the ` Op`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -38,6 +38,43 @@ static llvm::cl::opt<std::string> clEnableAMDAIEUkernels(
...
+// Hence we need to find the func.func that coresponds to the export op symbol
+// and return its parent aie.device Op. This is what we will pass to the
+// `aie2xclbin` tool for artifact generation per entry point
+static xilinx::AIE::DeviceOp getDeviceOpFromEntryPoint(ModuleOp moduleOp,
+                                                       StringRef exportOpName) {
+  xilinx::AIE::DeviceOp deviceOp;
+
+  moduleOp.walk([&](func::FuncOp funcOp) {
+    if (funcOp.getSymName() == exportOpName) {
+      deviceOp = dyn_cast_or_null<xilinx::AIE::DeviceOp>(funcOp->getParentOp());
```
newling, line 65: Could use `funcOp->getParentOfType<xilinx::AIE::DeviceOp>()` (recursive though). 

https://github.com/llvm/llvm-project/blob/aae025a01b423ce43e554bc93da15016fa647beb/mlir/include/mlir/IR/Operation.h#L238

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -164,131 +205,195 @@ LogicalResult AIETargetBackend::serializeExecutable(
...
-  }
+  FlatbufferBuilder builder;
+  iree_amd_aie_hal_xrt_ExecutableDef_start_as_root(builder);
+  SmallVector<iree_amd_aie_hal_xrt_XclbinDef_ref_t> xclbinRefs;
+  SmallVector<iree_amd_aie_hal_xrt_AsmInstDef_ref_t> asmInstrRefs;
+
+  // Per entry-point data.
+  // Note that the following vectors should all be of the same size and
+  // element at index #i is for entry point with ordinal #i!
+  SmallVector<std::string> entryPointNamesFb;
```
newling, line 260: nit

```suggestion
  SmallVector<std::string> entryPointNamesFb (ordinalCount);
```

etc rather than resize

newling (approved):
This .fbs stuff is vaguely familiar but I also feel like I'm missing context, so my comments are all very shallow. Looks fine though

nirvedhmeshram:
> I don't have much context, but the codes generally LGTM. Is this working currently? It's good to add a test.

Yes I wanted to add a e2e test but couldnt find a easy way to do so with the current CI infrastructure here is a example matmul sequence I tested locally
```
!A_TYPE = tensor<32x32xf32>
!B_TYPE = tensor<32x32xf32>
!C_TYPE = tensor<32x32xf32>
!D_TYPE = tensor<32x16xf32>
func.func @two_mm(%lhs : !A_TYPE,
    %rhs : !B_TYPE, %rhs_2 : !D_TYPE) -> !D_TYPE {
  %empty = tensor.empty() : !C_TYPE
  %empty_2 = tensor.empty() : !D_TYPE
  %cst = arith.constant 0.0 : f32
  %fill = linalg.fill ins(%cst : f32) outs(%empty : !C_TYPE) -> !C_TYPE
  %fill_2 = linalg.fill ins(%cst : f32) outs(%empty_2 : !D_TYPE) -> !D_TYPE
  %2 = linalg.matmul ins(%lhs, %rhs : !A_TYPE, !B_TYPE)
      outs(%fill : !C_TYPE) -> !C_TYPE
  %3 = linalg.matmul ins(%2, %rhs_2 : !A_TYPE, !D_TYPE)
      outs(%fill_2 : !D_TYPE) -> !D_TYPE
  return %3 : !D_TYPE
}
```
The main blocker is that the signing process needs to account for the new directory structure that happens after the linking optimization. I am not too inclined to fix it becuase once we update the driver we wont have to sign anything so maybe we can add a test after that? In the mean time I have opned this issue
https://github.com/nod-ai/iree-amd-aie/issues/397

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
nirvedhmeshram: Good point, this is just the pattern I saw accross all the other backends, I think the *Target.cpp files just call the top level pass pipelines and they are implemented in the Passes.cpp file. e.g
https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/SPIRV/Passes.cpp#L663-L681
https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/LLVMCPU/Passes.cpp#L776-L789

nirvedhmeshram:
> Not sure what happened in CI, looks like an xclbin was created but then the `cp` failed because it couldn't see an xclbin

It was because the directory structure / file names are different with the link optimization, I changed that if there is only one dispatch. Lets see if that helps.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -164,131 +206,193 @@ LogicalResult AIETargetBackend::serializeExecutable(
...
     // The xclbin kernel name, appended with instance name suffix (`:MLIRAIEV1`,
     // 10 chars) is required by the xclbinutil to have a length smaller or equal
     // to 64 right now. To have some additional wiggle room for suffix changes,
-    // we use the 48 first characters for the kernel name. This is okay to do
-    // for now because we are only supporting single entry point.
-    entryPointNames.emplace_back(exportOp.getSymName().substr(0, 48));
+    // we use the 42 first characters for the kernel name. We suffix ordinal
+    // number to the name to gaurantee uniqueness within the executable.
+    std::string entryPointName = llvm::join_items(
+        "_", exportOpName.substr(0, 42), std::to_string(ordinal));
```
makslevental, line 238: nit: wouldn't it be better to check and fail that the name is less than 42 chars?

nirvedhmeshram: Not sure I follow, do you mean if the user and/or IREE's naming scheme on top of that gives a large name we dont compile?

nirvedhmeshram:
@newling could you PTAL at the new CI side changes?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
nirvedhmeshram: I think I see what you meant, you are talking of the resulting name, not the initial name? I added a check that the resulting name is less than 50 characters which leaves enough wiggle room for the utility tools to suffix more things to it.

build_tools/ci/cpu_comparison/run_test.sh
```diff
@@ -130,6 +122,8 @@ fi
 
 
 source $XRT_DIR/setup.sh
+# Circumvent xclbin security (no longer needed as of April 2024 XDNA driver)
```
yzhang93, line 125: Is this the reason we don't need to do signing?

yzhang93 (approved):
Nice! As you are bumping AIR in this PR, can you bump it to the latest again? Thanks

build_tools/ci/cpu_comparison/run_test.sh
newling: Yup. This line will be removed too as soon as we can update the CI machine driver

newling (commented):
If you remove [build_tools/ci/rm_xclbin.sh](https://github.com/nod-ai/iree-amd-aie/pull/395/files#diff-ea9667334ebb5ac03ae832dd096a1d24eca842f0398326e2e72c624e46bc006e) can you please update the test `cpu_comparison/run_test.sh`  too?

nirvedhmeshram:
> If you remove [build_tools/ci/rm_xclbin.sh](https://github.com/nod-ai/iree-amd-aie/pull/395/files#diff-ea9667334ebb5ac03ae832dd096a1d24eca842f0398326e2e72c624e46bc006e) can you please update the test `cpu_comparison/run_test.sh` too?

I already did? Is something missing in it?

PR #396, opened 2024-06-07
[UnrollAndDistributeWorkgroup] Remove workgroup dependency and rename

jtuyls:
First PR in a sequence to remove `amdaie.workgroup` before the `AMDAIECreateAIEWorkgroup` pass. Right now, local `amdaie.workgroup`  operations are inserted first and then consolidated into a single `amdaie.workgroup` operation. This flow is confusing and not necessary. As mentioned in the PR that added `AMDAIECreateAIEWorkgroup` (https://github.com/nod-ai/iree-amd-aie/pull/316): 

> I noticed that it's probably not necessary to create/insert workgroups before this pass, so I will probably get it out of InsertAIEWorkgroup in a follow-up. This will also resolve the unclarity of having both a InsertAIEWorkgroup and CreateAIEWorkgroup pass. This inconsistency is the result of some rewrites/refactoring after getting in the earlier transformations.

This PR removes the unnecessary dependency of the `AMDAIEUnrollAndDistributeWorkgroup` pass on `amdaie.workgroup`.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -8,32 +8,50 @@
...
 
+// Used to annotate loops that should be unrolled.
+static const llvm::StringLiteral kAMDAIELoopUnroll = "amdaie.unroll";
+
 namespace {
 
-/// Convert scf.forall ops within a workgroup to scf.for ops
-/// TODO(jornt): use upstream `forallToFor` function once merged.
-LogicalResult workgroupForallToFor(ModuleOp moduleOp) {
+/// Convert local scf.forall ops chosen for parallel distribution to scf.for
```
yzhang93, line 27: Nit: maybe "inner" scf.forall is a better description?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -8,32 +8,50 @@
...
+    if (!isa<mlir::gpu::GPUThreadMappingAttr>(*mapping.begin()))
+      return WalkResult::advance();
+
+    SmallVector<Operation *> results;
+    if (failed(scf::forallToForLoop(rewriter, forallOp, &results))) {
+      forallOp.emitOpError() << "failed to transform scf.forall to scf.for";
+      return WalkResult::interrupt();
+    }
+    // Set attribute to unroll this loop later in this pass.
+    for (Operation *res : results) {
```
yzhang93, line 45: Nit: I would avoid using the same variable name "res" here.

yzhang93 (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
jtuyls: I didn't use `inner` because when we get to include block ids as well, it won't be only the inner ones anymore.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
jtuyls: Great catch! Thanks

PR #398, opened 2024-06-07
Fix the memory space attribute for new allocation

yzhang93:
- Previously, the new allocations are always set as AMDAIEMemSpace::Local, since the others have already be bufferized with the correct memory space before comprehensive-bufferization. However, for the newly added test case, there's truncf op to convert f32 back to bf16. The previous memref allocation data type in L2 is f32,  so we'll need a new allocation with data type bf16  in L2 memory space. This PR fixed such problem which had the wrong memory space.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -77,10 +76,15 @@ void appendVectorizationToPipeline(OpPassManager &funcPassManager) {
 static FailureOr<Value> aieComprehensiveBufferizeAllocationFn(
     OpBuilder &builder, Location loc, MemRefType memRefType,
     ValueRange dynamicSizes, unsigned alignment) {
+  int64_t numDims = memRefType.getShape().size();
+  AMDAIEMemSpace memSpace = AMDAIEMemSpace::Local;
+  if (clUsePipeline == AIEPassPipeline::PackPeelPipeline && numDims == 4) {
```
nirvedhmeshram, line 81: is guarding it for the `AIEPassPipeline::PackPeelPipeline` necessary?

nirvedhmeshram (approved):
LGTM!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
yzhang93: Yes. First, the number of dimensions is 2 instead of 4 in L2 memory for pad-pack pipeline. In addition, I think all the allocations have already been assigned memory space before the comprehensive bufferization step for pad-pack pipeline.

PR #399, opened 2024-06-07
[FuseLogicalObjectFifoIntoWorkgroup] Remove workgroup dependency and rename

jtuyls:
Second PR in a sequence to remove `amdaie.workgroup` before the `AMDAIECreateAIEWorkgroup` pass. Right now, local `amdaie.workgroup`  operations are inserted first and then consolidated into a single `amdaie.workgroup`  operation. This flow is confusing and not necessary. As mentioned in the PR that added `AMDAIECreateAIEWorkgroup` (https://github.com/nod-ai/iree-amd-aie/pull/316):

> I noticed that it's probably not necessary to create/insert workgroups before this pass, so I will probably get it out of `InsertAIEWorkgroup` in a follow-up. This will also resolve the unclarity of having both a `InsertAIEWorkgroup` and `CreateAIEWorkgroup` pass. This inconsistency is the result of some rewrites/refactoring after getting in the earlier transformations.

This PR removes the unnecessary dependency of the `FuseLogicalObjectFifoIntoWorkgroup` pass on `amdaie.workgroup` .


yzhang93 (approved):
Nice and clean codes!

PR #406, opened 2024-06-11
[InsertAIEWorkgroup] Refactor to only insert cores and rename

jtuyls:
Third and final PR in a sequence to remove `amdaie.workgroup` before the `AMDAIECreateAIEWorkgroup` pass. Right now, local `amdaie.workgroup`  operations are inserted first and then consolidated into a single `amdaie.workgroup`  operation. This flow is confusing and not necessary. As mentioned in the PR that added `AMDAIECreateAIEWorkgroup` (https://github.com/nod-ai/iree-amd-aie/pull/316):

> I noticed that it's probably not necessary to create/insert workgroups before this pass, so I will probably get it out of `InsertAIEWorkgroup` in a follow-up. This will also resolve the unclarity of having both a `InsertAIEWorkgroup` and `CreateAIEWorkgroup` pass. This inconsistency is the result of some rewrites/refactoring after getting in the earlier transformations.

This PR removes the `amdaie.workgroup` insertion and renames `AMDAIEInsertAIEWorkgroupPass` to `AMDAIEInsertCoresPass`.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCores.cpp
```diff
@@ -30,74 +40,36 @@ void getAttributeMapping(SmallVector<scf::ForallOp> forallOps,
...
 /// Insert core ops inside innermost forall ops around computational ops and
 /// add synchronization ops along the way to synchronize with surrounding
 /// dma ops.
-LogicalResult insertCoreOpsInWorkgroup(mlir::ModuleOp moduleOp) {
+LogicalResult insertCoreOps(mlir::ModuleOp moduleOp) {
   IRRewriter rewriter(moduleOp.getContext());
-  WalkResult res = moduleOp->walk([&](AMDAIE::WorkgroupOp workgroupOp) {
-    // Find innermost forall loops
-    SmallVector<scf::ForallOp> innermostForallLoops =
-        getInnermostForallLoops(workgroupOp);
```
yzhang93, line 72: This utility seems not be used in other places, and can be cleaned as well?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCores.cpp
```diff
@@ -30,74 +40,36 @@ void getAttributeMapping(SmallVector<scf::ForallOp> forallOps,
...
-      return WalkResult::interrupt();
-    }
-    scf::ForallOp innermostForall = innermostForallLoops[0];
-    if (!innermostForall.isNormalized()) {
-      innermostForall.emitOpError()
+  WalkResult res = moduleOp->walk([&](scf::ForallOp forallOp) {
+    // Advance if no thread mapping.
+    SmallVector<Attribute> mapping =
+        llvm::to_vector(forallOp.getMapping()->getValue());
+    if (!isa<mlir::gpu::GPUThreadMappingAttr>(*mapping.begin()))
```
yzhang93, line 52: Add comment to say assuming the innermost forall ops are with thread mapping attribute.

yzhang93 (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCores.cpp
jtuyls: Thanks for the catch!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCores.cpp
jtuyls: Done

PR #410, opened 2024-06-12
[AMD-AIE] Use emitOpError in consumer fusion pass and segregate the lit test

Abhishek-Varma:
-- This commit uses proper pass failure instead of debug statements.


nirvedhmeshram:
Can you elaborate why we should fail if the input IR doesn't have the specific loop nest the pass is looking for. The subsequent passes, which may include this same pass but with a further lowered IR may be able co compile to success?

Abhishek-Varma:
> Can you elaborate why we should fail if the input IR doesn't have the specific loop nest the pass is looking for. The subsequent passes, which may include this same pass but with a further lowered IR may be able co compile to success?

Hi @nirvedhmeshram - the pass itself doesn't make any assumption on the loop nest structure and just tries to fetch the first `scf.for*` in the reverse order. I get your point though and now `LLVM_DEBUG` vs `emitOpError`'s usage is similar to what we have today with `iree-amdaie-fuse-fill-into-loop` pass which should be acceptable now.

nirvedhmeshram (approved):
LGTM, may be update the PR title as now I think the addition from PR is that this is making two separate test files for scf.for and scf.for_all?

Abhishek-Varma:
> LGTM, may be update the PR title as now I think the addition from PR is that this is making two separate test files for scf.for and scf.for_all?

Well, that and couple of `LLVM_DEBUG` statements were changed to `emitOpError`, yes - I've changed the PR title to reflect the same.

@yzhang93 - since this is a follow-up patch can you please review this as well?

yzhang93 (approved):
LGTM. Thanks!

PR #415, opened 2024-06-13
Benchmark: matmul ukernel vs direct codegen

newling:
End to end script. Running locally (nuc50)

Direct codegen 
```
Benchmark summary
=================
m: 2048
k: 4096
n: 2048
number of operations (2*m*n*k): 3.43597e+10
execution times for all 10 runs [s]:
    0.0586853 0.0539645 0.0539398 0.0539536 0.0540862 0.0540921 0.054008 0.0540939 0.0540261 0.0540442
teraops/second over all 10 runs:
    0.585491 0.63671 0.637002 0.636839 0.635277 0.635208 0.636197 0.635187 0.635984 0.635771
mean time over runs: 0.0544894 [s]
minimum time over runs: 0.0539398 [s]
max teraops/second: 0.637002 [teraops/second]
```

Using ukernel 
```
Benchmark summary
=================
m: 2048
k: 4096
n: 2048
number of operations (2*M*N*K): 3.43597e+10
execution times for all 10 runs [s]:
    0.0219439 0.0181417 0.0181282 0.0177506 0.0187402 0.0183464 0.0187545 0.0181528 0.0177086 0.0180813
teraops/second over all 10 runs:
    1.5658 1.89397 1.89537 1.93569 1.83348 1.87283 1.83208 1.89281 1.94029 1.90029
mean time over runs: 0.0185748 [s]
minimum time over runs: 0.0177086 [s]
max teraops/second: 1.94029 [teraops/second]
```

So the _ukernel approach is currently 3x faster_. This is a lower bound though (i.e. core ukernel probably more than 3x faster). Consider:
```
total-time-ukernel = time-in-ukernel + other-time
total-time-dcg = time-in-dcg + other-time
```
Where `other-time` is the same in the 2 experiments, as only the instruction memory is different (identical DMA data movement). We observed that
```
total-time-ukernel / total-time-dcg = 1/3
```
so that 
```
time-in-ukernel / time-in-dcg = 1/3 - 2/3 * (other-time / time-in-dcg) < 1/3
```
as `other-time` is the same in both approaches (date movement between DDR <-> memtile <-> core is identical)

I think on this phoenix machine, theoretical max is `4 tops/second`. So ukernel approach is 50% of theoretical max. 

Two extremes: 

1) all time in ukernel. i.e. other-time = 0. Then time-in-dcg = 3 * time-in-ukernel
2) 50% of time in ukernel (i.e. ukernel itself is 100% efficient). i.e. other-time = time-in-ukernel.  Then time-in-dcg = 5 * time-in-ukernel. 

So performance of ukernel is between 3x and 5x better than dcg.

### DIFF
--- a/build_tools/ci/matmul_ukernel_vs_codegen_benchmark/run_test.sh
+++ b/build_tools/ci/matmul_ukernel_vs_codegen_benchmark/run_test.sh
@@ -1,0 +1,253 @@
+#!/bin/bash
+#
+# Copyright 2024 The LLVM Project
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+set -euox pipefail
+
+# The usual boilerplate for getting the AIE directories setup.
+if [ "$#" -lt 2 ] || [ "$#" -gt 6 ]; then
+
+   # The expected parameters are
+   #    1) <output-dir>            (required)
+   #    2) <iree-install-dir>      (required)
+   #    3) <mlir-aie-install-dir>  (optional)
+   #    4) <peano-install-dir>     (optional)
+   #    5) <xrt-dir>               (optional)
+   #    6) <vitis-install-dir>     (optional)
+    echo -e "Illegal number of parameters: $#, expected 2,3,4,5, or 6 parameters." \
+            "\n The parameters are as follows:" \
+            "\n     1) <output-dir>               (required)" \
+            "\n     2) <iree-install-dir>         (required)" \
+            "\n     3) <mlir-aie-install-dir>     (optional)" \
+            "\n     4) <peano-install-dir>        (optional)" \
+            "\n     5) <xrt-dir>                  (optional)" \
+            "\n     6) <vitis-install-dir>        (optional)" \
+            "\n Example, dependent on environment variables:" \
+            "\n     ./run_test.sh  " \
+            "results_dir_tmp  \$IREE_BUILD_OR_INSTALL_DIR  \$MLIR_AIE_INSTALL_DIR  " \
+            "\$PEANO_INSTALL_DIR  /opt/xilinx/xrt  \$VITIS_INSTALL_PATH"
+    exit 1
+fi
+
+# All generated files should end up in the output directory.
+OUTPUT_DIR=`realpath "$1"`
+DIRECT_OUTPUT_DIR="${OUTPUT_DIR}/direct"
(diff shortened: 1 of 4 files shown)


build_tools/ci/matmul_ukernel_vs_codegen_benchmark/test.cpp
```diff
@@ -0,0 +1,261 @@
...
+  for (int i = 0; i < nIters; ++i) {
+
+    bo_instr.sync(XCL_BO_SYNC_BO_TO_DEVICE);
+    bo_a.sync(XCL_BO_SYNC_BO_TO_DEVICE);
+    bo_b.sync(XCL_BO_SYNC_BO_TO_DEVICE);
+    bo_c.sync(XCL_BO_SYNC_BO_TO_DEVICE);
+
+    // We will time the run, and print it.
+    auto start = std::chrono::high_resolution_clock::now();
+    auto run = kernel(bo_instr, instr_v.size(), bo_a, bo_b, bo_c);
```
nirvedhmeshram, line 172: was this kernel generated with latest iree-amd-aie, if the answer is yes then this would be the wrong way to run it. You can copy the changes done in the changes done by https://github.com/Xilinx/mlir-aie/pull/1517

build_tools/ci/matmul_ukernel_vs_codegen_benchmark/run_test.sh
```diff
@@ -0,0 +1,253 @@
...
+echo "Executing command: $GENERATE_COMMAND"
+eval $GENERATE_COMMAND
+
+THIS="$(cd $(dirname $0) && pwd)"
+
+BASE_COMPILATION_FLAGS="-iree-hal-target-backends=amd-aie \
+-iree-amd-aie-peano-install-dir=${PEANO} \
+-iree-amd-aie-mlir-aie-install-dir=${MLIR_AIE_INSTALL} \
+-iree-amd-aie-vitis-install-dir=${VITIS} \
+-iree-amd-aie-show-invoked-commands"
```
jsetoain, line 147: This is missing the path to IREE:

```bash
-iree-amd-aie-install-dir=$IREE_INSTALL_DIR
```

newling: Ok thanks. This is a recent flag addition afaik. Maybe I should ask/push for this PR to be landed so it doesn't go stale..

jsetoain: Yes, very recent (it just started failing for me). Having this test in CI wouldn't be entirely out of the question, I don't think. It's a good reference for people trying to do some performance analysis (like myself).

PR #418, opened 2024-06-14
merge kernels in existing XCLBIN

nirvedhmeshram:
This PR uses a new  xclbin generation tool from https://github.com/Xilinx/mlir-aie/pull/1561 that allows us to merge kernel PDIs into the same XCLBIN.
A mlir-aie wheel bump is needed to use the tool.
An e2e example which forms multiple (three) dispatches is added.


build_tools/ci/cpu_comparison/test_files/three_matmuls.mlir
```diff
@@ -0,0 +1,31 @@
+// This test shows arbitory matmuls that would have producer consumer relationships
```
newling, line 1:
```suggestion
// This test shows arbitrary matmuls that would have producer consumer relationships
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -300,18 +302,34 @@ LogicalResult AIETargetBackend::serializeExecutable(
...
+                                       "--npu-insts-name",
+                                       npuInstPath,
+                                       "--xclbin-kernel-name",
+                                       entryPointNamesFb[ordinal],
+                                       "--tmpdir",
+                                       entryPointWorkDir,
+                                       "--xclbin-kernel-id",
+                                       ordinalHex};
+    cmdArgs = cmdArgsBase;
+    bool AttemptingMerge = false;
```
newling, line 326:
```suggestion
    bool attemptingMerge = false;
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -350,11 +368,24 @@ LogicalResult AIETargetBackend::serializeExecutable(
     {
       SmallVector<StringRef> cmdEnvRefs{cmdEnv.begin(), cmdEnv.end()};
       int result = llvm::sys::ExecuteAndWait(cmdArgs[0], cmdArgs, cmdEnvRefs);
-      if (result != 0)
+      if (result != 0 && AttemptingMerge) {
```
newling, line 371: Can we detect before hand if the same kernel appears twice? Just because it'll be better to know why we failed. My concern is that we'll stop being able to merge for some other reason, and not know about it.

newling (commented):
Very cool that the e2e test works! Just to confirm: these matmuls form 3 different dispatches?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
nirvedhmeshram: I am not sure I correctly understand the case you are thinking about. But if I understand correctly you are thinking some kernel cannot be merged in the existing XCLBIN and if the next kernel is same as the one that couldnt be we will just not be able to merge?

If what I interpret is correct then a few reasons why this is not of concern.
1. IREE has deduplication passes after dispatch formation so same kernel is never appearing twice.
2. The success of the merge is independent of the content in the kernel. The only case that I am aware of when it will fail is when we exceed the number of PDIs (we have one PDI per kernel for now until we get to super kernels) that xclbin support. I believe the limit is 16. So my idea here is the 17th kernel will fail and then start a new xclbin, and the process repeats for the 33rd kernel and so on. A bit hard to test and we arent at that scale anyway now so havent worked on testing that.

nirvedhmeshram:
> Very cool that the e2e test works! Just to confirm: these matmuls form 3 different dispatches?

Yes that is correct.

nirvedhmeshram:
On further investigating I realized that with the wheel the CI is using, no actual merging is taking place, you can see this error in the CI test added
```
Expected<T> must be checked before access or destruction.
Expected<T> value was in success state. (Note: Expected<T> values in success mode must still be checked prior to being destroyed).
```
This causes the merging to fail and it goes through the regular new xclbin per dispatch path. 
This is not what I see when I build my own mlir-aie. However, given I am seeing same performance with and without this PR I think I will hit pause on this for now and come back to it later perhaps after this PR lands, 
https://github.com/nod-ai/iree-amd-aie/pull/422

PR #421, opened 2024-06-16
Restore AIE/AIEX passes temporarily (2/n)

makslevental:
This PR is part of a stack of PRs that refactor the dependency on MLIR-AIE. See https://github.com/nod-ai/iree-amd-aie/issues/430 for more information.

This PR adds all the boilerplate for the passes we'll ultimately need (and rewrite): `aie-assign-lock-ids`, `aie-assign-buffer-descriptor-ids`, `aie-assign-buffer-addresses-basic`, `aie-pathfinder`, `aie-localize-locks`, `aie-objectstateful-transform`, `aiex-dma-to-npu`. This PR also adds all the tests from upstream that'll serve as that unit-tests going forward.

Note, `aie-assign-buffer-addresses-basic` is based on the previous version of what mlir-aie currently has now and we take that one because it is dramatically simpler.

Also note, there's not much point in looking at the implementations of the passes closely until this PR https://github.com/nod-ai/iree-amd-aie/pull/427, at which point the implementation is the one that will be final-ish (i.e., up until that PR everything is mostly copy-pasted from mlir-aie, including bugs 😞).


compiler/plugins/target/AMD-AIE/aie/AIEAssignBufferAddressesBasic.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+      std::sort(buffers.begin(), buffers.end(), [](BufferOp a, BufferOp b) {
+        return a.getAllocationSize() > b.getAllocationSize();
+      });
+
+      // Leave room at the bottom of the address range for stack
+      int stacksize = 0;
+      int address = 0;
+      if (auto core = tile.getCoreOp()) {
+        stacksize = core.getStackSize();
```
newling, line 76:
```suggestion
        int stacksize = core.getStackSize();
```
and remove previous version, keeping local to scope

compiler/plugins/target/AMD-AIE/aie/AIEAssignBufferAddressesBasic.cpp
```diff
@@ -0,0 +1,103 @@
...
+      }
+
+      for (auto buffer : buffers) {
+        buffer.setAddress(address);
+        address += buffer.getAllocationSize();
+      }
+
+      if (address > maxDataMemorySize) {
+        InFlightDiagnostic error =
+            tile.emitOpError("allocated buffers exceeded available memory\n");
```
newling, line 87:
```suggestion
            tile.emitOpError("allocated buffers exceeded available memory (") << address << ">" << maxDataMemorySize << ")\n";
```
and add a test for this. Example of 'negative' test: https://github.com/nod-ai/iree-amd-aie/blob/9773f55fc84ff0c12616da4d17719d362555893e/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/pack_to_dma_failures.mlir#L4

compiler/plugins/target/AMD-AIE/aie/AIEAssignBufferAddressesBasic.cpp
```diff
@@ -0,0 +1,103 @@
...
+      if (tile.isMemTile())
+        maxDataMemorySize = targetModel.getMemTileSize();
+      else
+        maxDataMemorySize = targetModel.getLocalMemorySize();
+      SmallVector<BufferOp, 4> buffers;
+      device.walk<WalkOrder::PreOrder>([&](BufferOp buffer) {
+        if (buffer.getTileOp() == tile) buffers.push_back(buffer);
+      });
+
+      std::sort(buffers.begin(), buffers.end(), [](BufferOp a, BufferOp b) {
```
newling, line 68: Why largest to smallest?

newling (changes requested):
Eventually how many mlir dialects do you envisage here? amdaie & aie & aiex seems a bit unneccessary

Maybe create Passes.td generated to avoid boilerplate of aie/AIEAssignBufferAddressesBasic.cpp? That would also be a good place to add a sentence on what the pass does (as obvious as it should be now). 

The pass AIEAssignBufferAddressesBasic is indeed nice and simple. But I guess we'll definitely want to make fine-grained decisions on which bank to allocate in eventually as in https://github.com/Xilinx/mlir-aie/blame/875648d47ede10c2405181325a325a91c7bad463/lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp thoughts? 

I don't know understand why you have added tests for passes which only exist in mlir-aie. I would bring the tests in if/when the pass being tested lives in iree-amd-aie.

makslevental:
> Eventually how many mlir dialects do you envisage here? amdaie & aie & aiex seems a bit unneccessary


Only `amdaie`. All analysis from `aiex` and `aie` will get folded into utility code (that manipulates non-AST representations).

> Maybe create Passes.td generated to avoid boilerplate of aie/AIEAssignBufferAddressesBasic.cpp? That would also be a good place to add a sentence on what the pass does (as obvious as it should be now).

I didn't do this because the pass won't stay a pass - it is exactly the kind of simple utility code that has no need for all the boilerplate involved with transforming AST objects (it's just a loop...). 

> The pass AIEAssignBufferAddressesBasic is indeed nice and simple. But I guess we'll definitely want to make fine-grained decisions on which bank to allocate in eventually as in https://github.com/Xilinx/mlir-aie/blame/875648d47ede10c2405181325a325a91c7bad463/lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp thoughts?

Too complex but not complex enough (currently there are bugs). The right way to do this kind of thing is to write down a [CPSAT problem and solve it](https://arxiv.org/abs/2203.00448). That was planned future work when I was in `mlir-aie` and now becomes planned future work here 🙂.

> I don't know understand why you have added tests for passes which only exist in mlir-aie. I would bring the tests in if/when the pass being tested lives in iree-amd-aie.

Because they will serve as unit tests for the forthcoming PRs that vendor all the mlir-aie passes and refactor them. See https://github.com/nod-ai/iree-amd-aie/pull/425, https://github.com/nod-ai/iree-amd-aie/pull/426, https://github.com/nod-ai/iree-amd-aie/pull/427 (which aren't ready for review).

compiler/plugins/target/AMD-AIE/aie/AIEAssignBufferAddressesBasic.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+  void getDependentDialects(DialectRegistry &registry) const override {
+    registry.insert<func::FuncDialect>();
+    registry.insert<AIEDialect>();
+  }
+
+  void runOnOperation() override {
+    DeviceOp device = getOperation();
+    int counter = 0;
+    device.walk<WalkOrder::PreOrder>([&](BufferOp buffer) {
```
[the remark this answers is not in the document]
makslevental: Since this will all go away in 4 PRs, I'm gonna leave as is.

compiler/plugins/target/AMD-AIE/aie/AIEAssignBufferAddressesBasic.cpp
makslevental: Why not indeed; this has no point in this context (when the buffers aren't freed/reused). Elsewhere this heuristic has some sense (see "Greedy by Size" [here](https://arxiv.org/abs/2001.03288)) but here simple "bump allocation" is sufficient (I've changed to bump allocation).

compiler/plugins/target/AMD-AIE/aie/CMakeLists.txt
```diff
@@ -186,3 +186,89 @@ iree_tablegen_library(
...
+    "AIEXTransformPasses.cpp"
+    "${IREE_MLIR_AIE_SOURCE_DIR}/lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp"
+    "${IREE_MLIR_AIE_SOURCE_DIR}/lib/Dialect/AIEX/Transforms/AIEXToStandard.cpp"
+  DEPS
+    ::defs
+    ::AIEXDialectIR
+    ::AIEXTransformPassHeaders
+)
+
+add_subdirectory(test)
```
nirvedhmeshram, line 274: nit: newline

newling (approved):
LGTM, thanks for the clarifications and changes. Note of reference: most of this code is temporary so I'm not reviewing it.

PR #424, opened 2024-06-16
Vendor all target emission utils (6/n)

makslevental:
This PR is part of a stack of PRs that refactor the dependency on MLIR-AIE. See https://github.com/nod-ai/iree-amd-aie/issues/430 for more information.

This PR vendors all necessary target emission utils that we need (and none that we don't 😄). Namely `AIETargetBCF`, `AIETargetCDODirect`, `AIETargetLdScript`, and `AIETargetNPU`. Just as for the previous "vendoring" PR, it's not worth it to bikeshed the code yet because it'll be refactored in a subsequent PR (e.g., `AIETargetNPU` will disappear and be absorbed into `AIEDmaToNpu`).

@nirvedhmeshram @newling this is now ready for review.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETargetBCF.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETargetBCF.cpp
@@ -1,0 +1,135 @@
+// Copyright 2024 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "AIETargets.h"
+#include "aie/Dialect/AIE/IR/AIEDialect.h"
+#include "aie/Passes.h"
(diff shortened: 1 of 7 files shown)


newling (approved):

nirvedhmeshram:
This may be a nit level comment but something to think on in the subsequent PRs. I think for a new reader it will be hard to demarcate which files relate to top level IREE targets like AIETarget and AIETargetDirect and the rest of the low level targets. Can we distinguish by changing directory structure or maybe changing the top level target names to AMDAIETarget* ?

makslevental:
> This may be a nit level comment but something to think on in the subsequent PRs. I think for a new reader it will be hard to demarcate which files relate to top level IREE targets like AIETarget and AIETargetDirect and the rest of the low level targets. Can we distinguish by changing directory structure or maybe changing the top level target names to AMDAIETarget* ?

Ya it's true (I agree). My plan is to roll all of that utility code into `AIETargetDirect` itself. I haven't done it in the higher up "refactor" PR because I'm still weighing pros/cons so maybe as a stopgap (until we resolve pros/cons) I will indeed rename.

PR #428, opened 2024-06-17
Change tiling and packing sizes to make outer dimension 2x2

yzhang93:
- This PR changes the packing strategy to offload the large inner tile sizes to outer dimensions. This avoids the large allocations in L1, and is required in the objectFifo pipeline `DistributeCoresAndObjectFifos` pass.
- This PR also bumps the MLIR-AIR to include a fix to support the new packing strategy.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
@@ -561,7 +561,13 @@ void addMLIRAIRAIELoweringPasses(OpPassManager &passManager, bool packPeel) {
   passManager.addPass(xilinx::air::createAIRLoweringPass());
   {
     xilinx::air::AffineLoopOptPassOptions options;
-    const std::vector<unsigned> tile_sizes = {2, 2};
+    std::vector<unsigned> tile_sizes;
+    if (packPeel) {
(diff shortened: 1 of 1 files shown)


Abhishek-Varma (approved):
LGTM! I'll let @erwei-xilinx review this too.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -561,7 +561,7 @@ void addMLIRAIRAIELoweringPasses(OpPassManager &passManager, bool packPeel) {
   passManager.addPass(xilinx::air::createAIRLoweringPass());
   {
     xilinx::air::AffineLoopOptPassOptions options;
-    const std::vector<unsigned> tile_sizes = {4, 4};
+    const std::vector<unsigned> tile_sizes = {2, 2};
```
erwei-xilinx, line 564: I think while {2,2} works with both pad-pack and pack-peel pipelines through AIR, {4,4} can also work if pad-pack; pack-peel requires {2,2}. 

{4,4} is faster due to having less NPU instructions. So it might be beneficial for performance if we do a condition here between pad-pack and pack-peel, so that pad-pack can still use {4,4} for now.

erwei-xilinx (approved):
Thanks. Looks great to me.

PR #443, opened 2024-06-18
Prepare for testing non-matmul workloads

newling:
Addresses the main aspect of https://github.com/nod-ai/iree-amd-aie/pull/394 that I found potentially difficult to use, namely that a single function (run_test) did 2 different things: supports matmul specific mlir file generation via flags like --m --n --k, but also accepts generic mlir files which have nothing to do with matmul. This PR makes this test infrastructure easier to scale IMO. See discussions in https://github.com/nod-ai/iree-amd-aie/pull/394


nirvedhmeshram (approved):
LGTM!

PR #444, opened 2024-06-18
Hoist the bias allocation

newling:
The elementwise input is promoted to local memory after the pass to hoist statically bound allocations. It is the only tensor which is promoted after the hoisting pass. This PR is just a tidy-up, so that all of the allocations are hoisted before bufferization.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
@@ -269,7 +269,7 @@ void addPackPeelBasedPassPipeline(OpPassManager &funcPassManager,
     funcPassManager.addPass(createAMDAIEFusePackIntoLoopPass(fusePackOptions));
   }
 
-  // Promote the elementwise input to local memory
+  // Promote the elementwise input to local memory, and hoist the allocation.
   {
     AMDAIEBufferizeToAllocationOptions bufferizeOptions;
     bufferizeOptions.memorySpace = 2;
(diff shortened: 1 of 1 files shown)


yzhang93:
> The elementwise input is promoted to local memory after the pass to hoist statically bound allocations. It is the only tensor which is promoted after the hoisting pass. This PR is just a tidy-up, so that all of the allocations are hoisted before bufferization.

It is not the only tensor that is promoted after the first hoisting pass. There are cases that the allocation is created during comprehensive-bufferization https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp#L76. So the hoisting pass should be kept after bufferization pass.

yzhang93:
@newling I think you edit my comment instead of replying it. My point is there is new allocation created during comprehensive-bufferization(https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp#L76), so in any case we need to do hoisting after comprehensive-bufferization. https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp#L294. What you added is redundant.

newling:
> @newling I think you edit my comment instead of replying it. My point is there is new allocation created during comprehensive-bufferization(https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp#L76), so in any case we need to do hoisting after comprehensive-bufferization. https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp#L294. What you added before bufferization is redundant.

As I replied before, I understand that there is allocation during comprehensive bufferization and that this PR is redundant in the sense that this bias allocation will be lowered later, my point is that the IR is more complex without it (most allocations hoisted, 1 which hasn't, until comprehensive bufferization). But I will close as you don't seem to agree :)

PR #450, opened 2024-06-20
Adding error message whenever the xclbin tool fails to produce xclbin

josemonsalve2:
I noticed that whenever the xclbin compilation process fails (say that iree is not able to find the binary), it just fails but it is hard to find the reason. 

This just makes sure that the error message is displayed as well.


nirvedhmeshram (approved):
LGTM!

Issue #447, opened 2024-06-19
Performance and tracing issues when looking at multi dispatch execution


nirvedhmeshram:
I tried running [this](https://gist.github.com/nirvedhmeshram/784bfb71619187cd70ea932749067bc8) sequence of matmuls IR through the amd-aie hal target and XRT driver. The following [trace](https://sharkpublic.blob.core.windows.net/sharkpublic/nirvedh/xrt_detailed.tracy) was obtained.
There are two observed issues
1. First dispatch took 10 ms (which is expected as this is not on the vectorization path) but the second and third dispatch took 16s each which seems to be a big performance issue.
2. The observation in 1.  was done by looking at the `iree_hal_xrt_direct_command_buffer_dispatch` stats however when we start analyzing real world models we would want to have dispatch specific information about this, when observing traces on cpu backend I do see this information so why is it not there for amd-aie? I did include the following flag `--iree-hal-executable-debug-level=3`in my testing.

nirvedhmeshram:
@ScottTodd maybe you have some thoughts on point 2. I am wondering if its because of it being a external plugin, its a problem?

ScottTodd:
See https://github.com/iree-org/iree/pull/16757 . Executables need to track any debug information they want to be available at runtime, then the runtime needs to tell tracing about that debug information.

fifield:
> 1. First dispatch took 10 ms (which is expected as this is not on the vectorization path) but the second and third dispatch took 16s each which seems to be a big performance issue.

16s is suspicious because it was ([until yesterday](https://github.com/amd/xdna-driver/pull/126)) equal to a timeout in the driver, meaning you may have a hanging kernel: https://github.com/amd/xdna-driver/pull/51, https://github.com/amd/xdna-driver/blob/f839c4e1de4eb790d5307253649e18912e15f8e2/src/driver/amdxdna/aie2_ctx.c#L21

nirvedhmeshram:
> > 1. First dispatch took 10 ms (which is expected as this is not on the vectorization path) but the second and third dispatch took 16s each which seems to be a big performance issue.
> 
[...]

PR #453, opened 2024-06-24
Fix bug in artifact creation of the LX6 instructions

nirvedhmeshram:
There was a bug where were appending LX6 instructions to an existing vector for subsequent entry points rather than making a new vector for each entry point. This change fixes that and hence fixes https://github.com/nod-ai/iree-amd-aie/issues/447 which was indeed a kernel time out due to bad artifacts.
Also adds a multi-dispatch e2e test to CI.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
@@ -253,7 +253,6 @@ LogicalResult AIETargetBackend::serializeExecutable(
 
   SmallString<128> aie2xclbin(options.mlirAieInstallDir);
   llvm::sys::path::append(aie2xclbin, "bin", "aie2xclbin");
-  std::vector<uint32_t> npuInstrs;
   std::unique_ptr<llvm::MemoryBuffer> xclbinIn;
 
   FlatbufferBuilder builder;
@@ -359,6 +358,8 @@ LogicalResult AIETargetBackend::serializeExecutable(
 
     std::ifstream instrFile(static_cast<std::string>(npuInstPath));
     std::string line;
+    // Vector to store LX6 instructions.
+    std::vector<uint32_t> npuInstrs;
     while (std::getline(instrFile, line)) {
       std::istringstream iss(line);
       uint32_t a;
--- a/build_tools/ci/cpu_comparison/run_test.sh
+++ b/build_tools/ci/cpu_comparison/run_test.sh
@@ -326,6 +326,9 @@ function run_test() {
 # Example of running a test directly from an .mlir file with a function.
 run_test --test_file ${THIS_DIR}/test_files/matmul_int32.mlir
 
+# An example of an arbitrary graph with three matmuls which form three dispatches.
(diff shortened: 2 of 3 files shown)


makslevental:
Is this *the* kernel lag issue we've all been facing? I'm wondering how I was seeing the seeing the same thing even though I wasn't reusing the same buffer/vector in my own path (in xaiepy).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -359,6 +358,8 @@ LogicalResult AIETargetBackend::serializeExecutable(
 
     std::ifstream instrFile(static_cast<std::string>(npuInstPath));
     std::string line;
+    // Vector to store LX6 instructions.
```
newling, line 361: FYI: I've heard that AIE2p has LX7 instructions (not a request for change!)

newling (approved):
Nice find!

In the future we could add a test that there is exactly 1xclbin generated for a test like this, I can try and add that at some stage.

nirvedhmeshram:
> Is this _the_ kernel lag issue we've all been facing? I'm wondering how I was seeing the seeing the same thing even though I wasn't reusing the same buffer/vector in my own path (in xaiepy).

I think that might be a different issue, this one is only hit when we have multiple dispatches in one executable in iree xrt driver.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
makslevental: I believe it's true that forthcoming gens will have LX7 (and higher?) but the instructions themselves don't actually have anything to do with the LX arch (which is Tensilica's naming scheme for their archs and not connected to our firmware that runs _these_ instructions). But you're right the comment will be out of date.

nirvedhmeshram:
> Nice find!
> 
> In the future we could add a test that there is exactly 1xclbin generated for a test like this, I can try and add that at some stage.

Yes that would be useful, there is one challenge though, we will create multiple XCLBINs as we create them first and then merge it into the previous one, so the old xclbin artifacvt is still present even though we wont use it at runtime. 

Btw jut to be clear, this test currently will produce three xclbins and use all of them at runtime, we need something like https://github.com/nod-ai/iree-amd-aie/pull/418 for it to use 1 xclbin at runtime. As I said previously it will still produced three XCLBIN's at compile time.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
nirvedhmeshram: Lets have a clean up PR for this, I noticed you did LX[0-9] in some places, maybe we should say LX6/LX7 instead?

makslevental: I think the safest term would be `command processor` but who cares (honestly when I first started it was useful knowing that it was LX6 because there's a datasheet type thing out there for that so it clarified what people were really talking about).

PR #458, opened 2024-06-25
Fix for extra qualification C++ compiler warning

newling:
I would've liked to add Werror to CI, but it wasn't obvious what the best/canonical way to do this is, so for now just fixing this warning.


makslevental:
Thanks for catching this (sorry). What alarm did this trip? I didn't notice...

makslevental (approved):
Thanks

newling:
> Thanks for catching this (sorry). What alarm did this trip? I didn't notice...

no worries I think this one is specific to my version of clang++, I didn't see the warning in CI. The only warning in CI is from mlir-air.

PR #420, opened 2024-06-16
Add `amd-aie-direct` HAL target (3/n)

makslevental:
This PR is part of a stack of PRs that refactor the dependency on MLIR-AIE. See https://github.com/nod-ai/iree-amd-aie/issues/430 for more information.

This PR adds a new target `amd-aiet-direct`, directly references AIEVec passes (for eventual lowering of core code to chess LLVM IR) and also (simultaneously) vendors `XCLBinGen.cpp` from `mlir-aie` (second part used to be broken out in https://github.com/nod-ai/iree-amd-aie/pull/422 but on @newling's suggestion I squashed). Note, `XCLBinGen` and `AIETargetDirect` are basically carbon copies of what currently exists except `AIETargetDirect` doesn't shell out and calls `aie2xclbin` directly. Note also that `XCLBinGen` will undergo further revision up to for the chess backend (but that'll have to wait until I get home).

To test, I added some lines to `print_ir_aie2xclbin.sh`. I'm happy to bike shed this but probably not too hard.

PR #459, opened 2024-06-25
Transition AIETarget to use internal aie2xclbin function

nirvedhmeshram:
Since https://github.com/nod-ai/iree-amd-aie/pull/420 added a internal utility to go from AIE to XCLBIN it makes sense for all flows to use that.
This utility doesn't add any AIE lowering passes so the `TranslationPassPipeline`s are responsible for that. Hence the passes needed for the IR lowered through the AIR flow to be further lowered to AIE are added in its lowering pipeline.

There were some minor additions needed to the utility to get it working e2e for the AIE flow.

We no longer need to have a separate `aie2xclbin-disable-threading` as we are passing the existing context and the `mlir-disable-threading` flag gets automatically propagated in this case. And it actually  gives an error to try changing it if the passed context is multi-threaded.

With this PR the AIE lowering is becoming a part of the IREE `TranslationPassPipeline` so the pack-peel lit test that was just passing the air-to-aie conversion had to be watered down to reflect what can actually be lowered to AIE. 

Unfortunately, we still also need the mlir-aie wheel as the internal utility is still expecting a mlir-aie install for me_basic.o library [here](https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp#L302-L304)


makslevental:
> Unfortunately, we still also need the mlir-aie wheel as the internal utility is still expecting a mlir-aie install for me_basic.o library [here](https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp#L302-L304)

The thing is that `me_basic.o` doesn't really make sense, [at least for CI](https://github.com/Xilinx/mlir-aie/blob/main/aie_runtime_lib/AIE2/me_basic.c#L96-L97): 

```c++
static inline void _init() {
  // constructors are called in reverse order of the list
  for (thunk *t = &_ctors_end; t-- != chess_copy(&_ctors_start);)
    (*t)();
}

void _fini() {
  // destructors in forward order
  for (thunk *t = &_dtors_start; t != chess_copy(&_dtors_end); ++t)
    (*t)();
}
```

So like even though I only vaguely understand what this code accomplishes (constructors? destructors?) I'm pretty it can't possibly apply to our CI executions because that's through Peano. How the linking still succeeds I have no idea either (possibly symbol resolution is deferred to runtime and then since there is no runtime linker on the single cores and the code path is never taken there's no issue 🤷). Regardless, we should adapt LDScript to be more flexible wrt the inclusion of `me_basic.o` (which we now have the freedom to do...) depending on whether we're using the chess backend or Peano backend. I'm not fluent enough in LD linker scripts to do it blind (without access to an E2E smoketest on my local machine) so while I'm happy to make the change, it'll have to wait until ~Monday.

nirvedhmeshram:
looks like more lowering passes from AIE are needed (I tried a minimal set first) but it fails for large problem sizes.

compiler/plugins/target/AMD-AIE/aie/CMakeLists.txt
```diff
@@ -304,6 +304,8 @@ iree_cc_library(
     "AIEObjectFifoStatefulTransform.cpp"
     "AIEDmaToNpu.cpp"
     "AIEXToStandard.cpp"
+    # Passes needed by AIR-AIE lowering
+    "${IREE_MLIR_AIE_SOURCE_DIR}/lib/Dialect/AIE/Transforms/AIENormalizeAddressSpaces.cpp"
```
makslevental, line 308: Where did you find you needed this pass? I couldn't find any tests that failed without it? Anyway can you bring it into our source tree like the others.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -581,6 +584,17 @@ void addMLIRAIRAIELoweringPasses(OpPassManager &passManager, bool packPeel) {
...
+  // Now lower using the AIE passes used by the mlir-air/mlir-aie flow.
+  passManager.addPass(createLowerAffinePass());
+  OpPassManager &devicePM = passManager.nest<xilinx::AIE::DeviceOp>();
+  devicePM.addPass(createAIEAssignLockIDsPass());
+  devicePM.addPass(createAIEObjectFifoStatefulTransformPass());
+  devicePM.addPass(createAIEAssignBufferDescriptorIDsPass());
+  devicePM.addPass(createAIEAssignBufferAddressesBasicPass());
+  devicePM.addPass(createAIEPathfinderPass());
+  devicePM.addPass(createAIELocalizeLocksPass());
+  passManager.addPass(createConvertSCFToCFPass());
```
makslevental, line 597: In order to keep the two paths separate (until we're ready to shift to using only `amd-aie-direct`) you can't actually use the in-tree passes. To put it simply: if you put this here today, tomorrow I will have to revert this commit because tomorrow (or very soon) I will be removing/deleting those passes.

makslevental: So it's very awkward I know but if you want to do this you basically need to restore the CMakes as done here https://github.com/nod-ai/iree-amd-aie/pull/421/files and then write `devicePM.addPass(AIE::createAIEPathfinderPass());` etc.

makslevental:
> looks like more lowering passes from AIE are needed (I tried a minimal set first) but it fails for large problem sizes.

```
<unknown>:0: error: 'aie.tile' op allocated buffers exceeded available memory (99328>65536)
```

This is (I believe but @jtuyls can authoritatively answer) some fail in objectfifostatefultransform rather than a lack of a pass (indeed the data memory on a tile is 64K and so whoever/whomever/whatever has delegated that much work to some tile is mistaken).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
nirvedhmeshram: wait, I think these are in tree passes no?

compiler/plugins/target/AMD-AIE/aie/CMakeLists.txt
nirvedhmeshram: This is needed in  the llvm conversion, otherwise the buffers have address spaces on them which peano doesnt like, will in tree this.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
nirvedhmeshram: ah you dont want me to use in-tree got it. then I have to make my own lib for it..

makslevental: Yes what you currently have written are calls to the "creators" of the in-tree passes and what I'm saying is tomorrow (soon) those will no longer exist and this code will break.

compiler/plugins/target/AMD-AIE/aie/CMakeLists.txt
nirvedhmeshram: is this one okay to use from in-tree with you though?

makslevental: Well I left this comment before understanding what you're doing so the answer is no (not because it's okay/not-okay but for the reason I give below). But also I don't actually think you need this? I tested extensively (included E2E with numerics) and I never needed this pass (which does nothing but set a memory address space attribute which is never then queried/used in a subsequent analysis).

nirvedhmeshram: Hmm, in my testing, peano was failing to lower to assembly code without this pass due to the memory addressspace difference when not using this pass. That being said my peano is I believe 3 months old (didnt update it since it wasnt giving me issues) so once I get the CI green, I will remove this pass and if it stays green after then that means it was my old peano issue.

nirvedhmeshram:
> > looks like more lowering passes from AIE are needed (I tried a minimal set first) but it fails for large problem sizes.
> 
> ```
> <unknown>:0: error: 'aie.tile' op allocated buffers exceeded available memory (99328>65536)
> ```
> 
> This is (I believe but @jtuyls can authoritatively answer) some fail in objectfifostatefultransform rather than a lack of a pass (indeed the data memory on a tile is 64K and so whoever/whomever/whatever has delegated that much work to some tile is mistaken).

I think I know whats going on, there is a large size peel-pack test that worked till iree translation but not in the actual AIE lowering, now that the AIE lowering is moving into the iree translation we have problems. I will test this hypothesis by adding one of the failing tests to CI and seeing what happens.

nirvedhmeshram:
Yup, I think this failure is expected when we are moving the translation pipeline to lower through AIE,
https://github.com/nod-ai/iree-amd-aie/actions/runs/9683545497/job/26719580993?pr=463#step:6:4990

compiler/plugins/target/AMD-AIE/aie/CMakeLists.txt
nirvedhmeshram: Ya, its happening on CI, so I will keep the pass, open to opinions on where you want it for now though
https://github.com/nod-ai/iree-amd-aie/actions/runs/9686354878/job/26729179745#step:6:268

nirvedhmeshram:
@newling @makslevental this is ready for review.

compiler/plugins/target/AMD-AIE/aie/CMakeLists.txt
makslevental: Just if you can do what I originally proposed that would be good (just copy the file to the same place as the rest of the things and update the header). But I'll just reiterate: if you actually want to make this happen then you should not use those same passes because I will delete/coalesce/condense them very soon (next week when I get back).

nirvedhmeshram: I am confused, you want me to move the AIENormalizeAddressSpaces passes with the rest of your passes but then you are saying you may delete it when you refactor? I do want this path to be stable..

nirvedhmeshram: I made my own library with the "upstream" mlir-aie passes that includes the AIENormalizeAddressSpaces pass.

makslevental: > I am confused, you want me to move the AIENormalizeAddressSpaces passes with the rest of your passes but then you are saying you may delete it when you refactor?

Well I was saying that 

1. If you want to continue to use my passes then yes move over `AIENormalizeAddressSpaces`
2. But don't use my passes because they'll go away

> I made my own library with the "upstream" mlir-aie passes that includes the AIENormalizeAddressSpaces pass.

Yup that's good (I'll move over `AIENormalizeAddressSpaces` to my dir myself soon).

nirvedhmeshram: Cool, thanks, makes sense!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -581,6 +586,19 @@ void addMLIRAIRAIELoweringPasses(OpPassManager &passManager, bool packPeel) {
 
   passManager.addPass(xilinx::airrt::createAIRRtToNpuPass());
   passManager.addPass(createCanonicalizerPass());
+
+  // Now lower using the AIE passes used by the mlir-air/mlir-aie flow.
+  passManager.addPass(createLowerAffinePass());
+  OpPassManager &devicePM = passManager.nest<xilinx::AIE::DeviceOp>();
```
newling, line 592: Can the other aie2xclbin flags also be removed now? aie2xcbin-print-ir-before-all, etc.

build_tools/ci/print_ir_aie2xclbin/print_ir_aie2xclbin.sh
```diff
@@ -131,7 +131,6 @@ ${SOURCE_MLIR_FILE} \
 --aie2xclbin-print-ir-after-all \
```
newling, line 131: I suspect that this flag is not actually doing anything: as the flag `--mlir-print-ir-after-all` is also passed, the printing is happening still.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -12,6 +12,7 @@
 
 #include "AMDAIETargets.h"
 #include "aie/Conversion/AIEVecToLLVM/AIEVecToLLVM.h"
```
newling, line 14: Here or later, we should absorb this change : https://github.com/Xilinx/mlir-aie/pull/1583 (related to https://github.com/nod-ai/iree-amd-aie/issues/467)

newling (changes requested):
Thanks! Only the one comment about the other aie2xclbin-* flags, which might also be removable. 

At some point we can completely remove large parts (or all) of print_ir_aie2xclbin.sh

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
nirvedhmeshram: I thought about this, but no, only the context is shared (which is what the threading is lin ked to). The other passes use a new pass manager so the flags have to be set separately for them. Of course there might be a way to parse the top level flags but thats something to think on seperately.

build_tools/ci/print_ir_aie2xclbin/print_ir_aie2xclbin.sh
nirvedhmeshram: I did try this and the printing for the aie2xclbin passes didnt happen without the flag.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
nirvedhmeshram: Ya those changes should just be made on the files we have moved into iree-amd-aie. Not the scope of this PR though.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ iree_cc_library(
     iree::target::amd-aie::aie::AIEDialectIR
     iree::target::amd-aie::aie::AIEXDialectIR
     iree::target::amd-aie::aie::AIEPasses
+    iree::target::amd-aie::aie::AIEPassesFromMLIRAIE
```
makslevental, line 27: Can you remove this - `AIETargets` has `iree::target::amd-aie::aie::AIEPasses` which are the "in-tree" versions (which you already realized/know I'm sure).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -686,6 +685,8 @@ static LogicalResult generateUnifiedObject(MLIRContext *context,
 
   pm.addNestedPass<AIE::DeviceOp>(
       mlir::iree_compiler::AMDAIE::createAIELocalizeLocksPass());
+  pm.addNestedPass<AIE::DeviceOp>(
+      xilinx::AIE::createAIENormalizeAddressSpacesPass());
```
makslevental, line 689: Can you remove this - I'll readd when I bring over `AIENormalizeAddressSpacesPass`.

makslevental (commented):
I'm not gonna approve since James already asked for changes (not sure how that conflict resolution works...) but basically looks good to me modulo the two small things I commented on.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
nirvedhmeshram: That will fail CI,

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/CMakeLists.txt
nirvedhmeshram: I need it for AIENormalizeAddressSpacesPass

nirvedhmeshram:
> I'm not gonna approve since James already asked for changes (not sure how that conflict resolution works...) but basically looks good to me modulo the two small things I commented on.

I cant make those exactly so let me know if you want me to bring in the pass myself.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/CMakeLists.txt
makslevental: OH. Whoops. Sorry you're right. It's unfortunate that we've "crossed the streams here" but okay we'll live with it.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
newling: Makes sense, future work.

newling (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/CMakeLists.txt
nirvedhmeshram: I was able to move it and localize locks pass into the translation pipeline. So atleast a little bit uncrossed :)

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
nirvedhmeshram: done (by moving elsewehre)

nirvedhmeshram:
I made two changes post approval so calling those out. Moved a couple of passes out of aie2xclbin utlity and into the translation pipeline. The reason for this was that in @makslevental 's flow those are not required. 
I renmaed the top level pipelines to demarcate the mlir-air and mli-aie pass pipelines, this way if the objectfifo work can just directly use the mlir-aie pipeline if it wants to.

nirvedhmeshram:
ouch the pass movment breaks microkernels.

nirvedhmeshram:
ok it was a pass ordering issue and wasn't directly related to micro-kernels.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -12,6 +12,7 @@
 
 #include "AMDAIETargets.h"
 #include "aie/Conversion/AIEVecToLLVM/AIEVecToLLVM.h"
+#include "aie/Dialect/AIE/Transforms/AIEPasses.h"
```
makslevental, line 15: nit: you can now remove this include

PR #460, opened 2024-06-25
Merge TargetNPU and DmaToNPU

makslevental:
This PR merges `AIEDmaToNPU` and `AMDAIETargetNPU` and  i.e., the pass that lowers `NpuDmaMemcpyNdOp` and etc and the actual "translate" utility that emits the NPU instructions. Moreso than just consolidating code, it begins the process of migrating to a more sane "runtime model" for the NPU instructions. 

What am I talking about? Instead of emitting the NPU instructions as a list/vector of `uint32_t`s (what the translation utility does/did), this PR does 

`device->setAttr("npu_instructions", DenseI32ArrayAttr::get(&getContext(), npuInstructions));`

i.e., stows the npu instructions on the device op itself. It looks like this:

```
aie.device(npu1_4col) {
  func.func @sequence() {
    return
  }
} {npu_instructions = array<i32: 100860160, 261, 2, 64, 0, 0, 119308, 0, -2147483645, 24, 0, 0, 67228180, 0, 196610, 24>}
```

Now this is a little different from how the various `gpu` devices work (where the instructions/ptx/etc are cached on a `gpu.object` op that's in the region/body of a `func` or something) but I opted to put it on the outer `aie.device` op purely as a MVP because it's relatively easy to retrieve from the caller[^1]. 

Feel free to bikeshed this and tell me you'd prefer the `npu_instructions` live in a `gpu.object` op (preferrably actually `gpu.object` because I prefer to reuse/recycle existing upstream ops/dialects etc rather than growing/expanding `AMDAIE` dialect). In particular maybe there's some "command buffer" attribute or op or something in IREE that's a more approriate home for this sequence of instructions? A while ago I was chit-chatting with @benvanik about relocatable command buffers in IREE but I haven't actually gotten around to finding/figuring them out.

[^1]: Where currently the caller does end up dumping the npu instructions to a text file that is then slurped up at runtime by XRT glue code (somewhere in [runtime/src/iree-amd-aie/driver/xrt](https://github.com/nod-ai/iree-amd-aie/tree/main/runtime/src/iree-amd-aie/driver/xrt) I think) but that'll change as soon as I start my next sprint of "refactoring" XRT 😄.


compiler/plugins/target/AMD-AIE/aie/AIEDmaToNpu.cpp
```diff
@@ -356,6 +480,39 @@ struct AIEDmaToNpuPass : ::impl::AIEDmaToNpuBase<AIEDmaToNpuPass> {
...
+    patterns.insert<ConvertNpuOp<NpuSyncOp>>(&getContext(), instructions,
+                                             appendSync, count);
+    patterns.insert<ConvertNpuOp<NpuWrite32Op>>(&getContext(), instructions,
+                                                appendWrite32, count);
+    patterns.insert<ConvertNpuOp<NpuAddressPatchOp>>(
+        &getContext(), instructions, appendAddressPatch, count);
+    patterns.insert<ConvertNpuOp<NpuWriteBdOp>>(&getContext(), instructions,
+                                                appendWriteBdShimTile, count);
+
+    if (failed(applyPartialConversion(device, target, std::move(patterns))))
```
newling, line 507: I think I prefer a non-pattern approach for this. It's not clear here what order the instructions get added (what order the do the patterns get run? I assume it is deterministic, but still). I'm actually surprised this change works with the end-to-end tests, I guess that order in which the pattern manager applies the patterns just happens to be correct for the use case?  

I like the idea of avoiding the llvm::switch that the previous impl had, maybe you can find a solution somewhere between the 2 approaches? We've also had a few discussions in the team with recent PRs about when pattern approaches are appropriate, and when a simple pass if good enough. The general consensus is that if you don't need a fixed-point algorithm (i.e. iterate more than once through the IR until no changes), then we should use a simple pass.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -828,45 +828,27 @@ LogicalResult xilinx::aie2xclbin(MLIRContext *ctx, ModuleOp moduleOp,
...
+          .asArrayRef();
+  std::vector<uint32_t> unsignedNpuInstructions(
+      signedNpuInstructionsAttr.begin(), signedNpuInstructionsAttr.end());
 
-    output->keep();
-    copy->erase();
-  }
+  std::string errorMessage;
+  auto output = openOutputFile(OutputNPU, &errorMessage);
+  if (!output) return moduleOp.emitOpError(errorMessage);
```
newling, line 848: Is the plan that this write-to-file will be removed too, eventually?

newling (commented):
Looks good, I think it makes sense to add the instructions as IR on device op (for now, and maybe somewhere else in IR in the future). 

Question/comment about the use of patterns.

compiler/plugins/target/AMD-AIE/aie/AIEDmaToNpu.cpp
```diff
@@ -24,16 +27,171 @@ using namespace xilinx::AIEX;
...
+    count++;
+    return success();
+  }
+};
 
 // Helper class to get a ShimDMAAllocationOp for a given <device, symbol name>
 // pair. An object of this class is invalidated if, for any symbol_name, a
-// ShimDMAAllocationOp that uses it changes, as the cache is not updated in this
-// case.
+// ShimDMAAllocationOp that uses it changes, as the cache is not updated in
```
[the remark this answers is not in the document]
makslevental: I never touch comments by hand (I don't write comments...) so it must be the `clang-format`.

compiler/plugins/target/AMD-AIE/aie/AIEDmaToNpu.cpp
makslevental: The issue is you need `ConversionPattern` in order to be able to have the `eraseOp` actually be applied (or something like that - I always get confused by `ConversionPattern` and it's actually being rewritten [upstream](https://discourse.llvm.org/t/rfc-a-new-one-shot-dialect-conversion-driver/79083)). I had to fiddle to the right thing to happen (the `eraseOp`). 

> I guess that order in which the pattern manager applies the patterns just happens to be correct for the use case?

There's no ordering implied here from the perspective of mlir-aie - it's another one of those spots in AIE dialect where it's just a 1-1 mapping (to register writes). 

**Underlining something here**: there might be an ordering necessitated by the architecture/device itself (e.g., maybe DMAs have to be programmed before locks? or vice-versa?). I have run into issues about ordering stuff (don't enable the cores before starting the DMAs...). **But 100%** whatever order you see in mlir-aie is completely accidental or at best inherited by osmosis from the external examples (some other team) the original code was copy-pasted from. 

Back to mlir itself: in principle the order in which these are applied is determined by the `benefit` (I believe). I don't have a strong opinion on it - I can see why you would prefer something that was more explicitly sequenced but on the other hand there's some (*some*) value in the declarative specification.

> The general consensus is that if you don't need a fixed-point algorithm (i.e. iterate more than once through the IR until no changes), then we should use a simple pass.

Yea I agree with this. Let me try to figure out if I can force the correct `eraseOp` behavior without relying on a `ConversionPattern`.

jtuyls:
> Feel free to bikeshed this and tell me you'd prefer the npu_instructions live in a gpu.object op (preferrably actually gpu.object because I prefer to reuse/recycle existing upstream ops/dialects etc rather than growing/expanding AMDAIE dialect).

In general I think it would be great to align AMDAIE more with upstream dialects like the GPU dialect, for example `gpu.module`/`gpu.launch` instead of the `aie.device` with the nested `func.func`. However I am not sure about completely reusing some of the dialect's operations/attributes unless those ops/attrs are widely applicable for AIE as well. My concern is that I think a lot of the core GPU dialect's operations are not applicable for AIE and it would be unclear/weird to interleave a few GPU ops/attrs between AMDAIE ops, especially when they are very simple like `gpu.object`. Just some thoughts, would love to hear other people's take on this as well.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
makslevental: Yes. Looking ahead, we will do this the same exact way to that GPU dialects do this - the runtime calls themselves are emitted into the IR (as `func.call` -> `llvm.call` -> C FFI calls) that is then executed for real real at runtime. So we will just emit a call somewhere (my rough/hazy vision is that IREE `hal` dialect/dispatch region will do this) that passes this sequence of instructions directly to some equivalent of `xrt::bo(npu_instructions, XCL_BO_CACHEABLE)`. @nirvedhmeshram I believe has been "hands-on" keyboard for the XRT connection so he can comment/paint a picture here.

compiler/plugins/target/AMD-AIE/aie/AIEDmaToNpu.cpp
newling: Ok, makes sense now that the order doesn't matter, if it's just register setting (I was confused, thought this was lx6 instructions...)

Sometimes it helps to collect all the ops to erase in a vector while walking the IR, and then remove them as a subsequent step (you've probably seen this design before...). Pointer confusion somewhere, or something.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
newling: That's a clear explanation, thanks

compiler/plugins/target/AMD-AIE/aie/AIEDmaToNpu.cpp
```diff
@@ -24,16 +27,171 @@ using namespace xilinx::AIEX;
...
+  instructions.resize(newSize, 0);
+  return llvm::MutableArrayRef<uint32_t>(instructions.data() + oldSize,
+                                         tailSize);
+}
+
+void appendSync(std::vector<uint32_t> &instructions, NpuSyncOp op) {
+  auto words = reserveAndGetTail(instructions, 4);
+
+  // XAIE_IO_CUSTOM_OP_TCT
+  words[0] = TXN_OPC_TCT;
```
jtuyls, line 53: What about using `aie-rt` to create the transactions instead of doing it manually?

compiler/plugins/target/AMD-AIE/aie/AIEDmaToNpu.cpp
makslevental: > I was confused, thought this was lx6 instructions

Here is your zen moment: the bulk of LX6 instruction are literally just `write32`, `read32`, `maskwrite32`. So these *are* LX6 instructions.

makslevental:
> However I am not sure about completely reusing some of the dialect's operations/attributes unless those ops/attrs are widely applicable for AIE as well. My concern is that I think a lot of the core GPU dialect's operations are not applicable for AIE and it would be unclear/weird to interleave a few GPU ops/attrs between AMDAIE ops, especially when they are very simple like `gpu.object`. Just some thoughts, would love to hear other people's take on this as well.

Yea I'll admit I'm not completely sure about the idea either - I would prefer to reuse absolutely as much upstream stuff as possible in order to prevent the "ghettoization"[^1] of AIE, but you're right we shouldn't force it.

[^1]: don't get mad, it's a real, [technical term](https://dictionary.cambridge.org/dictionary/english/ghettoize).

compiler/plugins/target/AMD-AIE/aie/AIEDmaToNpu.cpp
makslevental: So you and I chit-chatted about this right but I actually never saw it all the way through - does [all the transaction stuff in`aie-rt`](https://github.com/Xilinx/aie-rt/blob/1ad203de0b7f282b1c0659fd2ae3f218652c7274/driver/src/global/xaiegbl.h#L579-L585) actually refer to the transaction interpreter in the FW? But that's a good reminder - I'll check it out.

compiler/plugins/target/AMD-AIE/aie/AIEDmaToNpu.cpp
newling: Yeah... makes sense. 

So I'm guessing one of the LX6 instructions is the one which toggles a lock somewhere, setting the AIE in motion, tick tocking away until we've got the result of our kernel. 

And it's important that that instruction appears after all the instructions which initialize BDs, DMAs, etc. 

Is that the point you're making above?

makslevental: Exactly correct - I hit that issue earlier this year with stuff I was doing in mlir-aie+python where I was trying to configure more at runtime. 

This one (program BDs/DMAs before enabling cores) is not such a bad gotcha - it's intuitive. The thing that keeps me up at night is that the DMA/BDs/semaphores can have race conditions. I.e if you sequence your acquires/releases wrong you will get a deadlock. Some of this is documented in the arch spec but some of it is not (ie *bugs in the RTL/hardware..............*).

compiler/plugins/target/AMD-AIE/aie/AIEDmaToNpu.cpp
jtuyls: Yeah, I was wondering about the state of it. And yes, these submit transactions refer to the transaction interpreter in the FW: https://gitenterprise.xilinx.com/ai-engine/aie-ipu/blob/master/dpu/aie_controller/include/ps/interpreter_op_impl.h#L867

jtuyls:
> > However I am not sure about completely reusing some of the dialect's operations/attributes unless those ops/attrs are widely applicable for AIE as well. My concern is that I think a lot of the core GPU dialect's operations are not applicable for AIE and it would be unclear/weird to interleave a few GPU ops/attrs between AMDAIE ops, especially when they are very simple like `gpu.object`. Just some thoughts, would love to hear other people's take on this as well.
> 
> Yea I'll admit I'm not completely sure about the idea either - I would prefer to reuse absolutely as much upstream stuff as possible in order to prevent the "ghettoization"[1](#user-content-fn-1-1f5cf488933fe9f20ebbba4d70339a7e) of AIE, but you're right we shouldn't force it.
> 
> ## Footnotes
> 1. don't get mad, it's a real, [technical term](https://dictionary.cambridge.org/dictionary/english/ghettoize). [↩](#user-content-fnref-1-1f5cf488933fe9f20ebbba4d70339a7e)

I will admit I would have no issue with starting to reuse some of these ops/attrs if they were called: `device.module`, `device.launch`, `device.object` or something similar. Maybe some of these `gpu` ops really belong in a separate dialect that can be reused across different accelerator devices..

compiler/plugins/target/AMD-AIE/aie/AIEDmaToNpu.cpp
makslevental: I mean I knew/know that - that `XAIE_IO_CUSTOM_OP` whatever hits that path in the firmware. What I'm saying I don't know is if the stuff in `aie-rt` that goes by the same name is actually about the "transaction interpreter" in the firmware. 

It's an easy thing to check (just look in the source of `aie-rt` to see what happens eg when you call `SubmitSerializedTransaction`) I just haven't done it yet.

makslevental:
> I will admit I would have no issue with starting to reuse some of these ops/attrs if they were called: `device.module`, `device.launch`, `device.object` or something similar. Maybe some of these `gpu` ops really belong in a separate dialect that can be reused across different accelerator devices..

We can always RFC but I'm not a fan of those - too much debating. Anyway I think GPU today means GPGPU and we're basically doing that same kind of stuff right? Tiling, matmuls, vectorized single core, etc. Is it really that different? But if you do think AIE is too different then it's fine - we can just copy paste some td (for ops we don't currently have) and leave a note "mostly just like `gpu` dialect from upstream".

compiler/plugins/target/AMD-AIE/aie/AIEDmaToNpu.cpp
jtuyls: > It's an easy thing to check (just look in the source of aie-rt to see what happens eg when you call SubmitSerializedTransaction) I just haven't done it yet.

Oh, I think I see what you mean. I don't think `SubmitSerializedTransaction` would work (I hope it would though). What should work is exporting the transaction with `XAie_ExportSerializedTransaction`: https://github.com/Xilinx/aie-rt/blob/1ad203de0b7f282b1c0659fd2ae3f218652c7274/driver/src/global/xaiegbl.h#L582, which should give you the instruction bytestream like you have it here: https://github.com/nod-ai/iree-amd-aie/pull/460/files#diff-bcbdafe57dbdd0686e4a2a056c5018c13dc4164bba50861a04e92b0066268b3cR490

makslevental: Yea so this is actually exactly why I never finished this same sprint in `mlir-aie` and actually *exactly what gave me the ideas that I'm implementing here today*. I.e., if we remove everything below objectfifo then we only have `aie-rt` APIs and we can just call them and "record" them using `SerializedTransaction` and then send them as NPU/IPU instructions. The problem of course is what's mentioned up-thread here (not everything can be a runtime instruction) and CDO object and `SerializedTransaction` aren't 1-1 ([cdo has its own stupid op code suffixes](https://github.com/Xilinx/bootgen/blob/master/cdo-driver/cdo_driver.h#L27-L31)) so converting between isn't completely trivial.

jtuyls (approved):

compiler/plugins/target/AMD-AIE/aie/AIEDmaToNpu.cpp
jtuyls: > The problem of course is what's mentioned up-thread here (not everything can be a runtime instruction) and CDO object and SerializedTransaction aren't 1-1 ([cdo has its own stupid op code prefixes](https://github.com/Xilinx/bootgen/blob/master/cdo-driver/cdo_driver.h#L27-L31)) so converting between isn't completely trivial.

I am not following this. Why do we need to translate between CDO and transactions? The serialized transaction should just be usable with the current flow by using it to create the `sdt::vector<uint32_t> instructions` object and keeping everything else the same?

compiler/plugins/target/AMD-AIE/aie/AIEDmaToNpu.cpp
newling: Makes sense. I'm not concerned that there are race conditions and deadlocks, but I haven't actually written any bare metal programs yet...

I still think that we should use a non-pattern approach here for 2 reasons
1) the instruction sequence might change if upstream pattern ordering changes, making the lit test fail which is going to confuse some unfortunate person bumping our llvm in 2 years time
2) there's no fixed point algo here

So let's get back to reconsidering this sometime!

makslevental: I tried to do this this morning before landing. I did a `walk` and erase thing and it failed because there were `uses`. I could've pushed harder ofc but that would mean doing a toposort or slice analysis or something. Which is fine but I felt like I'd just be reproducing some/most of what the `PatternRewriter` infra already does. I also did think about explicitly sequencing the patterns by using `benefit` but then didn't because you have to think a little bit about whether you want to support later inserting a pattern underneath. Anyway combined with the fact that like everything else it'll go away (mostly) I felt like it wasn't worth pursuing. 

So I'll make you a deal: if this code is still here by end of July sure let's redesign but I don't think it will be.

newling: Deal.

compiler/plugins/target/AMD-AIE/aie/AIEDmaToNpu.cpp
makslevental: Yes but using `SerializedTransaction` just here for these few ops is +/- no net gain really. Ideally you want to use `SerializedTransaction` for *everything*. But you can't because some things *have to* go into a CDO and some things can be either/or.

makslevental: Anyway you're right - maybe the net gain is simply starting the gradual transition to using `SerializedTransaction` in as many places as possible. Lemme create and issue for myself to do what you said so I remember.

PR #461, opened 2024-06-26
[Conv2d] Add lowering pipeline and tiling strategy for Conv2d

yzhang93:
- This is the first PR to add C++ pipeline for Conv2d ops. This PR added a new lowering pipeline for Conv2d (`conv_2d_nchw_fchw` / `conv_2d_nhwc_hwcf`). The tiling strategy is following the tile sizes shown in [transform dialect](https://github.com/nod-ai/iree-amd-aie/blob/main/tests/transform_dialect/conv_fill_spec_pad.mlir), and some dimensions are hardcoded for now. We can explore more on the tiling strategy once simple example working e2e.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -422,74 +498,46 @@ static LogicalResult setRootConfigForPadPackPipeline(
...
-      utils::IteratorType::parallel, utils::IteratorType::parallel,
-      utils::IteratorType::reduction};
-  SmallVector<utils::IteratorType> opIteratorTypes =
-      genericOp.getIteratorTypesArray();
-  if (matmulTransposeIteratorTypes != opIteratorTypes) {
-    return false;
+static LogicalResult setRootConfigForConvDecomposePipeline(
+    mlir::FunctionOpInterface entryPointFn, linalg::LinalgOp linalgOp,
+    AIEConfig cfg) {
+  auto maybeInstructionSize = getMatmulInstructionSize(linalgOp);
```
Abhishek-Varma, line 504: Expand `auto` here and elsewhere.

Here: `auto` -> `FailureOr<std::array<uint32_t, 3>>`

Abhishek-Varma (approved):
Nice. LGTM! One trivial comment to address.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
newling: I prefer auto ;) Do we have a coding style reference?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -297,6 +301,78 @@ static SmallVector<int64_t> setInnerPermB(bool isMatmulTransposeB) {
   return innerPermB;
 }
 
+/// TODO(avarma): This currently is skipping checking for ext* ops.
+static bool bodyMatcherForMatmulTranspose(Value yieldVal, Block *body) {
```
newling, line 305: Did this function need to move upwards in the .cpp? It's just that it makes reviewing a bit trickier, as I'm not sure if this function has changed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -556,6 +604,20 @@ static LogicalResult setRootConfig(mlir::FunctionOpInterface entryPointFn,
...
+                                       AIEPassPipeline passPipeline,
+                                       AIEConfig cfg) {
+  assert(!getLoweringConfig<IREE::Codegen::LoweringConfigAttr>(convOp) &&
+         "expected lowering_config is not set");
+  auto linalgOp = cast<linalg::LinalgOp>(convOp.getOperation());
+
+  // Current tiling strategy is based on llvm-cpu ConvTileAndDecomposeExpert.
+  if (passPipeline == AIEPassPipeline::ConvDecomposePipeline)
+    return setRootConfigForConvDecomposePipeline(entryPointFn, linalgOp, cfg);
+  return linalgOp.emitError("Unhandled pass pipeline in setRootConfig.");
```
newling, line 618:
```suggestion
  return linalgOp.emitError("Unhandled pass pipeline in setConvRootConfig.");
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.h
```diff
@@ -18,6 +18,7 @@ namespace mlir::iree_compiler::AMDAIE {
 enum class AIEPassPipeline : int32_t {
   PackPeelPipeline = 0,
   PadPackPipeline = 1,
+  ConvDecomposePipeline = 2,
   None = 2
```
newling, line 22: None = 2 (is that what I'm seeing)? 

I think it's not necessary to give all the AIEPassPipeline names explicit numbers, having just PackPeelPipeline = 0 is enough, the others will be contiguous from there.

newling (changes requested):
Looks good, only minor comments. The PR title is 

[Conv2d] Add lowering pipeline and tiling strategy for Conv2d

But it only adds the tiling strategy AFAICT?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
nirvedhmeshram: The compiler follows llvm standards so this
https://releases.llvm.org/10.0.0/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable

It is subjective on if it makes the code more readable in this case :) 
Since the llvm standard leaves room for interpretation we can also look athe google standard (which the non compiler part of the project follow)
https://google.github.io/styleguide/cppguide.html#Type_deduction

erwei-xilinx (approved):
The tiling recipe looks great to me. Thanks.

yzhang93:
> Looks good, only minor comments. The PR title is
> 
> [Conv2d] Add lowering pipeline and tiling strategy for Conv2d
> 
> But it only adds the tiling strategy AFAICT?

Yeah, my intention for this first PR was just to add a new lowering pipeline `conv-decompose`. The tiling strategy need to be paired with the pipeline. It doesn't include the building passes for the pipeline yet. Sorry for the confusion.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.h
yzhang93: Looks like we are not even using the values. I have deleted the assigned numbers.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: I think this is a good guideline. Thanks for sharing it. 
```Don’t “almost always” use auto, but do use auto with initializers like cast<Foo>(...) or other places where the type is already obvious from the context. Another time when auto works well for these purposes is when the type would have been abstracted away anyways, often behind a container’s typedef such as std::vector<T>::iterator.```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
newling: Thanks. I have no strong preference so if you think there's a good reason to move it, then go ahead.

PR #462, opened 2024-06-26
Replace `AIETargetModel` with `AMDAIEDeviceModel` via `aie-rt`

makslevental:
# Summary

This PR replaces AIE's `AIETargetModel` with the canonical representation/metadata for the devices in `aie-rt` (named `AMDAIEDeviceModel`). There's a lot going on here - some of it is code motion - some of it is new - and some of it is dealing with `mlir-aie`'s idiosyncracies. Before diving in, I want to emphasize the goal(s) here:

1. Another brick in the wall (out of the wall?) in the process of decoupling from `mlir-aie`;
2. `aie-rt` should be the canonical representation/model of the device - it has all the register maps, it is maintained (despite appearances), and it's basically dead simple to understand (just some arithmetic implemented in C). 
3. Having said that, we should provide a (very thin) layer around it; that's what `iree_aie_runtime` is doing.

The approach taken here is basically two phase:

1. In this PR we simply provide all of the same APIs that `mlir-aie`'s `AIETargetModel` provides;
2. In a follow-up PR, when we refactor and eliminate the passes that consume the target/device model (`amdaie-pathfinder` and `amd-aie-objectstateful-transform`, i.e., the in-tree passes), we will also refactor the device model and eliminate superfluous APIs[^1].

Note, this PR is 9 separate commits, 1 for each pass/target-emission utility that is affected. I am pushing them individually to minimize the surface area (both for reviewers/cognitive-load and for CI to test).

# Testing strategy

We test in two ways:

1. By adding a `DeviceModelTest.cpp` that compares `AMDAIEDeviceModel` against `AIETargetModel`;
2. By aiming to preserve agreement with existing `aie` dialect tests as much as possible (tests checked in https://github.com/nod-ai/iree-amd-aie/pull/421).

The reason for this is because currently, all flows/paths above this layer implicitly assume things about `AIETargetModel`. In particular, things related to "virtualization"/partitioning. More on this in a moment. Unfortunately, in the process of setting up `DeviceModelTest`, several bugs (in both [`mlir-aie`](https://github.com/Xilinx/mlir-aie/issues/1567) and [`aie-rt`](https://github.com/Xilinx/aie-rt/issues/4)) were discovered, thereby complicating the interpretation of the test. The bugs in `aie-rt` are minor and I am working with them to submit patches. The bugs in `mlir-aie` are structural and I've decided we will just live with the inconsistency for a little bit, rather than trying to fix there first. This isn't unreasonable because as I mentioned above, we will soon be completely decoupled anyway and, more importantly, **these differences do not affect numerical accuracy** (the `run_matmul` tests pass in CI and locally).

These inconsistencies are "recorded" as essentially XFAILs in `DeviceModelTest`; they look like this: 

```cpp
// mlir-aie reports true when it should be false
const std::set<std::tuple<int, int, int, int>> IsLegalMemtileConnectionFails{
    // srcPort, srcChan, dstPort, dstChan
    {CTRL, 0, DMA, 0},
    // trace
    {TRACE, 0, CTRL, 0},
    {TRACE, 0, DMA, 0},
    {TRACE, 0, DMA, 1},
    {TRACE, 0, DMA, 2},
    {TRACE, 0, DMA, 3},
    {TRACE, 0, DMA, 4},
    {TRACE, 0, NORTH, 0},
    {TRACE, 0, NORTH, 1},
    {TRACE, 0, NORTH, 2},
    {TRACE, 0, NORTH, 3},
    {TRACE, 0, NORTH, 4},
    {TRACE, 0, NORTH, 5},
};
```

If anyone knows a better way to handle XFAILs like this in `gtest` feel free to recommend.

# Bonus: Virtualization/partitioning

### Prior to https://github.com/nod-ai/iree-amd-aie/pull/511, this was more relevant because we diverged from `AIETargetModel` wrt partitioning. We no longer do, so this info isn't important for reviewing the PR but I'm leaving it in case anyone is interested.

This is important to understand at some level but probably the details are best left to me or anyone else that's responsible for debugging. But to review this PR you'll need to at least understand the basics. Firstly, note, these are actually two different things:

1. Virtualization: configuration instructions (i.e., register writes) can pretend the design will execute in column 1 but in reality the design can execute in any column >=1 and the device will adjust the instructions accordingly;
2. Partitioning: [clock-gating](https://en.wikipedia.org/wiki/Clock_gating), i.e., enabling/disabling power to certain columns.

They overlap because we almost always do both: the designs/kernels are written as if they're configuring column 0 but they actually run in column 1 *and* we don't use the entire array (e.g., we only use a single column or we use columns 1,2,3,4 and not 0). 

The former is entirely handled by some black magic dance between the driver and the firmware (I haven't tracked it all the way down yet); in **our** code virtualization is only evident in exactly [one line of code](https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp#L453): 

```
"partition": {
  "column_width": 4,
  "start_columns": [
    1
  ]
},
```

This entry/attribute of `aie_partition.json` is [queried by the driver](https://github.com/amd/xdna-driver/blob/6cd2e9b5b92157511378b83c884504b8985a2561/src/include/uapi/drm_local/amdxdna_accel.h#L389). 

The latter (partitioning) is handled/touched/considered by `aie-rt` in a small number of places ([XAie_PmSetColumnClk](https://github.com/Xilinx/aie-rt/blob/release/main_aig/driver/src/pm/xaie_clock.c#L154-L165)). **In addition**, `aie-rt` uses the partition information to determine/report tile types depending on device generation ([_XAie_GetTTypefromLoc](https://github.com/Xilinx/aie-rt/blob/63e1958e1fa9dad72a39f184c59b2657ba53af4b/driver/src/device/xaie_device_aie.c#L130-L129), [_XAieMl_GetTTypefromLoc](https://github.com/Xilinx/aie-rt/blob/63e1958e1fa9dad72a39f184c59b2657ba53af4b/driver/src/device/xaie_device_aieml.c#L62-L61), [_XAie2Ipu_GetTTypefromLoc](https://github.com/Xilinx/aie-rt/blob/c25d0939d3265ce6d213f87ec4adc17b5678ea4d/driver/src/device/xaie_device_aie2ipu.c))[^2].

[^1]: APIs like `isNorth/South/East/West` [that don't actually do what they say they do](https://github.com/Xilinx/mlir-aie/issues/1021).
[^2]: This is why a previous version of this PR (prior to https://github.com/nod-ai/iree-amd-aie/pull/511, i.e., to me finding the pleasant surprise that `aie-rt` supports IPU) had a hack to tweak which tiles in row 0 are considered ShimNOC and which are considered ShimPL (on `aie2` the pattern is apparently `ShimPL, ShimPL, ShimNOC, ShimNOC`, while on `aie2IPU` the pattern is `ShimPL, ShimNOC, ShimNOC, ShimNOC`).


compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIETargetCDODirect.cpp
```diff
@@ -423,32 +354,34 @@ struct AIEControl {
...
 
     // StreamSwitch (switchbox) configuration
-    for (auto switchboxOp : targetOp.getOps<SwitchboxOp>()) {
+    for (auto switchboxOp : device.getOps<SwitchboxOp>()) {
       int32_t col = switchboxOp.colIndex();
       int32_t row = switchboxOp.rowIndex();
       XAie_LocType tileLoc = XAie_TileLoc(col, row);
-      assert(targetModel.isNPU() && "Only NPU currently supported");
+      //      assert(device.getDevice() == AIEDevice::npu &&
+      //             "Only NPU currently supported");
```
jtuyls, line 372: Delete or re-enable?

iree_compiler_plugin.cmake
```diff
@@ -8,6 +8,9 @@ find_package(Boost REQUIRED)
 include_directories(${Boost_INCLUDE_DIRS})
 
 set(IREE_AMD_AIE_SOURCE_DIR "${CMAKE_CURRENT_LIST_DIR}")
+set(IREE_MLIR_AIR_SOURCE_DIR "${IREE_AMD_AIE_SOURCE_DIR}/third_party/mlir-air/mlir")
```
jtuyls, line 11: Out of curiosity, why removing these lines from the CMakeLists and adding them here and here: [iree_runtime_plugin.cmake](https://github.com/nod-ai/iree-amd-aie/pull/462/files#diff-c01404c842f5f36da999f4af30e03786a5c294d6b03b2830a5809cd07c74abfa)?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -43,50 +50,170 @@ void configureHeader();
...
+#define XAIE_MEM_TILE_ROW_START 1
+#define XAIE_NUM_MEM_TILE_ROWS 1
+#define XAIE_NUM_SHIM_TILE_ROWS 1
+#define XAIE_PARTITION_BASE_ADDR 0x0
+#define XAIE_SHIM_ROW 0
+
+#define NPI_ADDR 0x0
+#define NUM_LOCKS 16
+#define MEM_TILE_LOCK_ID_INCR 64
+#define BASE_ADDR_A_INCR 0x80000
```
jtuyls, line 69: Should these be added as a config to `aie-rt` itself? It seems like there is no existing AIE2IPU config inside that repo?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -43,50 +50,170 @@ void configureHeader();
 void insertNoOpCommand(unsigned int numPadBytes);
 }
 
-#define OSTREAM_OP(O_TYPE, TYPE) O_TYPE &operator<<(O_TYPE &os, const TYPE &s);
+#define XAIE2_BASE_ADDR 0x40000000
+#define XAIE1_BASE_ADDR 0x20000000000
+#define XAIE1_COL_SHIFT 23
+#define XAIE1_ROW_SHIFT 18
+#define XAIE2_BASE_ADDR 0x40000000
```
jtuyls, line 57: Duplicated `XAIE2_BASE_ADDR `. Also, it might be more clear to use `XAIE2IPU_BASE_ADDR` as aie-rt uses   `0x20000000000` for AIE2, VCK5000 I guess? (https://github.com/Xilinx/aie-rt/blob/1ad203de0b7f282b1c0659fd2ae3f218652c7274/driver/src/lite/xaie_lite_hwcfg.h#L72) ?

runtime/src/iree-amd-aie/aie_runtime/test/DeviceModelTest.cpp
```diff
@@ -0,0 +1,476 @@
...
+#include "aie/Dialect/AIE/IR/AIETargetModel.h"
+#include "gtest/gtest-param-test.h"
+#include "gtest/gtest.h"
+#include "iree-amd-aie/aie_runtime/iree_aie_runtime.h"
+#include "llvm/Support/FormatVariadic.h"
+
+#define NPU_NUM_COLS 4
+#define NPU_NUM_ROWS 6
+
+// operator<< for various types
```
jtuyls, line 25:
```suggestion
```
?

stephenneuendorffer:
For the record, I fundamentally disagree with statement 2: "aie-rt should be the canonical representation/model of the device - it has all the register maps, it is maintained (despite appearances), and it's basically dead simple to understand (just some arithmetic implemented in C)."

If aie-rt were just the arithmetic, I would agree, but unfortunately, it's the arithmetic bundled inside a runtime API with a bunch of stuff that is fundamentally not needed in the context of a compiler system.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -43,50 +50,170 @@ void configureHeader();
...
+
+enum class AMDAIETileType : uint8_t {
+  AIETILE = 0U,
+  SHIMNOC = 1U,
+  SHIMPL = 2U,
+  MEMTILE = 3U,
+  MAX = 4U
+};
+
+struct AMDAIEDeviceModel {
```
jtuyls, line 109: I think this struct could use some documentation.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -32,9 +32,16 @@
 
 extern "C" {
 #include "xaiengine.h"
+#include "xaiengine/xaie_device_aieml.h"
+#define s8
+#define u8
+#define u16
+#define s32
+#define u32
+#define u64
```
jtuyls, line 41: Are these needed?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc
```diff
@@ -4,6 +4,21 @@
...
+#include "xaiengine.h"
+#include "xaiengine/xaie_device_aie.h"
+#include "xaiengine/xaie_device_aie2ipu.h"
+#include "xaiengine/xaie_device_aieml.h"
+#define s8
+#define u8
+#define u16
+#define s32
+#define u32
+#define u64
```
jtuyls, line 19: Are these needed?

makslevental:
> If aie-rt were just the arithmetic, I would agree, but unfortunately, it's the arithmetic bundled inside a runtime API with a bunch of stuff that is fundamentally not needed in the context of a compiler system.

I don't know where there's room to disagree here - one can look at the implementation of `aie-rt` and see with one's own eyes it is just arithmetic on relative register addresses combined with masks. There's no runtime API whatsoever if one chooses any backend other than `__AIEBAREMETAL__`.

> bunch of stuff that is fundamentally not needed in the context of a compiler system

It would certainly be helpful if you could point this "bunch of stuff".

stephenneuendorffer:
> > If aie-rt were just the arithmetic, I would agree, but unfortunately, it's the arithmetic bundled inside a runtime API with a bunch of stuff that is fundamentally not needed in the context of a compiler system.
> 
> I don't know where there's room to disagree here - one can look at the implementation of `aie-rt` and see with one's own eyes it is just arithmetic on relative register addresses combined with masks. There's no runtime API whatsoever if one chooses any backend other than `__AIEBAREMETAL__`. Since we are using the `__AIECDO__` backend, we incur no such costs/complexities.

All the complexity is there, but it is at least complexity that someone else owns.

> > bunch of stuff that is fundamentally not needed in the context of a compiler system
> 
> It would certainly be helpful if you could point out this "bunch of stuff".

Compare this, which implements the math without aie-rt:
https://github.com/Xilinx/mlir-aie/blob/main/lib/Targets/AIETargetAirbin.cpp

runtime/src/iree-amd-aie/aie_runtime/test/DeviceModelTest.cpp
```diff
@@ -0,0 +1,476 @@
...
+INSTANTIATE_TEST_SUITE_P(
+    VerifyAIERTAIE2MemTileConnectivity,
+    AMDAIENPUDeviceModelParameterizedMemtileConnectivityNPU4ColTestFixture,
+    ::testing::Combine(::testing::Range(0, (int)MEMTILE_CONNECTIVITY.size()),
+                       ::testing::Range(0,
+                                        (int)MEMTILE_CONNECTIVITY[0].size())));
+
+#define MAX_CHANNELS 6
+
+// Figure 6-9: Stream-switch ports and connectivity matri
```
jtuyls, line 459:
```suggestion
// Figure 6-9: Stream-switch ports and connectivity matrix
```

makslevental:
> All the complexity is there, but it is at least complexity that someone else owns.

I don't know what this means - can you please point to a "runtime API" in `aie-rt` when one uses the `__AIECDO__` backend.

> Compare this, which implements the math without aie-rt: https://github.com/Xilinx/mlir-aie/blob/main/lib/Targets/AIETargetAirbin.cpp

1. That's not a pointer to something in `aie-rt` that highlights the "bunch of stuff that is fundamentally not needed in the context of a compiler system" you're implying exists;
2. Seeing as I wrote `AIETargetAirbin`, I think I am uniquely qualified to make the comparison; `AIETargetAirbin` reimplements (incompletely) only the Elf parsing components of `aie-rt` so of course it should be much simpler.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc
```diff
@@ -92,3 +142,323 @@ STRINGIFY_2TUPLE_STRUCT(XAie_Packet, PktId, PktType)
...
+uint32_t AMDAIEDeviceModel::getCoreTileLocalMemorySize() {
+  if (devInst.DevProp.DevGen == XAIE_DEV_GEN_AIE)
+    return 32768;
+  else if (devInst.DevProp.DevGen == XAIE_DEV_GEN_AIEML ||
+           devInst.DevProp.DevGen == XAIE_DEV_GEN_AIE2IPU)
+    return 65536;
+  else
+    llvm::report_fatal_error(llvm::Twine("unknown AIE DevGen: ") +
+                             std::to_string((int)devInst.DevProp.DevGen));
+  // bytes
```
jtuyls, line 312:
```suggestion
```
?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc
```diff
@@ -92,3 +142,323 @@ STRINGIFY_2TUPLE_STRUCT(XAie_Packet, PktId, PktType)
...
+  return (isMemSouth && !isMemTile(memCol, memRow)) || isMemNorth ||
+         isMemWest || isMemEast;
+}
+
+bool AMDAIEDeviceModel::isLegalMemtileConnection(uint8_t col, uint8_t row,
+                                                 StrmSwPortType srcBundle,
+                                                 uint8_t srcChan,
+                                                 StrmSwPortType dstBundle,
+                                                 uint8_t dstChan) {
+  // this isn't correct but for agreement with mlir-aie...
```
jtuyls, line 345:
```suggestion
  // TODO(max): this isn't correct but for agreement with mlir-aie...
```

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc
```diff
@@ -92,3 +142,323 @@ STRINGIFY_2TUPLE_STRUCT(XAie_Packet, PktId, PktType)
...
+  if (srcBundle == dstBundle and srcBundle != DMA) return true;
+  assert(isMemTile(col, row) && "expected memtile");
+  AMDAIETileType tileType = getTileType(col, row);
+  assert(tileType != AMDAIETileType::MAX && "invalid tile");
```
jtuyls, line 349:
```suggestion
  AMDAIETileType tileType = getTileType(col, row);
  assert(tileType == AMDAIETileType::MEMTILE && "expected memtile");
```

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
makslevental: [There is](https://github.com/Xilinx/aie-rt/blob/release/main_aig/driver/src/global/xaie2ipugbl_reginit.c) but these things are (for the most part) not part of that config. I have tried to find as many of these constants as possible (or find their representation in terms of more fundamental constants) but I believe the philosophy here is that `aie-rt` is parameterized by these values. Anyway I will go over these constants another time and I see if I can reduce their number even further.

iree_compiler_plugin.cmake
makslevental: Well `IREE_MLIR_AIE_SOURCE_DIR` is now used in a couple of places so I just moved both. I can put this one back since I guess it still remains only used in one place.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
makslevental: Ah good job - you found the magic header that has more of these constants. Although I think I've tried to enable `#ifdef XAIE_FEATURE_LITE` before and it disables some other useful headers/utilities.

runtime/src/iree-amd-aie/aie_runtime/test/DeviceModelTest.cpp
```diff
@@ -0,0 +1,476 @@
...
+}
+
+#include "aie/Dialect/AIE/IR/AIEDialect.h"
+#include "aie/Dialect/AIE/IR/AIETargetModel.h"
+#include "gtest/gtest-param-test.h"
+#include "gtest/gtest.h"
+#include "iree-amd-aie/aie_runtime/iree_aie_runtime.h"
+#include "llvm/Support/FormatVariadic.h"
+
+#define NPU_NUM_COLS 4
```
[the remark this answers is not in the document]
makslevental: That's a mistake from before I found `aie2ipu` config. Let me change.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
makslevental: Combing with your pointer `xaie_lite_hwcfg.h` I guess it's correct there's no such config in `aie-rt`. I propose we carry it locally for now and upstream/PR it to them.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
makslevental: Yes - otherwise we leak the same defines from `aie-rt` which I would prefer not pop up as  suggestions in anyone's editors. But I just realized I'm doing this wrong and these should just be `#undef`s.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc
makslevental: Same as above - these need to be `undef`ined.

runtime/src/iree-amd-aie/aie_runtime/test/DeviceModelTest.cpp
makslevental: Without this `using` you do not get the `operator<<` overloads in `iree_aie_runtime.h`. I should expand this comment to say more.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIETargetCDODirect.cpp
```diff
@@ -79,27 +65,18 @@ StrmSwPortType toStrmT(WireBundle w) {
...
-#define XAIE_SHIM_ROW 0
-#define XAIE_MEM_TILE_ROW_START 1
-#define XAIE_PARTITION_BASE_ADDR 0x0
-
-#define NPI_ADDR 0x0
 #define NUM_LOCKS 16
-#define EVEN_BD_NUM_START 0
-#define ODD_BD_NUM_START 24
 #define MEM_TILE_LOCK_ID_INCR 64
 #define BASE_ADDR_A_INCR 0x80000
```
jtuyls, line 74: Can be removed as well?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
jtuyls: Yeah, makes sense

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIETargetCDODirect.cpp
makslevental: Good catch. All of the `#define`s need to be moved out from the various places they've been injected. Will do that.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
jtuyls: I see, the `#undef` makes sense, thanks.

jtuyls (dismissed):
LGTM. Just some minor comments above. It might be useful to have someone else go through it in detail as well though.

runtime/src/iree-amd-aie/aie_runtime/test/DeviceModelTest.cpp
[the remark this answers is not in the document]
makslevental: Actually I take it back - when you set a partition config you actually reduce the number of cols that the device supposedly has access to: https://github.com/Xilinx/aie-rt/blob/38fcf1f9eb7c678defaf8c19ffb50a679b644452/driver/src/global/xaiegbl.c#L120

So I renamed - basically this is an artifact of the fact that I don't know how to parameterize the ranges in `gtest` using the fixture itself (like I should be getting the range of cols from the device model itself using `cols()`).

compiler/plugins/target/AMD-AIE/aie/test/nd_dma_distribute_AIE2.mlir
```diff
@@ -103,10 +103,10 @@
 
 module @ndDMAObjFifoAIE2 {
  aie.device(xcve2302) {
-    %tile10 = aie.tile(1, 0)
-    %tile11 = aie.tile(1, 1)
-    %tile22 = aie.tile(2, 2)
-    %tile23 = aie.tile(2, 3)
+    %tile10 = aie.tile(2, 0)
```
nirvedhmeshram, line 106: There is no FileCheck in the run command for this test (not sure if accidentally missing or just a disabled test?)

makslevental: good catch

nirvedhmeshram (approved):
LGTM

compiler/plugins/target/AMD-AIE/aie/CMakeLists.txt
```diff
@@ -287,6 +297,7 @@ iree_cc_library(
     AMDAIEXToStandard.cpp
   DEPS
```
jtuyls, line 298: Shouldn't `AMDAIEDialect` be added to the dependencies as well?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -43,50 +36,155 @@ void configureHeader();
...
+
+  bool operator==(const TileLoc& rhs) const {
+    return std::tie(col, row) == std::tie(rhs.col, rhs.row);
+  }
+
+  bool operator!=(const TileLoc& rhs) const { return !(*this == rhs); }
+
+  operator XAie_LocType() const { return XAie_TileLoc(col, row); }
+  TileLoc(XAie_LocType loc) : col(loc.Col), row(loc.Row) {}
+  TileLoc(int col, int row) : col(col), row(row) {}
```
newling, line 54: Can never remember C++ rule for uninitialised class int members, do you want TileLoc() : TileLoc(0,0) {} or TileLoc() = delete ?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -43,50 +36,155 @@ void configureHeader();
...
+};
+
+enum class AMDAIEDevice : uint32_t {
+  xcvc1902 = 1,
+  xcve2302 = 2,
+  xcve2802 = 3,
+  npu = 4,
```
newling, line 61: Having 2 enums with the same value is strange, can we just remove one of them?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -43,50 +36,155 @@ void configureHeader();
...
+  npu1 = 4,
+  npu1_1col = 5,
+  npu1_2col = 6,
+  npu1_3col = 7,
+  npu1_4col = 8,
+};
+
+enum class AMDAIETileType : uint8_t {
+  AIETILE = 0U,
```
newling, line 70: The C++ spec says the values increment from 1 after the first one by default, you don't need to enumerate all values here

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
newling: Definitely needs documentation! Also, where is the class implemented?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -43,50 +36,155 @@ void configureHeader();
...
+struct AMDAIEDeviceModel {
+  XAie_Config configPtr;
+  XAie_DevInst devInst;
+
+  explicit AMDAIEDeviceModel(uint8_t aieGen, uint64_t baseAddr,
+                             uint8_t colShift, uint8_t rowShift,
+                             uint8_t devNColumns, uint8_t devNRows,
+                             uint8_t memTileRowStart, uint8_t nMemTileRows,
+                             uint8_t nShimTileRows, int partitionNumCols,
+                             int partitionStartCol = 1, bool aieSim = false,
```
newling, line 86: I don't like default arguments, they get annoying when you want to change APIs, and buggy when 2 operands with defaults have the same type (bool and bool) -- I'd prefer if these were explicit if not too big a pain please

compiler/plugins/target/AMD-AIE/aie/CMakeLists.txt
makslevental: There's no use of `AMDAIEDialect` here yet.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
makslevental: Added `xaie_hwcfg.c/h`.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
makslevental: Added `xaie_hwcfg.c/h`.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -43,50 +36,155 @@ void configureHeader();
...
+  int col, row;
+  inline bool operator<(const TileLoc& rhs) const {
+    return std::tie(col, row) < std::tie(rhs.col, rhs.row);
+  }
+
+  bool operator==(const TileLoc& rhs) const {
+    return std::tie(col, row) == std::tie(rhs.col, rhs.row);
+  }
+
+  bool operator!=(const TileLoc& rhs) const { return !(*this == rhs); }
```
newling, line 50: Why not just throw in the other 3 while you're here >, >=, <=

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
makslevental: `mlir-aie` has `npu1` which is ... strange for an enum that signifies "the whole device", hence just `npu`. The intent is to remove `npu1_*` when we ditch the old passes that reference those enum variants.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
makslevental: Even for `class`? Still I prefer the explicit enumeration because e.g. for 

```
typedef enum{
	CORE,
	DMA,
	CTRL,
	FIFO,
	SOUTH,
	WEST,
	NORTH,
	EAST,
	TRACE,
	UCTRLR,
	SS_PORT_TYPE_MAX
} StrmSwPortType;
```

I am constantly counting on my numbers (not really) when the enum gets serialized to an integer.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
makslevental: Fair enough

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
makslevental: Hmm I thought I did have those; lemme implment `std::less`.

makslevental: Oh duh - it's already there (just above) using `std::tie`.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
makslevental: definitely `delete`. Thanks.

runtime/src/iree-amd-aie/aie_runtime/export.h
```diff
@@ -0,0 +1,23 @@
+//
+// Created by mlevental on 7/9/24.
```
jtuyls, line 2: Why this specific line and not add the typical copyright?

makslevental: need to make my editor insert the IREE header instead of that default...

makslevental: Sorry for constantly leaving this...

runtime/src/iree-amd-aie/aie_runtime/xaie_hwcfg.c
```diff
@@ -0,0 +1,166 @@
+#include "xaiengine/xaiegbl_defs.h"
```
jtuyls, line 1: Add the copyright

runtime/src/iree-amd-aie/aie_runtime/xaie_hwcfg.h
```diff
@@ -0,0 +1,79 @@
+#ifndef XAIE_HWCFG_H
```
jtuyls, line 1: Add the copyright

runtime/src/iree-amd-aie/aie_runtime/xaie_hwcfg.h
```diff
@@ -0,0 +1,79 @@
...
+extern uint64_t XAIE2IPU_NPI_BASEADDR;
+extern uint64_t XAIE2IPU_PARTITION_BASE_ADDR;
+
+extern int XAIE2IPU_MEM_TILE_LOCK_ID_INCR;
+extern uint64_t XAIE2IPU_ADDR_ARRAY_OFF;
+
+extern int XAIEIPU_MEM_TILE_LOCK_ID_INCR;
+extern uint64_t XAIEIPU_ADDR_ARRAY_OFF;
+
+#endif
```
jtuyls, line 79: Needs an empty line at the end

runtime/src/iree-amd-aie/aie_runtime/xaie_hwcfg.c
```diff
@@ -0,0 +1,166 @@
...
+#undef XAIE_NUM_NOC_INTR_OFFSET
+#undef XAIE_CORE_TILE_NUM_EVENT_STATUS_REGS
+#undef XAIE_SHIM_TILE_NUM_EVENT_STATUS_REGS
+#undef XAIE_MEM_TILE_NUM_EVENT_STATUS_REGS
+#undef XAIE_DEV_SINGLE_GEN
+#undef XAIE_LITE_HWCFG_H
+
+// AIE Gen IPU
+#define IPU_AIE_BASEADDR 0x40000000;
+#define IPU_AIE_NPI_ADDR 0x0;
```
jtuyls, line 112: Are these needed? Can't the existing `XAIE2IPU_BASE_ADDR` and `XAIE2IPU_NPI_BASEADDR` be reused?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
newling: Now I'm not sure actually, maybe the spec doesn't say increment by default (sorry for the noise). As for serializing though, shouldn't you always be going with static_cast<int>(AMDAIETileType)(AIETILE), i.e. you never actually need to use an raw integer? I guess for debugging it can be useful though.

compiler/plugins/target/AMD-AIE/aie/CMakeLists.txt
jtuyls: Right, nvm

runtime/src/iree-amd-aie/aie_runtime/xaie_hwcfg.c
makslevental: Well `XAIE2IPU*` don't exist (those are just my names for these things). But `XAIE_BASE_ADDR`, `XAIE_NPI_BASEADDR` for IPU don't exist either because I can't find where some key headers are stashed https://github.com/Xilinx/aie-rt/blob/release/main_aig/driver/src/lite/xaie_lite_hwcfg.h#L113-L123

```
#ifdef __AIEIPU__
#include <platform-hw-config.h>
#endif

#ifndef XAIE_BASE_ADDR
#define XAIE_BASE_ADDR			IPU_AIE_BASEADDR
#endif

#ifndef XAIE_NPI_BASEADDR
#define XAIE_NPI_BASEADDR		IPU_AIE_NPI_ADDR
#endif
```

`platform-hw-config.h` is in [XRT](https://github.com/Xilinx/XRT/blob/master/src/runtime_src/tools/xclbinutil/aie-pdi-transform/libinclude/platform-hw-config.h#L10) (:facepalm:)

```
#ifdef _ENABLE_IPU_LX6_
#include <ipu.h>
extern aie2_manager_t aie2_manager_info;
```

but I can't find `ipu.h` (and git blame person hasn't responded).

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
makslevental: `printf`s and the like serialize to integers (well at least they did before I implementd `operator<<` all over the place...).

newling (approved):

runtime/src/iree-amd-aie/aie_runtime/xaie_hwcfg.c
jtuyls: I see, strange where that file comes from.. but thanks for the explanation.

jtuyls (approved):

PR #470, opened 2024-06-27
Add missing build dependency edge

newling:
This resolves the build issue with error message

```
"iree/third_party/llvm-project/mlir/include/mlir/Dialect/EmitC/IR/EmitC.h:28:10: fatal error: 'mlir/Dialect/EmitC/IR/EmitCEnums.h.inc' file not found"
```

Observed when building with the plugin without having first built IREE without the plugin. Error while building translation unit

```
mlir-aie/lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```


nirvedhmeshram (approved):
Thanks.

PR #472, opened 2024-06-27
[Conv2d] Add building passes for conv-decompose pipeline

yzhang93:
- This PR made some changes to the existing passes for `conv-decompose` pipeline working through pad based approach. The lowered IR after bufferization is here https://gist.github.com/yzhang93/cc60d4d3f0b439c96bcc0ebfc6726e51.
- Will add e2e examples to ci tests once the lowering works through MLIR-AIR and MLIR-AIE.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/tile_and_fuse_using_scf_forall.mlir
```diff
@@ -63,6 +80,55 @@ func.func @matmul_static(%arg0: tensor<8x16xi32>, %arg1 : tensor<16x8xi32>) -> t
...
+}
+
+//      TILE-LEVEL-1: @conv_2d_nchw_fchw
+//      TILE-LEVEL-1:   scf.forall
+// TILE-LEVEL-1-SAME:   {
+//      TILE-LEVEL-1:       linalg.fill
+//      TILE-LEVEL-1:       scf.forall
+// TILE-LEVEL-1-SAME:       {
+//      TILE-LEVEL-1:           linalg.conv_2d_nchw_fchw
+//      TILE-LEVEL-1:       }
```
yzhang93, line 127: @MaheshRavishankar @Abhishek-Varma Any idea why `linalg.fill` is not fused into the forall loop after tiling for this Conv2d test? I think it worked for matmul tests. This is not a critical issue at the moment, since I'm using a separate pass `createAMDAIEFuseFillIntoForallPass()` for fusion.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
```diff
@@ -159,20 +159,23 @@ void AMDAIEBufferizeToAllocationPass::runOnOperation() {
   mlir::FunctionOpInterface funcOp = getOperation();
   linalg::LinalgOp linalgOp;
   funcOp->walk<WalkOrder::PostOrder, ReverseIterator>([&](linalg::LinalgOp op) {
+    if (!isElementwise(op) && !linalg::isaContractionOpInterface(op) &&
+        !linalg::isaConvolutionOpInterface(op)) {
+      return WalkResult::advance();
+    }
+    if (isa<linalg::FillOp>(op)) {
```
nirvedhmeshram, line 166: I think this should be redundant as the first if statement should take care of advancing for fill?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/tile_and_fuse_using_scf_forall.mlir
nirvedhmeshram: My guess would be the linalg.copy that is in the input IR, can you check without it?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
yzhang93: Actually it's not the case. `isElementwise(op)` returns true for `linalg.fill`.

newling (approved):
Really nice code reuse!

PR #473, opened 2024-06-27
Add objectFifo cpp pipeline

jtuyls:


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -36,7 +36,9 @@ static llvm::cl::opt<AIEPassPipeline> clUsePipeline(
             "Use the IREE lowering to AIR dialect through pack operation"),
         clEnumValN(AIEPassPipeline::PadPackPipeline, "pad-pack",
                    "Use the IREE lowering to AIR dialect through "
-                   "pad and pack operations")),
+                   "pad and pack operations"),
+        clEnumValN(AIEPassPipeline::ObjectFifoPipeline, "objectFifo",
+                   "Use the IREE lowering to objectFifos")),
```
yzhang93, line 41: This is kind of confusing to put objectFifo pipeline with pack-peel pipelines, since it's actually using pack-peel pipeline. Maybe add another flag just to switch between objectfifo and MLIR-AIR?

yzhang93 (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -437,6 +441,46 @@ void buildAMDAIETransformPassPipeline(OpPassManager &variantPassManager) {
   });
 }
 
+void addAMDAIEObjectFifoLoweringPasses(OpPassManager &passManager) {
+  passManager.addPass(createEraseHALDescriptorTypeFromMemRefPass());
+  passManager.addPass(memref::createFoldMemRefAliasOpsPass());
+  passManager.addPass(createAMDAIEPackToDmaPass());
+  passManager.addPass(xilinx::air::createCopyToDmaPass());
```
nirvedhmeshram, line 448: Dont have to do this in this PR, but it would be good to just have a pass that lowers memcopy to dma because the `air::createCopyToDmaPass` does a lot of other stuff looking for subviews that we (hopefully) dont need.

yzhang93:
@jtuyls This is the change needed to make ci test pass. https://github.com/nod-ai/iree-amd-aie/pull/476

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
jtuyls: Good point! I created an issue for this: https://github.com/nod-ai/iree-amd-aie/issues/480

Abhishek-Varma (commented):
For sake of completeness it'd be nice to change the `RUN` command of [matmul_peeled_objectfifo.mlir](https://github.com/nod-ai/iree-amd-aie/blob/c8d095bc1089f02dea1706af121ce66624f0849f/tests/samples/matmul_peeled_objectfifo.mlir#L1)

jtuyls:
> For sake of completeness it'd be nice to change the `RUN` command of [matmul_peeled_objectfifo.mlir](https://github.com/nod-ai/iree-amd-aie/blob/c8d095bc1089f02dea1706af121ce66624f0849f/tests/samples/matmul_peeled_objectfifo.mlir#L1)

This test starts on some intermediate IR so I don't think I can use the full pipeline. I will add another one that uses `iree-compile`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
jtuyls: Yes, exactly

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -528,10 +539,14 @@ void buildAMDAIETransformPassPipeline(OpPassManager &variantPassManager) {
...
-    addMLIRAIRLoweringPasses(modulePassManager, true);
+  if (clUseBackend == AIEBackend::ObjectFifo) {
+    addAMDAIEObjectFifoLoweringPasses(modulePassManager);
+  } else if (clUseBackend == AIEBackend::AIR) {
+    if (clUsePipeline == AIEPassPipeline::PadPackPipeline) {
+      addMLIRAIRLoweringPasses(modulePassManager, false);
+    } else if (clUsePipeline == AIEPassPipeline::PackPeelPipeline) {
+      addMLIRAIRLoweringPasses(modulePassManager, true);
+    }
   }
```
newling, line 550: 2 design change suggestions
```
if (clUseBackend == AIEBackend::ObjectFifo) {
    addAMDAIEObjectFifoLoweringPasses(modulePassManager);
 } else if (clUseBackend == AIEBackend::AIR) {
     bool isPackPeel = clUsePipeline == AIEPassPipeline::PackPeelPipeline;
     addMLIRAIRLoweringPasses(modulePassManager, isPackPeel);
 } else{ 
    assert(false, "Only 2 backends supported here"); 
 }
 ```

**Review by newling — COMMENTED** [review]
I'm happy to do some redesign work later, but I think we're introducing too many concepts, what is the difference between "backend" and "pipeline" here? 
   
I think it'd be clearer to have 2 "pipelines" (they're just pipelines right?)

 --iree-amdaie-tile-and-bufferizer-pipeline=a
 --iree-amdaie-lower-to-aie-pipeline=b
 
 where a in A, b in B, still with
 A = {pack-peel, pad-pack, default-conv}
 B = {air, objectFifo}
 
We should also consider which pairs {a,b} in AxB are actually supported, and gate with clear errors for the other cases. Again, happy to postpose til  a later PR.

**Comment by jtuyls** [comment]
> I'm happy to do some redesign work later, but I think we're introducing too many concepts, what is the difference between "backend" and "pipeline" here?
> 
> I think it'd be clearer to have 2 "pipelines" (they're just pipelines right?)
> 
> --iree-amdaie-tile-and-bufferizer-pipeline=a --iree-amdaie-lower-to-aie-pipeline=b
> 
> where a in A, b in B, still with A = {pack-peel, pad-pack, default-conv} B = {air, objectFifo}
> 
> We should also consider which pairs {a,b} in AxB are actually supported, and gate with clear errors for the other cases. Again, happy to postpose til a later PR.

Good point. I think these changes belong in this PR itself, so happy to still make them

File: build_tools/ci/run_matmul_test.sh [hunk 4]
```diff
@@ -369,6 +375,7 @@ function run_matmul_test() {
   set +e
 
   compilation_flags="--iree-hal-target-backends=${target_backend} \
+                      --iree-amdaie-use-aie-backend=${aie_backend} \
```
yzhang93, line 378: I think it shouldn't be called `aie-backend`.

Maybe we should adopt James' suggestion and use flags like

```
--iree-amdaie-tile-pipeline
--iree-amdaie-lower-to-aie-pipeline
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -704,10 +768,10 @@ void addMLIRAIRLoweringPasses(OpPassManager &passManager) {
 void addMLIRAIELoweringPasses(OpPassManager &passManager) {
   passManager.addPass(createLowerAffinePass());
   OpPassManager &devicePM = passManager.nest<xilinx::AIE::DeviceOp>();
-  devicePM.addPass(createAMDAIEAssignBufferAddressesBasicPass());
   devicePM.addPass(createAMDAIEAssignLockIDsPass());
-  devicePM.addPass(createAMDAIEAssignBufferDescriptorIDsPass());
   devicePM.addPass(createAMDAIEObjectFifoStatefulTransformPass());
+  devicePM.addPass(createAMDAIEAssignBufferDescriptorIDsPass());
```
jtuyls, line 773: @makslevental It looks like I need to reorder the passes here to avoid a failure

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
jtuyls: This wasn't needed anymore after the addition of the `ConvDecompose` pipeline: https://github.com/nod-ai/iree-amd-aie/pull/487/files#diff-42f0d0bb098689f25ee68e8f05ec6c2eaa89ce41a6394d7d99c1d1c912943b38R548

build_tools/ci/run_matmul_test.sh
jtuyls: Done

jtuyls:
@yzhang93 Could you check again?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
makslevental: Surprising - not sure how statefultransform collides with bdids but I'm not surprised. Sorry for the hiccup (I reordered these this morning and did think for a second whether it would break but it didn't so 🤷‍♂️).

build_tools/ci/run_matmul_test.sh
```diff
@@ -485,7 +492,7 @@ run_matmul_test \
     --peano_install_path "${PEANO}" \
     --mlir_aie_install_path "${MLIR_AIE_INSTALL}" \
     --vitis_path  "${VITIS}" \
-    --pipeline "pad-pack" \
+    --tile_pipeline "pad-pack" \
```
newling, line 495: I've been trying to make this specific test explicitly list all the options, please add

```
--lower_to_aie_pipeline "air" \
```

as one additional line.

newling (approved):
Great to see an end-to-end example working! Just the one minor request in the run_matmul_test.sh from my end

build_tools/ci/run_matmul_test.sh
```diff
@@ -369,7 +375,8 @@ function run_matmul_test() {
   set +e
 
   compilation_flags="--iree-hal-target-backends=${target_backend} \
-                      --iree-amdaie-use-pipeline=${pipeline} \
+                      --iree-amdaie-use-lower-to-aie-pipeline=${lower_to_aie_pipeline} \
+                      --iree-amdaie-use-tile-pipeline=${tile_pipeline} \
```
yzhang93, line 379: I think we don't have to keep `use` in the flag just to make the flag concise. I'm also okay if you want to keep it as is.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -33,22 +33,34 @@
 
 namespace mlir::iree_compiler::AMDAIE {
 
+/// Command line option for selecting the lowering pipeline to use to generate
+/// AIE instructions.
+static llvm::cl::opt<LowerToAIEPassPipeline> clUseLowerToAIEPipeline(
+    "iree-amdaie-use-lower-to-aie-pipeline",
+    llvm::cl::desc("Pick the lowering pipeline to use"),
```
yzhang93, line 40: This description should be more concrete and different from the one below.

yzhang93 (approved):
Thanks for making changes to the pipelines! Happy to see the first e2e test passed the ci.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
jtuyls: No worries, now we will have an e2e test actually using most of these passes :) so it should bark now when broken

build_tools/ci/run_matmul_test.sh
jtuyls: Done

build_tools/ci/run_matmul_test.sh
jtuyls: Updated it

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
jtuyls: Added a more descriptive comment to the one below as that one had an unrelated comment.

PR #474, opened 2024-06-27
Changing the map_range validate usage to MAPPING_SCOPED

josemonsalve2:
This PR changes the usage rules for the map_range function in the XRT driver. It uses `IREE_HAL_BUFFER_USAGE_MAPPING_SCOPED` rather than `IREE_HAL_BUFFER_USAGE_MAPPING` to relax the need for `IREE_HAL_BUFFER_USAGE_MAPPING_ACCESS_RANDOM`.


josemonsalve2:
I am not sure if I am missing something, but the IREE_HAL_BUFFER_USAGE_MAPPING_ACCESS_RANDOM does not seem to be necessary here. It is conservative, but the SCOPED will also require synchronization. I do not understand why it wouldn't be enough. Please feel free to reject this PR if there is a strong justification.

nirvedhmeshram (approved):
Thanks you are right, we would never do random access on the buffer from the host while the device is doing something on it. Thanks!

PR #475, opened 2024-06-27
Remove AIE Serializer

daveliddell:
As `AIESerializer` and `iree-aie-translate` are obsolete, this PR removes them.


yzhang93 (approved):
Thanks!

PR #478, opened 2024-06-28
Bump iree, mlir-air

daveliddell:
Bump iree to 4294a5b
Bump mlir-air to 4559217
mlir-aie remains unchanged, as mlir-air's pin of mlir-aie is unchanged.

Submodule changes that required adaptation in iree-amd-aie:

- AIR DmaToChannel pass moved from Conversion to Transform


yzhang93 (approved):
Thanks!

Issue #477, opened 2024-06-28
Update disabled NPU tests


daveliddell:
The following tests were awaiting an update to mlir-aie:

- iree/target/amd-aie/aie/test/aiert_insts.mlir
- iree/target/amd-aie/aie/test/dma_to_npu.mlir
- iree/target/amd-aie/aie/test/dma_to_npu_issue_token.mlir

That update has landed with PR https://github.com/nod-ai/iree-amd-aie/pull/465, but with PR https://github.com/nod-ai/iree-amd-aie/pull/460, these tests may now be obsolete or need updating.

daveliddell:
Hi @makslevental, would you kindly look over the disabled tests at your convenience and decide what you would like to do with them?

PR #479, opened 2024-06-28
Rename in-tree AIE passes to AMDAIE

makslevental:
This PR renames all of the in-tree `AIE*` passes to `AMDAIE*` in order to prevent collision with the temporarily added **out-of-tree** `AIE*` passes added in https://github.com/nod-ai/iree-amd-aie/pull/459. 

Note, in order to do this and actually change/affect/tweak the CLI args, you/we/I have to add the minimal amount of boilerplate that comes from tablegen for each pass.

Just want to stress/reiterate: this will **all** disappear completely as soon as we move to using **only** in-tree lowerings.

Fixes https://github.com/nod-ai/iree-amd-aie/issues/477 by moving them to use in-tree passes.


nirvedhmeshram:
@makslevental since you asked offline if I see any existing dependency issues. Just wanted to say I could not find any so I think this is safe to land.

PR #488, opened 2024-07-01
[AMD-AIE] Add support for vector.contract in insert-cores pass

Abhishek-Varma:
-- This commit adds suppport for vector.contract in `--iree-amdaie-insert-cores`
   pass.

Co-authored-by: Vivian Zhang <vivian.zhang@amd.com>

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCores.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCores.cpp
@@ -20,6 +20,7 @@
 #include "iree-amd-aie/Transforms/Passes.h"
 #include "iree-amd-aie/Transforms/Transforms.h"
 #include "iree/compiler/Codegen/TransformStrategies/GPU/Common.h"
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
 
 #define DEBUG_TYPE "iree-amdaie-insert-cores"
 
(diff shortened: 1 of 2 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCores.cpp
```diff
@@ -111,6 +112,16 @@ LogicalResult insertCoreOps(mlir::ModuleOp moduleOp) {
       } else if (auto linalgOp = dyn_cast<linalg::LinalgOp>(op)) {
         rewriter.setInsertionPoint(endOp);
         rewriter.moveOpBefore(linalgOp, endOp);
+      } else if (isa<vector::ContractionOp>(op)) {
+        // Because vector.contract op is surrounded by vectorized loop nest, we
+        // need to traverse to the outermost loop to move the entire vectorized
+        // computation within amdaie.core op.
+        Operation *currOp = op;
+        while (currOp->getParentOp() != forallOp) {
+          currOp = currOp->getParentOp();
```
nirvedhmeshram, line 121: I think you could just use currOp = forallOp? Also does it make sense to even continue the walk if the whole loop is already moved inside the amdaie.core op?

nirvedhmeshram: actually looking for at the code, I see you want and an op whose parent is the forallOp, but from the IR it looks like the consequence is the whole loop nest gets moved, so what I was suggesting would work too?

yzhang93 (commented):
Thanks! LGTM. I'd like @jtuyls to review this.

jtuyls (approved):
LGTM

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCores.cpp
Abhishek-Varma: Discussed offline. Marking resolved.

PR #490, opened 2024-07-01
Use only in-tree AIE/AIEX passes

makslevental:
Let's pull the bandaid off and use only in-tree passes. My apologies @nirvedhmeshram for giving you the run-around last week just do throw it away :(

Note, after this PR we should be fixing/patching in-tree passes first and upstreaming (those fixes) second.


nirvedhmeshram (approved):
Thanks!

newling (approved):

PR #493, opened 2024-07-02
[AMD-AIE] Add support for vectorization in distribute-cores-and-objectfifo

Abhishek-Varma:
-- This commit adds support for vectorized matmul in `--iree-amdaie-distribute-cores-and-objectfifos` pass.

Co-authored-by: Vivian Zhang <vivian.zhang@amd.com>

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
@@ -12,6 +12,7 @@
 #include "iree/compiler/Codegen/TransformStrategies/GPU/Common.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/SCF/Transforms/Transforms.h"
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
 #include "mlir/IR/Iterators.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
@@ -97,7 +98,12 @@ LogicalResult distributeLocalMemory(ModuleOp moduleOp) {
     if (dmaUsers.empty()) return WalkResult::advance();
     LLVM_DEBUG(llvm::dbgs() << "DMA users: " << dmaUsers.size() << "\n");
 
+    SmallVector<Operation *> allocOpUsers;
     for (Operation *userOp : allocOp->getUsers()) {
+      if (isa<memref::SubViewOp, vector::TransferReadOp,
+              vector::TransferWriteOp>(userOp)) {
+        allocOpUsers.push_back(userOp);
+      }
       auto subviewOp = dyn_cast<memref::SubViewOp>(userOp);
       if (!subviewOp) continue;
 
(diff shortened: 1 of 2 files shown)


newling (changes requested):
[cmake change] 

I prefer listing files in C++, especially if we own them. Mostly because you do things like comment them out, and find where they're used, more easily. 

In https://github.com/nod-ai/iree-amd-aie/pull/421 @makslevental used GLOB, but I thought that was ok because we don't really "own" those tests (and will remove them later, I think). 

When it comes to globing source files, the general consensus from what I've read on the internet (and been told previously when I've suggested using GLOB!) is that it's not best practice, either because it can slow down configuration time, or require additional invocations of cmake configuration to get the build to refresh correctly 

If you can point to a style guide for this project, or example of IREE doing this, I'll approve the change though.

Abhishek-Varma:
> I prefer listing files in C++, especially if we own them. Mostly because you do things like comment them out, and find where they're used, more easily.
> If you can point to a style guide for this project, or example of IREE doing this, I'll approve the change though.

Sure, my preferred way for the changes made in this revision majorly stems out of - adding/renaming/deleting passes involves the extra step of not only renaming the main file but also reflecting that in the CMakeLists.txt and re-arranging it within CMakeLists.txt as well.

The GLOB way is just a nice way to take the onus of adding/renaming/re-arranging/deleting away - we just focus on creating/deleting/renaming the pass and other infra should just adapt without us having to worry about the same (in my opinion). And currently all the passes we add/own are within `Transforms/` - so the changes made in this revision shouldn't really be an issue.

A similar reasoning can be extended for the `.mlir` lit test files as well.

There's no style guide that I'm aware of (yet) that prefers using either file names directly or the GLOB way.

I didn't understand the use-case of finding where a pass is used based on CMakeLists.txt - could you elaborate a bit more and help understand your point? ( We can discuss about this offline - as I'm deciding to remove the last two commits anyway based on the following point )

> require additional invocations of cmake configuration to get the build to refresh correctly

I understood this point well and agree on this as I experienced this previously with GLOB - I see this as a strong reason to remove the last two commits as this would be a heavy price to pay for being lazy (the adding/renaming/re-arranging/deleting aspect :-P)

newling:
Thank you @Abhishek-Varma

> I didn't understand the use-case of finding where a pass is used based on CMakeLists.txt 

I just meant finding which library a .cpp file is part of, or which test suite a .mlir file is in. You can just grep a file's name, find out which CMakeLists.txt file it is in

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -115,52 +121,71 @@ LogicalResult distributeLocalMemory(ModuleOp moduleOp) {
...
-        // We have to discard non-zero offsets as subview has been replaced by a
-        // dedicated allocated memref.
-        SmallVector<int64_t> allocShape(type.getShape());
-        (void)discardAllNonZeroOffsets<CopyOpOperateOn::Source>(
-            rewriter,
-            cast<AMDAIE::DoublyStridedOpInterface>(newDmaOp.getOperation()),
-            allocShape);
+    if (!memrefToNew.contains(allocOp)) return WalkResult::advance();
+
+    memref::AllocOp newAlloc = memrefToNew[allocOp];
```
jtuyls, line 128: Could you add a comment here to explain the next block?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -115,52 +121,71 @@ LogicalResult distributeLocalMemory(ModuleOp moduleOp) {
...
+    for (Operation *userOp : allocOpUsers) {
+      if (auto subviewOp = dyn_cast<memref::SubViewOp>(userOp)) {
+        rewriter.replaceAllUsesWith(subviewOp, newAlloc);
+        toBeErased.push_back(subviewOp);
+      } else if (auto transferReadOp =
+                     dyn_cast<vector::TransferReadOp>(userOp)) {
+        rewriter.setInsertionPoint(transferReadOp);
+        Value transferRead = rewriter.create<vector::TransferReadOp>(
```
jtuyls, line 136: Now, you have `transferReadOp` and `transferRead` and the difference between the two is unclear from the naming. So, I would call this `newTransferReadOp` instead.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -115,52 +121,71 @@ LogicalResult distributeLocalMemory(ModuleOp moduleOp) {
...
+            transferReadOp.getLoc(), transferReadOp.getType(), newAlloc,
+            transferReadOp.getIndices());
+        rewriter.replaceAllUsesWith(transferReadOp, transferRead);
+        toBeErased.push_back(transferReadOp);
+      } else if (auto transferWriteOp =
+                     dyn_cast<vector::TransferWriteOp>(userOp)) {
+        rewriter.setInsertionPoint(transferWriteOp);
+        rewriter.create<vector::TransferWriteOp>(
```
jtuyls, line 144: Don't we need a `rewriter.replaceAllUsesWith` here as well?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -238,6 +263,7 @@ class AMDAIEUnrollLocalLoops : public OpRewritePattern<scf::ForOp> {
   /// scf.for. This assumption could be removed in the future.
   template <typename Iterator>
   LogicalResult hoistDmaOps(PatternRewriter &rewriter, scf::ForOp forOp) const {
+    if (forOp->getParentOfType<AMDAIE::CoreOp>()) return failure();
```
jtuyls, line 266: Why is this change needed?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -491,20 +518,71 @@ LogicalResult insertLogicalObjectFifoAccess(ModuleOp moduleOp) {
...
                 std::get<1>(value));
-            linalgOp->setOperand(idx, accessOp);
-          } else if (auto type =
-                         llvm::dyn_cast<MemRefType>(operand.get().getType())) {
-            Value memref = operand.get();
+            op->setOperand(idx, accessOp);
+          } else if (isa<memref::AllocOp, memref::SubViewOp,
+                         vector::TransferReadOp, arith::ExtFOp>(
+                         operand.get().getDefiningOp())) {
+            OpOperand *opOperand = &operand;
```
jtuyls, line 525: What's needed for these vector ops that's different than for the linalg? I.e. can we not just walk all the vector read, write and contraction ops and perform the same operation as for the linalg, i.e. looping through the operands and inserting acces operations?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -491,20 +518,71 @@ LogicalResult insertLogicalObjectFifoAccess(ModuleOp moduleOp) {
...
+                  rewriter.getUnknownLoc(), logicalObjectFifo,
+                  AMDAIE::MemoryAccess::None);
+            }
+            newMemrefToLogicalObjectFifoAccess[memref] = accessOp;
+            opOperand->getOwner()->setOperand(opOperand->getOperandNumber(),
+                                              accessOp);
+          }
+        }
+
+        // Handle vector.transfer_write.
```
jtuyls, line 561: Could you clarify in the comment what's needed to handle `vector.transfer_write` specifically?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -480,9 +506,10 @@ LogicalResult insertLogicalObjectFifoAccess(ModuleOp moduleOp) {
     });
 
     WalkResult res = coreOp->walk([&](Operation *op) {
-      if (auto linalgOp = dyn_cast<linalg::LinalgOp>(op)) {
-        for (auto &&[idx, operand] :
-             llvm::enumerate(linalgOp->getOpOperands())) {
+      DenseMap<Value, AMDAIE::LogicalObjectFifoAccessOp>
+          newMemrefToLogicalObjectFifoAccess;
```
jtuyls, line 510: Could you add a comment explaining what this new map is used for? Also, why are the memref's considered `new`? They're just the memref operands, no?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -491,20 +518,71 @@ LogicalResult insertLogicalObjectFifoAccess(ModuleOp moduleOp) {
...
+              std::tuple<AMDAIE::LogicalObjectFifoFromMemrefOp,
+                         AMDAIE::MemoryAccess>
+                  value = memrefToLogicalObjectFifo[operand.get()];
+              accessOp = rewriter.create<AMDAIE::LogicalObjectFifoAccessOp>(
+                  rewriter.getUnknownLoc(), std::get<0>(value),
+                  std::get<1>(value));
+            }
+            transferWriteOp->setOperand(/*destOperandNumber=*/1, accessOp);
+          } else {
+            return WalkResult::interrupt();
```
jtuyls, line 585: Can you emit an error as well here, before the return?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -491,20 +518,71 @@ LogicalResult insertLogicalObjectFifoAccess(ModuleOp moduleOp) {
...
+                  AMDAIE::MemoryAccess::None);
+            }
+            newMemrefToLogicalObjectFifoAccess[memref] = accessOp;
+            opOperand->getOwner()->setOperand(opOperand->getOperandNumber(),
+                                              accessOp);
+          }
+        }
+
+        // Handle vector.transfer_write.
+        if (auto contractOp = dyn_cast<vector::ContractionOp>(op)) {
```
jtuyls, line 562: Can we not walk the `TransferWriteOp`'s directly and skip the `ContractionOp`?

jtuyls (changes requested):

newling (approved):
Thanks for changing cmake. Leaving to others to review actual pass (I don't know how to unrequest changes, so just approving)

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
Abhishek-Varma: `vector.transfer_write` doesn't return any value. So, it won't be needed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
Abhishek-Varma: The hoisting of DmaOp concept shouldn't be needed for a vectorized loop nest. We can bail out quite early for such cases.

I've now added that as a comment.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
Abhishek-Varma: You're correct. Very good point! I've adapted the code to do that and it looks concise/clean. Thanks!

Based on this I've marked other comments as resolved - since those snippets aren't needed anymore.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
Abhishek-Varma: Done.
Yes, they're memref operands - the `new` was added when I started working on the fix and because a "similar" map existed, I just added that as a prefix. :-P

I've updated the name now.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -115,52 +121,77 @@ LogicalResult distributeLocalMemory(ModuleOp moduleOp) {
...
-            rewriter,
-            cast<AMDAIE::DoublyStridedOpInterface>(newDmaOp.getOperation()),
-            allocShape);
+    if (!memrefToNew.contains(allocOp)) return WalkResult::advance();
+
+    // Traverse the following users of the AllocOp and replace uses of current
+    // AllocOp within them by a new AllocOp :-
+    // 1. vector.transfer_read.
+    // 2. vector.transfer_write.
+    // 3. memref.subview - only for this op we would replace its uses with a new
```
jtuyls, line 132: You also replace the uses for `vector::TransferReadOp`: https://github.com/nod-ai/iree-amd-aie/pull/493/files#diff-1bb5dc575fe681a4b367fc1f2515f08c6361b6b22c98bd7ef50a88efaed49cdbR145 ?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -479,32 +514,53 @@ LogicalResult insertLogicalObjectFifoAccess(ModuleOp moduleOp) {
...
                 value = memrefToLogicalObjectFifo[operand.get()];
             auto accessOp = rewriter.create<AMDAIE::LogicalObjectFifoAccessOp>(
                 rewriter.getUnknownLoc(), std::get<0>(value),
                 std::get<1>(value));
-            linalgOp->setOperand(idx, accessOp);
-          } else if (auto type =
-                         llvm::dyn_cast<MemRefType>(operand.get().getType())) {
+            memrefToLogicalObjectFifoAccess[operand.get()] = accessOp;
+            op->setOperand(idx, accessOp);
+          } else if (operand.get().getDefiningOp() &&
```
jtuyls, line 538: Why do we need to specifically check alloc ops?

I.e. why did you replace:

```
auto type = llvm::dyn_cast<MemRefType>(operand.get().getType())
```

with:

```
operand.get().getDefiningOp() && isa<memref::AllocOp>(operand.get().getDefiningOp())
```
?

The former looks more general?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
Abhishek-Varma: I've updated the comment to reflect what I'm doing better now. Could you check now?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
Abhishek-Varma: Forgot to revert to the previous condition check. Thanks for pointing that out!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
jtuyls: I see, thanks, this clarification is helpful

jtuyls (approved):
LGTM. Thanks.

Issue #485, opened 2024-06-28
Create global and local allocations


yzhang93:
The goal is to create local allocations which could be a reference to the global allocation for each tile to access the memory. The prototype is as below

We leave the `memref.alloc` op as a global variable:
`%alloc = memref.alloc() : memref<1x1x8x4x8x4xi32, 2 : i32>`

And create a new pointer pointing to the global allocation as
`%ptr = amdaie.addressof %alloc : memref<1x1x8x4x8x4xi32, 2 : i32> -> !amdaie.ptr`

To load the memory:
`%local_alloc = amdaie.load %ptr : !amdaie.ptr -> memref<1x1x8x4x8x4xi32, 2 : i32>

And then replace all uses of %alloc with %local_alloc.

@MaheshRavishankar Could you take a look and see if I understand this correctly?  And do I miss anything?
@jtuyls It would be helpful if you could add some context of how we are going to use the local allocations on different tiles.

MaheshRavishankar:
So we dont technically need a "global variable". We can leave the `alloc` as it is today, just move it to the top of the function. 



> And create a new pointer pointing to the global allocation as
> `%ptr = amdaie.addressof %alloc : memref<1x1x8x4x8x4xi32, 2 : i32> -> !amdaie.ptr`

you dont need a new type. You can just use
```
%memref_ref = amdaie.reference_to %alloc : memref<..., 2: i32>
```
`%memref_ref` is the same type as `%alloc`, but the reference is not hoistable across the `scf.forall` boundary. This will make the `amdaie.load` isntruction unnecessary. You just have to replace all `%alloc` with `%memref_ref`.

PR #497, opened 2024-07-02
Add pass to create references to allocations in L1

yzhang93:
- This PR is created to solve https://github.com/nod-ai/iree-amd-aie/issues/485.
- This pass should be added after all `BufferizeToAllocation` passes and before `HoistStaticallyBoundAllocations` pass. The generated IR after bufferization for a matmul example: https://gist.github.com/yzhang93/25288d8299e75aeba47775d761dbadbe.


compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
```diff
@@ -781,4 +781,26 @@ def AMDAIE_CircularDmaCpyNdOp: AMDAIE_DmaCpyNdBaseOp<"circular_dma_cpy_nd", [Pur
   let hasVerifier = 0;
 }
 
+def AMDAIE_ReferenceToOp: AMDAIE_Op<"reference_to"> {
+  let summary = "The operation that creates reference to a memref allocation";
+  let description = [{
+    This operation creates a reference to a global memref allocation. It has the same type
+    as the allocation it refers to, but has a local scope that is not hoistable across the AIE tiles.
```
jtuyls, line 788: Last week we said `is not hoistable across loops ('scf.forall/scf.for')`? Not sure what 'hoistable across the AIE tiles' means?

newling:
Can you please provide a motivation for this? Without context, 

```
 %4 = amdaie.reference_to %alloc_0 : memref<1x1x4x8x4x8xi32, 2 : i32> -> memref<1x1x4x8x4x8xi32, 2 : i32>
 ```
 
 doesn't appear to do anything.

File: compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td [hunk 2]
```diff
@@ -781,4 +781,26 @@ def AMDAIE_CircularDmaCpyNdOp: AMDAIE_DmaCpyNdBaseOp<"circular_dma_cpy_nd", [Pur
   let hasVerifier = 0;
 }
 
+def AMDAIE_ReferenceToOp: AMDAIE_Op<"reference_to"> {
+  let summary = "The operation that creates reference to a memref allocation";
+  let description = [{
+    This operation creates a reference to a global memref allocation. It has the same type
```
jtuyls, line 787: Maybe add rationale for what this is needed to the description

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECreateReferenceToAllocation.cpp
```diff
@@ -0,0 +1,69 @@
...
+      const AMDAIECreateReferenceToAllocationPass &pass){};
+  void runOnOperation() override;
+};
+
+void AMDAIECreateReferenceToAllocationPass::runOnOperation() {
+  MLIRContext *context = &getContext();
+  mlir::FunctionOpInterface funcOp = getOperation();
+  IRRewriter rewriter(context);
+
+  funcOp->walk([&](memref::AllocOp allocOp) {
```
jtuyls, line 39: What are the assumptions on the `memref.alloc` ops? We are assuming here that they are already completely hoisted to the global scope? I think we should make sure that the `reference_to` ops are inserted locally and while the `memref.alloc` ops should be global.

Related, I think we have some test cases with nested loops, even a peeled version, to make sure the output is in the shape we need it to be (references local, allocs global).

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECreateReferenceToAllocation.cpp
jtuyls: > What are the assumptions on the memref.alloc ops? We are assuming here that they are already completely hoisted to the global scope? I think we should make sure that the reference_to ops are inserted locally and while the memref.alloc ops should be global.

Nvm, I see this should be inserted before `HoistStaticallyBoundAllocations`

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
```diff
@@ -781,4 +781,26 @@ def AMDAIE_CircularDmaCpyNdOp: AMDAIE_DmaCpyNdBaseOp<"circular_dma_cpy_nd", [Pur
...
+    %1 = amdaie.reference_to %0 : memref<1x1x8x4x8x4xi32, 2 : i32>
+    ```
+
+  }];
+
+  let arguments = (ins AnyMemRef:$memref);
+
+  let results = (outs AnyMemRef:$output);
+
+  let assemblyFormat = "$memref attr-dict `:` type($memref) `->` type($output)";
```
Abhishek-Varma, line 803: Since it has the same type as the allocation it refers to, perhaps we can keep it just (even matches the example given in the doc comment of the op) ? :-
```
let assemblyFormat = "$memref attr-dict `:` type($output)";
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/test/roundtrip.mlir
```diff
@@ -272,3 +272,15 @@ func.func @workgroup() {
   }
   return
 }
+
+// -----
+
+// CHECK-LABEL: @reference_to()
+// CHECK:       %[[ALLOC:.+]] = memref.alloc()
+// CHECK:       amdaie.reference_to
+// CHECK-SAME:  %[[ALLOC]]
```
Abhishek-Varma, line 281: Any reason to match `%[[ALLOC]]` use via `CHECK-SAME` ?

I believe it can be combined with the previous `CHECK` statements as `amdaie.reference_to %[[ALLOC]]` ?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECreateReferenceToAllocation.cpp
yzhang93: Exactly. Could you check a complete IR here?https://gist.github.com/yzhang93/25288d8299e75aeba47775d761dbadbe.

jtuyls: Cool, maybe it's useful to add a test where we run both passes?

Abhishek-Varma (commented):
Nice start on this PR! A few comments from my end.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
MaheshRavishankar: I think it would work if you add the `SameOperandAndResultType` trait to the op definition

yzhang93: Thanks @MaheshRavishankar! It works with this trait.

jtuyls (approved):

PR #499, opened 2024-07-03
Use upstream Xilinx/aie-rt

makslevental:


jtuyls (commented):
Can we avoid adding these binary elf files used in the tests? E.g. generating them instead?

makslevental:
> Can we avoid adding these binary elf files used in the tests? E.g. generating them instead?

Obv I can but

1. Then I would also be testing peano and linking and the lowering pipelines;
2. In point of fact, one thing I was double checking here (obliquely) was that my local (up to date) peano and peano on the runner are producing valid code. Locally I could not (with this patch) get the matmul tests to pass numerics (and still cannot). So I generated these locally and watched to see whether the runner matmul tests would fail. I say obliquely because I really need to compare the elfs generated by the runner to these and/or execute these and check numerics.

Ok that I realize that doesn't make complete sense but basically what I'm saying is since all this stuff is so brittle, when something breaks, we have basically no way of knowing where exactly a break happens. Now usually you could/narrow the tests by reproducing The environment but the issue for this PR was **I have no idea what version of peano is on the runner** (that's why one of the commits was just to print time stamps for that directory, which only revealed it's from December). So the only way to debug that was this extremely narrow unit test. Going forward we should obviously pin to a hash of peano and that'll be possible soon (when they're satisfied with https://github.com/Xilinx/llvm-aie/pull/109). At that point I'll be able to delete these.

makslevental:
Will move stuff around to remove binaries from commit.

runtime/src/iree-amd-aie/aie_runtime/CMakeLists.txt
```diff
@@ -107,30 +106,252 @@ iree_install_targets(
...
 include(${_aie_rt_source_dir}/fal/cmake/collect.cmake)
-# We would prefer not to actually include the aie-rt directory
-# (i.e., prefer not to run their CMake) but there are configure/install
-# commands in the CMake there that perform the "installs" of the
-# build/include/xaiengine headers.
+set(XAIENGINE_BUILD_SHARED OFF CACHE BOOL "" FORCE)
 add_subdirectory(${_aie_rt_source_dir}/driver/src iree_aie_runtime)
 
-get_target_property(_aie_runtime_compile_options aienginev2 COMPILE_OPTIONS)
+# https://github.com/Xilinx/aie-rt/issues/4
```
nirvedhmeshram, line 115: Seems like it would be better to make a fork at nod-ai/aie-rt and carry patches especially since the upstream repository doesnt seem to have much activity and reviews from the repo maintainers? Can do it in a later PR if you think its a good idea though.

nirvedhmeshram (approved):
Thanks!

runtime/src/iree-amd-aie/aie_runtime/CMakeLists.txt
makslevental: Nah no way - My plan is to apply lots of serious pressure on that team to start accepting PRs through that repo. I have recruited @kumardeepakamd and others :)

makslevental: Just to emphasize/underscore _why_ not fork: that repo/utility is for all intents and purposes the canonical representation of the device. Well it's not because there are various forks around because the team is has drifted in and out of public activity but it absolutely _needs to be_ canonical. With the push to upstream drivers and such to Linux FSF, I think it can happen.

makslevental: Also btw the repo is active just the main/default branch isn't.

PR #501, opened 2024-07-03
Vendor `bootgen` and `xclbinutil`

makslevental:
This PR adds executable targets to our CMakes to build both `xclbinutil` and `bootgen`; previously these were fetched/gotten from a distro of `mlir-aie` and/or env. We already have these same submodules (`XRT` and `bootgen`) so the only thing we're really vendoring here is the CMake. The advantages of this are :

1. After this PR and a follow-up taking care of `me_basic.o` (once https://github.com/Xilinx/llvm-aie/pull/109 lands) we will no longer need a distro of `mlir-aie` (but obv still need the submodule for a little while longer);
    * Also we will be able to run lit tests for xclbin generation (i.e., during the build test step in CI);
3. We eliminate one dependency on OpenSSL (within `XRT`'s xclbinutil);
4. We will eventually move to not shelling out to these utilities and just calling the APIs directly (like [xaiepy](https://github.com/nod-ai/prototype-aie-toolchain/blob/main/xaiepy/xclbinutil.cpp#L58-L70)) so this was going to happen anyway[^1].

Note, all of the CMake shenanigans are very necessary because the respective code bases are quite questionable...............

Note also that it's clear `XCLBinGenConfig` needlessly duplicates many of the options in `AMDAIEOptions`. I'll get rid of the former in the a follow-up PR (see https://github.com/nod-ai/iree-amd-aie/issues/504).

[^1]: Indeed `bootgen` is already here as `bootgen-lib`.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
@@ -20,6 +20,8 @@ struct AMDAIEOptions {
   // TODO(MaheshRavishankar): Remove this dependency.
   std::string mlirAieInstallDir;
 
+  std::string amdAieInstallDir;
+
   // Path to Peano installation directory.
   std::string peanoInstallDir;
 
@@ -52,6 +54,11 @@ struct AMDAIEOptions {
         "iree-amd-aie-mlir-aie-install-dir", mlirAieInstallDir,
         llvm::cl::cat(category),
         llvm::cl::desc("Path to MLIR-AIE installation directory"));
+
+    binder.opt<std::string>(
+        "iree-amd-aie-install-dir", amdAieInstallDir, llvm::cl::cat(category),
+        llvm::cl::desc("Path to AMDAIE installation directory (typically the "
+                       "IREE install directory)"));
 
     binder.opt<std::string>(
         "iree-amd-aie-peano-install-dir", peanoInstallDir,
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/CMakeLists.txt
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/CMakeLists.txt
@@ -28,7 +28,7 @@ iree_cc_library(
 
 add_dependencies(iree_target_amd-aie_Target_AIETargets
         iree-amd-aie_aie_runtime_amdaie_xclbinutil
-        iree-amd-aie_aie_runtime_bootgen)
+        iree-amd-aie_aie_runtime_amdaie_bootgen)
 
 iree_cc_library(
   NAME
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/CMakeLists.txt
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/CMakeLists.txt
@@ -26,6 +26,10 @@ iree_cc_library(
     iree::target::amd-aie::Transforms
 )
 
+add_dependencies(iree_target_amd-aie_Target_AIETargets
+        iree-amd-aie_aie_runtime_amdaie_xclbinutil
+        iree-amd-aie_aie_runtime_bootgen)
+
 iree_cc_library(
   NAME
     Target
--- a/runtime/src/iree-amd-aie/aie_runtime/CMakeLists.txt
+++ b/runtime/src/iree-amd-aie/aie_runtime/CMakeLists.txt
@@ -22,6 +22,7 @@ cmake_minimum_required(VERSION 3.23)
 # We use our own, slightly modified, FindOpenSSL because of issues in
 # CMake's distribution of the same for versions prior to 3.29.
 # https://gitlab.kitware.com/cmake/cmake/-/issues/25702
+set(OPENSSL_USE_STATIC_LIBS TRUE CACHE BOOL "" FORCE)
 find_package(OpenSSL)
 if(NOT DEFINED OPENSSL_FOUND OR NOT ${OPENSSL_FOUND})
   list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}")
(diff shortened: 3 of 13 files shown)


makslevental:
After an inordinant amount of thrashing, this PR is ready for review @daveliddell @newling @nirvedhmeshram

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/CMakeLists.txt
```diff
@@ -26,6 +26,10 @@ iree_cc_library(
     iree::target::amd-aie::Transforms
 )
 
+add_dependencies(iree_target_amd-aie_Target_AIETargets
+        iree-amd-aie_aie_runtime_amdaie_xclbinutil
+        iree-amd-aie_aie_runtime_bootgen)
+
```
makslevental, line 32: One thing that I'm not handling correctly here is making these utilities available in the build directory (i.e., in `build/tools`). I tried to get it to work using IREE's CMake functions but couldn't get it. Can someone give me a tip on how this is done.

runtime/src/iree-amd-aie/aie_runtime/CMakeLists.txt
```diff
@@ -35,16 +36,47 @@ if(NOT DEFINED OPENSSL_FOUND OR NOT ${OPENSSL_FOUND})
...
 message(STATUS "OpenSSL include directories:" ${OPENSSL_INCLUDE_DIR})
 
+# https://stackoverflow.com/a/49216539/9045206
+macro(remove_flag_from_target _target _flag)
+    get_target_property(_target_cxx_flags ${_target} COMPILE_OPTIONS)
+    if(_target_cxx_flags)
+        list(REMOVE_ITEM _target_cxx_flags ${_flag})
+        set_target_properties(${_target} PROPERTIES COMPILE_OPTIONS "${_target_cxx_flags}")
+    endif()
+endmacro()
```
makslevental, line 46: Only a single use in this PR but will get more use in a follow-up.

runtime/src/iree-amd-aie/aie_runtime/NoOpXclBinSignature.cxx
```diff
@@ -0,0 +1,6 @@
+#include "XclBinSignature.h"
```
jtuyls, line 1: Should we add a Copyright?

jtuyls (commented):
LGTM, but maybe one nit below. Will leave it to others to explicitly approve

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
```diff
@@ -53,6 +55,11 @@ struct AMDAIEOptions {
         llvm::cl::cat(category),
         llvm::cl::desc("Path to MLIR-AIE installation directory"));
 
+    binder.opt<std::string>(
+        "iree-amd-aie-install-dir", amdAieInstallDir, llvm::cl::cat(category),
+        llvm::cl::desc("Path to AMDAIE installation directory (typically the "
+                       "IREE install directory)"));
+
```
nirvedhmeshram, line 62: It would be nice if we didnt have to add yet another flag to our compile flag. Especially this one from the user point of view will look unnecessary as it would be the same path they are running iree-compile from. But I am guessing this is both necessary and temporary so its okay. Another option is that we can expect the user to add certain directories to their PATH and make it a part of our README then this flag wont be needed but you can do whichever you think is better since if this is temporary it doesnt matter much.   However, we need to advertise this change in a few places before we can land this.
1. A full example iree-compile command using peano in the description of this PR.
2. In our e2e wiki here https://github.com/nod-ai/iree-amd-aie/wiki/End-to-end-walk-through#e2e-compilation-command
3. Any internal resources where we show an e2e command, @daveliddell can help with that one.
As long as 1. and 2. are done we are good to land this PR in my opinion.

nirvedhmeshram (approved):
Thanks! I have a minor request about documentation but looks good overall.
Also just making sure this is working e2e for some matmul example for Chess as well locally for you since we dont test that in CI but it is something we definitely care about.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
makslevental: > add certain directories to their PATH

I thought about this ofc and decided that this is the path to madness (that mlir-aie finds itself in with bash wrappers on wrappers on wrappers). And yes this flag is indeed temporary because the shell outs to `bootgen` and `amdaie_xclbinutil` will disappear.

Re updating the docs: this change is actually specifically/exactly to help @daveliddell put together e2e flows for Linux foudation dliverables. So I think he'll be able to tell me all the places that need to be updated.

runtime/src/iree-amd-aie/aie_runtime/NoOpXclBinSignature.cxx
makslevental: Doesn't hurt I guess

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
daveliddell: Yes, I will take responsibility for documenting the public dev build instructions and testing them.  I can also update the e2e wiki.  The playbook instructions will probably be in flux, as the public instructions should eventually replace them.  However, work remaining includes testing on Windows and coming up with more definitive machine (as opposed to workspace) setup instructions.

makslevental:
> Also just making sure this is working e2e for some matmul example for Chess as well locally for you since we dont test that in CI but it is something we definitely care about.

Chess shouldn't be affected at all here but sure I can test that locally.

makslevental:
Adding the full `iree-compile` command (that I basically pulled from `run_matmul_test.sh`):

```
iree-compile \
  test_matmuls/mm_test1_bf16_f32_m64_n64_k64_ir.mlir \
  --iree-hal-target-backends=amd-aie \
  --iree-amdaie-lower-to-aie-pipeline=air \
  --iree-amdaie-tile-pipeline=pad-pack \
  --iree-amd-aie-peano-install-dir=/tmp/llvm-aie \
  --iree-amd-aie-mlir-aie-install-dir=/tmp/mlir_aie \
  --iree-amd-aie-install-dir=/tmp/iree-install \
  --iree-amd-aie-vitis-install-dir=/opt/Xilinx/Vitis/2023.2 \
  --iree-hal-dump-executable-files-to=test_matmuls \
  --iree-amd-aie-show-invoked-commands \
  -o test_matmuls/mm_test1_bf16_f32_m64_n64_k64.vmfb
```

I'll note here though that these tests currently **do not pass with Peano tip** so the `iree-amd-aie-peano-install-dir=/tmp/llvm-aie` is the same that we have here in CI. This is next on my TODO list (update our CI to use tip Peano).

makslevental:
Doing 

```
export PATH=/opt/Xilinx/Vitis/2023.2/aietools/bin:$PATH
export VITIS=/opt/Xilinx/Vitis/2023.2
```

and adding `--iree-amd-aie-enable-chess=true` to the above command works.

Of course this is all very annoying but I have fixes/QoL improvements for this too in upcoming PRs.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
makslevental: All set - I'll hit `Save page` on the e2e docs as soon as this PR merges.

PR #503, opened 2024-07-04
Update mlir-air and mlir-aie version; add bf16 conv2d board test

erwei-xilinx:

PR #505, opened 2024-07-04
[AMD-AIE] Support implicit source/target addressing in amdaie.npu.dma_cpy_nd

Abhishek-Varma:
-- This commit adds a fix for supporting implicit source/target addressing in amdaie.npu.dma_cpy_nd ops as part of `--iree-amdaie-lower-to-aie` pass.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
@@ -366,14 +366,14 @@ LogicalResult circularDmaToAIE(IRRewriter &rewriter,
 namespace {
 
 /// Utility to get the static offsets, sizes and strides for
-/// `AIEX::NpuDmaMemcpyNdOp`.
-LogicalResult getStaticDims(Operation *op,
-                            const SmallVector<OpFoldResult> &offsets,
-                            const SmallVector<OpFoldResult> &sizes,
-                            const SmallVector<OpFoldResult> &strides,
-                            SmallVectorImpl<int64_t> &staticOffsets,
-                            SmallVectorImpl<int64_t> &staticSizes,
-                            SmallVectorImpl<int64_t> &staticStrides) {
+/// `AIEX::NpuDmaMemcpyNdOp` with explicit addressing.
+LogicalResult getStaticDimsForExplicitAddressing(
+    Operation *op, const SmallVector<OpFoldResult> &offsets,
+    const SmallVector<OpFoldResult> &sizes,
+    const SmallVector<OpFoldResult> &strides,
+    SmallVectorImpl<int64_t> &staticOffsets,
+    SmallVectorImpl<int64_t> &staticSizes,
+    SmallVectorImpl<int64_t> &staticStrides) {
   if (offsets.size() > staticOffsets.size()) {
     return op->emitError() << "size of `offsets` should be smaller or equal to "
                               "size of `staticOffsets`";
@@ -401,6 +401,31 @@ LogicalResult getStaticDims(Operation *op,
   return success();
 }
 
+/// Utility to get the static offsets, sizes and strides for
+/// `AIEX::NpuDmaMemcpyNdOp` with implicit addressing.
+LogicalResult getStaticDimsForImplicitAddressing(
+    Operation *op, MemRefType memrefType, SmallVectorImpl<int64_t> &staticSizes,
+    SmallVectorImpl<int64_t> &staticStrides) {
+  SmallVector<int64_t> shapeArr;
+  for (auto shape : memrefType.getShape()) shapeArr.push_back(shape);
+  int64_t sizeIndex = staticSizes.size() - 1;
+  for (int64_t i = shapeArr.size() - 1; i >= 0; i--) {
+    staticSizes[sizeIndex--] = shapeArr[i];
+  }
+  int64_t strideIndex = staticStrides.size() - 1;
+  // Since aiex.npu.dma_cpy_nd expects 3-dimensional stride attribute, it has
+  // the 4-th implicit dimension as 1. Hence we store that in `prevStride`.
+  int64_t prevStride = 1;
+  for (int64_t i = shapeArr.size() - 1; i >= 0; i--) {
+    staticStrides[strideIndex] = shapeArr[i] * prevStride;
+    prevStride = staticStrides[strideIndex--];
+  }
+  for (int64_t i = strideIndex; i >= 0; i--) {
+    staticStrides[i] = staticStrides[i + 1];
+  }
+  return success();
+}
+
 /// Convert the `amdaie.npu.dma_cpy_nd` operation to `aiex.npu.dma_memcpy_nd`.
 LogicalResult npuDmaCpyNdOpToAIE(IRRewriter &rewriter,
                                  AMDAIE::NpuDmaCpyNdOp dmaOp,
@@ -408,16 +433,47 @@ LogicalResult npuDmaCpyNdOpToAIE(IRRewriter &rewriter,
                                  IRMapping &mapper, IRMapping &bindingsMapper) {
   rewriter.setInsertionPoint(dmaOp);
   // Convert bidirectional `amdaie.npu.dma_cpy_nd` op into two halves.
-  if (dmaOp.hasSourceAddressing()) {
+
+  // TODO(avarma): The logic for checking implicit L3 addressing should just be
+  // part of AMDAIEOps.td. Confirm and address during PR.
+  MemRefType sourceMemrefType =
+      cast<LogicalObjectFifoType>(dmaOp.getDmaCpyNdOp().getSourceType())
+          .getElementType();
+  Attribute sourceMemSpaceAttr = sourceMemrefType.getMemorySpace();
+  MemRefType targetMemrefType =
+      cast<LogicalObjectFifoType>(dmaOp.getDmaCpyNdOp().getTargetType())
+          .getElementType();
+  Attribute targetMemSpaceAttr = targetMemrefType.getMemorySpace();
+  if (dmaOp.hasSourceAddressing() || !sourceMemSpaceAttr) {
+    // DmaOp has source addressing on L3 either explicitly or implicitly.
     SmallVector<Value> empty;
-    SmallVector<int64_t, 4> staticOffsets(4, 1);
-    SmallVector<int64_t, 4> staticSizes(4, 1);
-    SmallVector<int64_t, 4> staticStrides(4, 1);
-    if (failed(getStaticDims(dmaOp, dmaOp.getSourceMixedOffsets(),
-                             dmaOp.getSourceMixedSizes(),
-                             dmaOp.getSourceMixedStrides(), staticOffsets,
-                             staticSizes, staticStrides))) {
-      return failure();
+    SmallVector<int64_t, 4> staticOffsets;
+    SmallVector<int64_t, 4> staticSizes;
+    SmallVector<int64_t, 3> staticStrides;
+    if (dmaOp.hasSourceAddressing()) {
+      SmallVector<int64_t, 4> _staticOffsets(4, 1);
+      SmallVector<int64_t, 4> _staticSizes(4, 1);
+      SmallVector<int64_t, 3> _staticStrides(4, 1);
+      if (failed(getStaticDimsForExplicitAddressing(
+              dmaOp, dmaOp.getSourceMixedOffsets(), dmaOp.getSourceMixedSizes(),
+              dmaOp.getSourceMixedStrides(), _staticOffsets, _staticSizes,
+              _staticStrides))) {
+        return failure();
+      }
+      staticOffsets = _staticOffsets;
+      staticSizes = _staticSizes;
+      staticStrides = _staticStrides;
+    } else {
+      SmallVector<int64_t, 4> _staticOffsets(4, 0);
+      SmallVector<int64_t, 4> _staticSizes(4, 0);
+      SmallVector<int64_t, 3> _staticStrides(4, 1);
+      if (failed(getStaticDimsForImplicitAddressing(
+              dmaOp, sourceMemrefType, _staticSizes, _staticStrides))) {
+        return failure();
+      }
+      staticOffsets = _staticOffsets;
+      staticSizes = _staticSizes;
+      staticStrides = _staticStrides;
     }
 
     AMDAIE::CircularDmaCpyNdOp dmaCpyNd = dmaOp.getDmaCpyNdOp();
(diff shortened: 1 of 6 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
```diff
@@ -300,26 +300,37 @@ module {
...
-// CHECK-SAME:    issue_token = true
-// CHECK-SAME:    @[[OBJ1]]
-// CHECK-NEXT:    aiex.npu.dma_wait
-// CHECK-SAME:    @[[OBJ1]]
+// CHECK-SAME:            %[[ARG0]][1, 1, 0, 32][1, 1, 32, 32][1, 1, 64]
+// CHECK-SAME:            issue_token = true
+// CHECK-SAME:            metadata = @[[OBJ1]]
+// CHECK-NEXT:    aiex.npu.dma_wait {symbol = @[[OBJ1]]}
+// CHECK:         aiex.npu.dma_memcpy_nd
+// CHECK-SAME:            %[[ARG0]][0, 0, 0, 0][0, 0, 32, 64][2048, 2048, 64]
```
Abhishek-Varma, line 320: Hi @jtuyls - could you check if it looks okay to you?

The former one is for the explicit target addressing whereas the latter for implicit.

Similarly for the two cases below for source addressing.

jtuyls: Yes, this output looks good

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -403,23 +403,79 @@ LogicalResult getStaticDims(Operation *op,
...
 /// Convert the `amdaie.npu.dma_cpy_nd` operation to `aiex.npu.dma_memcpy_nd`.
 LogicalResult npuDmaCpyNdOpToAIE(IRRewriter &rewriter,
                                  AMDAIE::NpuDmaCpyNdOp dmaOp,
                                  SmallVector<Operation *> &toBeErased,
                                  IRMapping &mapper, IRMapping &bindingsMapper) {
   rewriter.setInsertionPoint(dmaOp);
   // Convert bidirectional `amdaie.npu.dma_cpy_nd` op into two halves.
-  if (dmaOp.hasSourceAddressing()) {
+
+  // TODO(avarma): The logic for checking implicit L3 addressing should just be
```
jtuyls, line 439: What do you mean with 'checking implicit L3 addressing'?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -403,23 +403,79 @@ LogicalResult getStaticDims(Operation *op,
...
-    SmallVector<int64_t, 3> staticStrides(3, 1);
-    if (failed(getStaticDims(dmaOp, dmaOp.getSourceMixedOffsets(),
-                             dmaOp.getSourceMixedSizes(),
-                             dmaOp.getSourceMixedStrides(), staticOffsets,
-                             staticSizes, staticStrides))) {
-      return failure();
+    SmallVector<int64_t, 4> staticOffsets;
+    SmallVector<int64_t, 4> staticSizes;
+    SmallVector<int64_t, 3> staticStrides;
+    if (dmaOp.hasSourceAddressing()) {
```
jtuyls, line 455: Is this underscore prefix a common pattern for variable naming in MLIR/LLVM codebases? I would avoid it as there are some C++ caveats around naming collisions with prefix underscores.

jtuyls: Also, why are these additional variables with leading underscores needed? Can't we just work with `staticOffsets` etc directly?

jtuyls (commented):
Not sure whether this is ready for review as you I saw a TODO which I think you want to address in this PR? 

Left a few preliminary comments and I think it would good to add an e2e test as well.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
Abhishek-Varma: So I was trying to refer to the following case :-
```
     %dma_target_l3 = amdaie.circular_dma_cpy_nd(%obj0[] [] [], %obj1[] [] []) : (!amdaie.logicalobjectfifo<memref<32x64xi32>>, !amdaie.logicalobjectfifo<memref<32x32xi32, 1>>)
    controlcode {
        %npu_dma_1 = amdaie.npu.dma_cpy_nd %dma_target_l3([] [] [], [] [] [])
        amdaie.npu.dma_wait(%npu_dma_1, S2MM)
    }
```

Where the access pattern is made implicit in `%npu_dma_1` and I'm trying to see through the `npu.dma_cpy_nd` source if the addressing on L3 is in source or target of `circular_dma_cpy_nd`.

Could you suggest a better way to word the same in comment?

Also, I wanted to confirm if the TODO's intention is valid - as in I wanted to have an API (defined in `AMDAIEOps.td`) within `npu.dma_cpy_nd` that'd answer if the op has implicit/explicit access/addressing patterns. If that's okay, I would like to add that within the same PR and use it in the appropriate place in the fix.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
Abhishek-Varma: Yes, good point - I shall address that.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
jtuyls: > Also, I wanted to confirm if the TODO's intention is valid - as in I wanted to have an API (defined in AMDAIEOps.td) within npu.dma_cpy_nd that'd answer if the op has implicit/explicit access/addressing patterns. If that's okay, I would like to add that within the same PR and use it in the appropriate place in the fix.

What's the difference between that new api and the existing `hasSourceAddressing`?

jtuyls: > Could you suggest a better way to word the same in comment?

I think the wording 'checking implicit L3 addressing' is ok! I am just not sure about what exact api you're thinking of?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -403,23 +403,79 @@ LogicalResult getStaticDims(Operation *op,
...
+    Operation *op, MemRefType memrefType, SmallVectorImpl<int64_t> &staticSizes,
+    SmallVectorImpl<int64_t> &staticStrides) {
+  SmallVector<int64_t> shapeArr;
+  for (auto shape : memrefType.getShape()) shapeArr.push_back(shape);
+  int64_t sizeIndex = staticSizes.size() - 1;
+  for (int64_t i = shapeArr.size() - 1; i >= 0; i--) {
+    staticSizes[sizeIndex--] = shapeArr[i];
+  }
+  int64_t strideIndex = staticStrides.size() - 1;
+  // Since aiex.npu.dma_cpy_nd expects 3-dimensional stride attribute, it has
```
jtuyls, line 418: Note the MLIR-AIE bump in this PR which will make the last dimension explicit: https://github.com/nod-ai/iree-amd-aie/pull/503

Abhishek-Varma: Thanks for pointing that out!
I'll accordingly make changes to this revision (and the lit test).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
jtuyls: I think `hasImplicitSourceAddressing()` should just be `!hasExplicitSourceAddressing()`. If you also want to check for L3 access, I think it would be better to make that explicit in the function name and imo better to make that a separate function: `hasSourceOnL3` or maybe better `int64_t getSourceMemorySpaceInt()`

Abhishek-Varma: Oh okay, cool.

Then I can add `getSourceMemorySpaceInt` and `getTargetMemorySpaceInt` instead for `npu.dma_cpy_nd` op and make use of those APIs here.

jtuyls: Btw, for consistency, I used `uint64_t getMemorySpaceAsUInt()` before: https://github.com/nod-ai/iree-amd-aie/blob/741facf6ab616529978f077bd75b1a941b7385ac/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td#L504. And also note that I used `0` for L3

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -401,23 +401,71 @@ LogicalResult getStaticDims(Operation *op,
...
+  for (unsigned i = 0, n = staticOffsets.size(); i < n; i++) {
+    staticOffsets[i] = 0;
+  }
+  // 2. Static sizes.
+  SmallVector<int64_t> shapeArr;
+  for (auto shape : memrefType.getShape()) shapeArr.push_back(shape);
+  int64_t sizeIndex = staticSizes.size() - 1;
+  for (int64_t i = shapeArr.size() - 1; i >= 0; i--) {
+    staticSizes[sizeIndex--] = shapeArr[i];
+  }
```
jtuyls, line 421: You can use `std::copy()` here.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -436,16 +484,28 @@ LogicalResult npuDmaCpyNdOpToAIE(IRRewriter &rewriter,
...
     SmallVector<Value> empty;
     SmallVector<int64_t, 4> staticOffsets(4, 1);
     SmallVector<int64_t, 4> staticSizes(4, 1);
-    SmallVector<int64_t, 4> staticStrides(4, 1);
-    if (failed(getStaticDims(dmaOp, dmaOp.getTargetMixedOffsets(),
-                             dmaOp.getTargetMixedSizes(),
-                             dmaOp.getTargetMixedStrides(), staticOffsets,
-                             staticSizes, staticStrides))) {
-      return failure();
+    SmallVector<int64_t, 3> staticStrides(4, 1);
```
jtuyls, line 492:
```suggestion
    SmallVector<int64_t, 4> staticStrides(4, 1);
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -401,23 +401,71 @@ LogicalResult getStaticDims(Operation *op,
...
     SmallVector<Value> empty;
     SmallVector<int64_t, 4> staticOffsets(4, 1);
     SmallVector<int64_t, 4> staticSizes(4, 1);
-    SmallVector<int64_t, 4> staticStrides(4, 1);
-    if (failed(getStaticDims(dmaOp, dmaOp.getSourceMixedOffsets(),
-                             dmaOp.getSourceMixedSizes(),
-                             dmaOp.getSourceMixedStrides(), staticOffsets,
-                             staticSizes, staticStrides))) {
-      return failure();
+    SmallVector<int64_t, 3> staticStrides(4, 1);
```
jtuyls, line 452:
```suggestion
    SmallVector<int64_t, 4> staticStrides(4, 1);
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -436,16 +484,28 @@ LogicalResult npuDmaCpyNdOpToAIE(IRRewriter &rewriter,
         empty, empty, staticOffsets, staticSizes, staticStrides,
         objFifo.getName(), 0, issueToken);
   }
-  if (dmaOp.hasTargetAddressing()) {
+  if (dmaOp.hasTargetAddressing() || dmaOp.getTargetMemorySpaceAsUInt() == 0) {
+    // DmaOp has target addressing on L3 either explicitly or implicitly.
```
jtuyls, line 488: You're not really ensuring here that the addressing is on L3 side. This condition would still succeed if `dmaOp.hasTargetAddressing() == true` and `dmaOp.getTargetMemorySpaceAsUInt() != 0`, so it might be good to reflect that in the comment.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -401,23 +401,71 @@ LogicalResult getStaticDims(Operation *op,
...
+
 /// Convert the `amdaie.npu.dma_cpy_nd` operation to `aiex.npu.dma_memcpy_nd`.
 LogicalResult npuDmaCpyNdOpToAIE(IRRewriter &rewriter,
                                  AMDAIE::NpuDmaCpyNdOp dmaOp,
                                  SmallVector<Operation *> &toBeErased,
                                  IRMapping &mapper, IRMapping &bindingsMapper) {
   rewriter.setInsertionPoint(dmaOp);
   // Convert bidirectional `amdaie.npu.dma_cpy_nd` op into two halves.
-  if (dmaOp.hasSourceAddressing()) {
+  if (dmaOp.hasSourceAddressing() || dmaOp.getSourceMemorySpaceAsUInt() == 0) {
```
jtuyls, line 447: You're not really ensuring here that the addressing is on L3 side. This condition would still succeed if `dmaOp.hasSourceAddressing() == true` and `dmaOp.getSourceMemorySpaceAsUInt() != 0`, so it might be good to reflect that in the comment.

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -401,23 +401,55 @@ LogicalResult getStaticDims(Operation *op,
...
 
+/// Utility to get the static sizes and strides for `AIEX::NpuDmaMemcpyNdOp`
+/// with implicit addressing.
+LogicalResult getStaticDimsForImplicitAddressing(
+    Operation *op, MemRefType memrefType, SmallVectorImpl<int64_t> &staticSizes,
+    SmallVectorImpl<int64_t> &staticStrides) {
+  // 1. Static sizes.
+  SmallVector<int64_t> shapeArr = llvm::to_vector(memrefType.getShape());
+  std::copy(shapeArr.begin(), shapeArr.end(),
+            staticSizes.end() - shapeArr.size());
```
jtuyls, line 412: I think we should check that `staticSizes` is large enough. And it would be good to add a test case for this as well.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -401,23 +401,55 @@ LogicalResult getStaticDims(Operation *op,
...
+  // 2. Static strides.
+  int64_t strideIndex = staticStrides.size() - 1;
+  // For contiguous access the strides for the last dimension is always 1.
+  int64_t prevStride = 1;
+  for (int64_t i = shapeArr.size() - 1; i >= 0; i--) {
+    staticStrides[strideIndex] = prevStride;
+    prevStride = staticStrides[strideIndex--] * shapeArr[i];
+  }
```
jtuyls, line 420: I think we should check sizes here as well to ensure `strideIndex` doesn't become negative? And add a testcase for this.

jtuyls (changes requested):

Abhishek-Varma:
Hi @jtuyls - I've addressed your changes.

Note: For the size/stride check only one lit test would suffice because staticSize/staticStride size is same - so we only need to check on shapeArr.size() being greater than that at the beginning of static metadata formation.

jtuyls:
> Hi @jtuyls - I've addressed your changes.
> 
> Note: For the size/stride check only one lit test would suffice because staticSize/staticStride size is same - so we only need to check on shapeArr.size() being greater than that at the beginning of static metadata formation.

That's not really guaranteed from the function's perspective right? Someone could change the sizes of the size and stride vectors so they're not the same anymore?

Abhishek-Varma:
> > Hi @jtuyls - I've addressed your changes.
> > Note: For the size/stride check only one lit test would suffice because staticSize/staticStride size is same - so we only need to check on shapeArr.size() being greater than that at the beginning of static metadata formation.
> 
> That's not really guaranteed from the function's perspective right? Someone could change the sizes of the size and stride vectors so they're not the same anymore? You could add an assert to ensure the sizes are the same.

Sure. I've added an assert now within the function.

jtuyls (approved):

Issue #513, opened 2024-07-08
Naming based issue with upstream iree in Bootgen


nirvedhmeshram:
I am having trouble lowering a simple torch dialect module like this through the backend
```
module @module {
func.func @matmul(%arg0: !torch.vtensor<[256,256],f32>, %arg1: !torch.vtensor<[256, 256],f32>) -> !torch.vtensor<[256, 256],f32> {
  %4 = torch.aten.mm %arg0, %arg1: !torch.vtensor<[256,256],f32>, !torch.vtensor<[256,256],f32>  -> !torch.vtensor<[256,256],f32>
  return %4 : !torch.vtensor<[256,256],f32>
}  
}
```
The reason is that we name the dispatch with a `$` sign like `'configured_module_matmul$async_dispatch_0.mlir'` and this is passed as a symbol to the aie.device op and then eventually to bootgen which doesnt like it
```
****** Bootgen v2024.1
  **** Build date : Jun 18 2024-14:12:05
    ** Copyright 1986-2022 Xilinx, Inc. All Rights Reserved.
    ** Copyright 2022-2024 Advanced Micro Devices, Inc. All Rights Reserved.

WARNING: bad character! '$' (0x24)

ERROR:BootGen - syntax error
   Line #9, "three_matmul_torch/module_torch.aten.cat$async_dispatch_0_amdaie_xclbin_fb/design.bif".
... h.aten.cat$async_dispatch_0_amdaie_xclbin_fb/aie_cdo_elfs.bin
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

nirvedhmeshram:
Although strictly bootgen should allow the `$` in the time being I will santize the names to not have the `$`

stephenneuendorffer:
This would seem to be more of a backend legalization question, rather than a fronted problem, though?

nirvedhmeshram:
> This would seem to be more of a backend legalization question, rather than a fronted problem, though?

Yeah, I will basically sanitize the names in the backend to legalize them.

makslevental:
> Although strictly speaking bootgen should allow the $ in the time being I will sanitize the names to not have the $

bootgen uses bison/yacc for various things (parsing/serializing) so I'm guessing the requirement comes from there (yacc/bison might not allow) and thus we don't have much choice here.

PR #521, opened 2024-07-09
Sanitize entrypoint and module names to support what bootgen allows

nirvedhmeshram:
An optional `--function` argument is added to the cpu_comparison ci test as in presence of $ signs the export name is not automatically detected.
Fixes: https://github.com/nod-ai/iree-amd-aie/issues/513


nirvedhmeshram:
I can see the error the CI is producing locally for a simple case, looks like `$` means something special to IREE so I will have to investigate this further.

newling (approved):
Thanks, LGTM

PR #525, opened 2024-07-10
Ensure input values are reproducible in tests

newling:
Useful for 2 reasons:
- Same numbers in CI and between users when a test fails
- Allows us (in the future) to safely cache the baseline/golden values for tests, which is useful for 2 reasons:
-- Don't need to rely on llvm-cpu for baseline/golden values
-- Can make testing faster (computing results for large convs/matmuls on CPU can be slow).


build_tools/ci/cpu_comparison/input_generator.py
```diff
@@ -95,22 +93,43 @@ def get_numpy_type(element_type):
         raise ValueError("Invalid or unsupported element type: " + element_type)
 
 
+def verify_determinism():
+    """
+    Assert that the approach we use is deterministic across space and time...
+    we don't want OS, numpy version, etc, influencing random values.
+    """
+    seed = 1
+    rng = np.random.Generator(np.random.MT19937(np.random.SeedSequence(1)))
```
makslevental, line 102: I think  you meant

```suggestion
    rng = np.random.Generator(np.random.MT19937(np.random.SeedSequence(seed)))
```

build_tools/ci/cpu_comparison/input_generator.py
```diff
@@ -95,22 +93,43 @@ def get_numpy_type(element_type):
...
     # Random integer values in range [lower_bound, upper_bound)
     # will be generated for the input data.
     lower_bound = 0
     upper_bound = 10
 
-    # Fix the seed for each input, based on the input number.
-    np.random.seed(1 + input_number)
+    verify_determinism()
+    # We have now verified that approach below is deterministic across platforms.
+    seed = input_number
```
makslevental, line 124: This was something I was thinking about when debugging flaky tests last week (something == the seed). Do we print/report this number (the seed) somewhere in the test logs/outputs?

makslevental: Yea I guess I noticed that it is fixed but I think the "right" approach is to "free" it but log/report the seed.

newling: So pass the seed in as an argument somewhere? One seed for each operand? Like, generate the numpy operands "standalone"? 

That's kind of what happens here: https://github.com/nod-ai/iree-amd-aie/blob/9f809c6c27730894275919347f2c44ff40e4a505/build_tools/ci/cpu_comparison/run_test.sh#L294 

You still need to run that script to create the input arrays

makslevental: yea sure cool good enough - so then we can move to "actually random in CI" + "print the seed for reproducibility" ?

newling: We've agreed to disagree for now on using different seeds between CI runs, but I've changed the PR now and made it easier for a developer to set the seed in any way they want now. @makslevental can you accept this PR if you're ok with the compromise!

makslevental: discussed offline - tabled until future bikeshedding sessions.

daveliddell: Wouldn't freeing the seed mean you can't tell whether you have a failure because your code change caused it or because it found a failing seed?  I would think that for CI, we'd want to prioritize detecting breaking code changes, while for nightly regressions, we'd want to do random seed testing.

daveliddell (approved):
LGTM!

build_tools/ci/cpu_comparison/input_generator.py
newling: Yeah that's what I think, except maybe even nightly regressions should keep frozen seeds -- IMO if you think the coverage of a test isn't good enough, increase the size of the random tensor rather than freeing the seed. Or improve the random distribution you're using.

PR #526, opened 2024-07-10
[WiP] Enable compilation with latest Peano

jsetoain:
There are (at least) two issues that prevent the current AMD AIE plugin to work with the latest version of Peano (LLVM AIE single core compiler):

* Peano now provides initialization code through crt0, so linking against `me_basic.o` is unnecessary and creates redefined symbols.

* Peano expect the entry point to be named `main`, not `_main`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,6 +112,7 @@ jobs:
           python3 -m venv .venv
           source .venv/bin/activate
           pip install https://github.com/Xilinx/mlir-aie/releases/download/latest-wheels/mlir_aie-0.0.1.2024070222+76460fe-py3-none-manylinux_2_35_x86_64.whl
+          pip install https://github.com/Xilinx/llvm-aie/releases/download/nightly/llvm_aie-18.0.0.2024071001+6bfc265c-py3-none-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl
 
           pip install -r tests/matmul/requirements.txt
 
--- a/build_tools/ci/cpu_comparison/run_test.sh
+++ b/build_tools/ci/cpu_comparison/run_test.sh
@@ -88,7 +88,7 @@ fi
 
 # Parameter 4) <peano-install-dir>
 if [ -z "${4-}" ]; then
-  PEANO=/opt/llvm-aie
+  PEANO=`realpath .venv/lib/python3.10/site-packages/llvm-aie`
 else
   PEANO=`realpath "$4"`
 fi
--- a/build_tools/ci/print_ir_aie2xclbin/print_ir_aie2xclbin.sh
+++ b/build_tools/ci/print_ir_aie2xclbin/print_ir_aie2xclbin.sh
@@ -39,7 +39,7 @@ mkdir -p ${OUTPUT}
 # The CI case:
 if [ "$#" -eq 2 ]; then
   echo "Assuming that this is the 'CI case' as 2 parameters were provided."
-  PEANO=/opt/llvm-aie
+  PEANO=`realpath .venv/lib/python3.10/site-packages/llvm-aie`
   XRT=/opt/xilinx/xrt
   VITIS=/opt/Xilinx/Vitis/2023.2
   MLIR_AIE=`realpath .venv/lib/python3.10/site-packages/mlir_aie`
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -96,7 +96,7 @@ fi
 
 # Parameter 4) <peano-install-dir>
 if [ -z "${4-}" ]; then
-  PEANO=/opt/llvm-aie
+  PEANO=`realpath .venv/lib/python3.10/site-packages/llvm-aie`
 else
   PEANO=`realpath "$4"`
 fi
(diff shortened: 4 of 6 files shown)


jsetoain:
I think this might be a good thing to have just in case folks want to check integration with Peano. With this patch, I'm able to compile and run the matmul examples using direct code generation with the latest Peano.

newling:
Very nice! Maybe we should 

1) install latest peano on CI
2) land this PR with (with an update to the CI files in iree-amd-aie to make them point at the new version installed in 1). 
3) update HOWTO and README

jsetoain:
> Very nice! Maybe we should
> 
>     1. install latest peano on CI
> 
>     2. land this PR with (with an update to the CI files in iree-amd-aie to make them point at the new version installed in 1).
> 
>     3. update HOWTO and README

Yes! Thanks! I also want to figure out what's the issue with μkernel code path; it compiles, but it hangs and times out when I try to run it. I have _no clue_ whatsoever...

jtuyls:
> > Very nice! Maybe we should
> > ```
> > 1. install latest peano on CI
> > 
> > 2. land this PR with (with an update to the CI files in iree-amd-aie to make them point at the new version installed in 1).
> > 
> > 3. update HOWTO and README
> > ```
> 
> Yes! Thanks! I also want to figure out what's the issue with μkernel code path; it compiles, but it hangs and times out when I try to run it. I have _no clue_ whatsoever...

Maybe we should wait with updating the CI machine until the ukernel works as well.

makslevental:
> Peano expect the entry point to be named main, not _main.

Here's the line you want to patch: 
https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIETargetLdScript.cpp#L161

jsetoain:
> Maybe we should wait with updating the CI machine until the ukernel works as well.

For sure, we won't land this until everything relevant is in working order, but it's useful to me (and I assume, others) to have an integration point with the latest Peano. When the time comes, I move this out of draft and it can land.

makslevental:
> I have no clue whatsoever...

It's gotta be the `_main`/`main` thing?

jsetoain:
> > Peano expect the entry point to be named main, not _main.
> 
> Here's the line you want to patch: https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIETargetLdScript.cpp#L161

Yup, that's there. I also patched a similar one in: https://github.com/nod-ai/iree-amd-aie/blob/9f809c6c27730894275919347f2c44ff40e4a505/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIETargetBCF.cpp#L122

Not sure that one's needed, but it's not breaking anything downstream and I don't want to forget about it.

jsetoain:
> > I have no clue whatsoever...
> 
> It's gotta be the `_main`/`main` thing?

It's for sure not that (or not only that). The `_main/main` change happened late last week, and I've been experiencing this for a couple of weeks at least, before that change.

makslevental:
> It's for sure not that (or not only that). The `_main/main` change happened late last week, and I've been experiencing this for a couple of weeks at least, before that change.

Only other thing I can think of is lock intrinsics are borked but then no ukernels would work?

newling:
> Screenshot from 2024-07-10 14-53-56
> 
> yea so this is what I was seeing (not a hang) when I was testing tip Peano (ie all zeros always).

With ukernel? I'd just like to confirm that the ukernel test failing is a CI test, because there are currently no safety rails with our ukernel pipeline and if the tiling isn't done "just right" then you get bad numbers / crashes.

jsetoain:
> yea so this is what I was seeing (not a hang) when I was testing tip Peano (ie all zeros always).

How do you know it's not timing out? When it hangs, the driver must have some sort of watchdog that ends the execution after a bit over 16s. The test ending and returning all zeroes does match observed behavior.

Anyhow, I guess there's no way around it, I'll have to take a deep dive into the µkernel architecture. It's very odd that just by updating the back-end compiler, the direct code generation works, but the µkernel doesn't. I suspect it's some silly convention change that needs to be accounted for in that pipeline.

makslevental:
> the driver must have some sort of watchdog

you sure about that? lol. 

anyway there is a timeout but it causes an error (easy to hit/see if you misconfigure a sync). that's not what you see here (unless our xrt bindings are failing gracefully? @nirvedhmeshram. 

anyway locally there's no lag/delay - the results just come back (quickly) as zeros.

makslevental:
dunno why I didn't think of doing this before but here's the info on which version is here in CI

```
(base) mlevental@mlevental-F7BSC:~/Downloads/LLVM-AIE/llvm-aie-3/llvm-aie/bin$ ./clang --version
clang version 17.0.0 (https://gitenterprise.xilinx.com/XRLabs/llvm-aie 5a044b2342f51cf939e032c349c91c2731c812a2)
```

makslevental:
https://github.com/Xilinx/mlir-aie/pull/1612

makslevental:
after fiddling with a related issue today, i believe the problem is since we're not linking `crt0`, we're not correctly setting the stack pointer https://github.com/Xilinx/llvm-aie/blob/aie-public/libc/startup/baremetal/aie2/crt0.S#L13. we need to rewrite the LD script emission (which is currently specifically tuned to chess) to be more like https://github.com/Xilinx/llvm-aie/blob/aie-public/libc/startup/baremetal/aie2/ldfile. I'll get to this soon.

makslevental:
Closed in favor of https://github.com/nod-ai/iree-amd-aie/pull/595

PR #528, opened 2024-07-10
Add AMDAIEDevice to the executable target config

jtuyls:
This enables passes pre-`aie.device` to retrieve the target AIE device, which I will use in later PRs to lookup hardware-related parameters from the `AMDAIEDeviceModel`.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
@@ -16,6 +16,7 @@
 #include "aie/Target/LLVMIR/Dialect/XLLVM/XLLVMToLLVMIRTranslation.h"
 #include "air/Dialect/AIR/AIRDialect.h"
 #include "air/Dialect/AIRRt/AIRRtDialect.h"
+#include "iree-amd-aie/IR/AMDAIEAttrs.h"
 #include "iree-amd-aie/IR/AMDAIEDialect.h"
 #include "iree-amd-aie/Transforms/Passes.h"
 #include "iree-dialects/Dialect/LinalgTransform/Passes.h"
@@ -47,6 +48,25 @@
 #define DEBUG_TYPE "aie-target"
 
 namespace mlir::iree_compiler::AMDAIE {
+
+/// Command line option for selecting the target AIE device.
+static llvm::cl::opt<AMDAIEDevice> clAMDAIETargetDevice(
+    "iree-amdaie-target-device",
+    llvm::cl::desc("Sets the target device architecture."),
+    llvm::cl::values(
+        clEnumValN(AMDAIEDevice::xcvc1902, "xcvc1902", "The xcvc1902 device"),
+        clEnumValN(AMDAIEDevice::xcve2302, "xcve2302", "The xcve2302 device"),
+        clEnumValN(AMDAIEDevice::xcve2802, "xcve2802", "The xcve2802 device"),
+        clEnumValN(AMDAIEDevice::npu1, "npu1", "Default Phoenix NPU"),
+        clEnumValN(AMDAIEDevice::npu1_1col, "npu1_1col",
+                   "Phoenix NPU with a single column"),
+        clEnumValN(AMDAIEDevice::npu1_2col, "npu1_2col",
+                   "Phoenix NPU with two columns"),
+        clEnumValN(AMDAIEDevice::npu1_3col, "npu1_3col",
+                   "Phoenix NPU with three columns"),
+        clEnumValN(AMDAIEDevice::npu1_4col, "npu1_4col",
+                   "Phoenix NPU with four columns")),
+    llvm::cl::init(AMDAIEDevice::npu1_4col));
 
 static llvm::cl::opt<std::string> clEnableAMDAIEUkernels(
     "iree-amdaie-enable-ukernels",
@@ -142,8 +162,10 @@ class AIETargetBackend final : public IREE::HAL::TargetBackend {
     auto addConfig = [&](StringRef name, Attribute value) {
       configItems.emplace_back(StringAttr::get(context, name), value);
     };
-    // Set target arch
-    addConfig("target_arch", StringAttr::get(context, "chip-tbd"));
+    // Set target device
+    addConfig(
+        "target_device",
+        StringAttr::get(context, AMDAIE::stringifyEnum(clAMDAIETargetDevice)));
     // Set microkernel enabling flag.
     addConfig("ukernels", StringAttr::get(context, clEnableAMDAIEUkernels));
     auto configAttr = b.getDictionaryAttr(configItems);
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
@@ -12,6 +12,16 @@
 
 namespace mlir::iree_compiler::AMDAIE {
 
+std::optional<AMDAIEDevice> getConfigAMDAIEDevice(
+    IREE::HAL::ExecutableTargetAttr targetAttr) {
+  if (!targetAttr) return std::nullopt;
+  auto config = targetAttr.getConfiguration();
+  if (!config) return std::nullopt;
+  std::optional<StringAttr> attr = config.getAs<StringAttr>("target_device");
+  if (!attr) return std::nullopt;
+  return AMDAIE::symbolizeEnum<AMDAIEDevice>(attr.value().getValue());
+}
+
 namespace {
 
 /// Generate a DenseMap key we can use for the element types (alternatives
@@ -23,7 +33,7 @@ constexpr uint32_t getElementTypeKey(uint32_t a, uint32_t b, uint32_t c) {
 
 /// Map from (LHS bitwidth, RHS bitwidth, Accumulator bitwidth) to the AIE
 /// instruction size (m, n, k) for the integer types with those bitwidths.
-const auto& getIntegerMatmulInstructionSizeMap() {
+const auto &getIntegerMatmulInstructionSizeMap() {
   // Sanity check.
   static_assert(getElementTypeKey(1, 2, 3) == 1 + 2 * 256 + 3 * 65536);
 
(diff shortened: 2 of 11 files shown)


makslevental:
I was wondering about how this could/should work. LGTM but can you get rid of https://github.com/nod-ai/iree-amd-aie/blob/main/runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h#L59-L68 then.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -48,6 +49,25 @@
...
+        clEnumValN(AMDAIEDevice::npu1, "npu1", "Default Phoenix NPU"),
+        clEnumValN(AMDAIEDevice::npu1_1col, "npu1_1col",
+                   "Phoenix NPU with a single column"),
+        clEnumValN(AMDAIEDevice::npu1_2col, "npu1_2col",
+                   "Phoenix NPU with two columns"),
+        clEnumValN(AMDAIEDevice::npu1_3col, "npu1_3col",
+                   "Phoenix NPU with three columns"),
+        clEnumValN(AMDAIEDevice::npu1_4col, "npu1_4col",
+                   "Phoenix NPU with four columns")),
+    llvm::cl::init(AMDAIEDevice::npu1_4col));
```
nirvedhmeshram, line 69: I am wondering that shouldn't the codegen tiling strategy decide this rather than having a seperate target for each?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
```diff
@@ -1,15 +1,23 @@
...
 
-// CHECK: module
+// expected-error @+1 {{No AMDAIEDevice found in the target attribute configuration}}
 module {
 }
 
 // -----
 
+// CHECK: module
+#executable_target_amdaie_xclbin_fb = #hal.executable.target<"amd-aie", "amdaie-xclbin-fb", {target_device = 8 : i32, ukernels = "none"}>
```
nirvedhmeshram, line 10: I think we should use the text of the target rather tahn `8` ?

tests/samples/matmul_peeled_objectfifo_e2e.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: iree-compile --iree-hal-target-backends=amd-aie --compile-to=executable-sources %s | iree-opt --pass-pipeline="builtin.module(hal.executable(hal.executable.variant(iree-hal-translate-target-executable-variants{target=amd-aie})))" --iree-amdaie-lower-to-aie-pipeline=objectFifo --iree-amdaie-tile-pipeline=pack-peel --split-input-file | FileCheck %s
+// RUN: iree-compile --iree-hal-target-backends=amd-aie --compile-to=executable-sources --iree-amdaie-target-device=npu1_4col %s | iree-opt --pass-pipeline="builtin.module(hal.executable(hal.executable.variant(iree-hal-translate-target-executable-variants{target=amd-aie})))" --iree-amdaie-lower-to-aie-pipeline=objectFifo --iree-amdaie-tile-pipeline=pack-peel --split-input-file | FileCheck %s
```
nirvedhmeshram, line 1: nit: its default value so not required?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
nirvedhmeshram: Especially because we may be mixing and matching dispatches that are 1,2,4 columns?

jtuyls: It's not just the number of columns, but also the physical device, for example strix would be `npu2_?` I believe. I guess the number of columns can be removed at some point.

tests/samples/matmul_peeled_objectfifo_e2e.mlir
jtuyls: I wanted to explicitly test it

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
jtuyls: How to do that?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
jtuyls: However, a bit more work is needed to decouple the device from the number of columns.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
makslevental: nah I disagree - if you use text you won't be able to interop with eventual C usages of the enum.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
makslevental: The number of columns in that name should be considered legacy (from mlir-aie) and will be deprecated. For us (in-tree, today) those enum variants are only queried in a single place: https://github.com/nod-ai/iree-amd-aie/blob/main/runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc#L444-L448

```
    case AMDAIEDevice::npu1:
    case AMDAIEDevice::npu1_1col:
    case AMDAIEDevice::npu1_2col:
    case AMDAIEDevice::npu1_3col:
    case AMDAIEDevice::npu1_4col:
```

so the only bit of work left is to promote `partitionNumCols, partitionStartCol` to be the canonical attributes (rather than `device`).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
nirvedhmeshram: I think it
1. makes the command line obscure for developers
2. different from every other iree backend that take a string attribute here. 
3. Leaking lower level workings that may change to high level API which will have to change each time.
Anyway looks like it will require changing the device attribute to string and then having a mapping function but you guys can make the call if you like it this way we can leave it.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
nirvedhmeshram: Ya npu2 as a target makes sense. I was only asking about the columns. Can codegen still make a 2 column design with the npu1_4col target? In that case this is okay.

nirvedhmeshram: I didnt see Max's reply so in that case I am thinking partitionNumCols, partitionStartCol will be choosen by our codegen startegy so this makes sense thanks.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
makslevental: I mean re CLI vs purely IR based I don't have a strong opinion. I was only talking about int vs string for the attribute. 

> different from every other iree backend that take a string attribute here.

Ignorant on that - if conventionally it's a string then I'm slightly more sympathetic

> Leaking lower level workings that may change to high level API which will have to change each time.

I guess that's true but I don't see it changing (you'd have to be a sadist to like reorder an enum :angry:). 

> then having a mapping function

High-level I think what I really don't like is the proliferation/profusion of these banal kinds of things (like enums) that aren't unified/have to be constantly synced up. Like we (via mlir-aie) already have this all over the place because stuff like `WireBundle`, `DMAChannel`, etc that they declared/defined as MLIR attributes when they already existed as perfectly good enums in `aie-rt`. So now we have stuff like [this in a bunch of places](https://github.com/nod-ai/iree-amd-aie/blob/main/runtime/src/iree-amd-aie/aie_runtime/test/DeviceModelTest.cpp#L33-L44). Now you could say "well just promote that to a header" but I'm working in the other direction (removing `WireBundle` and "canonicalizing" on `StrmSwPortType`).

nirvedhmeshram: > I guess that's true but I don't see it changing (you'd have to be a sadist to like reorder an enum 😠).
Wont it change when you decouple the row/columns ?

Anyway I would feel awkward telling the front end SHARK tem to use `8` if they want phoenix and `9` for strix but its NBD.

jtuyls: Based on the discussion here, I updated it to use a `StringAttr` purely for the IR, while keeping the `AMDAIEDevice` as is. Could you guys have another look?

nirvedhmeshram: Nice, I think this way it is string and is not talking on any extra mapping that Max was concerned about so I love it!

nirvedhmeshram (approved):
LGTM

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
makslevental: > Wont it change when you decouple the row/columns ?

No - the values would stay the same and still be required, just that `partitionStart` and `partitionCols` would _also_ be required.

> Anyway I would feel awkward telling the front end SHARK team to use `8` if they want phoenix and `9` for strix but its NBD.

Fair enough. Anyway Jorn threaded the needle so :partying_face:.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
```diff
@@ -12,6 +12,16 @@
...
 namespace mlir::iree_compiler::AMDAIE {
 
+std::optional<AMDAIEDevice> getConfigAMDAIEDevice(
+    IREE::HAL::ExecutableTargetAttr targetAttr) {
+  if (!targetAttr) return std::nullopt;
+  auto config = targetAttr.getConfiguration();
+  if (!config) return std::nullopt;
+  std::optional<StringAttr> attr = config.getAs<StringAttr>("target_device");
+  if (!attr) return std::nullopt;
+  return AMDAIE::symbolizeEnum<AMDAIEDevice>(attr.value().getValue());
```
makslevental, line 22: :sunglasses: yea I forgot mlir gives symbolize and stringify.

makslevental (approved):
LGTM!

PR #510, opened 2024-07-05
Cleanup `amdaie-objectFifo-stateful-transform`

makslevental:
This is a refactor prior to the follow-up removal of this and other passes (in order to reduce the diff in the follow-up). 
We remove `aie.shim_dma` support from the pass and `XFAIL`tests that are legacy/erroneous from `mlir-aie`. E.g., `aie.device(npu1_4col)` does not support `aie.shim_dma` (recall we use runtime instructions to program shim DMAs on IPU/NPU).

Note, there are a lot of lines changed; probably easiest to first skim the entire file as a whole rather than as a diff.

Feel free to nitpick/bikeshed - it'll help me "robustify" the final version in the follow-up.

For posterity/interested parties a rough outline of what `amdaie-objectFifo-stateful-transform` does:

1. `splitFifo`: "split" fifos (i.e., create more fifos) into producer and consumer fifos (attributes of those fifos are determined by arcane rules relating to "depth" and also number of "acquires");
    * Note this produces fifos from a tile to itself o.O (e.g., `aie.objectfifo @mem_in_0_cons(%tile_2_2, {%tile_2_2}, 2 : i32)`);
2. `createBuffersAndLocks`: as the name suggests, creates the `aie.buffer`s and `aie.lock`s corresponding to fifos (again, arcane rules);
3. `createFlowsAndTileDMAs`: as the name suggests, creates the `aie.mem`/`aie.memtile_dma` and `aie.flows`s corresponding to fifos (again, arcane rules);
4. `unrollForLoops`: unrolls inner loops containing `aie.objectfifo.acquire`s to LCM of corresponding fifo depths;
5. `replaceReleaseOp`: replaces `aie.objectfifo.release` with `aie.use_lock(..., Release)`;
6. `replaceObjectAcquireOp`: replaces `aie.objectfifo.acquire` with `aie.use_lock(..., Acquire)`;
7. replaces `aie.objectfifo.subview` with `aie.buffers` created in step 2;
8. replaces `aie.objectfifo` create ops (which are ultimately anchored on symbols) with `memref.global` ops with the same symbols.

Note, replace here doesn't mean the existing ops are erased (i.e., all the aforementioned `replace*` steps actually just add/insert ops). Thus

9. erases all the `aie.objectfifo*` ops.

PR #531, opened 2024-07-11
Cleanup `amdaie-create-pathfinder-flows`

makslevental:
This PR refactors/cleans up `amdaie-create-pathfinder-flows` with the same ultimate goal as the previous such clean up (https://github.com/nod-ai/iree-amd-aie/pull/510): removal and replacement with pure `aie-rt` calls against simple repesentations (no dialect ops). 

The bulk of the change here is moving all of the pure router stuff into `iree_aie_runtime`. I'm also removing some legacy stuff related to PL shim tiles ~~(a couple TODOs remaining on this)~~.

Note, if you're not famliiar with this pass you're better off reading the descriptions/discussion when I refactored this prior (to remove Boost dep) in [mlir-aie](https://github.com/Xilinx/mlir-aie/pull/761) or when I added an [ILP implementation of the same feature](https://github.com/Xilinx/mlir-aie/pull/761). If you want something official, you can look at [Frank–Wolfe algorithm](https://aequilibrae.com/python/V.0.7.0/traffic_assignment.html#frank-wolfe-fw) for traffic assignment[^1].

[^1]: If you squint you will see that perturbing the `*_COEFF`s in our impl is basically gradient descent.


compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferDescriptorIDs.cpp
```diff
@@ -132,7 +132,10 @@ struct AMDAIEAssignBufferDescriptorIDsPass : mlir::OperationPass<DeviceOp> {
           Block *nextBlock = block.getSuccessor(0);
           if (!blockBdIdMap.contains(nextBlock))
             assert(nextBlock->getOperations().size() == 1 &&
-                   isa<EndOp>(nextBlock->getOperations().front()) &&
+                   // for some reason i can't stick both of ops in a single
+                   // isa<...>
+                   (isa<EndOp>(nextBlock->getOperations().front()) ||
+                    isa<DMAStartOp>(nextBlock->getOperations().front())) &&
```
makslevental, line 138: just a fix from a branch in mlir-aie i had lying around.

compiler/plugins/target/AMD-AIE/aie/AMDAIECreatePathFindFlows.cpp
```diff
@@ -573,22 +158,25 @@ LogicalResult DynamicTileAnalysis::runAnalysis(DeviceOp &device) {
...
   }
 
   // add existing connections so Pathfinder knows which resources are
   // available search all existing SwitchBoxOps for exising connections
   for (SwitchboxOp switchboxOp : device.getOps<SwitchboxOp>()) {
     for (ConnectOp connectOp : switchboxOp.getOps<ConnectOp>()) {
-      if (!pathfinder->addFixedConnection(connectOp))
+      auto sb = connectOp->getParentOfType<SwitchboxOp>();
+      // TODO: keep track of capacity?
+      if (sb.getTileOp().isShimNOCTile()) continue;
```
jtuyls, line 173: Could you elaborate? What needs to happen for Shim tiles?

compiler/plugins/target/AMD-AIE/aie/AMDAIECreatePathFindFlows.cpp
```diff
@@ -1359,40 +568,38 @@ bool AMDAIEPathfinderPass::reconnectConnectOps(const OpBuilder &builder,
...
           connect.getSourceChannel() == problemChan) {
         candidate = connect;
         foundCandidateForFixup = true;
       }
       if (connect.getSourceBundle() == problemBundle &&
-          connect.getSourceChannel() == emptyChan) {
+          connect.getSourceChannel() == emptyChan)
         hasEmptyChannelSlot = false;
-      }
     }
```
jtuyls, line 597: Style guide says to keep braces here: https://google.github.io/styleguide/cppguide.html#Formatting_Looping_Branching

compiler/plugins/target/AMD-AIE/aie/AMDAIECreatePathFindFlows.cpp
```diff
@@ -1320,36 +529,36 @@ bool AMDAIEPathfinderPass::attemptFixupMemTileRouting(
...
     return false;  // Problem is not about n-s routing
   int problemSouthChannel;
-  if (problemConnect.getSourceBundle() == WireBundle::South) {
+  if (problemConnect.getSourceBundle() == StrmSwPortType::SOUTH)
     problemSouthChannel = problemConnect.getSourceChannel();
-  } else if (problemConnect.getDestBundle() == WireBundle::South) {
+  else if (problemConnect.getDestBundle() == StrmSwPortType::SOUTH)
     problemSouthChannel = problemConnect.getDestChannel();
-  } else
+  else
```
jtuyls, line 543: Style guide says to keep braces: https://google.github.io/styleguide/cppguide.html#Formatting_Looping_Branching

compiler/plugins/target/AMD-AIE/aie/test/unit_broadcast.mlir
```diff
@@ -24,17 +24,17 @@
 // CHECK:             aie.connect<East : 0, DMA : 0>
 // CHECK:           }
 // CHECK:           %[[SWITCHBOX_2_0:.*]] = aie.switchbox(%[[TILE_2_0]]) {
-// CHECK:             aie.connect<South : 3, North : 0>
 // CHECK:             aie.connect<South : 3, East : 0>
+// CHECK:             aie.connect<South : 3, North : 0>
```
jtuyls, line 28: `CHECK-DAG` would avoid these reordering issues.

compiler/plugins/target/AMD-AIE/aie/test/unit_simple_flows_shim.mlir
```diff
@@ -1,30 +1,6 @@
 
 // RUN: iree-opt --split-input-file --amdaie-create-pathfinder-flows %s | FileCheck %s
 
-// CHECK-LABEL:   aie.device(xcvc1902) {
```
jtuyls, line 4: Why remove this test?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
```diff
@@ -0,0 +1,319 @@
...
+                             maxCapacity);
+          (void)graph.connect(westernNeighbor, thisNode, edges.back());
+        }
+      }
+    }
+  }
+}
+
+// Add a flow from src to dst can have an arbitrary number of dst locations due
+// to fanout.
```
jtuyls, line 75:
```suggestion
/// Add a flow from src to dst can have an arbitrary number of dst locations due
/// to fanout.
```

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
```diff
@@ -0,0 +1,319 @@
...
+
+#define DEBUG_TYPE "iree-aie-runtime-router"
+
+#define OVER_CAPACITY_COEFF 0.02
+#define USED_CAPACITY_COEFF 0.02
+#define DEMAND_COEFF 1.1
+
+namespace mlir::iree_compiler::AMDAIE {
+
+void Pathfinder::initialize(int maxCol, int maxRow,
```
jtuyls, line 24: Add docstring.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
```diff
@@ -0,0 +1,319 @@
...
+      std::find_if(graph.begin(), graph.end(), [&](const SwitchboxNode *sb) {
+        return sb->col == dstCoords.col && sb->row == dstCoords.row;
+      });
+  assert(matchingDstSb != graph.end() && "didn't add flow destinations");
+  flows.push_back({PathEndPointNode{*matchingSrcSb, srcPort},
+                   std::vector<PathEndPointNode>{{*matchingDstSb, dstPort}}});
+}
+
+// Keep track of connections already used in the AIE; Pathfinder algorithm will
+// avoid using these.
```
jtuyls, line 112:
```suggestion
/// Keep track of connections already used in the AIE; Pathfinder algorithm will
/// avoid using these.
```

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
```diff
@@ -0,0 +1,319 @@
...
+           ch.bundle == getConnectingBundle(sourceBundle);
+  });
+  if (matchingCh != edges.end())
+    return matchingCh->fixedCapacity.insert(connectOp.src.channel).second ||
+           true;
+
+  return false;
+}
+
+static constexpr double INF = std::numeric_limits<double>::max();
```
jtuyls, line 142: Move to top of file?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
```diff
@@ -0,0 +1,319 @@
...
+      // add this flow to the proposed solution
+      routingSolution[src] = switchSettings;
+    }
+  } while (!isLegal());  // continue iterations until a legal routing is found
+
+  return routingSolution;
+}
+
+std::optional<std::map<PathEndPoint, SwitchSettings>> findPaths(
+    int maxIterations);
```
jtuyls, line 309: What does this do?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.h
```diff
@@ -0,0 +1,147 @@
...
+  static nodes_iterator nodes_end(
+      mlir::iree_compiler::AMDAIE::SwitchboxGraph *DG) {
+    return DG->end();
+  }
+};
+}  // namespace llvm
+
+namespace mlir::iree_compiler::AMDAIE {
+
+struct Pathfinder {
```
jtuyls, line 128: Could use a docstring

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -60,6 +60,97 @@ struct TileLoc {
...
+  Switchbox sb;
+  Port src;
+  Port dst;
+
+  bool operator==(const Connect& rhs) const {
+    return std::tie(src, dst) == std::tie(rhs.src, rhs.dst);
+  }
+};
+
+struct Channel {
```
jtuyls, line 106: What's the difference between `Channel` and `DMAChannel`. This can be confusing.

compiler/plugins/target/AMD-AIE/aie/AMDAIECreatePathFindFlows.cpp
makslevental: The next line tries to add an existing (i.e., hand-written) connection and errors if that fails. I don't know/remember why shim noc tiles are/were treated differently (they're given special treatment in various places for unclear reasons) but the TODO was the thought that even if we don't try to add those connections (as actual routes) we might consider checking if capacity has been exceeded.  

Regardless this doesn't apply to us because we never handwrite anything (so the loop could be deleted) but you might at some point want to (in order to debug?). Figuring out why/how shim nocs are treated differently I'll leave to this [follow-up](https://github.com/nod-ai/iree-amd-aie/issues/540) that I wasn't planning but is now necessary since we want to use control packets.

compiler/plugins/target/AMD-AIE/aie/AMDAIECreatePathFindFlows.cpp
makslevental: I guess that style guide says always keep braces? MLIR's convention is to drop them for single line conditonals and loops. If you really want I'll change it back but I think keeping things from being stretched out is beneficial (very small benefit but nonetheless).

compiler/plugins/target/AMD-AIE/aie/test/unit_broadcast.mlir
makslevental: image

compiler/plugins/target/AMD-AIE/aie/test/unit_simple_flows_shim.mlir
makslevental: It uses PLIO: `aie.flow(%t23, North : 0, %t22, PLIO : 0)`

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
makslevental: The whole algo is basically 

1. try to route using dijkstra
2. succes? cool
3. fail and less than max iterations ? dump congestion coefficients goto 1 else bail and report error

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
makslevental: Channel is a connection between switchboxes, DMAChannel is a dma port (ie direction plus index). I don't remember if these were the names prior - feel free to suggest new names.

makslevental:
Let's not do too much more here - I wasn't planning on refactoring this further prior to the removal but after seeing https://github.com/Xilinx/mlir-aie/pull/1614 I decided to incorporate those changes too (so there will be another refactor here soon).

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
jtuyls: I meant, isn't this the same function as above?

compiler/plugins/target/AMD-AIE/aie/AMDAIECreatePathFindFlows.cpp
jtuyls: > I guess that style guide says always keep braces? 

Nope, it doesn't. 

```
// OK - fits on one line.
if (x == kFoo) { return new Foo(); }

// OK - braces are optional in this case.
if (x == kFoo) return new Foo();

// OK - condition fits on one line, body fits on another.
if (x == kBar)
  Bar(arg1, arg2, arg3);

// Bad - `if ... else` statement is missing braces.
if (x) DoThis();
else DoThat();

// Bad - `do ... while` statement is missing braces.
do DoThis();
while (x);
```

IREE follows this style guide, so I think we should be consistent on this.

>  I think keeping things from being stretched out is beneficial (very small benefit but nonetheless).

In the case of  of if/else there's no stretching out by adding braces?

compiler/plugins/target/AMD-AIE/aie/test/unit_broadcast.mlir
jtuyls: `sed -i 's/CHECK/CHECK-DAG/g' unit_broadcast.mlir`

jtuyls: Not saying to make the change in this PR though..

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
jtuyls: `SwitchBoxConnection`, `SwitchConnection`?

compiler/plugins/target/AMD-AIE/aie/AMDAIECreatePathFindFlows.cpp
```diff
@@ -573,22 +158,25 @@ LogicalResult DynamicTileAnalysis::runAnalysis(DeviceOp &device) {
...
     pathfinder->addFlow(srcCoords, srcPort, dstCoords, dstPort);
   }
 
   // add existing connections so Pathfinder knows which resources are
   // available search all existing SwitchBoxOps for exising connections
   for (SwitchboxOp switchboxOp : device.getOps<SwitchboxOp>()) {
     for (ConnectOp connectOp : switchboxOp.getOps<ConnectOp>()) {
-      if (!pathfinder->addFixedConnection(connectOp))
+      auto sb = connectOp->getParentOfType<SwitchboxOp>();
+      // TODO: keep track of capacity?
```
jtuyls, line 172:
```suggestion
      // Don't add handwritten connections for Shim NOC tiles.
      // TODO: keep track of capacity so that even if we don't try to add those
      // connections (as actual routes) we might consider checking if capacity 
      // has been exceeded.
```

compiler/plugins/target/AMD-AIE/aie/test/unit_broadcast.mlir
makslevental: You can't check-dag everywhere- otherwise you'll allow these to float between switchboxes (that's why I didn't do it last time I wrote these either)

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
makslevental: Lolwut I have no idea how that happened. Will delete.

compiler/plugins/target/AMD-AIE/aie/test/unit_broadcast.mlir
jtuyls: Yeah, fair enough. Doing it completely right would indeed be some work

compiler/plugins/target/AMD-AIE/aie/AMDAIECreatePathFindFlows.cpp
makslevental: > This exception does not apply to multi-keyword statements like if ... else or do ... while.

So basically only single `if`s right?

jtuyls (approved):
LGTM

Issue #520, opened 2024-07-09
Replace `AMDAIETargetModel` with `AMDAIEDeviceModel `


jtuyls:
Replace the constants in https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIETargetModel.h with calls to the `AMDAIEDeviceModel ` once https://github.com/nod-ai/iree-amd-aie/pull/462 has been merged.

makslevental:
Can go here https://github.com/nod-ai/iree-amd-aie/blob/e4f47318470cf6295068592005f47d6f97bdda10/runtime/src/iree-amd-aie/aie_runtime/xaie_hwcfg.c

jtuyls:
Done with https://github.com/nod-ai/iree-amd-aie/pull/534

PR #534, opened 2024-07-11
Replace AMDAIETargetModel with AMDAIEDeviceModel

jtuyls:
Addresses: https://github.com/nod-ai/iree-amd-aie/issues/520


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaLoopSubsumption.cpp
```diff
@@ -349,22 +386,29 @@ class SubsumeLoopIntoDMA
     //   be used for DMA access patterns.
     size_t sourceMaxNbDims{0};
     size_t targetMaxNbDims{0};
+    size_t shimNbIntraDims = (size_t)deviceModel.getDmaProp<uint8_t>(
```
newling, line 389:
```suggestion
    uint8_t shimNbIntraDims = deviceModel.getDmaProp<uint8_t>(
```
Doesn't this work? Seems like then you add a uint8_t to a size_t, which the compiler shouldn't warn about ever as there's no risk

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaLoopSubsumption.cpp
```diff
@@ -33,6 +34,33 @@
...
+// 2. Inter-iteration access pattern. Specified via an iteration 'stride',
+// 'size' and 'current_iteration' ('stride' is the same as 'stepsize' and 'size'
+// is the same as 'wrap' in buffer descriptor lingo). Here, 'current_iteration'
+// keeps track of the current execution iteration of the buffer descriptor and
+// is incremented after buffer descriptor execution. the 'stride' is the offset
+// to be used for each execution of the buffer descriptor, relative to the
+// previous one. When 'iteration_current' is equal to 'size', the
+// 'iteration_current' is reset to zero.
+//
+// Although DMAs will can a different number of intra-iteration dimensions, all
```
newling, line 60:
```suggestion
// Although DMAs can have a different number of intra-iteration dimensions, all
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaLoopSubsumption.cpp
```diff
@@ -33,6 +34,33 @@
...
+// DMAs have a single inter-iteration dimension for now.
```
newling, line 61:
```suggestion
// DMAs have a single inter-iteration dimension (at least in AIE2 and AIE2p)
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaLoopSubsumption.cpp
```diff
@@ -329,6 +357,15 @@ class SubsumeLoopIntoDMA
...
   LogicalResult matchAndRewrite(AMDAIE::DoublyStridedOpInterface op,
                                 PatternRewriter &rewriter) const override {
+    // Get the device model.
+    auto targetAttr = IREE::HAL::ExecutableTargetAttr::lookup(op);
+    std::optional<AMDAIEDevice> device = getConfigAMDAIEDevice(targetAttr);
+    if (!device)
+      return op.emitOpError()
+             << "No AMDAIEDevice found in the target attribute configuration";
+    AMDAIE::AMDAIEDeviceModel deviceModel =
+        AMDAIE::getDeviceModel(AMDAIE::AMDAIEDevice::npu1_4col);
```
newling, line 367: afaict we're not yet at the point of thinking about supporting different array sizes and architectures, but maybe we should say that explicitly in a meeting (or something)

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -68,6 +68,33 @@ enum class AMDAIETileType : uint8_t {
   MAX = 4
 };
 
+/// Enum of DMA properties, using the offset within the `XAie_DmaMod` structure
+/// as underlying value.
```
newling, line 72: Can you motivate why this isn't just 0,1,2,3,... in the comment ? This design (I've not seen it before) makes me slightly nervous about layouts changing depending on the compiler. That said, I do like how it ties the underlying class in. What is MAX useful for? These are definitely not static members of XAie_DmaMod (seems like that'd be UB https://en.cppreference.com/w/cpp/named_req/StandardLayoutType)

makslevental: > What is MAX useful for?

:(

newling: lol not THAT max

jtuyls: > Can you motivate why this isn't just 0,1,2,3,... in the comment ? This design (I've not seen it before) makes me slightly nervous about layouts changing depending on the compiler.

Could you elaborate on what you mean with layouts changing depending on the compiler? The idea of using `offsetof` is exactly to accommodate the XAie struct changing. 

> What is MAX useful for?

It can be useful to perform checks.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaLoopSubsumption.cpp
jtuyls: This is a mistake, I meant to do:

```
AMDAIE::AMDAIEDeviceModel deviceModel = AMDAIE::getDeviceModel(device.value());
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaLoopSubsumption.cpp
jtuyls: That works as well

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaLoopSubsumption.cpp
jtuyls: Thanks, updated

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaLoopSubsumption.cpp
jtuyls: Thanks, updated

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
newling: I'm not sure if the compiler can rearrange the members to improve alignment. For example in

```
struct X{
 char a; 
 float b;
 char c;
uint16_t d;
 };
 ```
 
 Can the compiler effectively rearrange this so that it is 
 
 ```
 struct X {
 float b;
 char a; 
 char c;
 uint16_t d;
  };
 
 So that b is not offset by 1 byte.

newling (approved):
LGTM. I'm not 100% sure that `enum class AMDAIEDmaProp` is completely future proof, but I guess it reduces boilerplate and maintenance for now, so I'm onboard

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
jtuyls: I see, this is not allowed I believe. Would introduce many issues if it were.

newling: Ok I did a bit of reading and indeed it's not allowed. Padding is allowed, but no reording.

jtuyls: Yeah, from the `offsetof` doc:

> The macro offsetof expands to an integral constant expression of type [std::size_t](https://en.cppreference.com/w/cpp/types/size_t), the value of which is the offset, in bytes, from the beginning of an object of specified type to its specified subobject, including [padding bits](https://en.cppreference.com/w/cpp/language/object#Object_representation_and_value_representation) if any.

> Given an object o of type type and static storage duration, o.member shall be an lvalue constant expression that refers to a subobject of o. Otherwise, the behavior is undefined. Particularly, if member is a [static data member](https://en.cppreference.com/w/cpp/language/static), a [bit-field](https://en.cppreference.com/w/cpp/language/bit_field), or a [member function](https://en.cppreference.com/w/cpp/language/member_functions), the behavior is undefined.

>If type is not a [PODType](https://en.cppreference.com/w/cpp/named_req/PODType)(until C++11)[standard-layout](https://en.cppreference.com/w/cpp/language/data_members#Standard-layout) type(since C++11), the result of offsetof is undefined(until C++17)use of the offsetof macro is conditionally-supported(since C++17).

>
> The expression offsetof(type, member) is never [type-dependent](https://en.cppreference.com/w/cpp/language/dependent_name#Dependent_types) and it is value-dependent if and only if type is dependent.

PR #535, opened 2024-07-11
[xclbingen] check peano version for whether to use me_basic.o

makslevental:
Stop-gap measure to support both CI peano and tip peano (goes away after https://github.com/nod-ai/iree-amd-aie/pull/526).


daveliddell (approved):
LGTM!  Thanks, Max!

daveliddell:
Tested locally with recent version of peano.  Works great!

PR #538, opened 2024-07-12
[AMD-AIE][Objectfifo] Add support for ukernel in insert-cores pass

Abhishek-Varma:
-- This commit is the first fix aimed towards enabling ukernel support in objectfifo backend.
-- It does so by ensuring that the information carried in the ukernel's function declaration is relayed over to amdaie.core op when it gets created as part of `--iree-amdaie-insert-cores` pass.
-- For the relaying aspect, this commit also adds an optional attribute to amdaie.core op - `link_with`.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCores.cpp
```diff
@@ -122,6 +122,27 @@ LogicalResult insertCoreOps(mlir::ModuleOp moduleOp) {
...
+        // Symbol table.
+        StringRef fnName = callOp.getCallee();
+        auto fnDecl = dyn_cast_or_null<func::FuncOp>(
+            SymbolTable::lookupSymbolIn(moduleOp, fnName));
+        assert(fnDecl && "expected function declaration");
+        assert(fnDecl->hasAttr("link_with") &&
+               "expected 'link_with' construct for the function declaration");
+        // From the declaration of the function, we extract the value of
+        // attribute "link_with" and attach it to amdaie.core op.
+        // TODO(avarma): What to do when more than one func.call has different
```
jtuyls, line 136: This seems like like a pretty important feature we might need in the future and we could support it with a `Variadic`? But up to you on whether you want to implement it now.

jtuyls (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCores.cpp
Abhishek-Varma: Absolutely! In fact, this would need changes to be thought of in the `lower-to-ukernel` + the actual ukernel `.o` formation.

If we club all possible ukernels within a single `.o` then the above issue should never be hit.
Also for single dispatches too this doesn't become an issue.

So, will table this for later when we gradually mature this and as and when the need arises.

PR #543, opened 2024-07-12
Fix passes to support conv_2d_nhwc_hwcf_q ops

yzhang93:
This PR fix passes and add tests for `linalg.conv_2d_nhwc_hwcf_q` ops. 

- In particular, the `AMDAIEPadPass` is refactored to work with `linalg.conv_2d_nhwc_hwcf_q` which has 4 input operands and 1 output operand.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
```diff
@@ -45,31 +45,29 @@ static LogicalResult applyBufferizeToAllocation(RewriterBase &rewriter,
...
-    SmallVector<Value> operands;
-    for (Value operand : linalgOp->getOperands()) {
-      if (isMatmulInDefChain(operand)) continue;
-      operands.push_back(operand);
-    }
-    return operands;
+  SmallVector<Value> operands;
+  for (Value operand : linalgOp->getOperands()) {
+    if (!isa<RankedTensorType>(operand.getType())) continue;
+    if (isMatmulProducerOfElementwise(linalgOp) && isMatmulInDefChain(operand))
```
Abhishek-Varma, line 51: Would it suffice if we just use `isMatmulInDefChain` instead ?

Trying to understand case which would align with the following but we do not want to `continue` :-
> since one of the elementwise op's input is the output of the matmul op and has already been promoted, there is no need to promote such operand again.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPad.cpp
```diff
@@ -19,108 +19,106 @@ namespace mlir::iree_compiler::AMDAIE {
...
-  options.paddingValues = paddingValues;
-  SmallVector<bool> packPaddings(linalgOp->getNumOperands(), true);
-  options.packPaddings = packPaddings;
+  return paddingValues;
+}
 
+static SmallVector<int64_t> getPaddingDimensions(linalg::LinalgOp &linalgOp) {
   SmallVector<int64_t> paddingDimensions;
-  for (unsigned i = 0, n = linalgOp.getNumLoops(); i < n; i++)
+  for (unsigned i = 0; i < linalgOp.getNumLoops(); i++) {
```
Abhishek-Varma, line 44: I'd prefer the first syntax.

Any particular reason for the new change being suggested?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
```diff
@@ -165,8 +165,10 @@ void AMDAIETileAndFusePass::runOnOperation() {
     // loops, and the first level of tiling is always using scf.forall and
     // mapped to blocks. Currently we are not using mapping attributes for
     // Conv2d ops, because there could be four parallel tiling dimensions.
+    // Somehow `linalg::isaConvolutionOpInterface()` doesn't work properly.
```
Abhishek-Varma, line 168: Ouch.

tests/samples/conv_pipeline_e2e.mlir
```diff
@@ -18,3 +18,24 @@ func.func @conv_2d_nhwc_hwcf(%arg0: tensor<2x14x14x32xi32>, %arg1: tensor<3x3x32
...
+}
+
+// CHECK-LABEL: hal.executable.export public @conv_2d_nhwc_hwcf_q_dispatch_0_conv_2d_nhwc_hwcf_q_2x12x12x64x3x3x32_i8xi8xi32xi32xi32
+//       CHECK:    aie.device(npu1_4col)
+//       CHECK:    aie.shim_dma_allocation
+//       CHECK:    aie.shim_dma_allocation
+//       CHECK:    aie.shim_dma_allocation
+//       CHECK:    func.func @conv_2d_nhwc_hwcf_q_dispatch_0_conv_2d_nhwc_hwcf_q_2x12x12x64x3x3x32_i8xi8xi32xi32xi32(%arg0: memref<3136xi32>, %arg1: memref<4608xi32>, %arg2: memref<2x12x12x64xi32>)
+//       CHECK:      aiex.npu.dma_memcpy_nd
+//       CHECK:      aiex.npu.dma_memcpy_nd
```
Abhishek-Varma, line 39: Nice!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -639,7 +640,8 @@ static LogicalResult setRootConfigImpl(mlir::FunctionOpInterface entryPointFn,
         // let it first crash for all the other ops and then consiously
         // add support for them, this way we can verify our work.
         // TODO (vivian): add support for other conv interface ops
-        .Case<linalg::Conv2DNhwcHwcfOp, linalg::Conv2DNchwFchwOp>([&](auto op) {
+        .Case<linalg::Conv2DNhwcHwcfOp, linalg::Conv2DNchwFchwOp,
+              linalg::Conv2DNhwcHwcfQOp>([&](auto op) {
```
Abhishek-Varma, line 644: I don't think we have lit test for Convs in lowering_strategy.mlir - it'd be nice to add at least one ConvOp there (if not present).

Abhishek-Varma (approved):
Nice @yzhang93 !

I've few change requests - although I'm hitting "approve" it'd be nice to address those (wherever applicable).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: Actually I added a test for conv2d lowering strategy in my previous PR. https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lowering_strategy_conv.mlir

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPad.cpp
yzhang93: It looks cleaner to me.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
yzhang93: Good point. It doesn't really need `isMatmulProducerOfElementwise(linalgOp)` check, but I'd add a check to make sure the linalgOp is an elementwise op and then the matmul op is in its def chain. So I change it to 

`if (isElementwise(linalgOp) && isMatmulInDefChain(operand))`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
newling: Can you elaborate, have you tried 

```
if (!isa<linalg::ConvolutionOpInterface>(consumerOp.getOperation()))
```

?

yzhang93: Yes, I tried it, and somehow `isa<linalg::ConvolutionOpInterface>` returns true for matmul ops. I think something is broken in upstream, but haven't got time to debug.

PR #544, opened 2024-07-12
Support chess-llvm-ir

makslevental:
This PR adds support for chess as an llvmir backend. Support isn't complete yet - bfloat16 intrinsics need to be mapped through AIEVec (our AIEVec) - but e2e examples on i32/f32 work. Note, chess is very slow so we should not add many such tests (this is mostly to test XCLBinGen rather than the codegen itself).

I have also upgraded CI aietools/chess to 2024.1 using a lightweight distro of `aietools` that I produced using `auditwheel`.


makslevental:
ping @newling @nirvedhmeshram

makslevental:
I went ahead and removed all the deps on mlir-aie wheel here.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -63,49 +66,121 @@ void applyConfigToPassManager(PassManager &pm, bool printIRBeforeAll,
...
-    SmallString<64> vitis_path(env_vitis);
-    SmallString<64> vitis_bin_path(vitis_path);
-    sys::path::append(vitis_bin_path, "bin");
-
-    SmallString<64> aietools_path(vitis_path);
-    sys::path::append(aietools_path, "aietools");
-    if (!sys::fs::exists(aietools_path)) {
-      aietools_path = vitis_path;
-      sys::path::append(aietools_path, "cardano");
+  if (vitisDir) {
```
jtuyls, line 85: Replacing with `if (!vitisDir) return std::nullopt;` and then unindenting the block below would be more clean imo.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -63,49 +66,121 @@ void applyConfigToPassManager(PassManager &pm, bool printIRBeforeAll,
...
+    return aieToolsPath.str().str();
   } else {
-    errs() << "VITIS not found ...\n";
+    return std::nullopt;
   }
 }
 
+static const std::string _CHESS_INTRINSIC_WRAPPER_LL{
+#include "chess_instrinsics_wrapper.ll"
+};
```
jtuyls, line 140: Put at the top of the file + add a comment to explain what this is?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/chess_instrinsics_wrapper.ll
```diff
@@ -0,0 +1,53 @@
+// This is actually a cpp file that contains a raw string that contains ll
```
jtuyls, line 1: Could you elaborate in the comment on what kind of stuff this file contains and why it's needed?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -364,21 +461,21 @@ static LogicalResult generateCDO(MLIRContext *context, ModuleOp moduleOp,
...
   applyConfigToPassManager(passManager, printIRBeforeAll, printIRAfterAll,
                            printIRModuleScope, timing);
-
   passManager.addNestedPass<AIE::DeviceOp>(
       mlir::iree_compiler::AMDAIE::createAMDAIEPathfinderPass());
-  if (failed(passManager.run(copy)))
-    return moduleOp.emitOpError(
-        "failed to run passes to prepare of XCLBin generation");
+  if (failed(passManager.run(copy))) {
+    llvm::errs() << "failed to run passes to prepare of XCLBin generation";
```
jtuyls, line 470:
```suggestion
    llvm::errs() << "failed to run passes to prepare for XCLBin generation";
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -758,67 +881,101 @@ static LogicalResult generateUnifiedObject(
...
-      llvmModule->print(llvmirStream, nullptr);
+      std::string inputLLStr;
+      auto inputLL = openInputFile(LLVMIRFile, &errorMessage);
+      if (!inputLL) {
+        llvm::errs() << "Failed to open input.ll file because: "
+                     << errorMessage;
+        return failure();
+      }
+      inputLLStr = std::string(inputLL->getBuffer());
+      std::string inputLLChessHackedStr = chesshack(inputLLStr);
```
jtuyls, line 911: This still uses the existing `chesshack` approach, right? Not the LLVM 15/17? dialect approach?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
makslevental: LOL

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
makslevental: Correct - I didn't do that here (and haven't done it) because I haven't quite decided where those libs should come from (LLVM submodule checked out to the right commit? `ExternalProject_Add` with a zip? Pre-built and kept somewhere?).

newling:
Thank you!

> I went ahead and removed all the deps on mlir-aie wheel here (mm.o kernel is now stored [here](https://github.com/nod-ai/iree-amd-aie/releases/tag/ukernels)).

Can we move mm.cc into iree-amd-aie, and create mm.o with test at compile time or in a test? Is that what mlir-aie does?

makslevental:
> Can we move mm.cc into iree-amd-aie, and create mm.o with test at compile time or in a test? 

I thought about doing that but then you'd have to selectively compile depending on whether someone passes ukernel somewhere? And then I saw this https://github.com/nod-ai/iree-amd-aie/blob/9191fa2f28708ad94c2acaebc1f020338b6102e5/build_tools/ci/run_matmul_test.sh#L405-L410 

Plus, then, we have to decide to we always compile with chess for peak perf? or do we compile with chess/peano alternatively? So then I just decided "maybe later".

> Is that what mlir-aie does?

We were getting it from their wheel and I don't remember what they do (I guess compile at build time of that wheel).

build_tools/ci/run_matmul_test.sh
```diff
@@ -411,8 +395,7 @@ function run_matmul_test() {
     if [ -f "${OUTPUT_DIR}/mm.o" ]; then
       echo "File 'mm.o' already exists in ${OUTPUT_DIR}."
     else
-      SRC_DIR="${mlir_aie_install_path}/aie_kernels/mm.o"
-      ln -s ${SRC_DIR} ${OUTPUT_DIR}/mm.o
+      wget $MM_KERNEL_URL -O  ${OUTPUT_DIR}/mm.o
```
nirvedhmeshram, line 398: do we need this? looks like we will definitely have the object where we need it so we can remove any reference to it in the CI script?

build_tools/ci/run_matmul_test.sh
```diff
@@ -768,3 +749,48 @@ run_matmul_test \
...
+###################################################################
+
+run_matmul_test \
+    --name_prefix "chess_i32_matmul" \
+    --lhs_rhs_type "i32" \
+    --acc_type "i32" \
+    --m "64" \
+    --n "64" \
+    --k "64" \
+    --num_repeat_runs "0" \
```
nirvedhmeshram, line 764: not sure if this is intentional since in the PR description its mentioned e2e works. with ` --num_repeat_runs "0"` we are not actually running only compiling?

makslevental: :facepalm: i thought it meant no repeats but just the first run. ok let me remove (and :crossed_fingers: everything still works).

makslevental: all good

build_tools/ci/run_matmul_test.sh
makslevental: are we? I don't see anywhere in aie2xclbin where we handle ukernels?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/chess_instrinsics_wrapper.ll
makslevental: replace with cpp file stolen from mlir-aie which has an adequate description

jtuyls (approved):
LGTM

makslevental:
> Can we move mm.cc into iree-amd-aie, and create mm.o with test at compile time or in a test?

I have a plan for this in the follow-up.

build_tools/ci/run_matmul_test.sh
nirvedhmeshram: my bad, I thought this was m[e_basic].o stuff, I get it now, thanks.

PR #545, opened 2024-07-12
Make aie2xclbin stop when any intermediate step fails 

newling:
Fix for bug introduced in https://github.com/nod-ai/iree-amd-aie/commit/7d530aff2146bc5d73e6b051d65f45757aa6c3b2 where aie2xclbin keeps running even after a tool fails. Tested locally.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
@@ -206,8 +206,8 @@ std::optional<std::string> runTool(
   }
 
   if (result != 0) {
-    llvm::errs() << "Failed to run tool: " << program << " "
-                 << llvm::join(args, " ") << " with error: " << errMsg << "\n";
+    llvm::errs() << "Failed to run tool: " << program << ". Error: '" << errMsg
+                 << "'\n";
     return {};
   }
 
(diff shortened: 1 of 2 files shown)


makslevental:
> The weird and worrying thing is that the e2e numerical tests passed for me locally when I gave an invalid path to peano -- nothing was being compiled yet the numerical test passed

Were you manually setting tempdir? If so, it's because the workdir isn't cleared and somewhere (haven't found it yet) there's some optimization that reuses existing (ie stale from a previous successful run, in your case) object files.

newling:
> > The weird and worrying thing is that the e2e numerical tests passed for me locally when I gave an invalid path to peano -- nothing was being compiled yet the numerical test passed
> 
> Were you manually setting tempdir? If so, it's because the workdir isn't cleared and somewhere (haven't found it yet) there's some optimization that reuses existing (ie stale from a previous successful run, in your case) object files.

Yes, that must be it 🧠

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -187,10 +187,20 @@ Expected<std::string> runTool(
     llvm::outs() << buffer.str();
   }
 
+  if (result != 0) {
+    return make_error<StringError>(
```
makslevental, line 191: might as well move this to right after the `sys::ExecuteAndWait`? doesn't really matter

makslevental (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
newling: Yeah maybe printing whatever stdouts captured first will be useful though

PR #546, opened 2024-07-13
Vendor `aievec` and `xllvm`

makslevental:
This PR vendors AIEVec and XLLVM (including ops/types/attrs etc.). Importantly, we only vendor the components related to translation to llvm ir (no codegen/optimization/etc). The reason for this is the goal of this sprint is to support chess via the llvm ir path i.e., we will add support for chess aie2 intrinsics here instead of mlir-aie.

Note we greatly reduce the number of supported aievec ops down to `matmul`, `shuffle`, `srs`, `ups`, `cast`.


jtuyls (approved):
LGTM. As this is a vendor PR, I think we don't need to review this in detail and can do that if/when refactoring in the future. (Like we have done for other vendor PRs)

PR #548, opened 2024-07-15
[AMD-AIE][Objectfifo] Add Ukernel support in distribute-cores-and-objectfifos

Abhishek-Varma:
-- This commit adds ukernel support in `--iree-amdaie-distribute-cores-and-objectfifos` pass.
-- Ukernel lowering involves creation of memref.extract_strided_data ops whose outputs (base/offset) are then fed into the ukernel function invocation.
-- This commit therefore adds support for the same.


jtuyls (approved):
LGTM

PR #549, opened 2024-07-15
[AMD-AIE][Objectfifo] Add Ukernel support in create-aie-workgroup

Abhishek-Varma:
-- This commit adds Ukernel support in `--iree-amdaie-create-aie-workgroup` pass.
-- The pass is used to create amdaie.workgroup and we can simply skip this processing for a private function declaration.


jtuyls (approved):
LGTM

PR #551, opened 2024-07-15
Add pass to assign BD IDs to NPU DMA ops and refactor `AMDAIEAssignBufferDescriptorIDsPass`

jtuyls:
Adds a pass to generate BD IDs for NPU DMA operations that avoids race conditions with BDs being reconfigured during execution by some DMA.

This only assigns BD IDs on the L3 side of NPU DMA operations as the backend can currently only handle Shim DMA instructions and MemTile/core DMA configurations are assumed to be part of the xclbin. Therefore, you will see the checks like `npuDmaOp.getSourceMemorySpaceAsUInt() == 0` and `npuDmaOp.getTargetMemorySpaceAsUInt() == 0` being used (memory space `0` is L3).

BD IDs are represented with an `amdaie.bd_id` operation, which takes a tile and an integer value as operands as every tile has its own set of BDs. This helps with guaranteeing/verifying correct reuse of the same id across DMA operations. This usage was derived from https://github.com/Xilinx/mlir-aie/discussions/1610


runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -294,6 +294,11 @@ struct AMDAIEDeviceModel {
 
   uint32_t getColumnShift() const;
   uint32_t getRowShift() const;
+
+  /// Return a map from channels to valid BD ids for the requested tile type.
+  /// TODO(jornt): find these ranges in the device model.
```
makslevental, line 299: it's here https://github.com/Xilinx/aie-rt/blob/c28d57972957d8032839f4da5d2e3b9dd01805ea/driver/src/global/xaie2ipugbl_reginit.c#L389-L391 but they don't distinguish between even/odd

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
```diff
@@ -302,6 +341,18 @@ def AMDAIE_NpuDmaCpyNdOp: AMDAIE_Op<"npu.dma_cpy_nd",
       return memSpace ? cast<IntegerAttr>(memSpace).getInt() : 0;
     }
 
+    BdIdOp getSourceBdIdOp() {
+      Value bdIdValue = getSourceBdId();
+      if (!bdIdValue || !bdIdValue.getDefiningOp()) return nullptr;
+      return dyn_cast<BdIdOp>(bdIdValue.getDefiningOp());
```
makslevental, line 347: is the reason to not make this a `cast` (and instead catch `null` outside) in order to emit a legible error message?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Utils/ChannelBdIdGenerator.cpp
```diff
@@ -0,0 +1,27 @@
...
+  }
+  uint32_t bdId = channelToValidBdIds[channel][0];
+  size_t index{1};
+  while (isBdIdAssigned(bdId) && index < channelToValidBdIds[channel].size()) {
+    bdId = channelToValidBdIds[channel][index++];
+  }
+  if (isBdIdAssigned(bdId)) return std::nullopt;
+  assignBdId(bdId);
+  return bdId;
+}
```
makslevental, line 25: Why not just put this into iree_aie_runtime instead of creating a new lib? I realize it's just basically a thin wrapper around two maps and possibly (superficially) not intimately related to the stuff in iree_aie_runtime but i think it makes sense to keep close to there since it fiddles with bds.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
```diff
@@ -420,7 +460,7 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
       // expected-error @+1 {{could not convert to AIEDialect ops}}
       amdaie.controlcode {
         // expected-error @+1 {{implicit source/target L3 memref has rank greater than the expected static offsets/sizes/strides rank (4)}}
-        %npu_dma_1 = amdaie.npu.dma_cpy_nd %dma_target_l3([] [] [], [] [] [])
+        %npu_dma_1 = amdaie.npu.dma_cpy_nd %dma_target_l3([] [] [] %bd_id_0, [] [] [])
```
makslevental, line 463: i don't think it's a good idea that these aren't delimited in some special way _and_ optional - you're gonna have parser ambiguities if/when you try to introduce another operand (which already happens for attrs in [mlir-aie](https://github.com/Xilinx/mlir-aie/pull/1137)). i guess you don't want to require them always because you don't always need to run the subsumption pass but then you _also_ don't want to special/weird syntax.

it's too bad you can't stick value into the attr dict.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -669,6 +675,11 @@ LogicalResult workgroupToAIE(IRRewriter &rewriter,
   int dmaId = 0;
   WalkResult res = workgroupOp.walk<WalkOrder::PreOrder>([&](Operation *op) {
     return TypeSwitch<Operation *, WalkResult>(op)
+        .Case<AMDAIE::BdIdOp>([&](auto bdIdOp) {
+          // BD ID ops are purely used for retrieving information in other ops
+          // so don't convert to AIE dialect.
+          return WalkResult::advance();
```
makslevental, line 681: i didn't notice - how do they disappear? dce?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -0,0 +1,179 @@
...
+    std::optional<int64_t> row = getConstantIntValue(tileOp.getRow());
+    if (col && row && deviceModel.isShimNOCTile(col.value(), row.value())) {
+      ChannelBdIdGenerator generator(
+          deviceModel.getChannelToValidBdIds(AMDAIETileType::SHIMNOC));
+      shimTileToGeneratorMap[tileOp.getResult()] = std::move(generator);
+    }
+  });
+
+  // Walk `amdaie.npu_dma_cpy_nd` and  `amdaie.dma_wait` operations and assign
+  // and release BD IDs when encountering the respective using the tile BD ID
```
makslevental, line 43: `the respective...`?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -0,0 +1,179 @@
...
+      AMDAIE::CircularDmaCpyNdOp inputDma = npuDmaOp.getDmaCpyNdOp();
+
+      // Utility to retrieve a TileOp from a vector of tile values, while doing
+      // appropriate verification.
+      auto getGeneratorTileOp = [&](const SmallVector<Value> &tiles,
+                                    AMDAIE::TileOp &tileOp) -> LogicalResult {
+        if (tiles.size() != 1) {
+          return npuDmaOp.emitOpError()
+                 << "operating on multiple tiles is not supported";
+        }
```
makslevental, line 57: is multiple tiles a planned extension or something?

makslevental (changes requested):
mostly fine - a few nittish type things to debate - but can you put a description somewhere (eg this PR itself) for the rules for assigning bdids re `npuDmaOp.getSourceMemorySpaceAsUInt()` and `npuDmaOp.getTargetMemorySpaceAsUInt()`.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
jtuyls: Yeah, they have the real numbers harcoded inside a verifier here: https://github.com/Xilinx/aie-rt/blob/1ad203de0b7f282b1c0659fd2ae3f218652c7274/driver/src/dma/xaie_dma_aieml.c#L1216

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
jtuyls: Yeah

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
jtuyls: In this pass, the whole IR is walked and used to construct the new `aie.device`. The old part of the IR is just deleted as a whole.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -0,0 +1,179 @@
...
+  Operation *parentOp = getOperation();
+
+  WalkResult res = parentOp->walk([&](AMDAIE::WorkgroupOp workgroupOp) {
+    if (failed(assignNpuDmaBdIds(workgroupOp))) {
+      return WalkResult::interrupt();
+    }
+    return WalkResult::advance();
+  });
+  if (res.wasInterrupted()) return signalPassFailure();
+}
```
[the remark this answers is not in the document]
jtuyls: No, this is just copied from other passes. I think basically every pass in `AMDAIE` does it like this.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -0,0 +1,179 @@
...
+        if (failed(getGeneratorTileOp(tiles, tileOp)))
+          return WalkResult::interrupt();
+        ChannelBdIdGenerator &generator =
+            shimTileToGeneratorMap[tileOp.getResult()];
+        // TODO(jornt): Temporarily use channel 0 for all DMAs. This should
+        // return correct results for Shim channels, however, for generality
+        // towards other DMAs and future hardware generations, channel
+        // assignment should happen before BD assignemnt. This requires more
+        // refactoring.
+        std::optional<uint32_t> bdId = generator.getAndAssignBdId(0);
```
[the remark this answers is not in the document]
jtuyls: I think `channel` belongs on the `CircularDmaCpyNd` operations which are referenced by the `NpuDmaCpyNd` ops.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
jtuyls: Multiple tiles currently only happens on the AIE core side to represent broadcasting data to multiple cores. It shouldn't come up here (LR) but that's why the logical objectfifo can have multiple tiles assigned and therefore `SmallVector<Value>`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
jtuyls: Good point. Let me try making them required, for example by having an unassigned value.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Utils/ChannelBdIdGenerator.cpp
jtuyls: Yeah, makes sense. Will move it there.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -0,0 +1,179 @@
...
+  WalkResult res = controlCodeOp->walk([&](Operation *op) {
+    if (auto npuDmaOp = dyn_cast<AMDAIE::NpuDmaCpyNdOp>(op)) {
+      AMDAIE::CircularDmaCpyNdOp inputDma = npuDmaOp.getDmaCpyNdOp();
+
+      // Utility to retrieve a TileOp from a vector of tile values, while doing
+      // appropriate verification.
+      auto getGeneratorTileOp = [&](const SmallVector<Value> &tiles,
+                                    AMDAIE::TileOp &tileOp) -> LogicalResult {
+        if (tiles.size() != 1) {
+          return npuDmaOp.emitOpError()
```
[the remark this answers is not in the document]
jtuyls: > also you're missing an error message here ("more than 1 tile <something>").

The error message is on the next line.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -0,0 +1,179 @@
...
+        rewriter.setInsertionPointAfter(tileOp);
+        auto bdIdOp = rewriter.create<AMDAIE::BdIdOp>(rewriter.getUnknownLoc(),
+                                                      tileOp, bdId.value());
+        rewriter.setInsertionPoint(npuDmaOp);
+        (void)rewriter.replaceOpWithNewOp<AMDAIE::NpuDmaCpyNdOp>(
+            npuDmaOp, npuDmaOp.getDma(), npuDmaOp.getTargetMixedOffsets(),
+            npuDmaOp.getTargetMixedSizes(), npuDmaOp.getTargetMixedStrides(),
+            npuDmaOp.getSourceMixedOffsets(), npuDmaOp.getSourceMixedSizes(),
+            npuDmaOp.getSourceMixedStrides(), nullptr, bdIdOp);
+      } else if (npuDmaOp.getTargetMemorySpaceAsUInt() == 0) {
```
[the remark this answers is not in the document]
jtuyls: > these conditionals aren't mutually exclusive

In principal you could create a NPU DMA operation with both source and target on L3/DDR, but that doesn't make much sense in the context of AIE. However, maybe this is not a good place to bake in any of those 'assumptions' and I can make it two separate `if` statements, but with a single `return WalkResult::advance();` at the end.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Utils/ChannelBdIdGenerator.cpp
jtuyls: Moved it.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
jtuyls: I think I like it more to have them optional so null means unassigned. Instead, I added a `bd_id =` delimiter.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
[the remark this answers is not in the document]
jtuyls: Done

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -0,0 +1,179 @@
...
+      return WalkResult::advance();
+    } else if (auto npuWaitOp = dyn_cast<AMDAIE::NpuDmaWaitOp>(op)) {
+      // Release BD ID used by input DMA op.
+      AMDAIE::NpuDmaCpyNdOp npuDmaOp = npuWaitOp.getDmaOp();
+      AMDAIE::BdIdOp bdIdOp;
+      if (npuDmaOp.getSourceBdId()) {
+        bdIdOp = cast<AMDAIE::BdIdOp>(npuDmaOp.getSourceBdId().getDefiningOp());
+      } else if (npuDmaOp.getTargetBdId()) {
+        bdIdOp = cast<AMDAIE::BdIdOp>(npuDmaOp.getTargetBdId().getDefiningOp());
+      }
```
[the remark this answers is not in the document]
jtuyls: Thanks, done

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
makslevental: Ya named kw is what I thought of too

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
makslevental: We can send them a PR to factor that out (but also that's a good util to use somewhere around here too)

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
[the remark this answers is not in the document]
makslevental: I generally like early returns in ifs because then you don't have to hunt for the exit. And also it prevents multiple ifs firing.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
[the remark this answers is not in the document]
makslevental: Hm really? I guess I haven't been paying attention to these ops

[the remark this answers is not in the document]
jtuyls: The connections (routing, channel/port selection) is set up based on the `CircularDmaCpyNd` ops after lowering to AIE dialect. As it's set up now, the NPU instructions shouldn't change the channels/DMA ports

makslevental (approved):
LGTM

PR #553, opened 2024-07-16
fix `deviceModel.isLegalMemtileConnection`

makslevental:
@erwei-xilinx noticed that `deviceModel.isLegalMemtileConnection` was returning legal for `(DMA, 0) -> (South, 4)` even though port/channel 4 isn't a legal channel for `South`. Closer reading of the [relevant fn](https://github.com/Xilinx/aie-rt/blob/d748ddc696702ac7e7e0d2012c93e251add84691/driver/src/stream_switch/xaie_ss_aieml.c#L102) in `aie-rt` shows

> The method assumes both ports exist, and the relevant tile is an MEM-Tile.

So we add a check to see if the ports/channels exist. This of course contradicts mlir-aie at the current commit but it's actually fixed in tip (maybe?).


jtuyls (approved):

erwei-xilinx:
Really appreciate the fix!

PR #554, opened 2024-07-16
[AMD-AIE][Objectfifo] Add Ukernel support in lower-to-aie pass

Abhishek-Varma:
-- This commit adds ukernel support in `--iree-amdaie-lower-to-aie` pass.
-- Majorly support for memref.extract_strided_metadata as well as func.call was added.
-- For func.call, because we're changing the function type (a consequence of the memref.extract_strided_metadata's type changing) - we are to reflect the same to the ukernel function declaration in the SymbolTable.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -302,6 +368,15 @@ LogicalResult coreToAIE(IRRewriter &rewriter, AMDAIE::CoreOp coreOp,
             .Case<linalg::LinalgOp>([&](auto linalgOp) {
               return coreLinalgOpToAIE(rewriter, linalgOp, mapper, toBeErased);
             })
+            .Case<memref::ExtractStridedMetadataOp>(
```
jtuyls, line 371: Why does the default implementation not work? That one should remap operands?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -222,6 +222,70 @@ LogicalResult coreLinalgOpToAIE(IRRewriter &rewriter, linalg::LinalgOp linalgOp,
...
+  LLVM_DEBUG(llvm::dbgs() << "Convert [func.call / function declaration]\n");
+  // Form new argument(s) and function type for the func.call op.
+  SmallVector<Value> newArgs;
+  SmallVector<Type> newArgTypes;
+  SmallVector<Type> newResultTypes;
+  for (Value operand : oldCallOp.getOperands()) {
+    Value newOperand = mapper.lookupOrDefault(operand);
+    newArgs.push_back(newOperand);
+    newArgTypes.push_back(newOperand.getType());
+  }
```
jtuyls, line 256: Use `remapOperands` utility instead?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -741,6 +816,9 @@ LogicalResult lowerToAIE(ModuleOp moduleOp) {
       static_cast<uint32_t>(device.value()));
 
   auto funcRes = moduleOp.walk([&](func::FuncOp funcOp) {
+    if (funcOp.isPrivate()) {
```
jtuyls, line 819: What do you think about creating the new functions here and then looking them up later?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -905,6 +983,19 @@ void AMDAIELowerToAIEPass::runOnOperation() {
           applyPatternsAndFoldGreedily(getOperation(), std::move(patterns)))) {
     return signalPassFailure();
   }
+
+  // All Ukernel related function declarations will be within aie.device, so
+  // delete the ones outside from the SymbolTable.
+  SmallVector<func::FuncOp> eraseCandidateFuncOps;
+  getOperation()->walk([&](func::FuncOp funcOp) {
+    if (funcOp.isPrivate() && !funcOp->getParentOfType<AIE::DeviceOp>()) {
+      eraseCandidateFuncOps.push_back(funcOp);
```
jtuyls, line 992: Isn't it possible to erase the func here?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
```diff
@@ -211,6 +211,61 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
...
+// CHECK:         %[[REINTERPRET:.+]] = memref.reinterpret_cast %[[ACCESS]]
+// CHECK:         linalg.fill ins(%{{.+}} : i32) outs(%[[REINTERPRET]] : memref<32x32xi32, 1>)
+// CHECK:         %[[BASE_BUFFER:.*]], %{{.+}}, %{{.+}}:2, %{{.+}}:2 = memref.extract_strided_metadata %[[REINTERPRET]] :
+// CHECK:         func.call @ukernel(%[[BASE_BUFFER]], %[[C0]]) : (memref<i32, 1>, index) -> ()
+// CHECK:         aie.end
+// CHECK:       } {link_with = "/path/to/ukernel.o"}
+// CHECK:       func.func @lower_to_aie_ukernel
+#executable_target_amdaie_xclbin_fb = #hal.executable.target<"amd-aie", "amdaie-xclbin-fb", {target_device = "npu1_4col", ukernels = "all"}>
+module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb} {
+  func.func private @ukernel(memref<i32, 2>, index) attributes {link_with = "/path/to/ukernel.o", llvm.bareptr = true}
```
jtuyls, line 231: It might be useful for testing purposes to add multiple uKernels in the test

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
Abhishek-Varma: So, this was based on a few similar tryst where I tried deleting/erasing the op being `walk`ed upon and for some reason it crashed.
I've since then tried capturing the candidates and then erasing them afterwards.

I'd indeed prefer what you're suggesting and will give that a try and report here.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
Abhishek-Varma: For creating a new declaration I'd need to know the updated operand types - which at this point I do not.

But since the operands would ALWAYS be `memref<elemType, 1>, index` - I can create them here by traversing each individual argument and creating `memref<elemType, 1>` for `memref` args, while letting the `index` args just be.
Followed by mapping the old one to the new.

But I didn't go for this because :-
1. The updated operand/arg types should just be inferred by the def-chain being updated.
    Eg: Updated `memref.extract_strided_metadata` return types -> `func.call` operand -> func def arguments.
2. Also, I believe the function declaration should be created within `aie.device` - which I cannot do in this step as for each `FuncOp` (public) we are creating a new `aie.device`.

jtuyls: Ok, thanks, makes sense.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
Abhishek-Varma: Done. It worked. :')

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
Abhishek-Varma: I cannot seem to do that because I need to create a new `FuncOp` instead and attach that to the new symbol (func declaration) from the SymbolTable.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
Abhishek-Varma: Because the `remapOpOperands` doesn't seem to be reflecting the necessary change on the inferred types.

Eg:
```
memref.extract_strided_metadata %1 : memref<32x32xi32, 2> -> memref<i32, 2>, index, index, index, index, index
```
In the above case `remapOpOperands` would change it to `memref<32x32xi32, 1> -> memref<i32, 2>, index, index, index, index, index` (Note the memory space mismatch).

It should indeed be `memref<32x32xi32, 1> -> memref<i32, 1>, index, index, index, index, index` which can only happen if I create a new `memref.extract_strided_metadata` op.

jtuyls: I see, yeah, this might become more clean when we refactor more of the lower-level passes, so that memory spaces don't have to change for example.

jtuyls (approved):
LGTM

PR #552, opened 2024-07-16
Update mlir-aie and mlir-air versions; enable support for conv2d with bigger image sizes

erwei-xilinx:

PR #557, opened 2024-07-16
Rolling back to using the previous mlir-aie wheel

erwei-xilinx:
`me_basic.o` is removed from `aie_runtime_lib` in a recent commit: https://github.com/Xilinx/mlir-aie/commit/1acb6fe98e1613bc5e1b6bb8cf02cbf555d2fd3a

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,7 +111,7 @@ jobs:
         run: |
           python3 -m venv .venv
           source .venv/bin/activate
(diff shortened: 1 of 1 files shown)


yzhang93:
I think I saw ci passed in your previous PR https://github.com/nod-ai/iree-amd-aie/pull/552. And it wouldn't be able to merge if it was not. 

We should look into the failure and fix it. Maybe @makslevental can help.

erwei-xilinx:
> I think I saw ci passed in your previous PR #552. And it wouldn't be able to merge if it was not.
> 
> We should look into the failure and fix it. Maybe @makslevental can help.

Make sense. Maybe the CI passed because it was pulling `me_basic.o` from cache. An issue is created for this issue: https://github.com/nod-ai/iree-amd-aie/issues/558

yzhang93 (approved):

PR #562, opened 2024-07-16
Initial support for depthwise convolution 

newling:
Needs some polish, and I'd like to add an e2e numerical test, so please add high-level comments only for now!

Some notes:
- uses the same pipeline as what we have for conv2d, although as this new pipeline may diverge in the future I've made a new pipeline (but have no strong opinion either way)
- I'd like to rename the "decompose-conv" pipeline to "decompose-conv2d" pipeline (depending on opinions on the point above)
- This is an inefficient way to implement depthwise convolution. But at least it is being vectorized. Note that it does not use mac instructions though.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -423,8 +425,15 @@ void addPadPackBasedPassPipeline(OpPassManager &funcPassManager,
   addAMDAIEBufferizePasses(funcPassManager);
 }
 
+void addDepthwiseConvolutionDecomposePassPipeline(
+    OpPassManager &funcPassManager, TilingConfig &tilingConfig) {
+  // Currently the pipelines are the same, but in the future they might diverge.
+  addConvDecomposePassPipeline(funcPassManager, tilingConfig);
```
yzhang93, line 431: High level question, what possible change we can make to depthwise convolution ops but not to regular conv2d ops?  If only the tiling strategy is changed then we shouldn't create a new pipeline.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -473,6 +474,36 @@ static LogicalResult setRootConfigForConvDecomposePipeline(
...
+    OC = instructionSize[1];
+  }
+  SmallVector<int64_t> tileSizeLevel0;
+  SmallVector<int64_t> tileSizeLevel1;
+  SmallVector<int64_t> tileSizeLevel2;
+  // Note: some of the tiling dimensions are hardcoded for now.
+  if (isa<linalg::DepthwiseConv2DNhwcHwcOp>(linalgOp)) {
+    tileSizeLevel0 = {0, OW, OC, 16, 0, 0};
+    tileSizeLevel1 = {1, OW, OC, 4, 0, 0};
+    tileSizeLevel2 = {0, 1, 0, 0, 0, 0};
```
yzhang93, line 496: These tile sizes look weird to me. It doesn't have any tiling for the last two dimensions? And we usually use for loops on reduction loops. It's better to prototype with the transform dialect first, so we make sure the IR after tiling and bufferization look correct.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
newling: Nothing obvious for now, I'm going to change this to use a single pipeline to be consistent with llvm-cpu, thanks for pointing me to that.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
newling: Yeah, this was tiled with 1 at some point in my experimenting, but I was hitting errors in mlir-air. I agree though -- to vectorize we need to tile to 3x3 kernel dimensions. I've asked Erwei to investigate, will make this either {1,1} or {1,0}.

newling:
Abandoning in favor of PR #565

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: Okay, I'll also take a look at the tiling strategy before @erwei-xilinx getting involved. We are currently working on changing the tiling strategy for Conv2d ops, so could borrow some to depthwise ops when it's in a good shape.

PR #563, opened 2024-07-16
[mlir-aie] remove stray AIE pass file

makslevental:
Just a file I forgot to vendor.


compiler/plugins/target/AMD-AIE/aie/AIENormalizeAddressSpaces.inc
```diff
@@ -0,0 +1,106 @@
+/*===- TableGen'erated file -------------------------------------*- C++ -*-===*\
```
jtuyls, line 1: Why not copy the tablegen file instead?

makslevental: We don't use `def Pat` anywhere right? Seemed weird to introduce it here just a for a thing that adds a default address space. Really I should just rewrite it (I'll do that now).

makslevental:
Closing since other work will subsume this.

Issue #556, opened 2024-07-16
ObjectFifo flow Matmul sweep


jtuyls:
Doing a sweep of matmul shapes [WIP] to check what works and what the remaining issues are:

# I32 matmul

## OK (Compiles and correct results)

|   Shape           | Compilation time (s) |
|  --------          |  ------- |
| 32x32x32        | 28.791    |
| 64x32x64        | 28.381    |
| 64x128x64      | 32.184    |
| 32x256x32      | 33.275    |
| 128x256x128  | 40.502    |
| 128x512x128  |  41.298   |
| 256x512x256  | 38.055    |
| 256x1024x256  | 37.735    |
| 512x512x512  | 38.484    |
| 512x1024x512  | 38.384    |
| 1024x1024x1024  | 42.301    |
| 2048x2048x2048  | 78.143    |
| 2048x4096x2048  | 121.291  |

## FAILS 

|   Shape           | Failure |
|  --------          |  ------- |
| 128x32x128        | Incorrect results |
| 256x32x256        | Incorrect results |
| 32x64x32        | <unknown>:0: error: ObjectFifoLinkOp does not support 'join' and 'distribute' at the same time |
| 64x64x64        | <unknown>:0: error: ObjectFifoLinkOp does not support 'join' and 'distribute' at the same time |
| 3072x4096x3072  | RESOURCE_EXHAUSTED; could not allocate memory for kernel or instr buffer; while invoking native function hal.executable.create; while calling import;  |
| 4096x4096x4096        | RESOURCE_EXHAUSTED; could not allocate memory for kernel or instr buffer; while invoking native function hal.executable.create; while calling import; |
| 8192x9728x2432        |  error on npu.dma_cpy_nd stride/size sizes (need to run again for exact error) |

PR #564, opened 2024-07-17
[CanonicalizeDoublyStridedOp] Fix for interleaved unit and linear dims

jtuyls:
Fixes an issue exposed by a 128x32x64 matmul: https://github.com/nod-ai/iree-amd-aie/issues/556.

In the case of a strided pattern like:

```
offsets: [0, 0, 0, 0] 
sizes: [2, 1, 64, 64]
strides: [4096, 64, 64, 1]
```

the unit dimension (size == 1) in the middle will block the recognition that this is a linear access pattern, resulting in the following canonicalized strided pattern:

```
offsets: [0, 0] 
sizes: [2, 4096]
strides: [4096, 1]
```

If the unit dimension is first removed, the strided pattern can be canonicalized further:

```
offsets: [] 
sizes: []
strides: []
```

meaning a complete linear access.

NOTE: with this fix the above matmul shape is still not functional, but exhibits the same behaviour as `128x32x128` etc

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECanonicalizeDoublyStridedOp.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECanonicalizeDoublyStridedOp.cpp
@@ -121,14 +121,16 @@ void AMDAIECanonicalizeDoublyStridedOpPass::runOnOperation() {
   Operation *parentOp = getOperation();
   IRRewriter rewriter(parentOp->getContext());
 
+  // Fold DMA unit dimensions. Needs to happen before folding linear dimensions
+  // to avoid blocking detection of linear dimension folding opportunities due
+  // to a unit dimension in between.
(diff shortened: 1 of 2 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECanonicalizeDoublyStridedOp.cpp
```diff
@@ -121,14 +121,16 @@ void AMDAIECanonicalizeDoublyStridedOpPass::runOnOperation() {
   Operation *parentOp = getOperation();
   IRRewriter rewriter(parentOp->getContext());
 
-  // Fold linear dimensions within a DMA op.
+  // Fold DMA unit dimensions. Needs to happen before folding linear dimensions
+  // to avoid blocking detection of linear dimension folding opportunities due
+  // to a unit dimension in between.
```
Abhishek-Varma, line 126: Nice!

Abhishek-Varma (approved):
LGTM!

Issue #541, opened 2024-07-12
Ability to run different pipelines on different dispatches (no hand-holding)


newling:
This task is to have an end to end model which is as follows (please excuse my python)
```python
A = np.empty([64,64])
B = np.empty([64,64])
C = np.matmul(A,B)
D = C.reshape([1,8,16,16])
E = np.empty([1,8,3,3])
F = torch.conv2D(D, E)
```

This model will correspond to 2 dispatches in IREE, the first with a matmul the second with a convolution. There will be an expand_shape inbetween them. 

To get this to generate code for AIE, the first dispatch (with the matmul) will need to use the pack-peel pipeline (or the pad-pack). The second dispatch (with the conv2d) will need to use the conv-decompose pipeline. 

Currently the only way we can set the pipeline to run on the untiled function is manually with a flag. The compiler needs to dynamically choose the pipeline to run, based on the operations in the function. 

The nearest thing I can see to an end-to-end test like this is the "three-matmuls" test: see [here](https://github.com/nod-ai/iree-amd-aie/blob/main/build_tools/ci/cpu_comparison/test_files/three_matmuls.mlir) and [here](https://github.com/nod-ai/iree-amd-aie/blob/91cab411445cfcb669a4f88411167cf857b5fd56/build_tools/ci/cpu_comparison/run_test.sh#L405). 

The task is to have an end-to-end numerical test of the above simple model.

PR #566, opened 2024-07-17
Add e2e tests that model switching costs and add func arg lowering for the air pipeline

nirvedhmeshram:
Another change we are making and will have in all iree-compile commands running something beyond a single dispatch is the use of ` --iree-scheduling-optimize-bindings=false` . This is needed because there is currently no way to pass runtime constants to generated kernel. In a typical iree backend they are used to pass two things
1. buffer offsets
2. dynamic shapes

with the above flag we wont generate offseted buffers and we dont support dynamic shapes yet so for now this flag allows us to run models where such offsets would otherwise occur.

It was also discovered that the lowering used by the air pipeline does not respect the binding number for the hal.binding ops so we lower them before the `AIRRtToNpuPass` pass can do it.


newling:
There are 2 tests which repeat the same matmul N times. Can you update the comments in the tests specifying what they're testing, so that it's clear they're both useful? Have you run this locally with success? CI has numerical errors which aren't zeros... is the offset thing not working? 

FYI I created a task which is _slightly_ related (compiler side though, not runtime) https://github.com/nod-ai/iree-amd-aie/issues/541

I think these are useful numerical tests, so happy include them here. But for benchmarking runtime overheads, should we also have some separate code, something along the lines of https://github.com/nod-ai/iree-amd-aie/pull/415 ?

nirvedhmeshram:
> There are 2 tests which repeat the same matmul N times. Can you update the comments in the tests specifying what they're testing, so that it's clear they're both useful? Have you run this locally with success? CI has numerical errors which aren't zeros... is the offset thing not working?

Yes I can update the comments, what they are testing is that what happens if you dont switch and just keep calling the same kernel, so they establish the baseline for the switching case, Ya the CI caught an error that I didnt locally becuase I was using uniform inputs, its a fun one where we are wrongly swapping argument locations on the hal.interface.bindings (kind of stuff that is tickled with these "serious" models :) )

> 
> FYI I created a task which is _slightly_ related (compiler side though, not runtime) #541
> 
> I think these are useful numerical tests, so happy include them here. But for benchmarking runtime overheads, should we also have some separate code, something along the lines of #415 ?

The way I am using these tests are with Tracy profiles which give me ns precision timing of all runtime functions,. Of course #415 is useful for the whole model performance benchmarking though, so we should have that for all these "models" too.

newling:
image
Just an FYI: I'm working on changing the script to allow golden values from other sources (arbitrary python) which might be useful if we don't fully trust the llvm-cpu backend.

nirvedhmeshram:
> Just an FYI: I'm working on changing the script to allow golden values from other sources (arbitrary python) which might be useful if we don't fully trust the llvm-cpu backend.
so far havent seen any issues but thats a good feature!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerFuncArgs.cpp
```diff
@@ -0,0 +1,87 @@
...
+    llvm::sort(subspanOps, [](IREE::HAL::InterfaceBindingSubspanOp a,
+                              IREE::HAL::InterfaceBindingSubspanOp b) {
+      return a.getBinding().getZExtValue() < b.getBinding().getZExtValue();
+    });
+    SmallVector<Type> inputTypes;
+    for (auto op : subspanOps) inputTypes.push_back(op.getType());
+
+    auto functionType = funcOp.getFunctionType();
+    auto newArgTypes = llvm::to_vector<6>(
+        llvm::concat<const Type>(functionType.getInputs(), inputTypes));
```
newling, line 51:
```suggestion
        llvm::concat<Type>(functionType.getInputs(), inputTypes));
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerFuncArgs.cpp
```diff
@@ -0,0 +1,87 @@
...
+namespace mlir::iree_compiler::AMDAIE {
+
+namespace {
+
+//===----------------------------------------------------------------------===//
+// Convert the module operation's contents to the AIE dialect
+//===----------------------------------------------------------------------===//
+
+LogicalResult lowerFuncArgs(ModuleOp moduleOp) {
+  IRRewriter rewriter(moduleOp.getContext());
```
newling, line 33:
```suggestion

```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerFuncArgs.cpp
```diff
@@ -0,0 +1,87 @@
...
+  auto funcRes = moduleOp.walk([&](func::FuncOp funcOp) {
```
newling, line 34:
```suggestion
  auto funcRes = moduleOp.walk([](func::FuncOp funcOp) {
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerFuncArgs.cpp
```diff
@@ -0,0 +1,87 @@
...
+    for (auto op : subspanOps) inputTypes.push_back(op.getType());
+
+    auto functionType = funcOp.getFunctionType();
+    auto newArgTypes = llvm::to_vector<6>(
+        llvm::concat<const Type>(functionType.getInputs(), inputTypes));
+    auto newFunctionType = FunctionType::get(funcOp.getContext(), newArgTypes,
+                                             functionType.getResults());
+    funcOp.setType(newFunctionType);
+    for (int i = 0; i < subspanOps.size(); ++i) {
+      auto newArg = funcOp.front().addArgument(inputTypes[i], funcOp->getLoc());
```
newling, line 56:
```suggestion
      auto newArg = funcOp.front().addArgument(inputTypes[i], subspanOps[i]->getLoc());
```
Super-nit. My thinking is that this location might be more informative in an error/note

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerFuncArgs.cpp
```diff
@@ -0,0 +1,87 @@
...
+  AMDAIELowerFuncArgsPass() = default;
+  AMDAIELowerFuncArgsPass(const AMDAIELowerFuncArgsPass &pass){};
+  void runOnOperation() override;
+};
+
+void AMDAIELowerFuncArgsPass::runOnOperation() {
+  if (failed(lowerFuncArgs(getOperation()))) {
+    return signalPassFailure();
+  }
+  LLVM_DEBUG(llvm::dbgs() << "Module after lowerFuncArgs: " << getOperation());
```
newling, line 78: Just print the new function signatures that change? Whole module might be too much text.

newling:
Thanks, looks good to me. I've left only minor suggestions, so accepting in advance. 

> It was also discovered that the lowering used by the air pipeline does not respect the binding number for the hal.binding ops so we lower them before the AIRRtToNpuPass pass can do it.

Nice fix! I would probably have done this as a separate PR, but they're both pretty small so fine.

newling (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerFuncArgs.cpp
nirvedhmeshram: I think the const is needed, I get
```
 error: cannot initialize return object of type 'mlir::Type *' with an rvalue of type 'const mlir::Type *
 ``
 I believe functionType.getInputs() is const Type

PR #570, opened 2024-07-18
[DmaLoopSubsumption] Fix for strided op in loop without induction var dependency

jtuyls:
Fixes an issue exposed by a 128x32x64 matmul: https://github.com/nod-ai/iree-amd-aie/issues/556.

This issue would come up in partial loop dependencies in case of `scf.forall`:

```
scf.forall (%arg2, %arg3) in (2, 6) {
  %1 = amdaie.npu.dma_cpy_nd %0([0, %arg3] [8, 16] [16, 1], [] [] [])
  amdaie.npu.dma_wait(%2, S2MM)
}
```

In this case, the outer iteration upon which the NPU DMA operation didn't depend ( `%arg2`) would not be considered, resulting in incorrect output IR:

```
%1 = amdaie.npu.dma_cpy_nd %0([0, 0, 0] [6, 8, 16] [1, 16, 1], [] [] [])
amdaie.npu.dma_wait(%2, S2MM)
```

However, the outer iteration should be considered as well, making sure the above DMA operation is executed twice, resulting in the below output IR:

```
%1 = amdaie.npu.dma_cpy_nd %0([0, 0, 0, 0] [2, 6, 8, 16] [0, 1, 16, 1], [] [] [])
amdaie.npu.dma_wait(%2, S2MM)
```

This PR fixes the issue by adding general support for subsuming loop iterations into strided operations without any loop induction variable dependency.

### DIFF
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -735,6 +735,7 @@ run_matmul_test \
     --acc_type "f32" \
     --m "8192" --k "2432" --n "7296"
 
+###################################################################
 # ObjectFifo Matmul tests
 ###################################################################
 
@@ -744,7 +745,7 @@ run_matmul_test \
     --tile_pipeline "pack-peel" \
     --lhs_rhs_type "i32" \
     --acc_type "i32" \
-    --m "128" --k "256" --n "128"
+    --m "32" --k "32" --n "32"
 
 run_matmul_test \
     --name_prefix "small" \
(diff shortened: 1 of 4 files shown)


nirvedhmeshram (approved):
Nice fixes, LGTM!

newling (approved):

Abhishek-Varma (approved):
LGTM!

jtuyls:
I noticed only the outermost dimension can be zero in HW, so I will create a follow-up to not perform the loop subsumption in that case. As that is a specific HW limitation, I think it's not needed to alter the general behaviour of this transformation for that, so will use a flag to enable/disable that limitation.

Issue #539, opened 2024-07-12
[Objectfifo] Bf16 Matmul Vectorization


Abhishek-Varma:
Here is the [e2e IR log](https://gist.github.com/Abhishek-Varma/db34e2e8ccc3fffd5d0ff36133aae4c6).

The issue is while running the following command during XCLBin's generation :-
```
llc input.opt.ll -O2 --march=aie2 --function-sections --filetype=obj -o input.o
```

Error: `LLVM ERROR: unable to legalize instruction: %3605:_(s16) = G_BITCAST %3601:_(<16 x s1>) (in function: core_1_3)`.

To reproduce you may use this [input.opt.ll](https://gist.github.com/Abhishek-Varma/cf89671edd36bb20688e7ff3a7bc9f87).

You may also use ToM branch + [this commit](https://github.com/nod-ai/iree-amd-aie/commit/b35e10d6288c45c0b670a4ebf85689e13760c4e4).

I had a talk with Gaeten Bossu (Peano dev) and the main issue seems to be :-
The mask load, it's turned into many scalar loads, that are unaligned. And that generates lots of instructions to legalize, and some of them aren't supported yet.

Abhishek-Varma:
[PR 572] solves this.

PR #572, opened 2024-07-18
[AMD-AIE][Objectfifo] Fix in_bounds property and indices for vectorized ops

Abhishek-Varma:
-- This commit adds a fix for setting the `in_bounds` to `true` for vector.transfer_read/transfer_write op as part of `--iree-amdaie-distribute-cores-and-objectfifos` pass.
-- This fix is indeed the right thing to do because after the vectorization passes are applied we indeed see those properties being set as `true`. And without this fix the `ConvertVectorToLLVM` pass was creating masked load/store intrinsics.
-- This solves https://github.com/nod-ai/iree-amd-aie/issues/539
-- This commit also adds a fix to the indices being formed for vector.transfer_read/transfer_write op.

Explanation for the fix :

I've updated the comments for the fix and adding here a detailed explanation to keep track of the issue being solved in this PR :-

INPUT:
```
scf.forall (%arg2, %arg3) in (2, 2) {
   ...
   scf.for %arg4 = %c0 to %c4 step %c1 {
        scf.for %arg5 = %c0 to %c4 step %c1 {
              scf.for %arg6 = %c0 to %c4 step %c1 {
                %22 = vector.transfer_read %alloc_1[%c0, %c0, %arg6, %arg4, %c0, %c0]
                %23 = vector.transfer_read %alloc[%c0, %c0, %arg5, %arg6, %c0, %c0]
                %24 = vector.transfer_read %alloc_2[%arg2, %arg3, %arg5, %arg4, %c0, %c0] :  vector<1x1x1x1x4x4xf32>
                ...
                vector.transfer_write %27, %alloc_2[%arg2, %arg3, %arg5, %arg4, %c0, %c0] :  vector<1x1x1x1x4x4xf32>
              }
         }
   }
}
```

`%24` and `%27` are dependent on a "yet-to-be-unrolled" loop.

So, if we simply relay the `getIndices()` of the "old" vector ops as is, then after the loop unrolling takes place in the same pass we will get instances like :-

```
UNROLL 1.
%24 = vector.transfer_read %alloc_2[0, 0, %arg5, %arg4, %c0, %c0]
...
vector.transfer_write %27, %alloc_2[0, 0, %arg5, %arg4, %c0, %c0]
 
UNROLL 2.
%24 = vector.transfer_read %alloc_2[0, 1, %arg5, %arg4, %c0, %c0]
...
vector.transfer_write %27, %alloc_2[0, 1, %arg5, %arg4, %c0, %c0]
 
UNROLL 3.
%24 = vector.transfer_read %alloc_2[1, 0, %arg5, %arg4, %c0, %c0]
...
vector.transfer_write %27, %alloc_2[1, 0, %arg5, %arg4, %c0, %c0]
 
UNROLL 4.
%24 = vector.transfer_read %alloc_2[1, 1, %arg5, %arg4, %c0, %c0]
...
vector.transfer_write %27, %alloc_2[1, 1, %arg5, %arg4, %c0, %c0]
```

So, to avoid those incorrect indices we will be needing the fix from this PR.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -136,24 +136,58 @@ LogicalResult distributeLocalMemory(ModuleOp moduleOp) {
...
+        auto newTransferReadOp = rewriter.create<vector::TransferReadOp>(
             transferReadOp.getLoc(), transferReadOp.getType(), newAlloc,
-            transferReadOp.getIndices());
-        rewriter.replaceAllUsesWith(transferReadOp, newTransferReadOp);
+            newIndices);
+        // The old vector.transfer_read has all `in_bounds` set as `true`. We
+        // therefore need to do the same on this new vector.transfer_read op
+        // else this would yield a masked load later on in the lowering stack.
+        SmallVector<bool, 4> bools(newTransferReadOp.getTransferRank(), true);
+        auto inBoundsAttr = rewriter.getBoolArrayAttr(bools);
```
newling, line 161: Does TransferReadOp have a build method which takes attributes? Something like 

```
  auto newTransferReadOp = rewriter.create<vector::TransferReadOp>(
            transferReadOp.getLoc(), transferReadOp.getType(), newAlloc, newIndices, tranferReadOp.getAttributes());
```

might work

newling (commented):
This function could do with some asserts in my opinion:

- in_bounds are always true
- All memrefs on the old L1 allocs are slices along core dimensions down to singleton dimensions (not introduced in this PR)
- A user is one of the expected types (what happens if its not one of subview, transfer_read, transfer_write, extract_strided_meta_data?) (not introduced in this PR)

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
yzhang93: I agree, it should get the attribute from the old transferReadOp.

Abhishek-Varma:
> This function could do with some asserts in my opinion:
> 
> * in_bounds are always true
> * All memrefs on the old L1 allocs are slices along core dimensions down to singleton dimensions (not introduced in this PR)
> * A user is one of the expected types (what happens if its not one of subview, transfer_read, transfer_write, extract_strided_meta_data?) (not introduced in this PR)

Hi @newling 

Regarding the asserts :-
1. Assert on `in_bounds` to be always true is perhaps best suited to be placed in the vectorization pass itself.
2. I didn't understand this. Could you help explain with an example and I can then see if I can add this as assert accordingly?
3. Assert on expected user will not required because it is already guarded here : https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp#L103-L109

I've addressed the comment pertaining to reusing metadata of the old `vector.transfer_read/transfer_write` op.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -136,24 +136,43 @@ LogicalResult distributeLocalMemory(ModuleOp moduleOp) {
...
+    ArrayRef<int64_t> newAllocShape = newAlloc.getType().getShape();
     for (Operation *userOp : allocOpUsers) {
       if (auto subviewOp = dyn_cast<memref::SubViewOp>(userOp)) {
         rewriter.replaceAllUsesWith(subviewOp, newAlloc);
         toBeErased.push_back(subviewOp);
       } else if (auto transferReadOp =
                      dyn_cast<vector::TransferReadOp>(userOp)) {
         rewriter.setInsertionPoint(transferReadOp);
-        Value newTransferReadOp = rewriter.create<vector::TransferReadOp>(
+        // For dimensions with size as 1 we set the indices as 0.
```
jtuyls, line 147: Why is this needed and what are the initial values in `transferReadOp.getIndices()` which are wrong then?

jtuyls (approved):
LGTM

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
jtuyls: Discussed offline and explanation was added

Abhishek-Varma:
> > This function could do with some asserts in my opinion:
> > 
> > * in_bounds are always true
> > * All memrefs on the old L1 allocs are slices along core dimensions down to singleton dimensions (not introduced in this PR)
> > * A user is one of the expected types (what happens if its not one of subview, transfer_read, transfer_write, extract_strided_meta_data?) (not introduced in this PR)
> 
> Hi @newling
> 
> Regarding the asserts :-
> 
> 1. Assert on `in_bounds` to be always true is perhaps best suited to be placed in the vectorization pass itself.
> 2. I didn't understand this. Could you help explain with an example and I can then see if I can add this as assert accordingly?
> 3. Assert on expected user will not required because it is already guarded here : https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp#L103-L109
> 
> I've addressed the comment pertaining to reusing metadata of the old `vector.transfer_read/transfer_write` op.

Hi @newling - since the assert related comments are orthogonal to this PR, I'm going ahead with merging this to unblock the bf16 e2e and we can tackle those (if any) in a separate PR/issue.

Issue #431, opened 2024-06-17
Matmul + Vectorization + Objectfifo


Abhishek-Varma:
We already have an e2e Matmul working with the in-flight Objectfifo backend.
The current in-flight branch being maintained by @jtuyls is https://github.com/nod-ai/iree-amd-aie/tree/jornt_cpp_pipeline.

Currently we're trying to support the same but with the vectorization switched on.
I'm maintaining that on top of @jtuyls 's branch : https://github.com/nod-ai/iree-amd-aie/tree/avarma_matmul_elem

Issues :-
- `insert-core` pass fixed by : [Fix of InsertCore for vectorization](https://github.com/nod-ai/iree-amd-aie/commit/c5a2aa0b6391e123a011ed76ade692a0ed6c3c5d)
- `IREEComprehensiveBufferize` but fixed by commenting `--canonicalize`
- `distribute-core-and-objectfifo-pass` fixed by : [Fix of DistributeCoreAndObjectFifo for vectorization](https://github.com/nod-ai/iree-amd-aie/commit/a2682cbcc59deff2c7d3c0b119fa36e3503b6854)

The current IR log state : [e2e Matmul + Vectorization + Objectfifo](https://gist.github.com/Abhishek-Varma/d695ee863700327e5d72aecc88911ae9)

yzhang93:
@Abhishek-Varma I've rebased your fix commits on top of Jorn's latest branch which is ahead of your branch by several important commits. I also fixed another problem in `distribute-core-and-objectfifo-pass`, and now the lowered IR looks correct to me after this pass. Also with the change, it no longer fails after `AccessToAcquireRelease`, but fails at a later pass `create-logical-objectfifo-link`.

Please refer to this new branch: https://github.com/nod-ai/iree-amd-aie/tree/objectfifo_vectorization
And the latest dump IR:
[matmul_objectfifo_vectorize.txt](https://github.com/user-attachments/files/15878388/matmul_objectfifo_vectorize.txt)

Abhishek-Varma:
Hi @yzhang93 - I didn't get any failure in `create-logical-objectfifo-link`.

The failure with the https://github.com/nod-ai/iree-amd-aie/tree/objectfifo_vectorization branch occurs at `lower-to-aie` pass instead (the final pass in the pipeline).

And that failure turned out to a red-herring, the main issue is at `distribute-core-and-objectfifo` pass which is caused due to the following structure we get due to `fold-memref-alias` pass :-

[...]

PR #573, opened 2024-07-18
[AMD-AIE][Objectfifo] Add e2e CI test for small shaped bf16 and i8 Matmul + lit test in lower-to-aie pass

Abhishek-Varma:
This PR contains 2 additions :-

-- Since because of [this commit in mlir-aie](https://github.com/erwei-xilinx/mlir-aie/commit/0504f7ad3662ce9a2044ea38f3a7f3df9b6708fc) we can now create `aiex.npu.dma_memcpy_nd` without having to linearize the arguments, I've added a lit test in `lower-to-aie` pass.
-- I've also added an e2e CI test for small shaped bf16 and i8 Matmul on Objectfifo.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -958,6 +958,88 @@ class MoveAllDependenciesIntoOp : public OpRewritePattern<OpTy> {
...
+        const std::vector<int64_t> newStaticOffsets = {0, 0, 0, oneDOffset};
+        SmallVector<int64_t> newStaticSizes;
+        for (int64_t size : dmaUser.getStaticSizes()) {
+          newStaticSizes.push_back(size);
+        }
+        SmallVector<int64_t> newStaticStrides;
+        for (int64_t stride : dmaUser.getStaticStrides()) {
+          newStaticStrides.push_back(stride);
+        }
+        newStaticSizes[newStaticSizes.size() - 1] /= 2;
```
nirvedhmeshram, line 1018: I will do a closer review later in the day but justa quick drive by comment, I think you want all the `2` used in divisions to be the variable `div`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -958,6 +958,88 @@ class MoveAllDependenciesIntoOp : public OpRewritePattern<OpTy> {
...
+          users.push_back(cast_r_user);
+      } else
+        users.push_back(user);
+    }
+    for (Operation *user : users) {
+      if (auto dmaUser = dyn_cast<AIEX::NpuDmaMemcpyNdOp>(user)) {
+        int oneDOffset = *getConstantIntValue(dmaUser.getMixedOffsets().back());
+        for (int j = dmaUser.getMixedOffsets().size() - 2; j > 0; j--)
+          oneDOffset += *getConstantIntValue(dmaUser.getMixedOffsets()[j]) *
+                        *getConstantIntValue(dmaUser.getMixedStrides()[j]);
```
yzhang93, line 1007: I think it should be `oneDoffset = 0` and
`oneDoffset += offsets[j] * strides[j] for j=0 to (dmaUser.getMixedOffsets().size() - 1)` ?

Here the innermost stride is always 1?

yzhang93 (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -958,6 +958,88 @@ class MoveAllDependenciesIntoOp : public OpRewritePattern<OpTy> {
...
+    unsigned int numElements = memrefTy.getNumElements() / div;
+    SmallVector<int64_t> shape{numElements};
+    MemRefType newMemrefTy =
+        MemRefType::get(shape, rewriter.getIntegerType(32));
+    argTypes[i] = newMemrefTy;
+    auto &entry = funcOp.front();
+    entry.insertArgument(i, newMemrefTy, rewriter.getUnknownLoc());
+    rewriter.setInsertionPointToStart(&entry);
+    // With memref shape collapsed to 1d, the multi-dimensional offset also
+    // needs to be collapsed.
```
nirvedhmeshram, line 992: The memref rank changing  should not determine what we do to the offsets (same with sizes/strides)  is independent of the rank. I believe we need to divide any stride 1 offsets by `div` and can leave everything as is?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
```diff
@@ -504,6 +504,76 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
 
 // -----
 
+// CHECK:       aie.device(npu1_4col) {
```
nirvedhmeshram, line 507: This test is for offset all zeros could you makea another  tests that has non zero offsets?

build_tools/ci/run_matmul_test.sh
```diff
@@ -770,6 +770,14 @@ run_matmul_test \
     --acc_type "i32" \
     --m "1536" --k "2048" --n "1536"
 
+run_matmul_test \
+    --name_prefix "small" \
+    --lower_to_aie_pipeline "objectFifo" \
+    --tile_pipeline "pack-peel" \
+    --lhs_rhs_type "bf16" \
+    --acc_type "f32" \
+    --m "64" --k "64" --n "64"
```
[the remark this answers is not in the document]
Abhishek-Varma: Hi @yzhang93,

Yes, you're indeed correct.

I've logged [128x128x256 findings here](https://github.com/nod-ai/iree-amd-aie/issues/431#issuecomment-2197181642).

The intention was then to first add support in order to make basic/small shape bf16 Matmul work on objectfifo.

With @jtuyls' support for varied shapes on i32 and after this PR lands, we can take a look at supporting other shapes for bf16 incrementally.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
Abhishek-Varma: I tried applying that logic but it didn't work.

This is the first row :-
```
64x64xf32=[384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 384 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256 256]
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -1001,6 +1001,82 @@ class MoveAllDependenciesIntoOp : public OpRewritePattern<OpTy> {
...
+  // This is in support of npu.dma_memcpy_nd which only allow 32bit integer
+  // types.
+  mlir::FunctionType funcType = funcOp.getFunctionType();
+  SmallVector<Type> argTypes(funcType.getInputs());
+  for (int i = 0, e = argTypes.size(); i < e; i++) {
+    auto memrefTy = dyn_cast<MemRefType>(argTypes[i]);
+    if (!memrefTy) continue;
+
+    unsigned int bitwidth = memrefTy.getElementTypeBitWidth();
+    if (bitwidth != 16 && bitwidth != 8) continue;
```
jtuyls, line 1024: Why specifically check for 16 and 8? Wouldn't this work for 4 bits as well?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -1001,6 +1001,82 @@ class MoveAllDependenciesIntoOp : public OpRewritePattern<OpTy> {
...
+    unsigned upcastingDivisor = 32 / bitwidth;
+
+    unsigned int div = 32 / bitwidth;
```
jtuyls, line 1027: I wouldn't use `div` as a name as it's a std function.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
```diff
@@ -665,7 +668,7 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
 // CHECK-SAME:          metadata = @[[OBJ2]]
 // CHECK-SAME:          memref<32x32xf32>
 // CHECK:           aiex.npu.dma_memcpy_nd
-// CHECK-SAME:          %[[RHS]][0, 0, 0, 0][1, 2, 32, 8][0, 8, 16, 1]
+// CHECK-SAME:          %[[RHS]][0, 0, 0, 34][1, 2, 32, 8][0, 8, 16, 1]
```
nirvedhmeshram, line 671: can you explain how you are reaching 34?, in the base example, you have 
` [%c0, %c1, %c2] [%c2, %c32, %c16] [%c16, %c32, %c1] `
this to me means linear offset is, 1*32+2*16 = 48, so now since you are stepping 2 elements, you get 24?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
nirvedhmeshram: to me it looks like you are doing nothing to the offsets in the linearization, so I wonder if leaving them unmodified, would work?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -1001,6 +1001,82 @@ class MoveAllDependenciesIntoOp : public OpRewritePattern<OpTy> {
...
+        SmallVector<int64_t, 4> newStaticOffsets = {0, 0, 0, oneDOffset};
+        SmallVector<int64_t, 4> newStaticSizes;
+        for (int64_t size : dmaUser.getStaticSizes()) {
+          newStaticSizes.push_back(size);
+        }
+        SmallVector<int64_t, 4> newStaticStrides;
+        for (int64_t stride : dmaUser.getStaticStrides()) {
+          newStaticStrides.push_back(stride);
+        }
+        newStaticSizes[newStaticSizes.size() - 1] /= upcastingDivisor;
```
nirvedhmeshram, line 1055: I dont follow why we need this, I thought the following for loop will take care of dividing the size of the dim with unit sride?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -1001,6 +1001,82 @@ class MoveAllDependenciesIntoOp : public OpRewritePattern<OpTy> {
...
+    // needs to be collapsed.
+    SmallVector<Operation *> users;
+    for (Operation *user : entry.getArgument(i + 1).getUsers()) {
+      if (auto dmaUser = dyn_cast<AIEX::NpuDmaMemcpyNdOp>(user)) {
+        int oneDOffset = 0;
+        for (int j = 0, n = dmaUser.getMixedOffsets().size(); j < n; j++)
+          oneDOffset += *getConstantIntValue(dmaUser.getMixedOffsets()[j]) *
+                        *getConstantIntValue(dmaUser.getMixedStrides()[j]);
+        rewriter.setInsertionPoint(dmaUser);
+        SmallVector<int64_t, 4> newStaticOffsets = {0, 0, 0, oneDOffset};
```
nirvedhmeshram, line 1046: actually I dont follow why we are restricted to make 4 dims always, we can just keep the same dims as the op already has?

build_tools/ci/run_matmul_test.sh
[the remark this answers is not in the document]
yzhang93: The log is quite old. Does it still have the same error? I'm asking because I'm not sure if the casting you implemented in this PR work in general or just work because the starting offsets are all 0.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
yzhang93: This is also my concern.

build_tools/ci/run_matmul_test.sh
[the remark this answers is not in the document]
Abhishek-Varma: Oh yes yes - it's still the same in `AIEAssignBufferAddresses` pass.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
Abhishek-Varma: Yes, it should.

But since `AIRRtToNpu` kept this guarding for `8` and `16` bitwidth only, I didn't change - but we should be able to modify this gradually.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
Abhishek-Varma: This is majorly because of restrictions in the lower stack. Here is [aiex.npu.dma_memcpy_nd's verify()](https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIEX/IR/AIEXDialect.cpp#L187-L193) - it makes certain assumptions on size/stride and ideally we would want all the metadata to be of same shape.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
Abhishek-Varma: So, I tried the offset suggestion (as per https://github.com/nod-ai/iree-amd-aie/pull/573#discussion_r1684537348) and it worked.

The point you made for the linearized offset to be 17 makes sense to me. Good catch! :)
I've therefore added => `original linearized offset / upcastDivisor` as the linearized offset modification and I've added that as part of the current revision's push. It worked too.

But both the above point "worked" because we are dealing with a zero offset case. We cannot verify the same for a non-zero case just yet because of https://github.com/nod-ai/iree-amd-aie/pull/573#discussion_r1684159453.

I believe we should revisit/modify the logic for metadata later when we're able to successfully compile a non-zero offset case for bf16 as it'd help us in verifying the implementation (CC: @yzhang93 ).

And currently since we have our first bf16 Matmul case working (I again tried locally with the set of changes that went in recently and affected the input IR), it'd be nice to unblock getting this in, in my opinion.

But if we have a metadata algorithm that we know will work irrespective and there's some other way that we can verify  the same - I'd be happy to know and include the same in this PR.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -971,6 +971,81 @@ class MoveAllDependenciesIntoOp : public OpRewritePattern<OpTy> {
   }
 };
 
+/// Utility to convert function arguments with memref element types' bitwidth
+/// either 8 or 16 to i32. This basically involves linearizing an N-d memref to
+/// 1-d i32 memref and adjusting the offset/size/stride metadata.
```
jtuyls, line 976:
```suggestion
/// Utility to convert function arguments with memref element types' bitwidth
/// either 4, 8 or 16 to i32. This basically involves linearizing an N-d memref to
/// 1-d i32 memref and adjusting the offset/size/stride metadata.
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
jtuyls: The check is good for me like this, but maybe it's useful to add tests for 8 and 4 bits as well, so at least we're sure this logic works for different bitwidths? 8 bit will for sure come up soon and 4 bit might come up as well for BW reasons.

yzhang93:
@Abhishek-Varma I just talked to @erwei-xilinx, and he mentioned bitwidth 8 / 16 is currently supported in mlir-aie and thus upcasting is not needed. Can you explore and see if we can directly use bf16 in the control codes?

erwei-xilinx:
> @Abhishek-Varma I just talked to @erwei-xilinx, and he mentioned bitwidth 8 / 16 is currently supported in mlir-aie and thus upcasting is not needed. Can you explore and see if we can directly use bf16 in the control codes?

Yeah here's one example: https://github.com/Xilinx/mlir-aie/blob/eef800b3bb4eb335c17390121ff8bcb3445cfa64/test/npu-xrt/add_21_i8_using_dma_op_with_padding/aie.mlir#L72

yzhang93:
> > @Abhishek-Varma I just talked to @erwei-xilinx, and he mentioned bitwidth 8 / 16 is currently supported in mlir-aie and thus upcasting is not needed. Can you explore and see if we can directly use bf16 in the control codes?
> 
> Yeah here's one example: https://github.com/Xilinx/mlir-aie/blob/eef800b3bb4eb335c17390121ff8bcb3445cfa64/test/npu-xrt/add_21_i8_using_dma_op_with_padding/aie.mlir#L72

Thanks! CC @makslevental as well, in case he has some insights about this.

Abhishek-Varma:
> @Abhishek-Varma I just talked to @erwei-xilinx, and he mentioned bitwidth 8 / 16 is currently supported in mlir-aie and thus upcasting is not needed. Can you explore and see if we can directly use bf16 in the control codes?
> Yeah here's one example: https://github.com/Xilinx/mlir-aie/blob/eef800b3bb4eb335c17390121ff8bcb3445cfa64/test/npu-xrt/add_21_i8_using_dma_op_with_padding/aie.mlir#L72

Hi @yzhang93 @erwei-xilinx  - sure! That's great! Must be a recent support in `mlir-aie` then because earlier it wasn't allowing `aiex.npu.dma_cpy_nd` to be created if bitwidth wasn't `32`. 

I'll take a look at this and see if we can get through objectfifo's lowering. Thanks!

Abhishek-Varma:
Hi @yzhang93 @erwei-xilinx - it worked! No issue in creating `aiex.npu.dma_memcpy_nd` ops anymore in `lower-to-aie`. Thanks!

jtuyls (approved):
LGTM

yzhang93 (approved):
Great, thanks!

PR #574, opened 2024-07-18
[DmaLoopSubsumption] Add option to only support repeat on outer dim

jtuyls:
Fixes an issue exposed by a 64x32x128 matmul: https://github.com/nod-ai/iree-amd-aie/issues/556.

As the hardware (AIE2(+)) doesn't support `stride == 0` with buffer descriptors, only a last repeat iteration can be achieved using the repeat count feature. This PR adjusts the DMA loop subsumption pass to optionally not perform any subsumption if this would be violated. For example, the following NPU DMA instruction is not supported on HW:

```
%1 = amdaie.npu.dma_cpy_nd %0([0, 0, %arg3] [8, 2, 16] [16, 0, 1], [] [] [])
```

due to `stride == 0` on the second (middle) dimension. However, the following is:

```
%1 = amdaie.npu.dma_cpy_nd %0([0, 0, %arg3] [2, 8, 16] [0, 16, 1], [] [] [])
```

but the outer repeat dimension needs to be moved to the first/most outer dimension for the AIE lowering passes. This is achieved with the logic added in `LowerToAIE`.

### DIFF
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -753,6 +753,15 @@ run_matmul_test \
     --tile_pipeline "pack-peel" \
     --lhs_rhs_type "i32" \
     --acc_type "i32" \
+    --m "64" --k "32" --n "128" \
+    --num_repeat_runs "10"
+
+run_matmul_test \
+    --name_prefix "small" \
+    --lower_to_aie_pipeline "objectFifo" \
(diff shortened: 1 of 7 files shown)


Abhishek-Varma (approved):
LGTM!

For my understand - wouldn't we always want to support AIE2(+) ? If yes, then we shouldn't be needing the pass flag option `only-zero-stride-on-outer-dim` ?

Issue #582, opened 2024-07-22
`aie.use_lock`s emitted in non-deterministic order


makslevental:
```diff
$ diff module_matmul_32x32_32xi32__dispatch_0_amdaie_xclbin_fb_backup/input.ll module_matmul_32x32_32xi32__dispatch_0_amdaie_xclbin_fb/input.ll
890d889
<   call void @llvm.aie2.release(i32 50, i32 1)
891a891
>   call void @llvm.aie2.release(i32 50, i32 1)
```

this

```
130:                                              ; preds = %48
  call void @llvm.aie2.release(i32 50, i32 1)
  call void @llvm.aie2.release(i32 48, i32 1)
  call void @llvm.aie2.release(i32 53, i32 1)
  ret void
```
vs
```
130:                                              ; preds = %48
  call void @llvm.aie2.release(i32 48, i32 1)
  call void @llvm.aie2.release(i32 50, i32 1)
  call void @llvm.aie2.release(i32 53, i32 1)
  ret void
```

jtuyls:
Should be fixed with: https://github.com/nod-ai/iree-amd-aie/pull/585

Issue #347, opened 2024-05-14
Pack-peel pipeline bf16 intermittent numerical error


yzhang93:
We've been noticing an intermittent numerical error with pack-peel pipeline and data type bf16. To reproduce the problem, one can run [this test](https://github.com/nod-ai/iree-amd-aie/blob/main/build_tools/ci/run_matmul_test.sh#L626) locally for multiple time and the test results and error are as below:

error: the actual and expected result matrices disagree at row 48, column 16.

actual value: 0
expected value: -54

left-hand side (rows 40..55 out of 0..63, columns 0..15 out of 0..127)
 0    1    2   -2    1    2    1    2    2   -2    0   -1    2   -1    2   -1
-2    1    1   -1    1    1   -1    1   -2   -2   -2    2    0    2   -1    1
 0   -1    1    2    0    2    1    1   -1   -2    1    0    2    2   -1    2
 2   -1    1    2    0    2    0   -2    1    2    2    2    1   -2   -2    2
-1   -1   -1    1    0    2    0    1   -2   -1    2    2   -2    1    1    1
-1    0    1    2   -2    2   -2    2    1   -2   -1    0    1   -2    0    1
 2   -1   -1    1    0    2    0    1    1    0   -2   -1    2    1   -1   -2
 1    2    0   -1   -1   -2    0    2    0    2    2    1   -2    0   -1    2
-2    2   -2    0    0    1    1   -2    1    0   -2   -1    1   -1   -1    2
 2   -2   -1    2    1   -2    1   -2   -1    1    1    2   -2    1    2   -1
 2    0   -2    0    1    2   -1    0    1   -1    1    1   -2    1    0   -1
-1    0    1   -2   -2   -2    0   -2    0   -1    1    2   -1   -2   -1    2
 0   -2   -1   -2    2    2   -2   -1   -2    2    1   -1   -2    0    1    2
 2    2    2   -1    1    2    1   -2   -2   -1   -1   -1    1   -2    0    0
-2    2    0   -2    2   -1   -1   -1    2    2    2   -1    0   -2    1    1
 1   -2   -2    1    0   -1   -1    0    0    2    0   -2   -2    2    2    0

right-hand side (rows 0..15 out of 0..127, columns 8..23 out of 0..63)
-1    2   -2    1    0    2   -2   -1    0    2    1    0    0   -2   -1   -1
-1    0    2    0    1    2    2    1   -2   -2   -2    2   -1   -2    1    0
[...]

PR #583, opened 2024-07-23
Increase `run_matmul_tests` repeats

makslevental:
To sniff out non-deterministic NA bugs.

### DIFF
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -444,13 +444,10 @@ function run_matmul_test() {
 
   # Renable exit on failure:
   set -e
-
-
   echo "**** Generating calls .vmfb file for ${name} ****"
   ${IREE_COMPILE_EXE} "${calls_ir}" \
       --iree-hal-target-backends=${target_backend} \
       -o "${calls_vmfb}"
-
 
   compiled_time=$(date +%s%3N)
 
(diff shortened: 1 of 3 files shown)


yzhang93:
What are non-deterministic NA bugs?

makslevental:
> What are non-deterministic NA bugs? 

NA=numerical accuracy - basically whenever you see the goofy 🦄 and 🐞 in the `run_matmul_test`.

> What caused failure happen?

I wish I knew :shrug: - today I thought it was this https://github.com/nod-ai/iree-amd-aie/issues/582 but @jtuyls thinks it's not. Anyway this is a step towards figuring it out.

yzhang93:
Okay, I think we've seen such problem before for some test cases (https://github.com/nod-ai/iree-amd-aie/issues/347) but definitely not all of the tests. If it happens to all tests now, we should probably look into the cause. I wonder if it is because we are switching to a new version of peano?

makslevental:
> Okay, I think we've seen such problem before for some test cases (#347) but definitely not all of the tests. If it happens to all tests now, we should probably look into the cause. 

I'm getting it sporadically in various tests but not all

> I wonder if it is because we are switching to a new version of peano?

We haven't switched yet (that's my TODO this week...).

jtuyls (approved):
LGTM. It doesn't seem like it makes anything fail consistently, but running a couple of repeats is a good practice imo as it's quite easy to leave AIE in a bad state after a single run or rely on indeterminism like producers/consumers that are not correctly synchronized could still yield correct results by accident.

makslevental:
offline discussion: we probably have a bug somewhere but in order to prevent blocking absolutely all forward progress (and/or driving people insane) this won't get merged as is. @jtuyls proposes we only increase all objectfifo tests to 10. He is very confident those will pass. I propose we raise them to 100 actually. Let's see :thinking:

build_tools/ci/run_matmul_test.sh
```diff
@@ -460,19 +459,20 @@ function run_matmul_test() {
...
       --max_elements_to_check=${max_elements_to_check}"
 
+  set +e
 
-
-  echo "**** Running '${name}' matmul test ${num_repeat_runs} times ****"
+  echo "**** Running '${name}' matmul test ${num_repeat_runs} times (command ${COMMAND}) ****"
   for i in $(seq 1 $num_repeat_runs); do
-    echo "Run number ${i} / ${num_repeat_runs} of command ${COMMAND}"
+    bash $THIS_DIR/reset_npu.sh
```
jtuyls, line 466: @makslevental What's the point of resetting the NPU every run? This would reduce testing coverage it wouldn't catch issues by the AIE/NPU being left in a bad state by some run.

PR #586, opened 2024-07-23
[DistributeCoresAndObjectFifos] Fix for L3 distribution across different tiles for same DMAs

jtuyls:
Fix for `32x64x32` and `64x64x64` matmul shapes: https://github.com/nod-ai/iree-amd-aie/issues/556.

Some legacy code in in the `DistributeCoresAndObjectFifos` pass that is not really needed is causing DMA operations across the peeled matmul that should map to the same physical channels/ports to be assigned to different tiles on either L2 and/or L3 instead. This is causing issues in the lowering stack as multiple DMAs writing to the same L2 buffer slices is currently not supported. Also, on larger examples this could potentially create resource issues.

### DIFF
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -802,6 +802,24 @@ run_matmul_test \
     --tile_pipeline "pack-peel" \
     --lhs_rhs_type "i32" \
     --acc_type "i32" \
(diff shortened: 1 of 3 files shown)


Abhishek-Varma (approved):
LGTM! Thanks @jtuyls !

PR #588, opened 2024-07-23
Make e2e test (cpu comparison) python based

newling:
Python is easier to write and maintain than bash. What finally pushed me into doing this is that I am working on a way to use python (pure numpy for example) to generate golden values, instead of the llvm-cpu backend of IREE. This resulted in a bash script with 4+ calls into python. Rather than do that, I think it's better to start in python, and call bash when necessary.


makslevental:
aww man i was just getting really good at bash :( jk god bless you for doing this :pray: 

but also btw i recently updated run_matmul https://github.com/nod-ai/iree-amd-aie/pull/583 can we not lose that (sorry i didn't look at the PR, maybe you already ported)

.github/workflows/ci.yml
```diff
@@ -115,6 +115,14 @@ jobs:
           source .venv/bin/activate
           pip install -r tests/matmul/requirements.txt
 
+      - name : E2E comparison of AIE to llvm-cpu
```
newling, line 118: Oops I moved this before the matmul tests just to make my CI testing faster, will move back to where it was to reduce noise.

build_tools/ci/cpu_comparison/input_generator.py
```diff
@@ -218,21 +239,5 @@ def generate_inputs(filename, write_dir, seed):
         )
 
     command_flags = "  ".join(input_args)
-    command_arg_filename = os.path.join(write_dir, name + "_input_args.txt")
-    with open(command_arg_filename, "w") as file:
-        file.write(command_flags)
-
+    return command_flags
 
-if __name__ == "__main__":
```
newling, line 226: These "if __name__" bits of code are no longer needed, as the python scripts are no longer called from bash

build_tools/ci/cpu_comparison/run_test.py
```diff
@@ -0,0 +1,471 @@
+# Copyright 2024 The IREE Authors
```
newling, line 1: This script is hopefully 1:1 equivalent with the run_test.sh file which is removed in this PR

yzhang93:
Overall LGTM. A feature request: can you add some print info about time spent for compilation and execution? Similar to this https://github.com/nod-ai/iree-amd-aie/blob/main/build_tools/ci/run_matmul_test.sh#L472

newling:
> Overall LGTM. A feature request: can you add some print info about time spent for compilation and execution? Similar to this https://github.com/nod-ai/iree-amd-aie/blob/main/build_tools/ci/run_matmul_test.sh#L472

Yes, good idea

yzhang93 (approved):

PR #590, opened 2024-07-24
[CI] Only reset NPU when running as root

jtuyls:
- Only try resetting NPU when root to facilitate easier local testing when not root
- Add missing copyright header

### DIFF
--- a/build_tools/ci/reset_npu.sh
+++ b/build_tools/ci/reset_npu.sh
@@ -1,4 +1,10 @@
-#!/usr/bin/env bash
+#!/bin/bash
+#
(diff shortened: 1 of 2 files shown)


Abhishek-Varma (approved):

makslevental:
This will then not work on CI? Just do a `|| true` instead.

jtuyls:
> This will then not work on CI because `github` runs tests. Just do a `|| true` instead.

As discussed offline, checking for the `GITHUB_ACTIONS` env variable now.

makslevental (approved):

PR #593, opened 2024-07-25
Port some aie-rt tests for educational/canary purposes

makslevental:
This PR adds a few tests ported from aie-rt's test suite for educational/canary purposes. Educational meaning they can serve as a brief/runnable intro (e.g., for on-boarding) to what's really going on under the hood and canary meaning they will fail in intelligible ways if aie-rt's APIs change in subtle ways.

1. `test_0335_aie_dma_tile_dma_packet_switch_mode` demos/illustrates how to use (configure) packet switching mode;
2. `test_1114_aie_stream_switch_packet_switch_control_packets` demos how to use control packets (in circuit or packet switched mode);
3. `test_transaction` demos how to use transaction recording/serialization (this isn't actually one of their tests).

The PR also adds `interpreter_op_impl.h` for purposes of printing but also, **more importantly**, as a mock/simulation of how the actual/real transaction interpreter operates in the FW. I plan to discuss this file (the real file) next week at on-site so @MaheshRavishankar consider this some reading material :)


nirvedhmeshram (approved):
Thanks! LGTM.

PR #595, opened 2024-07-26
Fix peano

makslevental:
I don't understand it but using their pre-built `crt` breaks us _but building the `crt` I literally [copy-pasted from them](https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/crt.c)_ works. And it is indeed the `crt` part of the patch not the `--target=aie2-none-unknown-elf` part.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,6 +108,9 @@ jobs:
         run: |
           mkdir iree-install
           tar -xf iree-dist.tar -C iree-install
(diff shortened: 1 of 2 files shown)


makslevental:
I'm guessing our ld script is wrong https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIETargetLdScript.cpp#L101-L113 (see https://github.com/Xilinx/llvm-aie/blob/aie-public/libc/startup/baremetal/aie2/ldfile#L37-L59). I'm going to investigate but we should land this as is in order to unblock ... everyone.

jtuyls (approved):
LGTM

PR #596, opened 2024-07-26
Ability to use numerics baseline other than llvm-aie

newling:
There are some pros and cons of using the llvm-cpu backend as a source of truth for testing the numerical correctness of the code generated for AIE backend. 

Pros: 
- easy (we don't need to write additional code to do convs / matmuls using numpy / torch)
- sort of tests 2 backends at once. If there is a discrepancy between llvm-cpu and aie backends, we might end up fixing a bug in llvm-cpu (less likely than fixing a bug in the aie backend)

Cons:
- additional compile time, and runtime: cannot benefit by sampling results, like IREE does (via run_matmul_test.sh)
- maybe we don't want to rely on llvm-cpu, would prefer to get the source of truth from torch directly, or somewhere else (onnx runtime?) 

This PR basically shows how we can switch the llvm-cpu backend for another source of truth. In this PR I use numpy to do a matmul to get the source the truth. 

Future things I'd like to do
1) ability to sample results. AIE in theory can do way more compute than CPU, so if we want to test large matmuls they might be very fast on AIE and slow on CPU. If we can (1) sample the result of the AIE matmul and (2) only compute the outputs at sample indices on CPU, we can have larger matmuls in CI. This is _exactly_ what happens under the hood in run_matmul_test.sh
2) ability to store golden values as serialized numpy files directly (now that we're confident the random number generation is deterministic across platforms, this is feasible)
3) Provide nicer (unicorn-level nice) feedback about where the numerical failure is. 

We might eventually consider deprecating run_matmul_test.sh (and move all tests here), if that reduces complexity in iree-amd-aie

### DIFF
--- a/build_tools/ci/cpu_comparison/run_test.py
+++ b/build_tools/ci/cpu_comparison/run_test.py
@@ -19,7 +19,7 @@ from input_generator import generate_inputs, verify_determinism, load_input
 from output_comparer import compare
 
 
-def matmul_from_input_line(input_args):
+def matmul_from_input_strings(input_args):
     """
     Input 'input_args' should be a list with two strings, of the form
 
@@ -485,7 +485,7 @@ def aie_vs_np_matmul(
 
     name = name_from_mlir_filename(test_file)
     input_args = generate_inputs(test_file, config.output_dir, seed)
-    numpy_output = matmul_from_input_line(input_args)
+    numpy_output = matmul_from_input_strings(input_args)
     aie_vs_baseline(
         config,
         test_file,
--- a/build_tools/ci/cpu_comparison/run_test.py
+++ b/build_tools/ci/cpu_comparison/run_test.py
@@ -15,7 +15,27 @@ import numpy as np
 
 from input_generator import generate_inputs, verify_determinism
 from matmul_template.matmul_generator import generate_matmul_test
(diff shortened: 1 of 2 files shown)


nirvedhmeshram (commented):
IIUC there is no actual aie_vs_golden test case added? do we want to do that here?

newling:
> IIUC there is no actual aie_vs_golden test case added? do we want to do that here?

aie_vs_golden is the base method which aie_vs_llvm_cpu and aie_vs_np_matmul, just for code reuse. Maybe "golden" is confusing in the name of this function, I'll change to aie_vs_provided or aie_vs_baseline

build_tools/ci/cpu_comparison/run_test.py
```diff
@@ -424,14 +469,90 @@ def run_test(
...
+    test_file,
+    use_ukernel=False,
+    pipeline="pad-pack",
+    function_name=None,
+    seed=1,
+    rtol=1e-6,
+    atol=1e-6,
+    n_repeats=1,
+):
+    """ """
```
nirvedhmeshram, line 532: maybe add a description string?

nirvedhmeshram (approved):
LGTM, just a minor doc request.

PR #597, opened 2024-07-26
[AMD-AIE][Objectfifo] Update KernelDispatch for targeting larger bf16 and i8 Matmul on Objectfifo

Abhishek-Varma:
In this PR I've updated the KernelDispatch logic to target `ObjectFifo` backend.

Currently we extract `scalingFactor` based on `LHS` type, but a better way is to use the accumulator's element type to do the same. And even better way is to see how to use element types of all the operands. This has already been acknowledged as [a comment in KernelDispatch.cpp](https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp#L159-L169).

In order to not hamper `AIR` backend, this PR therefore uses accumulator's element type to decide the `scalingFactor` for `objectFifo`.

e2e CI test for larger shaped `bf16` as well as `i8` has been added.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
@@ -169,7 +169,8 @@ FailureOr<ParameterSetting> ParameterSetting::create(linalg::LinalgOp linalgOp,
   // the element types of all tensors which need to be allocated in memory
   // simultaneously.
   FailureOr<unsigned> maybeScaleFactor =
-      getTilingScaleFactor(lhsType.getElementType());
+      isObjectFifo ? getTilingScaleFactor(initType.getElementType())
+                   : getTilingScaleFactor(lhsType.getElementType());
   if (failed(maybeScaleFactor)) {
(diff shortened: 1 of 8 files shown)


jtuyls (approved):
LGTM

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -168,7 +169,8 @@ FailureOr<ParameterSetting> ParameterSetting::create(linalg::LinalgOp linalgOp,
   // the element types of all tensors which need to be allocated in memory
   // simultaneously.
   FailureOr<unsigned> maybeScaleFactor =
-      getTilingScaleFactor(lhsType.getElementType());
+      isObjectFifo ? getTilingScaleFactor(initType.getElementType())
+                   : getTilingScaleFactor(lhsType.getElementType());
```
yzhang93, line 173: I think we had the same question previously that why the lhsType is taken for other pipelines. I think I tried to use the initType before but got failure for ukernel.

yzhang93 (approved):

PR #600, opened 2024-07-26
Depthwise convolution (part i)

newling:
This PR adds e2d support for i32 convolution, and partial support for i8 and bf16 (we cannot do e2e for types which need vectorization at the moment).

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
@@ -166,7 +166,6 @@ void AMDAIETileAndFusePass::runOnOperation() {
     // loops, and the first level of tiling is always using scf.forall and
     // mapped to blocks. Currently we are not using mapping attributes for
     // Conv2d ops, because there could be four parallel tiling dimensions.
-    // Somehow `linalg::isaConvolutionOpInterface()` doesn't work properly.
     // TODO (vivian): create AIE specific mapping attributes.
     if (!isa<linalg::ConvolutionOpInterface>(consumerOp.getOperation())) {
       if (tilingLevel == 0) {
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
@@ -8,6 +8,7 @@
 #include "iree/compiler/Codegen/Utils/Utils.h"
 #include "mlir/Dialect/GPU/IR/GPUDialect.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Linalg/IR/LinalgInterfaces.h"
 #include "mlir/Dialect/Linalg/Transforms/TilingInterfaceImpl.h"
 #include "mlir/Dialect/Linalg/Utils/Utils.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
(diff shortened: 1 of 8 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -465,6 +467,53 @@ static LogicalResult setRootConfigForConvDecomposePipeline(
...
+      return linalgOp.emitError(
+          "has an operand type with fewer than 32-bits, but no mac-width "
+          "could be determined.");
+    }
+
+    const uint16_t OC_1 = OC_0 / 4;
+
+    tileSizeLevel0 = {1, OH_0, OW_0, OC_0, 0, 0};
+    tileSizeLevel1 = {1, OH_1, OW_0, OC_1, 0, 0};
+    tileSizeLevel2 = {1, OH_1, OW_0, OC_1, 1, 1};
```
yzhang93, line 514: I think here we should use {0, 0, 0, 0, 1, 1}. It's good to use 0 if we don't want to tile certain dimensions, even though sometimes they are equivalent when the tile size equals to previous level of size. 

It's better we have the first two level tile sizes as [x x x x 0 0], and last level as [0 0 0 0 x x], because we know the first four dimensions are parallel dimensions (so use scf.forall for tiling) and the last two are reduction dimensions (use scf.for for tiling).

This style is more easy to understand strategy and also the same as the lowering configs for other backends in IREE.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
```diff
@@ -85,6 +85,17 @@ FailureOr<std::array<uint32_t, 3>> getAIEIntegerMatmulInstructionSize(
...
 }
 
+// The number of elements in a vector instruction for a given element type.
+FailureOr<uint32_t> getAIEMacWidth(Type inputType, Type outputType) {
+  // inputType = i8, outputType = i32:
+  if (inputType.isInteger(8) && outputType.isInteger(32)) return 32;
+  // inputType = i16, outputType = i32:
+  if (inputType.isInteger(16) && outputType.isInteger(32)) return 32;
+  // inputType = bf16, outputType = f32:
+  if (inputType.isBF16() && outputType.isF32()) return 16;
```
yzhang93, line 95: Why the first two get value from bitwidth of the output and the last get it from the input?

newling: This was info I got from @jsetoain. Javier can you point us to the source of truth, please? Also can you please check if this API design makes sense longer term.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
```diff
@@ -167,8 +168,7 @@ void AMDAIETileAndFusePass::runOnOperation() {
     // Conv2d ops, because there could be four parallel tiling dimensions.
     // Somehow `linalg::isaConvolutionOpInterface()` doesn't work properly.
     // TODO (vivian): create AIE specific mapping attributes.
-    if (!isa<linalg::Conv2DNhwcHwcfOp, linalg::Conv2DNchwFchwOp,
-             linalg::Conv2DNhwcHwcfQOp>(consumerOp)) {
+    if (!isa<linalg::ConvolutionOpInterface>(consumerOp.getOperation())) {
```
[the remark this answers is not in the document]
newling: Seems to work! Will remove the old comment

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
jsetoain: > Why the first two get value from bitwidth of the output and the last get it from the input?

I believe that function is not returning the bitwidth of the vector element, it's returning the number of elements. For whatever design reason, the output of a mul/mac for `f32` is 512-bits, instead of the full 1024-bit width used by the integer unit. Both end up in the accumulator register bank, but the floating point unit produces half as many results per op as the integer unit[^1].

I suggest a different name might be less confusing. Instead of "MacWidth", maybe "MacNumElements" or "MacNumLanes"?

[^1]: Source of true: [AMD's AI Engine-ML Intrinsics User Guide; Multiply-accumulate of bfloat16 datatypes](https://www.xilinx.com/htmldocs/xilinx2023_2/aiengine_ml_intrinsics/intrinsics/group__intr__gpvectorop__mul__bf16xbf16.html)

yzhang93 (approved):

Issue #581, opened 2024-07-22
Conv2d stride 2 test failed to legalize operation 'vector.extract_strided_slice'


yzhang93:
Conv2d with stride = 2 example:

```
func.func @conv_2d_nhwc_hwcf(%arg0: tensor<1x129x129x16xi8>, %arg1: tensor<3x3x16x32xi8>) -> tensor<1x64x64x32xi32> {
  %cst = arith.constant 0 : i32
  %0 = tensor.empty() : tensor<1x64x64x32xi32>
  %1 = linalg.fill ins(%cst : i32) outs(%0 : tensor<1x64x64x32xi32>) -> tensor<1x64x64x32xi32>
  %2 = linalg.conv_2d_nhwc_hwcf {dilations = dense<1> : vector<2xi64>, strides = dense<2> : vector<2xi64>} ins(%arg0, %arg1 : tensor<1x129x129x16xi8>, tensor<3x3x16x32xi8>) outs(%1 : tensor<1x64x64x32xi32>) -> tensor<1x64x64x32xi32>
  return %2 : tensor<1x64x64x32xi32>
}
```

Error:

```
error: failed to legalize operation 'vector.extract_strided_slice' that was explicitly marked illegal
 %13 = vector.extract_strided_slice %10 {offsets = [0, 0, 0], sizes = [1, 1, 8], strides = [1, 1, 1]} : vector<1x7x8xi8> to vector<1x1x8xi8>
```

Generated IR snippet:

```
%10 = vector.transfer_read %buf0[%c0, %c0, %c0, %c0], %c0_i8 {in_bounds = [true, true, true]} : memref<1x1x7x8xi8>, vector<1x7x8xi8>
%11 = vector.transfer_read %buf1[%c0, %c0, %c0, %c0], %c0_i8 {in_bounds = [true, true, true]} : memref<1x1x8x8xi8>, vector<1x8x8xi8>
%12 = vector.transfer_read %buf4[%c0, %c0, %c0, %c0], %c0_i32 {in_bounds = [true, true, true]} : memref<1x1x4x8xi32>, vector<1x4x8xi32>
%13 = vector.extract_strided_slice %10 {offsets = [0, 0, 0], sizes = [1, 1, 8], strides = [1, 1, 1]} : vector<1x7x8xi8> to vector<1x1x8xi8>
[...]

PR #601, opened 2024-07-26
Small cpu_comparison "pythonic" improvements

makslevental:
I was gonna start working on https://github.com/nod-ai/iree-amd-aie/issues/581, which is downstream of one of the comparisons, and I started using the script and then got distracted.

### DIFF
--- a/build_tools/ci/cpu_comparison/run_test.py
+++ b/build_tools/ci/cpu_comparison/run_test.py
@@ -5,7 +5,6 @@
 import argparse
 import os
 import re
-import shutil
 import subprocess
 import time
 import urllib.request
@@ -192,7 +191,6 @@ class TestConfig:
         peano_dir,
         xrt_dir,
         vitis_dir,
-        n_repeats,
         file_dir,
         iree_compile_exe,
         iree_run_exe,
@@ -205,7 +203,6 @@ class TestConfig:
         self.peano_dir = peano_dir
         self.xrt_dir = xrt_dir
         self.vitis_dir = vitis_dir
-        self.n_repeats = n_repeats
         self.file_dir = file_dir
         self.iree_compile_exe = iree_compile_exe
         self.iree_run_exe = iree_run_exe
@@ -293,7 +290,6 @@ class TestConfig:
         iree_run_exe:         {self.iree_run_exe}
         kernel_version:       {self.linux_kernel}
         output_dir:           {self.output_dir}
-        n_repeats:            {self.n_repeats}
         peano_commit_hash:    {self.peano_commit_hash}
         peano_dir:            {self.peano_dir}
         reset_npu_script:     {self.reset_npu_script}
@@ -382,7 +378,6 @@ def run_all(
     peano_dir,
     xrt_dir,
     vitis_dir,
-    n_repeats,
     return_on_fail,
     verbose,
 ):
@@ -404,9 +399,8 @@ def run_all(
        ./matmul_template. For example you might want to create ./conv_template
     """
 
-    if output_dir.exists():
-        shutil.rmtree(output_dir)
-    output_dir.mkdir()
+    if not output_dir.exists():
+        output_dir.mkdir()
     if not iree_install_dir.exists():
         raise RuntimeError(f"'{iree_install_dir}' is not a directory.")
     iree_compile_exe = find_executable(iree_install_dir, "iree-compile")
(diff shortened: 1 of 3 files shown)


build_tools/ci/cpu_comparison/run_test.py
```diff
@@ -114,83 +81,69 @@ def generate_aie_output(
...
+        if config.verbose:
+            print(f"Running {i}/{config.num_repeats} compares")
+        shell_out(config.reset_npu_script)
+        start = time.monotonic_ns()
+        shell_out(run_args, config.output_dir, config.verbose)
+        run_time += time.monotonic_ns() - start
+
+    print(f"Time spent in compilation: {compile_time // 1e6} [ms]")
+    print(
+        f"Time spent in running the model: {run_time // config.num_repeats // 1e6} [ms]"
```
newling, line 145:
```suggestion
        f"Mean time over {config.num_repeats} run(s) of model: {run_time // config.num_repeats // 1e6} [ms]"
```

newling (changes requested):
Thanks for the improvements. n_repeats vs num_repeats, not clear what's going on is num-repeats a global default? 

Please run black if you haven't already

makslevental:
> Thanks for the improvements. n_repeats vs num_repeats, not clear what's going on is num-repeats a global default?

err I'm blind - I didn't notice `n_repeats`.

> Please run black if you haven't already

I have black bound to `ctrl+shift+b` and I run it compulsively after every edit lol

makslevental:
> not clear what's going on is num-repeats a global default?

Since I didn't realize `n_repeats` existed, I patched in a different semantic for `num_repeat`s (just repeating the run). But actually the current/existing semantic you have for `n_repeats` is the right one for this test. So I'm gonna remove the changes associated with `num_repeats`.

newling:
> > not clear what's going on is num-repeats a global default?
> 
> Since I didn't realize `n_repeats` existed, I patched in a different semantic for `num_repeat`s (repeating only the run). But actually the current/existing semantic you have for `n_repeats` is the right one for this test. So I'm gonna remove the changes associated with `num_repeats`.

I was wondering about this actually, repeating the compile does hurt test time quite badly and I'm not sure it's worth it. Ideally we could have the option for both, and if we're repeating the compile it'd be nice if we could actually diff the artifacts (xclbin?) to see if compilation is deterministic. xclbin might contain timestamps or something which this not possible though. 

TL;DR: my current thinking is that if we have to choose one or the other, we shouldn't be repeating the compilation, just the run... FFS.

newling (commented):
n_repeats is still confusing: run_test takes as 2 arguments for the number of repeats, one via config (config.n_repeats) and one via n_repeats. 
 
In this PR, can you please remove all changes to n_repeats, and we can get back to it later? Everything else looks good to me

build_tools/ci/cpu_comparison/run_test.py
```diff
@@ -449,33 +403,22 @@ def run_all(argv):
...
-    assert_num_args(argv)
-
-    output_dir = os.path.realpath(argv[0])
-    os.makedirs(output_dir, exist_ok=True)
-    validate_directory(output_dir)
-
-    iree_install_dir = os.path.realpath(argv[1])
-    validate_directory(iree_install_dir)
+    if output_dir.exists():
+        shutil.rmtree(output_dir)
```
newling, line 407: This is dangerous! What if the user has something in output_dir that they don't want removed, for example maybe they accidentally provide the path to their home directory for output_dir.

newling (changes requested):

build_tools/ci/cpu_comparison/run_test.py
makslevental: is it more or less dangerous than the issue you had a little while ago where tests were passing even though nothing was compiling? i'm doing this by hand right now (ie in an outer script). anyway i can take it out and leave it to the user to make the decision.

newling: Maybe add a script flag which by default is "do not remove"

makslevental: too many script flags turns into "programming by script flags". i'll just leave it out.

newling: Ok. Yeah I'm not sure what a good solution is, sorry. Maybe walk the dir and remove xclbins, that would avoid the issue we had before when compilation was failing (although is seems arbitrary and complex).

makslevental:
@newling all set

newling (approved):
Thanks!

PR #605, opened 2024-07-29
Fixes for bugs in testing script 

newling:
- ability to not run reset_npu.sh 
- ability to not fail with peano_commit_hash not determined
- ability to set boolean flags to false

### DIFF
--- a/build_tools/ci/cpu_comparison/run_test.py
+++ b/build_tools/ci/cpu_comparison/run_test.py
@@ -524,11 +524,13 @@ if __name__ == "__main__":
     # problems with the default value of 1. It could be also be made nicer
     # by using type=bool, but this also has issues. So going with this
     # clunky design for now (feel free to improve).
+
+    cast_to_bool = lambda x: bool(x)
     parser.add_argument(
         "--return_on_fail",
         nargs="?",
         default=1,
-        type=int,
+        type=cast_to_bool,
         help=(
             "If 0, then the script will continue running even if a test fails, "
             "enumerating all failures. Otherwise the script will exit on the first failure."
@@ -539,7 +541,7 @@ if __name__ == "__main__":
         "--verbose",
         nargs="?",
         default=1,
-        type=int,
+        type=cast_to_bool,
         help="If 0, then print statements are suppressed, otherwise they are printed.",
     )
 
@@ -547,7 +549,7 @@ if __name__ == "__main__":
         "--reset_npu_between_runs",
         nargs="?",
         default=1,
-        type=int,
+        type=cast_to_bool,
         help=(
             "If 0 then the NPU is not reset between runs, otherwise it is reset. "
             "Resetting between runs can in theory help avoid certain types of "
--- a/build_tools/ci/cpu_comparison/run_test.py
+++ b/build_tools/ci/cpu_comparison/run_test.py
@@ -131,7 +131,8 @@ def generate_aie_output(
     ]
     if function_name:
         run_args += [f"--function={function_name}"]
-    shell_out(config.reset_npu_script)
+    if config.reset_npu_between_runs:
+        shell_out(config.reset_npu_script)
     start = time.monotonic_ns()
     shell_out(run_args, config.output_dir, config.verbose)
     run_time = time.monotonic_ns() - start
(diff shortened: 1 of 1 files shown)


makslevental:
There was an env check in the previous version of the script that got lost in the shuffle - there should be a check like `if os.getenv("GITHUB_ACTION"): reset_npu`.

newling:
> There was an env check in the previous version of the script that got lost in the shuffle - there should be a check like `if os.getenv("GITHUB_ACTION"): reset_npu`.
> 
> > not clear why I have to reset it, what is this testing?
> 
> it's not testing anything - it's resetting a flaky piece of hardware/flaky driver/flaky firmware so we can eliminate those from the set of possible causes for a failed test.

Ok, so when you test locally do you not reset the npu either? Do you think it's better to have a hard-coded rule for this (CI: reset, other: no reset) or a flag (like I try in this "PR")?

makslevental:
> Ok, so when you test locally do you not reset the npu either? Do you think it's better to have a hard-coded rule for this (CI: reset, other: no reset) or a flag (like I try in this "PR")?

Of course I think you should - locally I am setting `GITHUB_ACTION=true` and I have  

```
mlevental    ALL=(ALL) NOPASSWD: /usr/sbin/modprobe, /usr/sbin/rmmod, /usr/sbin/insmod, /bin/kmod
```

in `/etc/sudoers` (sorry I broadcast this in #shark-amd-aie when I landed the change but you were out I think).

> Do you think it's better to have a hard-coded rule for this (CI: reset, other: no reset) or a flag (like I try in this "PR")?

Up to you - the important thing is/was to make certain it does/did run in CI because that's where we were seeing all the flakes recently.

build_tools/ci/cpu_comparison/run_test.py
```diff
@@ -509,8 +516,13 @@ def run_all(
     parser.add_argument(
         "vitis_dir", nargs="?", default="/opt/Xilinx/Vitis/2024.1", type=abs_path
     )
+
+    # TODO(max) how do these get turned off?, how do I run with no verbose? 
     parser.add_argument("--return-on-fail", action="store_true", default=True)
     parser.add_argument("-v", "--verbose", action="store_true", default=True)
```
makslevental, line 522: d'oh - yea there's no way to turn these off. the only way to make optional flags that default to true actually work is `action="store_false"` (in which case they should be renamed). I guess drop the default but then add the explicit use of the flag in `ci.yml`?

build_tools/ci/cpu_comparison/run_test.py
```diff
@@ -509,8 +518,43 @@ def run_all(
...
+    # This (and other boolean flags) could be made more 'slick' by using
+    # `action='store_true'` in the `add_argument` call, but this has
+    # problems with the default value of 1. It could be also be made nicer
+    # by using type=bool, but this also has issues. So going with this
+    # clunky design for now (feel free to improve).
+    parser.add_argument(
+        "--return_on_fail",
+        nargs="?",
+        default=1,
+        type=int,
```
makslevental, line 531: what's the issue with bool and default=True?

newling: Trying to pass set it false at the command (=False, =0, ...) doesn't seem to do anything. No matter what I do it is True. 

With integers, I can pass it False / 0 at the command line and it actually makes it false.

makslevental, line 531:
```suggestion
    # clunky design for now (feel free to improve).
    cast_to_bool = lambda x: bool(x)
    parser.add_argument(
        "--return_on_fail",
        nargs="?",
        default=1,
        type=cast_to_bool,
```

makslevental: not sure - maybe something having to do with `add_argument` automatically injecting `action="store_true"` or something. anyway at least we should "cast"/"convert" to bool; see suggestion below.

build_tools/ci/cpu_comparison/run_test.py
```diff
@@ -509,8 +518,43 @@ def run_all(
...
+            "If 0, then the script will continue running even if a test fails, "
+            "enumerating all failures. Otherwise the script will exit on the first failure."
+        ),
+    )
+
+    parser.add_argument(
+        "--verbose",
+        nargs="?",
+        default=1,
+        type=int,
```
makslevental, line 542:
```suggestion
        type=cast_to_bool,
```

build_tools/ci/cpu_comparison/run_test.py
```diff
@@ -509,8 +518,43 @@ def run_all(
...
+        help="If 0, then print statements are suppressed, otherwise they are printed.",
+    )
+
+    parser.add_argument(
+        "--reset_npu_between_runs",
+        nargs="?",
+        default=1,
+        type=int,
```
makslevental, line 550:
```suggestion
        type=cast_to_bool,
```

makslevental (approved):
LGTM

PR #610, opened 2024-07-30
Allow 1 failed matmul test

makslevental:


jtuyls (commented):
I don't think we should be doing this. I would rather let the problematic tests fail. Edit: or add a flag and counter for tests that could potentially fail and check that max 1 of those fails.

newling (changes requested):
Alternatives: 
reduce num_repeats on tests if not already 1, or 
make the test compile only (there is an option to not run the vmfbs) and/or 
disable tests completely with follow-up tasks, or make a separate test if none of those is feasible. 

The idea of making CI fail just enough so we remember there's a problem, but not too often that it slows progress, doesn't seem sustainable.

newling:
It looks like this is the test which fails most frequently:

mm_chess_f32_matmul_f32_f32_m32_n32_k32

makslevental:
> mm_chess_f32_matmul_f32_f32_m32_n32_k32

you're reading the read-out incorrectly I believe - that's just the last test that runs. matmul_test just reportst at the end of the run how many tests have failed overall.

newling:
> you're reading the read-out incorrectly I believe - that's just the last test that runs. matmul_test just reportst at the end of the run how many tests have failed overall.

Yes you're right.

newling:
Lower repeat count: https://github.com/nod-ai/iree-amd-aie/pull/611

PR #611, opened 2024-07-30
Lower number of repeats in matmul tests

newling:
The number of times some of the objectFifo matmul tests are run was recently increased, to flush out our intermittent numerical errors. But the rate of failure in CI has gone above 50%, and this is slowing down developer speed across the board. No one is investigating the numerical issue now, so IMO it doesn't serve any purpose having this increased failure rate in CI. 

This PR lowers the number of repeats for tests (from 10 to 2, or 5 to 2).


jtuyls (approved):
LGTM. Will also help with CI time this and next week

PR #602, opened 2024-07-26
Compile chess ukernels at test time (rather than downloading)

newling:
Copy code from mlir-aie for generating ukernels with chess. Currently placed it in the build_tools/ci, we can consider making it part of core iree-amd-aie (with cmake) at later point (or now if people prefer). 

Files copies:
chess-clang : copied directly (except from removal of commented out code)
mm.cc : copied directly
zero.cc : copied directly (I'm not sure we need this though)
xchesscc_wrapper: simplified to only support AIE2, and reduce compile time warnings 

This PR also removes ukernels from run_matmul_test.sh, so that ukernels are only tested in cpu_comparison/run_test.sh. This is to reduce code complexity 

The one risk I see with this PR is that we become more dependent on chess -- downloading the ukernel meant that we could use ukernels without needing the heavy vitis machinery to start with

PR #617, opened 2024-07-31
Build mm.o from scratch

makslevental:
This PR enables building `mm.cc` from scratch instead of downloading from <somewhere>; see https://github.com/nod-ai/iree-amd-aie/pull/602 for more context.

**Note1**: currently we need to use chess to compile mm.cc always because peano doesn't distibute libcxx (which is needed for `<type_traits>`).

**Note 2**: this PR removes `AIETargetDirect.cpp` which is an NFC - I should've removed it a while ago since we've basically been "dog-fooding" the direct path for a couple of weeks now but now is as good a time as any.

**Note 3**: this PR also removes the ability to specify the kernel in the `aie.device` op. The reason to remove this is to prevent the linker script generation from including direct references to the ukernel objects and instead opting to pass them on the CLI (which enables reusing `mm.o` across compiles). There's really no need to directly call out which ukernel you plan to use anyway _since you're already referencing it by emitting the_ `func.call`. It's fine to pass a big kernel library on the command line (e.g., `xchescc` invocation) because the linker will DCE all function bodies you don't use.


makslevental:
BTW they (mlir-aie) is refactoring/updating mm.cc https://github.com/Xilinx/mlir-aie/pull/1631 (not sure if we should use their updated version or not).

jtuyls (approved):
LGTM

newling (commented):
Before this PR we could use ukernels without chess/vitis (that is what the example in run_test.py does). Is that still the case?

newling:
> Note 2: this PR removes AIETargetDirect.cpp which is an NFC - I should've removed it a while ago since we've basically been "dog-fooding" the direct path for a couple of weeks now but now is as good a time as any.

Please try not to mix patches...

makslevental:
> Before this PR we could use ukernels without chess/vitis (that is what the example in run_test.py does). Is that still the case?

No. If you want both that's a "have your cake and eat it too" kind of situation - like the prior was possible only because mm.o was cached somewhere. If you want to support both compile time building of mm.o _and_ use of cached mm.o that's a whole bunch of logic/threading/flags. Like it can be done ofc but I'm wondering if it's really necessary - in what situations do we have peano but not chess? Local dev? Because I'm assuming we certainly have both available for demos and perf runs right?

> Please try not to mix patches...

They're not completely orthogonal - making the changes that I did to `aie2xclbin` required making changes in both AIETarget and AIETargetDirect. And sending a stacked PR for a hipri item (Abishek's ask) seemed overly tedious.

newling:
> > Before this PR we could use ukernels without chess/vitis (that is what the example in run_test.py does). Is that still the case?
> 
> No. If you want both that's a "have your cake and eat it too" kind of situation - like the prior was possible only because mm.o was cached somewhere. If you want to support both compile time building of mm.o _and_ use of cached mm.o that's a whole bunch of logic/threading/flags. Like it can be done ofc but I'm wondering if it's really necessary - in what situations do we have peano but not chess? Local dev? Because I'm assuming we certainly have both available for demos and perf runs right?

I'm thinking ukernels could be provided pre-compiled by other teams, or might be created without chess. But ok, fine for now. So will mm.o will contain all ukernels in the future? Should probably be renamed to chess_microkernels.o or something in that case. 

> > Please try not to mix patches...
> 
> They're not completely orthogonal - making the changes that I did to `aie2xclbin` required making changes in both AIETarget and AIETargetDirect. And sending a stacked PR for a hipri item (Abishek's ask) seemed overly tedious.

PR #618, opened 2024-07-31
[CI] Add utility to run same matmul test on an array of shapes

jtuyls:


newling (approved):

newling:
Tangential thoughts:
- Let's try for fewer i32 and more i8 / bf16 tests where possible
- I/we should profile the time it takes to run the script, see if there are any easy pickings to make it faster

jtuyls:
> Tangential thoughts:
> 
> * Let's try for fewer i32 and more i8 / bf16 tests where possible
> * I/we should profile the time it takes to run the script, see if there are any easy pickings to make it faster

I  think more tests is better right now (we don't really have a lot of tests yet anyway). I32 has a reasonable coverage right now with different sizes and 50% or so of those tests have failed at some point, demonstrating that they are edge cases, so they are good to keep. Overall, I don't think we should be optimizing profile times too much right now as we're trying to get things functional and I don't think the current CI time is a real bottleneck for development. Also, some of these tests are quite slow for  now, but will become a lot faster over time with control code reduction, ukernels etc.

PR #621, opened 2024-07-31
Improvements to cpu_comparison testing

newling:
Changes to cpu_comparison testing:

1) Split AIE compile+run function into 2 functions (so we can repeat just the run step, future work)
2) Add a bit of structure the the tests, by partitioning them into Matmul/Conv/Smoke
3) Add 2 new flags. 
    - One to run only one of the above sets of test partition
    - One to only partially run the tests (just compile, don't run on AIE HW and do numerical comparison)
 4) Add smoke test to CI, which checks that some combination of the flags work as expected

### DIFF
--- a/build_tools/ci/cpu_comparison/run_test.py
+++ b/build_tools/ci/cpu_comparison/run_test.py
@@ -3,20 +3,16 @@
 # Copyright 2024 The IREE Authors
 
 import argparse
+from input_generator import generate_inputs, verify_determinism, load_input
+from matmul_template.matmul_generator import generate_matmul_test
+import numpy as np
 import os
+from pathlib import Path
 import re
 import subprocess
+from textwrap import dedent
 import time
 import urllib.request
-from pathlib import Path
-from textwrap import dedent
-
-import numpy as np
-
-from input_generator import generate_inputs, verify_determinism
-from matmul_template.matmul_generator import generate_matmul_test
-from input_generator import generate_inputs, verify_determinism, load_input
-from output_comparer import compare
 
 
 def matmul_from_input_strings(input_args):
@@ -60,8 +56,6 @@ def find_executable(install_dir: Path, executable_name):
 def shell_out(cmd: list, workdir=None, verbose=False):
     if workdir is None:
         workdir = Path.cwd()
-    if not isinstance(cmd, list):
-        cmd = [cmd]
     for i, c in enumerate(cmd):
         if isinstance(c, Path):
             cmd[i] = str(c)
@@ -87,7 +81,7 @@ def shell_out(cmd: list, workdir=None, verbose=False):
     return stdout_decode, stderr_decode
 
 
-def generate_aie_output(
+def generate_aie_artefacts(
     config,
     name,
     tile_pipeline,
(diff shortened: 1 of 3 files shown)


build_tools/ci/cpu_comparison/run_test.py
```diff
@@ -3,20 +3,17 @@
...
+from matmul_template.matmul_generator import generate_matmul_test
+import numpy as np
 import os
+from pathlib import Path
 import re
+from output_comparer import compare
 import subprocess
+from textwrap import dedent
 import time
 import urllib.request
```
makslevental, line 16: the reason these keep getting shuffled is because i have actually two formetters - black and one that clears out dead `import`s and alphabetizes. i'd prefer to keep them alpha but if we're not going all jump on board i'll stop running the second formetter (or at least reset these lines).

build_tools/ci/cpu_comparison/run_test.py
```diff
@@ -87,7 +84,7 @@ def shell_out(cmd: list, workdir=None, verbose=False):
     return stdout_decode, stderr_decode
 
 
-def generate_aie_output(
+def generate_aie_artefacts(
```
makslevental, line 87: Screenshot from 2024-08-01 17-53-03
https://www.youtube.com/watch?v=ddJqNY__9U8

jk i don't care

build_tools/ci/cpu_comparison/run_test.py
```diff
@@ -127,8 +125,21 @@ def generate_aie_output(
...
     shell_out(compilation_flags, config.output_dir, config.verbose)
     compile_time = time.monotonic_ns() - start
+    if config.verbose:
+        print(f"Time spent in compilation: {compile_time // 1e6} [ms]")
 
     aie_vmfb = config.output_dir / f"{name}_aie.vmfb"
+    if not aie_vmfb.exists():
+        raise RuntimeError(f"Failed to compile {test_file} to {aie_vmfb}")
+
+    return {"vmfb_path": aie_vmfb}
```
makslevental, line 135: i prefer [namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple) for these kinds of things (or dataclass). if you don't want to both with defining a type at top level you can do 

```suggestion
    return namedtuple("SOMETHING_THAT_WILL_SHOW_UP_IN_REPR/STR", ["vmfb_path"])(aie_vmfb)
```

then down below you'll just be able to `aie_artefacts.vmfb_path`.

But really I wouldn't go through this until we have actually multiple _artefacts_[^1] and just return `aie_vmfb`

[^1]: see what i did there

build_tools/ci/cpu_comparison/run_test.py
newling: My motivation for this was the duplicated imports (7 red lines, 6 green lines) the ordering was an after though, by all means run that formatter whenever you like

build_tools/ci/cpu_comparison/run_test.py
```diff
@@ -686,7 +784,37 @@ def all_tests(
...
+    parser.add_argument(
+        "--test_set",
+        type=str,
+        help=help_string,
+        default="All",
+    )
+
     args = parser.parse_args()
+
+    test_set_list = args.test_set.split(",")
```
makslevental, line 817: i'd put this directly into the `type=split_on_commas_lambda`

makslevental (approved):
LGTM (some small take it or leave it suggestions)

Issue #594, opened 2024-07-26
Matmul-elementwise fusion test failed to legalize operation 'vector.transpose'


yzhang93:
IR snippet:

```
%60 = vector.transfer_read %buf2[%c0, %c0, %c0, %c0, %c0, %c0], %c0_i32 {in_bounds = [true, true, true, true, true, true]} : memref<1x1x10x10x4x8xi32>, vector<1x1x10x10x4x8xi32>
%61 = vector.transfer_read %buf3[%c0, %c0, %c0], %c0_i32 {in_bounds = [true, true, true]} : memref<1x10x8xi32>, vector<1x10x8xi32>
%62 = vector.broadcast %61 : vector<1x10x8xi32> to vector<1x10x4x1x10x8xi32>
%63 = vector.transpose %62, [0, 3, 4, 1, 2, 5] : vector<1x10x4x1x10x8xi32> to vector<1x1x10x10x4x8xi32>
%64 = arith.addi %60, %63 : vector<1x1x10x10x4x8xi32>
%65 = arith.sitofp %64 : vector<1x1x10x10x4x8xi32> to vector<1x1x10x10x4x8xf32>
%66 = arith.mulf %65, %cst : vector<1x1x10x10x4x8xf32>
%67 = arith.divf %66, %cst_25 : vector<1x1x10x10x4x8xf32>
%68 = math.round %67 : vector<1x1x10x10x4x8xf32>
%69 = arith.addf %68, %cst_26 : vector<1x1x10x10x4x8xf32>
%70 = arith.cmpf ult, %69, %cst_27 : vector<1x1x10x10x4x8xf32>
%71 = arith.cmpf ugt, %69, %cst_28 : vector<1x1x10x10x4x8xf32>
%72 = arith.select %70, %cst_27, %69 : vector<1x1x10x10x4x8xi1>, vector<1x1x10x10x4x8xf32>
%73 = arith.select %71, %cst_28, %72 : vector<1x1x10x10x4x8xi1>, vector<1x1x10x10x4x8xf32>
%74 = arith.fptosi %73 : vector<1x1x10x10x4x8xf32> to vector<1x1x10x10x4x8xi8>
vector.transfer_write %74, %buf4[%c0, %c0, %c0, %c0, %c0, %c0] {in_bounds = [true, true, true, true, true, true]} : vector<1x1x10x10x4x8xi8>, memref<1x1x10x10x4x8xi8>
```

The above IR was vectorized from elementwise op

```
[...]

PR #622, opened 2024-07-31
Use correct peano opt flags

makslevental:
Related to https://github.com/nod-ai/iree-amd-aie/issues/594, we are getting stuff in `input.opt.ll` that we should not be getting because we are not disabling eg vectorization at the LLVM level.

Note, copy-pasta from [llvm-aie](https://github.com/Xilinx/llvm-aie/blob/0be095354faa49985cd031661853f6d9b9b787f2/clang/lib/Driver/ToolChains/AIE.cpp#L97-L121).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,9 +108,7 @@ jobs:
         run: |
           mkdir iree-install
           tar -xf iree-dist.tar -C iree-install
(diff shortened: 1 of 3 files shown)


newling (commented):
What's AA?

newling (approved):

nirvedhmeshram:
> What's AA?

Alias analysis
https://llvm.org/docs/Passes.html#id2

nirvedhmeshram (approved):
LGTM

makslevental:
> What's AA?

god i wish i could think of something funnier than alcoholic's anonymous

Issue #619, opened 2024-07-31
Failure running conv2d i32 with objectFifo


newling:
Checkpoints in IR:

IR Dump Before AssignTargetDevicesPass
IR Dump Before FoldMemRefAliasOps 
IR Dump Before AMDAIENormalizeLoopBounds
IR Dump Before AMDAIEInsertCores 

Core dumped in AMDAIEInsertCores. 

[first_run_conv.txt](https://github.com/user-attachments/files/16443802/first_run_conv.txt)

newling:
Update: probably need some sort of loop subsumption. Hoisting the dma_cpy_nd ops out of the scf.for loops

PR #623, opened 2024-07-31
[ObjectFifo][AMDAIEInsertCores] Fail gracefully when there is no thread mapping

newling:
This PR
1) improves failure mode encountered in https://github.com/nod-ai/iree-amd-aie/issues/619
2) changes logic to only process the inner-most forall loop(s) with thread mappings


yzhang93:
InsertCores pass needs thread mapping attribute to work. If not, it needs to emit error instead of silently continue.

Currently we are not using mapping attributes for Conv ops (https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp#L170), because there could be four parallel tiling dimensions. While for gpu mapping attributes there are only three dimensions [x, y, z].

MaheshRavishankar:
Fly by comment, I think the crash is probably closer to expected behavior. It shouldnt crash, but should probably be an error. What you need to do to map the four dimensions on two dimension is to linearize the parallel dimensions to 1D and then delinearize to 2D (just emit the indexing map that does this linearization/delinearization).

PR #626, opened 2024-08-01
Adding diagnostic message when finding amdaie_boot

josemonsalve2:
Adding an error message whenever amdaie_bootgen is not found.


josemonsalve2:
On a different note. Shouldn't this be initialized by default to a location. IDK if the best way would be to use the INSTALL_PREFIX, or if it would make sense to get the iree-compile path.

makslevental:
How does this happen? This is built and installed [here](https://github.com/nod-ai/iree-amd-aie/blob/269b49396aa077e9f65bb1a88e32fb5fe60c6400/runtime/src/iree-amd-aie/aie_runtime/CMakeLists.txt#L115) usually?

> On a different note. Shouldn't this be initialized by default to a location. IDK if the best way would be to use the INSTALL_PREFIX, or if it would make sense to get the iree-compile path.

You need something like an LLVMConfig/SiteConfig whatever for that (like basically the same thing that tells you where `llvm-*` is if you get a distro of LLVM. Quite a lot of boiler plate...

makslevental (approved):

josemonsalve2:
I just pushed a change using something similar to what other tools use

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -144,6 +145,28 @@ FailureOr<Path> findVitis(std::optional<Path> &vitisDir) {
   return *vitisDir;
 }
 
+static FailureOr<Path> findAieamdTool(std::string toolName,
```
makslevental, line 148: small nit 
```suggestion
static FailureOr<Path> findAMDAIETool(std::string toolName,
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -144,6 +145,28 @@ FailureOr<Path> findVitis(std::optional<Path> &vitisDir) {
...
+                                      const Path &amdAIEInstallDir) {
+  Path bootgenBin = "";
```
makslevental, line 150:
```suggestion
  Path toolBinExe = "";
```

PR #627, opened 2024-08-01
Don't append an empty vec to smallvec

makslevental:
whoops


MaheshRavishankar (approved):

makslevental:
@newling @nirvedhmeshram @jtuyls I'd like to figure out how this slipped past CI. The error was

```
LLVM ERROR: SmallVector unable to grow. Requested capacity (18446741108362260540) is larger than maximum value for size type (4294967295)
```

which is a [`report_fatal_error`](https://github.com/llvm/llvm-project/blob/3497211d80c22e1c488f1617db1d70413a2096ec/llvm/lib/Support/SmallVector.cpp#L70).

PR #631, opened 2024-08-01
Update mlir-air and mlir-aie versions

erwei-xilinx:


erwei-xilinx:
In mlir-aie, the operation containing the LX6 program has changed from `func.func` to `aiex.runtime_sequence`. https://github.com/Xilinx/mlir-aie/pull/1640

Cc: @makslevental, @jtuyls

newling (approved):
Thanks!

makslevental (changes requested):
Blocking for now before I have a chance to reviewv thoroughly, but I don't see the point of changing for the insertion site (lowertoaie) and the removal site (DmaToNPU). You could just leave both as is?

makslevental (approved):
My bad - it's air inserting and then in lowertoaie more stuff is inserted into the body. Ok that's fine.

PR #632, opened 2024-08-02
Port new mlir-aie pattern to iree-amd-aie

newling:
Pulls in 2 PRs from mlir-aie for the AIEVec dialect:

1) https://github.com/Xilinx/mlir-aie/pull/1655
2) https://github.com/Xilinx/mlir-aie/pull/1665

Required some changes to pass registration to get the testing working. 
I also did some refactoring specific to our simpler requirements (no AIE1 support, etc)

### DIFF
--- a/compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
+++ b/compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
@@ -90,25 +90,24 @@ static bool isGemmBTransposedContractionOp(vector::ContractionOp op) {
 // `vector.extract_strided_slice` selecting the subvector matching the
 // original `vector.transfer_read`.
 struct SplitUnalignedTransferReadPattern
-    : public OpConversionPattern<vector::TransferReadOp> {
-  using OpConversionPattern<vector::TransferReadOp>::OpConversionPattern;
+    : public OpRewritePattern<vector::TransferReadOp> {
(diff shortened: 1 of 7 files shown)


compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
```diff
@@ -16,15 +16,20 @@
 #include "AIEVecUtils.h"
 #include "Passes.h"
 #include "Utils.h"
+#include "aie/Dialect/AIEVec/AIEVecUtils.h"
```
makslevental, line 19: These are mlir-aie headers. We should not use absolutely any mlir-aie headers or code directly. I "ported" all the relevant stuff from them into `Utils.h` that's in this same dir.

makslevental (changes requested):

makslevental (approved):

makslevental:
> but its not obvious to me how to (seems like we can only test the pass `--convert-aievec-to-llvm`).

sorry I didn't see this - you can just add a `registerPass` thing [here](https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/aievec/Passes.h#L36) and then call it [here](https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/PluginRegistration.cpp#L41) and that'll expose it through CLI `iree-opt`.

PR #635, opened 2024-08-02
[aievec refactor] Remove unnecessary pass 

newling:
Suggestion from @jsetoain that this pass isn't needed with our compiler flow


makslevental:
i dig the cautious PR title lol

makslevental (approved):
gotta love `-247 +1`

jsetoain:
> Suggestion from @jsetoain that this pass isn't needed with our compiler flow

For the historical record:

That pass is needed to enable patterns that merge chains of macs (macs that accumulate over the same accumulator) into complex `vmac` operations (like MMA). These would commonly show-up when using the supervectorizer to vectorize gemm & conv1d coming from C (i.e.: represented as macs within a set of loops):

```mlir
    %0 = vector.load A[i]
    %1 = vector.load B[j]
    %2 = vector.load ACC[k]
    %3 = mac %0, %1, %2
    vector.store %3, ACC[k]
    // [...load next B...]
    %4 = vector.load ACC[k]
    %5 = mac %<...>, %<...>, %4
    vector.store %5, ACC[k]
```

The constant load/store of the accumulator hides the dependency in the macs. We remove them early to simplify conversion to complex aievec ops, but, actually, LLVM is very good at getting rid of these redundant load/store ops if they make it all the way to the backend.

IREE is generating code from a higher semantic level, so it can actually generate those complex ops directly, as opposed to let the optimization passes lift those macs up to MMAs and conv1d ops. I don't think this pass is doing anything for IREE.

This is subject to change in the future, of course, although if we find ourselves taking that route, I believe we should stop, take a step back, and re-think the whole thing.

Issue #430, opened 2024-06-17
[Tracking] Refactor MLIR-AIE dependency


makslevental:
TL;DR: This issue tracks ongoing work to refactor the dependency on MLIR-AIE

# Description:

The goal of this sprint of work is to remove the dependency on AIE and AIEX dialects completely (AIEVec will remain). This includes artifact generation. Thus, one final deliverable is a build configuration (of this repo/plugin) that does not need to clone/build `Xilinx/mlir-aie` at all (for platforms/deployments that don't need AIEVec). The "business" goal is a more stable device configuration and runtime layer/experience for higher level dialects.

# Development Plan

The work proceeds in roughly two necessary phases (for the MVP) and planned extension work:

1. **Vendoring/interning the relevant passes/parts of `mlir-aie`**;
    * `aie-rt` (the actual device configuration utility underneath `mlir-aie`) and `bootgen` (necessary for artifact construction) submodules;
    * `XCLBinGen`;
    * A minimal subset of AIE/AIEX passes necessary for supporting NPU.
      * `aie-assign-lock-ids`, `aie-assign-buffer-descriptor-ids`, `aie-assign-buffer-addresses-basic`, `aie-pathfinder`, `aie-localize-locks`, `aie-objectstateful-transform`, `aiex-dma-to-npu`;
      * Note, of these 7 passes, only `aie-pathfinder` and `aie-objectstateful-transform` perform any analysis
    * A minimal subset of translation utilities;
      * `aie-translate-cdo`, `aie-translate-bcf`, `aie-translate-ld-script`, `aie-translate-npu`.
2. **Refactor/merge/clean/DCE**;
    * Move all calls to `aie-rt` into a `iree_aie_runtime` library which emulates the various [`mlir_*_runtime` libs upstream](https://github.com/llvm/llvm-project/blob/main/mlir/lib/ExecutionEngine/CMakeLists.txt) (but still is only called at compile time; see planned work below); 
    * Merge all non-analysis passes into a single pass;
    * Simplify `aie-pathfinder` (remove legacy `d_ary_heap.h`) and `aie-objectstateful-transform`;
    * Re-design `AIETargetModel` as `AMDAIEDeviceModel` and base the latter on `aie-rt` (i.e., use `aie-rt` APIs for querying relevant device attributes/characteristics);
    * Re-design CDO emission to consume `objectfifo` directly instead of `buffer`, `dma`, `switchbox` etc;
[...]

PR #639, opened 2024-08-03
Delete MLIR-AIE submodule

makslevental:
This PR delete/removes the MLIR-AIE submodule. The load-bearing change is forking all of the `AIEOps.td` type things. The main reason for doing this now is https://github.com/Xilinx/mlir-aie/pull/1660, which does not fix the parser issue that broke us _but makes it permanent_.

In performing this fork we take the most minimal set of auxiliary/support impls/function/code. **Specifically, we take none of their verifiers**. The reason is, as usual, 1) we don't intend to continue to have most/many of these ops 2) they're now not an egress boundary i.e., they're purely "compiler state" ops/IR. Naturally, whichever ops we keep or iterate on, we should have verifiers for but I prefer 1) to add them incrementally rather than eagerly 2) add them in a separate PR (or PRs).

A couple things to call out:

1. The tests do not change (modulo `i32` -> `i8`). This is an indication that the fork/port/vendoring is successful;
    * This change, which brings in closer alignment with `aie-rt`, incurs that cost that `uint8_t`s need to printed with `std::to_string`[^3];
3. The prior vendored passes incur some minor changes because in doing the fork I removed all superfluous/redundant/confusing getters[^1];
4. Rather than adding patches to https://github.com/nod-ai/mlir-air I add a few local `sed`s. I did it this way because the patches would break an isolated build there[^2]. Not the prettiest thing but I'll keep up with it.

~~Note, there's a still bit to do (I want to rename the `xilinx::AIE` namespace and then use `typedef`s) but this is ready to review.~~

On second thought let's merge this and then move all the ops/stuff to our `mlir::iree_compiler::AMDAIE` namespace in a follow-up.

Fixes https://github.com/nod-ai/iree-amd-aie/issues/430

[^1]: Stuff like `getIDInt` when the existing attr getter is `getID`.
[^2]: Which never happens sure but just seems like it sprawls the hackiness rather than containing it to one place.
[^3]: Otherwise you print the ASCII character at the position rather the number itself.


jtuyls (approved):
LGTM

makslevental:
@nirvedhmeshram @newling @Abhishek-Varma you guys want to take a look or you're good with this?

build_tools/ci/build_test_cpp.sh
```diff
@@ -7,6 +7,7 @@ repo_root="$(cd $this_dir/../.. && pwd)"
 iree_dir="$(cd $repo_root/../iree && pwd)"
 build_dir="$repo_root/iree-build"
 install_dir="$repo_root/iree-install"
+rm -rf "$build_dir" "$install_dir"
```
nirvedhmeshram, line 10: Is this needed? (just thinking of some poor user who uses it and then deletes everything on what should have been an incremental change)

nirvedhmeshram (approved):
LGTM, just a minor question but you can address however and land.

build_tools/ci/build_test_cpp.sh
makslevental: Nah this was just to debug why CI was failing when local was passing. I'll remove

PR #641, opened 2024-08-04
Fix XRT bindings hangs and numerics issues

jtuyls:
Temporary fix as it seems to work reliably, but needs some follow ups to get rid of the global variable and output buffer sync inside the dispatch function.

Currently, I think there are two separate issues, which are fixed by this PR:

- Hangs, caused by the xrt::device getting destroyed while in use.
- Numerical issues, presumably caused by a missing sync on the output.

Locally, I can do 5000 runs without any issue on my debug shape 32x32x32.

### DIFF
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -209,6 +209,7 @@ function run_matmul_test() {
   # intermittently. It is also useful if a test is know to fail at runtime but
   # should still be checked to compile (set num_repeat_runs=0 in this case).
   local num_repeat_runs="1"
+  local num_corruption_repeat_runs="1"
 
   while [ "$#" -gt 0 ]; do
     case "$1" in
@@ -218,6 +219,10 @@ function run_matmul_test() {
         ;;
       --num_repeat_runs)
         num_repeat_runs="$2"
+        shift 2
+        ;;
+      --num_corruption_repeat_runs)
+        num_corruption_repeat_runs="$2"
         shift 2
         ;;
       --max_elements_to_check)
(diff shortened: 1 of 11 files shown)


makslevental:
I'll take a look on the flight today. Another thing I'm gonna do is try to enable ASAN build on CI which might definitively tell us where the problem is.

build_tools/ci/run_matmul_test.sh
```diff
@@ -209,6 +209,7 @@ function run_matmul_test() {
   # intermittently. It is also useful if a test is know to fail at runtime but
   # should still be checked to compile (set num_repeat_runs=0 in this case).
   local num_repeat_runs="1"
+  local num_corruption_repeat_runs="1"
```
nirvedhmeshram, line 212: add comment explaining this variable?

runtime/src/iree-amd-aie/driver/xrt/native_executable.cc
```diff
@@ -219,6 +220,8 @@ iree_status_t iree_hal_xrt_native_executable_create(
     instr->sync(XCL_BO_SYNC_BO_TO_DEVICE);
     iree_hal_xrt_kernel_params_t* params =
         &executable->entry_points[entry_ordinal];
+    params->xclbin = xclbin.release();
```
nirvedhmeshram, line 222: this should also be freed?

runtime/src/iree-amd-aie/driver/xrt/native_executable.h
```diff
@@ -21,6 +21,8 @@ extern "C" {
 
 // Object and launch parameters for a compute kernel.
 typedef struct iree_hal_xrt_kernel_params_t {
+  xrt::xclbin *xclbin;
+  xrt::device* device;
```
nirvedhmeshram, line 25: nit, `xrt::device *device` ?

runtime/src/iree-amd-aie/driver/xrt/direct_command_buffer.cc
```diff
@@ -323,9 +323,14 @@ static iree_status_t iree_hal_xrt_direct_command_buffer_dispatch(
   IREE_RETURN_AND_END_ZONE_IF_ERROR(
       z0, iree_hal_resource_set_insert(command_buffer->resource_set, 1,
                                        &executable));
+
+  // if (!kernel_params.device) std::cout << "NULLPTRd device" << std::endl;
+  xrt::device device = *kernel_params.device;
```
nirvedhmeshram, line 327: unused variable and hence ununsed struct member?

runtime/src/iree-amd-aie/driver/xrt/direct_command_buffer.cc
```diff
@@ -323,9 +323,12 @@ static iree_status_t iree_hal_xrt_direct_command_buffer_dispatch(
   IREE_RETURN_AND_END_ZONE_IF_ERROR(
       z0, iree_hal_resource_set_insert(command_buffer->resource_set, 1,
                                        &executable));
+
+  xrt::device device = *kernel_params.device;
   xrt::kernel kernel = *kernel_params.kernel;
   xrt::bo instr = *kernel_params.instr;
   uint32_t num_instr = kernel_params.num_instr;
+  instr.sync(XCL_BO_SYNC_BO_TO_DEVICE);
```
nirvedhmeshram, line 331: This is already done, in native_execitable code?

nirvedhmeshram (approved):
LGTM!

PR #643, opened 2024-08-04
[LoopSubsumption] Make pass fail when no device attribute available 

newling:
I guess it's a question on where we want to lie on the spectrum A <-> B where

A: continue when anything unexpected happens in a pass
B: fail when when anything expected happens in a pass

This PR moves closer to B. Also in PR: additional diagnostics for determining why subsumption doesn't happen. 

@jtuyls I'd like to use this pass for loop subsumption of amdaie.dma_cpy_nd (currently this pass only acts on amdaie.npu.dma_cpy_nd), does that make sense? Related to https://github.com/nod-ai/iree-amd-aie/issues/619


jtuyls (approved):
> A: continue when anything unexpected happens in a pass
B: fail when when anything expected happens in a pass

We should indeed fail in case there is no device attribute, but in most other cases I think we should just keep going.

PR #650, opened 2024-08-07
Adding conditional statement to remove target_include_directories when tests are not enabled

josemonsalve2:
When building iree-amd-aie with IREE_BUILD_TESTS=OFF, this line generates an error because the target is not being built:

```
CMake Error at /opt/iree/installers/AARInternal/iree-amd-aie/runtime/src/iree-amd-aie/aie_runtime/test/CMakeLists.txt:109 (target_include_directories):
  Cannot specify include directories for target
  "iree-amd-aie_aie_runtime_test_DeviceModelTestCppTest" which is not built
  by this project.
```

### DIFF
--- a/runtime/src/iree-amd-aie/aie_runtime/CMakeLists.txt
+++ b/runtime/src/iree-amd-aie/aie_runtime/CMakeLists.txt
@@ -500,7 +500,11 @@ endif()
 # iree-aie-runtime-static
 # ##############################################################################
 
-iree_add_all_subdirs()
+add_subdirectory(Utils)
+
+if (IREE_BUILD_TESTS)
+  add_subdirectory(test)
+endif()
 
 iree_tablegen_library(
   NAME


makslevental:
You should just remove this [iree_add_all_subdirs()](https://github.com/nod-ai/iree-amd-aie/blob/57238671a43d569c7ad8d3abec8a12ab7f3cabe2/runtime/src/iree-amd-aie/aie_runtime/CMakeLists.txt#L503) and instead do 

```
if(BUILD_TESTS)
  add_subdirectory(tests)
endif()
```

makslevental (changes requested):
You should just remove this [iree_add_all_subdirs()](https://github.com/nod-ai/iree-amd-aie/blob/57238671a43d569c7ad8d3abec8a12ab7f3cabe2/runtime/src/iree-amd-aie/aie_runtime/CMakeLists.txt#L503) and instead do 

```
if(BUILD_TESTS)
  add_subdirectory(tests)
endif()
```

josemonsalve2:
Do you mean in this file, or one level above?

In this file there is no `iree_add_all_subdirs()`. There could be one instead of the add_subdirectory line at the end.

makslevental:
> Do you mean in this file, or one level above?

yes one level above (that link I pasted goes to runtime/src/iree-amd-aie/aie_runtime/CMakeLists.txt)

josemonsalve2:
But wouldn't that also remove the Util folder?

makslevental:
> But wouldn't that also remove the Util folder?

You can then `add_subdirectory(Util)` unqualified (i.e., no guard).

makslevental (approved):

PR #651, opened 2024-08-07
[Testing] Command-line argument to take additional flags

newling:
I find myself changing run_test.py to add iree-compile flags often enough that it might be useful to have a flag to do this from the command-line


jtuyls (approved):
LGTM

PR #653, opened 2024-08-08
Replace logical objectfifo produce/consume with core in/out DMA operands

jtuyls:
Remove the partially redundant produce/consume ops as they have been largely replaced already with the `logicalobjectfifo.access` ops. The main use for them is to provide information on which cores are using which DMAs which can be handled by adding input and output DMA operands to the `amdaie.core` ops.


Abhishek-Varma (approved):
LGTM!

makslevental:
n00b question: does `!amdaie.logicalobjectfifo` (the type) actually do anything on top of `memref`? in `aie` there's the same thing and all it does is it forces you to do `objfifoT.getElement().getType()`.

PR #652, opened 2024-08-08
[ObjectFifo] Modify LowerToAIE pass to take flattened objectfifos

yzhang93:
This PR is to add `FlattenLogicalObjectFifo` pass https://github.com/nod-ai/iree-amd-aie/pull/638 into the pipeline, and modifies the tests accordingly.

PR #657, opened 2024-08-09
[DistributeCoresAndObjectFifos] Remove distributeSharedMemory function

yzhang93:
After some refactor of codes, distributeSharedMemory function is no longer needed and generated wrong allocations which caused test failure in https://github.com/nod-ai/iree-amd-aie/pull/652.


newling (approved):

jtuyls (commented):
For future reference, maybe add a description in the PR to describe why this is not needed anymore and actually creates issues.

PR #658, opened 2024-08-09
[AMDAIEInsertCore] Handle convolution case 

newling:
This is a small change to get convolution lowering through the insert core pass. Specifically, it makes sure that linalg::ConvolutionInterfaceOps are handled correctly (in the same way as vector::ContractionOp). I think the pass can be improved further at a later point (we should be able to lift only a subset of the nested loop into the core I think).

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCores.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCores.cpp
@@ -134,7 +134,7 @@ LogicalResult insertCoreOps(mlir::ModuleOp moduleOp) {
       // loop nest inside the amdaie.core op here. Currently look for a
       // subset of ops which we know should be in the core.
       // TODO(newling) improve this design.
-      auto insertInCore =
+      bool insertInCore =
           isa<linalg::LinalgOp>(op) || isa<vector::ContractionOp>(op) ||
           isa<memref::ExtractStridedMetadataOp>(op) || isa<func::CallOp>(op);
       if (insertInCore) {
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCores.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCores.cpp
@@ -18,8 +18,9 @@
 #include "iree-amd-aie/IR/AMDAIEOps.h"
 #include "iree-amd-aie/Transforms/AMDAIEOpUtils.h"
 #include "iree-amd-aie/Transforms/Passes.h"
-#include "iree-amd-aie/Transforms/Transforms.h"
 #include "iree/compiler/Codegen/TransformStrategies/GPU/Common.h"
+#include "mlir/Dialect/Linalg/IR/LinalgInterfaces.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
 
 #define DEBUG_TYPE "iree-amdaie-insert-cores"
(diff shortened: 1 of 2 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertCores.cpp
```diff
@@ -136,14 +128,29 @@ LogicalResult insertCoreOps(mlir::ModuleOp moduleOp) {
         //               As of now this hasn't turned up yet, so will table this
         //               for now.
         coreOp.setLinkWith(fnDecl->getAttrOfType<StringAttr>("link_with"));
+      }
+
+      // Some ops are surrrounded by scf.for loop nests. Place the entire
+      // loop nest inside the amdaie.core op here. Currently look for a
+      // subset of ops which we know should be in the core.
+      // TODO(newling) improve this design.
+      auto insertInCore =
```
jtuyls, line 137:
```suggestion
      bool insertInCore =
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_cores.mlir
```diff
@@ -279,3 +279,104 @@ module {
     return
   }
 }
+
+// -----
+
+// This is the starting IR currently generated for tiling a convd op.
```
jtuyls, line 285:
```suggestion
// This is the starting IR currently generated for tiling a conv2d op.
```

jtuyls (approved):
LGTM. Two small suggestions.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_cores.mlir
```diff
@@ -279,3 +279,104 @@ module {
...
+    %5 = amdaie.logicalobjectfifo.from_memref %4, {} : memref<2x12x12x64xi32> -> !amdaie.logicalobjectfifo<memref<2x12x12x64xi32>>
+    memref.assume_alignment %4, 64 : memref<2x12x12x64xi32>
+    scf.forall (%arg0, %arg1, %arg2) in (3, 3, 16) {
+      %6 = affine.apply #map(%arg2)
+      %7 = affine.apply #map(%arg1)
+      %8 = affine.apply #map(%arg0)
+      %alloc = memref.alloc() : memref<2x6x6x32xi32, 1 : i32>
+      %9 = amdaie.logicalobjectfifo.from_memref %alloc, {} : memref<2x6x6x32xi32, 1 : i32> -> !amdaie.logicalobjectfifo<memref<2x6x6x32xi32, 1 : i32>>
+      %10 = amdaie.logicalobjectfifo.from_memref %alloc, {} : memref<2x6x6x32xi32, 1 : i32> -> !amdaie.logicalobjectfifo<memref<2x6x6x32xi32, 1 : i32>>
+      %11 = amdaie.dma_cpy_nd(%9[] [] [], %1[%c0, %8, %7, %c0] [%c2, %c6, %c6, %c32] [%c6272, %c448, %c32, %c1]) : (!amdaie.logicalobjectfifo<memref<2x6x6x32xi32, 1 : i32>>, !amdaie.logicalobjectfifo<memref<2x14x14x32xi32>>)
```
Abhishek-Varma, line 339: This SSA isn't being used in the IR and the lit test can be made shorter. You may refer to either the [vectorization](https://github.com/nod-ai/iree-amd-aie/blob/0bf7b9f56b622844fb4cab8201664d2fdb0b4c43/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_cores.mlir#L154) or [ukernel](https://github.com/nod-ai/iree-amd-aie/blob/0bf7b9f56b622844fb4cab8201664d2fdb0b4c43/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_cores.mlir#L233) lit test and follow a similar structure.

Abhishek-Varma (commented):
Not blocking, but would prefer the comment pertaining to lit test addressed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_cores.mlir
newling: I agree, I'll strip this down a bit.

Issue #644, opened 2024-08-05
ObjectFifo Matmul + Elementwise


jtuyls:
The main failure in matmul + elementwise seems to be caused by too many connections being created. There are multiple ways this could be solved:

1. Packet routing, so streams can be reused for multiple connections.
2. Reprogramming DMAs and routes after matmul data has been moved in, to create new routes for elementwise data movement.
3. Combining connections together whenever possible. This could either be combining A and B connections, so the same streams are used for these, or combining the elementwise constant data connection with the A or B connection.

As there is no e2e support yet for the more general approaches 1 and 2, the current thought is to implement approach 3. This also still has some potential long term benefits as depending on how this is implemented and what operations are targeted, this could give very good performance as well.

For approach 3, we need following fixes and new transformations in the flow:

1. Fix access ops not at the right place (https://github.com/nod-ai/iree-amd-aie/pull/625) (@jtuyls)
2. Split temp memrefs in distribute pass (one 4x4x4x4x4 -> 4 separate 4x4x4x4 memrefs) (@Abhishek-Varma )
3. Linearize logical objectfifo memrefs (@yzhang93 )
4. Combine DMAs and insert additional reads after same number of accesses in both cores (@jtuyls)

jtuyls:
@Abhishek-Varma 

For 4, see the below snippet for a sample in/output:

NOTE: The circular DMA objectfifos need to be decoupled from the actual underlying memref argument, so a single circular DMA can be reused for multiple different memref arguments (see `ARG_NEW` in the expected output). I will have a look at how to update the ops to accomplish this.

```
%tile_0_2 = amdaie.tile(%c0, %c2)
%tile_1_2 = amdaie.tile(%c1, %c2)
%tile_0_3 = amdaie.tile(%c0, %c3)
%tile_1_3 = amdaie.tile(%c1, %c3)

%0 = amdaie.circular_dma_cpy_nd(%arg1[] [] [], %arg0[] [] []) : (!amdaie.logicalobjectfifo<memref<2x16xi32, 1>>, !amdaie.logicalobjectfifo<memref<2x16xi32>>)
%1 = amdaie.circular_dma_cpy_nd(%arg2[] [] [], %arg1[0] [16] [1]) : (!amdaie.logicalobjectfifo<memref<1x16xi32, 2>>, !amdaie.logicalobjectfifo<memref<2x16xi32, 1>>)
[...]

PR #659, opened 2024-08-09
[ObjectFifo] Create a new pass to split L2 buffers

Abhishek-Varma:
This PR includes 2 commits :-

**Commit 1**
-- This commit introduces a new pass `--iree-amdaie-split-buffers` to split L2 buffers for dealing with Matmul+Elementwise.
-- It addresses sub-action 2 as well from https://github.com/nod-ai/iree-amd-aie/issues/644

**Commit 2**
- Adds `--iree-amdaie-split-buffers` pass to the objectFifo pipeline.

NOTE: Let me know what can we name the pass instead of `iree-amdaie-split-buffers` - `iree-amdaie-split-l2-buffers` is a possibility but we clearly aren't splitting ALL L2 buffers.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -402,6 +402,12 @@ def AMDAIEPropagateDataLayout :
   let constructor = "mlir::iree_compiler::AMDAIE::createAMDAIEPropagateDataLayoutPass()";
 }
 
+def AMDAIESplitBuffers :
+  Pass<"iree-amdaie-split-buffers", "ModuleOp"> {
+  let summary = "Split buffers.";
```
yzhang93, line 407: Could you append the summary to include why and when this pass is needed?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitBuffers.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+  for (Operation *op : toBeErased) {
+    op->dropAllUses();
+    rewriter.eraseOp(op);
+  }
+}
+
+}  // namespace
+
+std::unique_ptr<Pass> createAMDAIESplitBuffersPass() {
```
jtuyls, line 141: I think it would be better to rename to `SplitLogicalObjectFifos`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_buffers.mlir
```diff
@@ -0,0 +1,500 @@
+// RUN: iree-opt --pass-pipeline="builtin.module(iree-amdaie-split-buffers,cse)" --split-input-file --verify-diagnostics %s | FileCheck %s
+
+// Glossary:
+// candidate core op : they are those amdaie.core ops which have at least three input dma ops.
+// non-candidate core op : they are those amdaie.core ops which have less than three input dma ops.
+
+// Test non-candidate core op.
```
jtuyls, line 7: Could you make this non-candidate core op test more minimal?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_buffers.mlir
```diff
@@ -0,0 +1,500 @@
...
+    memref.dealloc %alloc_0 : memref<1x2x32x32xi32, 1 : i32>
+    memref.dealloc %alloc_1 : memref<2x2x32x32xi32, 1 : i32>
+    memref.dealloc %alloc_2 : memref<128x128xi32>
+    return
+  }
+}
+
+// -----
+
+// Test one candidate core op.
```
jtuyls, line 80: Could you make the test more minimal? For example, there can be less strides, offsets, sizes (and use ints instead of index). Same for other tests below.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_buffers.mlir
```diff
@@ -0,0 +1,500 @@
...
+//  CHECK-SAME:         memref<128x128xi32> -> !amdaie.logicalobjectfifo<memref<128x128xi32>>
+//       CHECK:   scf.forall
+//       CHECK:       %[[DMA_CPY_ND_L3_TO_L2:.*]] = amdaie.dma_cpy_nd(%[[L2_OBJECTFIFO]]
+//  CHECK-SAME:                                                       %[[L3_OBJECTFIFO]]
+//       CHECK:       amdaie.logicalobjectfifo.from_memref
+//       CHECK:       amdaie.logicalobjectfifo.from_memref
+//       CHECK:       amdaie.dma_cpy_nd
+//       CHECK:       amdaie.dma_cpy_nd
+//       CHECK:       %[[L1_OBJECTFIFO:.*]] = amdaie.logicalobjectfifo.from_memref %[[L1_ALLOC]]
+//       CHECK:       %[[DMA_CPY_ND_L2_TO_L1:.*]] = amdaie.dma_cpy_nd(%[[L1_OBJECTFIFO]]
```
jtuyls, line 101: Could you add offsets, strides, sizes to the check as well? These  might need to be updated.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitBuffers.cpp
```diff
@@ -0,0 +1,145 @@
...
+        l3ToL2DmaOp.getLoc(), source, l3ToL2DmaOp.getTargetMixedOffsets(),
+        l3ToL2DmaOp.getTargetMixedSizes(), l3ToL2DmaOp.getTargetMixedStrides(),
+        l3ToL2DmaOp.getSource(), l3ToL2DmaOp.getSourceMixedOffsets(),
+        l3ToL2DmaOp.getSourceMixedSizes(), l3ToL2DmaOp.getSourceMixedStrides());
+
+    // Create new L2 -> L1 Input DmaOp.
+    rewriter.setInsertionPoint(l2ToL1DmaOp);
+    auto newL2ToL1DmaOp = rewriter.create<AMDAIE::DmaCpyNdOp>(
+        l2ToL1DmaOp.getLoc(), l2ToL1DmaOp.getTarget(),
+        l2ToL1DmaOp.getTargetMixedOffsets(), l2ToL1DmaOp.getTargetMixedSizes(),
```
jtuyls, line 108: The logic here doesn't take changes offsets into account and these might need to change. For example, dma1: [0] [16] [1], dma2: [16] [16] [1]. After splitting, the dma2 offset needs to become zero to be correct. I think more test cases are needed + think about complex cases, overlapping access patterns for example.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitBuffers.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include "iree-amd-aie/Transforms/Passes.h"
+#include "iree-amd-aie/Transforms/Transforms.h"
+#include "iree/compiler/Codegen/TransformStrategies/GPU/Common.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/SCF/Transforms/Transforms.h"
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/IR/Iterators.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "mlir/Transforms/LoopInvariantCodeMotionUtils.h"
```
jtuyls, line 17: Includes can be made more minimal I think, for example the `GreedyPatternRewriteDriver.h` is not used.

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_buffers.mlir
Abhishek-Varma: I've added the check for L2 buffers as L1/L3 buffers' access patterns will remain the same.

Could you check and let me know if it looks okay to you?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitBuffers.cpp
Abhishek-Varma: Good catch!

Here's what I'm currently doing : Since the new L2 buffer we're creating is linearized form of the L1 buffer's shape, I'm creating new offset/size/stride as per this linearized shape.
```
Eg: L1: <1x1x8x4x8x4>
    L2: <1024> (we get this by linearizing L1's buffer : 1x1x8x4x8x4 = 1024)
      
    So, because there is now 1:1 correspondence between L1 and L2
    I'm creating the offset/size/stride as [0,0,0,0][1,1,1,1024][0,0,1024,1]
      
    And since the same L2 buffer will be fed from L3, I'm using the same access pattern as above.
```

I might be wrong here but this linearized access pattern should also take care of overlapping access patterns ? Because of the 1:1 correspondence between L1 and L2 ?

Could you check and let me know if it looks okay to you?

jtuyls: > Since the new L2 buffer we're creating is linearized form of the L1 buffer's shape, I'm creating new offset/size/stride as per this linearized shape.

That's not correct as changing the offset/stride/access pattern changes the data packing and will therefore result in incorrect results.

> I might be wrong here but this linearized access pattern should also take care of overlapping access patterns ?

Were do you see an issue with overlapping access patterns?

Abhishek-Varma:
Previously L2's access patterns were incorrect after being split.

L3 -> L2 requires 4D access pattern. L2 -> L1 requires 6D access pattern. The previous revision was incorrectly forming 4D access pattern for both cases and was doing so by linearizing the L2 buffer.

The current revision takes care of the following :-
1. 4D access pattern for L3 -> L2.
2. 6D access pattern for L2 -> L1.
3. Correct offset (L2 for both `L3 -> L2` and `L2 -> L1`)
4. Correct size (L2 for both `L3 -> L2` and `L2 -> L1`)

I've pushed the current revision as a WIP because now I need to take a look at L3's case as well as strides in the access pattern.

Once verified, only then will I update the lit test case and clean up the code before asking for re-review. But attaching here the [IR snippet](https://gist.github.com/Abhishek-Varma/e95b0dc14646a8819007b634707016e3) - now instead of `<1024>`, I'm forming correct four splits (`<1x1x32x32>`) of the L2 buffer (`<2x2x32x32>`).

CC: @jtuyls

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjectFifos.cpp
```diff
@@ -0,0 +1,214 @@
...
+        cast<MemRefType>(
+            l3ToL2DmaOp.getTargetObjectFifo().getMemref().getType())
+            .getShape());
+    DenseMap<int64_t, int64_t> dimToOffsetMapForL3AsSource;
+    for (unsigned dim : splitDimensionsSet) {
+      std::optional<int64_t> constantOffset =
+          getConstantIntValue(staticL2AsSourceOffsets[dim]);
+      if (!constantOffset) {
+        l2ToL1DmaOp->emitOpError("found a non-constant value for offset");
+        return signalPassFailure();
```
jtuyls, line 111: Should we error out? Can't we just return if the offsets are non-constant and maybe add a debug message? And this could also be checked earlier in this function.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjectFifos.cpp
```diff
@@ -0,0 +1,214 @@
...
+      }
+      dimToOffsetMapForL3AsSource.insert(
+          {dim + 2,
+           constantOffset.value() *
+               (getConstantIntValue(staticL2AsTargetSizes[dim + 2]).value())});
```
jtuyls, line 116: WHat does this `dim + 2` do?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjectFifos.cpp
```diff
@@ -0,0 +1,214 @@
...
+    auto source = rewriter.create<AMDAIE::LogicalObjectFifoFromMemrefOp>(
+        rewriter.getUnknownLoc(), LogicalObjectFifoType::get(type),
+        newAllocOp.getResult(), sourceObjectFifo.getTiles());
+
+    SmallVector<OpFoldResult, 4> staticL3AsSourceOffsets =
+        l3ToL2DmaOp.getSourceMixedOffsets();
+    for (auto [dim, offsetToAdd] : dimToOffsetMapForL3AsSource) {
+      auto applyOp = cast<affine::AffineApplyOp>(
+          cast<Value>(staticL3AsSourceOffsets[dim]).getDefiningOp());
+      AffineMap affineMap = applyOp.getAffineMap();
```
jtuyls, line 150: What if the offset is not a `AffineApplyOp`?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjectFifos.cpp
```diff
@@ -0,0 +1,214 @@
...
+      AffineExpr affineExpr = affineMap.getResult(0);
+      AffineExpr newAffineExpr = affineExpr + offsetToAdd;
+      auto newAffineMap = AffineMap::get(/*dimCount=*/1, /*symbolCount=*/0,
+                                         {newAffineExpr}, context);
+      IRRewriter::InsertPoint oldInsertionPoint = rewriter.saveInsertionPoint();
```
jtuyls, line 155: Use an `InsertionGuard` instead?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjectFifos.cpp
```diff
@@ -0,0 +1,214 @@
...
+    auto newL2ToL1DmaOp = rewriter.create<AMDAIE::DmaCpyNdOp>(
+        l2ToL1DmaOp.getLoc(), l2ToL1DmaOp.getTarget(),
+        l2ToL1DmaOp.getTargetMixedOffsets(), l2ToL1DmaOp.getTargetMixedSizes(),
+        l2ToL1DmaOp.getTargetMixedStrides(), source,
+        llvm::ArrayRef(staticL2AsSourceOffsets),
+        llvm::ArrayRef(staticL2AsSourceSizes),
+        llvm::ArrayRef(staticL2AsSourceStrides));
+    rewriter.replaceOp(l2ToL1DmaOp, newL2ToL1DmaOp);
+    // We have to discard non-zero offsets as subview has been replaced by a
+    // dedicated allocated memref.
```
jtuyls, line 183: I think this is an old comment? Also, we should only erase offsets on the dimensions found above?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjectFifos.cpp
```diff
@@ -0,0 +1,214 @@
...
+    // Remove old dealloc.
+    memref::DeallocOp oldDeallocOp;
+    for (Operation *userOp : sourceAllocOp->getUsers()) {
+      if (auto deallocUser = dyn_cast<memref::DeallocOp>(userOp)) {
+        oldDeallocOp = deallocUser;
+      }
+    }
+    if (oldDeallocOp) {
+      rewriter.eraseOp(oldDeallocOp);
+    }
```
jtuyls, line 199: Add these `dealloc` ops to `toBeErased`?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logical_objectfifos.mlir
```diff
@@ -0,0 +1,348 @@
...
+//       CHECK:       }
+//       CHECK:   memref.dealloc %[[L2_ALLOC_0]] : memref<1x1x32x32xi32, 1 : i32>
+//       CHECK:   memref.dealloc %[[L2_ALLOC_1]] : memref<1x1x32x32xi32, 1 : i32>
+#map = affine_map<(d0) -> (d0 * 64)>
+#map1 = affine_map<(d0, d1, d2, d3, d4, d5, d6, d7, d8) -> (d0, d2, d5, d3, d6, d8)>
+#map2 = affine_map<(d0, d1, d2, d3, d4, d5, d6, d7, d8) -> (d2, d1, d4, d5, d8, d7)>
+#map3 = affine_map<(d0, d1, d2, d3, d4, d5, d6, d7, d8) -> (d0, d1, d4, d3, d6, d7)>
+#map4 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d1, d2, d3, d4, d5)>
+module {
+  func.func @split_l2_buffer_two_core_ops(%arg0: !amdaie.logicalobjectfifo<memref<1x1x4x8x4x8xi32, 2 : i32>>, %arg1: !amdaie.logicalobjectfifo<memref<1x1x8x4x8x4xi32, 2 : i32>>, %arg2: !amdaie.logicalobjectfifo<memref<1x1x8x8x4x4xi32, 2 : i32>>, %arg3: !amdaie.logicalobjectfifo<memref<2x2x32x32xi32, 1 : i32>>) {
```
jtuyls, line 109: As discussed offline, add all 4 cores.

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjectFifos.cpp
```diff
@@ -0,0 +1,226 @@
...
+  // applicable) from them.
+  // TODO: We will generalize this later.
+  moduleOp.walk([&](AMDAIE::CoreOp coreOp) {
+    SmallVector<Value> inputDmas = coreOp.getInputDmas();
+    if (inputDmas.size() != 3) return WalkResult::skip();
+    auto dmaCpyNdOp = inputDmas[2].getDefiningOp<AMDAIE::DmaCpyNdOp>();
+    assert(dmaCpyNdOp && "expected an amdaie.dma_cpy_nd op");
+    l2ToL1DmaOps.push_back(dmaCpyNdOp);
+    return WalkResult::advance();
+  });
```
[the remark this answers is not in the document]
Abhishek-Varma: Have added this as a separate function which we can abstract out in AMDAIEUtils.cpp (maybe?) when dealing with the reuse ?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjectFifos.cpp
Abhishek-Varma: I've used `emitRemark` instead. Let me know if it looks okay to you now?

I've not preponed this check and chose to keep it here because we need to check the same for L3->L2 and anyway need to use those constants at this part of the function for computing the new offset.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjectFifos.cpp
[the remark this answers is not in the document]
jtuyls: I was thinking the other way around, i.e. moving the splitting logic into a separate file `AMDAIELogicalObjFifoSplittingUtils.cpp` and then maybe rename the current pass from `AMDAIESplitLogicalObjectFifosPass` to `AMDAIESplitLogicalObjFifosForConnectionReuse` or something. The reason for this is that I am expecting that we will soon need a new transformation that uses a different check to split shared memory logical objectFifos across different memtiles for performance (`AMDAIESplitLogicalObjFifosAcrossSharedMem`). This pass can then reuse the splitting logic from `AMDAIELogicalObjFifoSplittingUtils.cpp`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjectFifos.cpp
```diff
@@ -0,0 +1,312 @@
...
+        l3ToL2DmaOp.getTargetMixedStrides();
+    SmallVector<int64_t, 4> l2ShapeAsTarget = llvm::to_vector(
+        cast<MemRefType>(
+            l3ToL2DmaOp.getTargetObjectFifo().getMemref().getType())
+            .getShape());
+    // We traverse through the split dimensions we captured earlier and for each
+    // such dimension we perform the following updates :-
+    // 1. Maintain a map: DIM -> CONST_OFFSET_TO_ADD. This is done with the
+    //    assumption that L3<->L2 is 4D and L2<->L1 is 6D. `DIM` here is split
+    //    dimension + 2. `CONST_OFFSET_TO_ADD` is the constant we get by
```
jtuyls, line 132: We should avoid this assumption.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjectFifos.cpp
```diff
@@ -0,0 +1,312 @@
...
+           . . . .
+               ___
+           . .|. .|
+           . .|. .|
+              -----
+
+      The following logic performs this computation of offsets for L3 source.
+    */
+    for (auto [dim, offsetToAdd] : dimToOffsetMapForL3AsSource) {
+      auto l3SourceOffsetVal = cast<Value>(staticL3AsSourceOffsets[dim]);
```
jtuyls, line 233: How are you sure it's a value?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjectFifos.cpp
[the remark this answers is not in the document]
Abhishek-Varma: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjectFifos.cpp
```diff
@@ -0,0 +1,312 @@
...
+        AffineExpr newAffineExpr = affineExpr + offsetToAdd;
+        auto newAffineMap = AffineMap::get(/*dimCount=*/1, /*symbolCount=*/0,
+                                           {newAffineExpr}, context);
+        newL3AsSourceOffsetVal = rewriter.create<affine::AffineApplyOp>(
+            loc, newAffineMap, applyOp.getMapOperands());
+      } else if (auto constantOffset = getConstantIntValue(l3SourceOffsetVal)) {
+        int64_t newOffset = *constantOffset + offsetToAdd;
+        newL3AsSourceOffsetVal = rewriter.create<arith::ConstantOp>(
+            loc, rewriter.getIndexAttr(newOffset));
+      } else {
```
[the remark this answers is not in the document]
Abhishek-Varma: I added that as a TODO in the last revision but have added handling of BlockArguments in the latest push.

Abhishek-Varma:
Here is the current state of Matmul + Elemwise : [e2e IR log with ToM + Splitting LogicalObjectFifos](https://gist.github.com/Abhishek-Varma/ae395f20a676d200d0f4f8b5ace92915).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,332 @@
...
+  4. Offset (2,2) - extract 2x2 memref
+        . . . .
+        . . . .
+            ___
+        . .|. .|
+        . .|. .|
+          -----
+
+  The following utility helps perform the computation of offsets for L3 source.
+*/
```
jtuyls, line 72:
```suggestion
/// This utility helps to perform the computation of offsets for L3 source.
///
/// Example:
/// For L3 -> L2 DmaCpyNd :-
/// From offset (0,0) we are extracting one 4x4 memref.
///                 _______
///               |. . . .|
///               |. . . .|
///               |. . . .|
///               |. . . .|
///               ---------
/// After split we will extract four 2x2 memrefs.
/// So, the corresponding offsets will be :-
/// 1. Offset (0,0) - extract 2x2 memref
///       ___
///     |. .|. .
///     |. .|. .
///     -----
///       . . . .
///       . . . .
/// 2. Offset (0,2) - extract 2x2 memref
///           ___
///       . .|. .|
///       . .|. .|
///         -----
///       . . . .
///       . . . .
/// 3. Offset (2,0) - extract 2x2 memref
///       . . . .
///       . . . .
///       ___
///     |. .|. .
///     |. .|. .
///     -----
/// 4. Offset (2,2) - extract 2x2 memref
///       . . . .
///       . . . .
///           ___
///       . .|. .|
///       . .|. .|
///         -----
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,332 @@
...
+        newL3AsSourceOffset = rewriter.getIndexAttr(newOffset);
+      } else {
+        return failure();
+      }
+    }
+  }
+  return newL3AsSourceOffset;
+}
+
+LogicalResult splitLogicalObjectFifos(
```
jtuyls, line 125: Maybe add a dosctring to explain the implementation steps.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,332 @@
...
+                                         {newAffineExpr}, context);
+      newL3AsSourceOffset =
+          rewriter
+              .create<affine::AffineApplyOp>(ownerOfBlockArg->getLoc(),
+                                             newAffineMap, l3SourceOffsetVal)
+              .getResult();
+    } else {
+      Operation *defOpOfL3SourceOffset = l3SourceOffsetVal.getDefiningOp();
+      Location loc = defOpOfL3SourceOffset->getLoc();
+      OpBuilder::InsertionGuard guard(rewriter);
```
jtuyls, line 101: `OpBuilder::InsertionGuard guard(rewriter);` can be moved to the top of this function.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
jtuyls: Also, doesn't this example belong to the documentation of the split function below?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,332 @@
...
+  for (unsigned dim : splitDimsForL2) {
+    staticL2AsTargetOffsets[dim] = zeroVal;
+    staticL2AsTargetSizes[dim] = oneVal;
+    l2ShapeAsTarget[dim] = 1;
+  }
+  SmallVector<unsigned> nonSplitDimsForL2;
+  for (unsigned dim = 0, n = staticL2AsTargetSizes.size(); dim < n; dim++) {
+    if (splitDimsSetForL2.contains(dim)) continue;
+    nonSplitDimsForL2.push_back(dim);
+  }
```
jtuyls, line 209: After `verifySplitDimensionConstraint`, `nonSplitDimsForL2` is just the following I think:
```
SmallVector<size_t> nonSplitDimsForL2(staticL2AsTargetSizes.size() - splitDimsForL2.size());
std::iota(nonSplitDimsForL2.begin(), nonSplitDimsForL2.end(), splitDimsForL2.size())
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,332 @@
...
+        << "expected alloc op as the defining op of source "
+           "logicalobjectfifo.from_memref";
+    return failure();
+  }
+  // We will now capture those dimensions where L2 memory was split. The way we
+  // do this is by checking all L2->L1 DmaOps' source offset and marking those
+  // dimensions which are not equal to at least one of the source offsets.
+  DenseSet<unsigned> splitDimsSetForL2;
+  SmallVector<unsigned> splitDimsForL2;
+  for (unsigned i = 1, n = l2ToL1DmaOps.size(); i < n; i++) {
```
jtuyls, line 147: We should check that all `l2ToL1DmaOps` have the same splitting dimensions.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,332 @@
...
+  } else {
+    auto l3SourceOffsetVal = cast<Value>(oldL3Offset);
+    if (auto blockArg = dyn_cast<BlockArgument>(l3SourceOffsetVal)) {
+      Operation *ownerOfBlockArg = blockArg.getOwner()->getParentOp();
+      OpBuilder::InsertionGuard guard(rewriter);
+      rewriter.setInsertionPointToStart(blockArg.getOwner());
+      AffineExpr affineExpr = rewriter.getAffineDimExpr(0);
+      AffineExpr newAffineExpr = affineExpr + offsetToAdd;
+      auto newAffineMap = AffineMap::get(/*dimCount=*/1, /*symbolCount=*/0,
+                                         {newAffineExpr}, context);
```
jtuyls, line 92: You could create a utility (lambda) for this as it's duplicated below in the else condition.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,332 @@
...
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/IR/Operation.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/Iterators.h"
+
+namespace mlir::iree_compiler::AMDAIE {
+
+/// Utility to verify that the split dimensions for L2 are contiguous.
+static LogicalResult verifySplitDimensionConstraint(
```
jtuyls, line 20: Suggestion: checkIsRangeFromZero

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,332 @@
...
+  if (auto l3SourceOffsetAttr = dyn_cast<Attribute>(oldL3Offset)) {
+    int64_t l3SourceOffsetIntVal =
+        cast<IntegerAttr>(l3SourceOffsetAttr).getInt();
+    int64_t newOffset = l3SourceOffsetIntVal + offsetToAdd;
+    newL3AsSourceOffset = rewriter.getIndexAttr(newOffset);
+  } else {
+    auto l3SourceOffsetVal = cast<Value>(oldL3Offset);
+    if (auto blockArg = dyn_cast<BlockArgument>(l3SourceOffsetVal)) {
+      Operation *ownerOfBlockArg = blockArg.getOwner()->getParentOp();
+      OpBuilder::InsertionGuard guard(rewriter);
```
jtuyls, line 87: Move insertion guard to top of function.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
Abhishek-Varma: I don't think that is required.

Eg:  Assume the following bizarre case occurs :-
```
     A. L2 -> L1 : [0, 0, 0, 0]
     B. L2 -> L1 : [0, 0, 0, 0]
     C. L2 -> L1 : [0, 1, 0, 0]
     D. L2 -> L1 : [1, 0, 0, 0]
```
In this part of the code I'm performing a `any_of`-like operation i.e "is any of the value in the same dimension different".
So, in the above case the splitting dimensions for ALL L2->L1 ops would be dimension 0 and dimension 1 - I don't think a check is required.

And then in another utility I'm simply checking if they're contiguous.

Eg:  Let's take another bizarre case :-
```
     A. L2 -> L1 : [0, 0, 0, 1]
     B. L2 -> L1 : [0, 0, 0, 0]
     C. L2 -> L1 : [1, 1, 0, 0]
     D. L2 -> L1 : [0, 0, 0, 0]
```
Here dimension 0, 1 and 3 would make the splitting dimension.

But later in the verification of the splitting dimensions being contiguous it'd bail out successfully.

Could you explain why there is a need for the check (and how exactly) ?

jtuyls: We should be prudent and bail out in these bizarre cases.

What about:

```
     A. L2 -> L1 : [0, 0, 1, 0]
     B. L2 -> L1 : [0, 0, 0, 0]
     C. L2 -> L1 : [1, 1, 0, 0]
     D. L2 -> L1 : [0, 0, 0, 0]
```

Here dimension 0, 1 and 2 would make the splitting dimension and dimensions are contiguous, but it's probably not going to go well.

> Could you explain why there is a need for the check (and how exactly) ?

We should be prudent and limit splitting buffers to the exact cases where we know it's going to work, i.e. all l2tol1dmas have the same splitting dimensions, contiguous range from zero, etc.

You can easily check it with a utility function that computes the dimensions for a single l2to1lDmaOp and then check that all of them have the same dimenions.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,332 @@
...
+      staticL3AsSourceOffsets[nonSplitdim] = *newOffset;
+    }
+    // Create new L3 -> L2 Dma Op.
+    rewriter.setInsertionPoint(l3ToL2DmaOp);
+    rewriter.create<AMDAIE::DmaCpyNdOp>(
+        l3ToL2DmaOp.getLoc(), source, llvm::ArrayRef(staticL2AsTargetOffsets),
+        llvm::ArrayRef(staticL2AsTargetSizes),
+        llvm::ArrayRef(staticL2AsTargetStrides), l3ToL2DmaOp.getSource(),
+        llvm::ArrayRef(staticL3AsSourceOffsets),
+        llvm::ArrayRef(staticL2AsTargetSizes),
```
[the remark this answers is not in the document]
jtuyls: For now, we should bail out in these cases, but still source sizes should be computed on themselves, no derived from target sizes.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
Abhishek-Varma: This is just about L3 -> L2 because for L2->L1 we do not modify non-split dimension offsets.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,368 @@
...
+      l2ToL1DmaOps[0].getSourceMixedOffsets();
+  LogicalObjectFifoFromMemrefOp sourceObjectFifo =
+      l2ToL1DmaOps[0].getSourceObjectFifo();
+  auto sourceAllocOp =
+      sourceObjectFifo.getMemref().getDefiningOp<memref::AllocOp>();
+  if (!sourceAllocOp) {
+    sourceObjectFifo->emitOpError()
+        << "expected alloc op as the defining op of source "
+           "logicalobjectfifo.from_memref";
+    return failure();
```
jtuyls, line 150:
```suggestion
    return sourceObjectFifo->emitOpError()
        << "expected alloc op as the defining op of source "
           "logicalobjectfifo.from_memref";
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+// Given a vector of L2->L1 Dma ops, split them via the following process :-
+// 1. Infer splitting dimension of L2 as a function of all L2->L1 Dma ops.
+// 2. Fetch and ensure a unique L3->L2 Dma op.
+// 3. For the split dimension inferred set offset = 0 and size as 1 for L2 and
+// L3.
+// 4. Now traverse each L2->L1 Dma op and perform the following :-
+//    a) Create a new L2 AllocOp based on the updated size (step 3 above) and
+//    create
+//       a logicalobjectfifo using the same.
```
jtuyls, line 132:
```suggestion
//    create a logicalobjectfifo using the same.
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,368 @@
...
+      }
+    }
+  }
+  std::sort(splitDimsForL2.begin(), splitDimsForL2.end());
+
+  if (failed(checkIsRangeFromZero(splitDimsForL2))) {
+    l2ToL1DmaOps[0]->emitRemark()
+        << "cannot split L2 logicalobjectfifo because of non-contiguous split "
+           "dimensions inferred";
+    return failure();
```
jtuyls, line 179: Why use `emitRemark`? Imo, this should be an error:

```suggestion
    return l2ToL1DmaOps[0]->emitOpError()
        << "cannot split L2 logicalobjectfifo because of non-contiguous split "
           "dimensions inferred";
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,368 @@
...
+  l3ToL2DmaOp = l3ToL2DmaOps[0];
+  if ((l3ToL2DmaOp.getTargetMixedOffsets().size() !=
+       l3ToL2DmaOp.getSourceMixedOffsets().size()) ||
+      (l3ToL2DmaOp.getTargetMixedSizes().size() !=
+       l3ToL2DmaOp.getSourceMixedSizes().size()) ||
+      (l3ToL2DmaOp.getTargetMixedStrides().size() !=
+       l3ToL2DmaOp.getSourceMixedStrides().size())) {
+    l3ToL2DmaOp->emitRemark() << "dimensionality of source and target's "
+                                 "offset/size/stride should be same";
+    return failure();
```
jtuyls, line 210:
```suggestion
    return l3ToL2DmaOp->emitOpError() << "dimensionality of source and target's "
                                 "offset/size/stride should be same";
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,368 @@
...
+  for (Operation *objFifoUserOp : sourceObjectFifo->getUsers()) {
+    if (auto dmaOp = dyn_cast<AMDAIE::DmaCpyNdOp>(objFifoUserOp);
+        dmaOp.getTargetObjectFifo() == sourceObjectFifo) {
+      l3ToL2DmaOps.push_back(dmaOp);
+      break;
+    }
+  }
+  if (l3ToL2DmaOps.size() == 0) {
+    sourceObjectFifo->emitRemark() << "no corresponding L3->L2 dma op found";
+    return failure();
```
jtuyls, line 195:
```suggestion
    return sourceObjectFifo->emitOpError() << "no corresponding L3->L2 dma op found";
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,368 @@
...
+  }
+  if (l3ToL2DmaOps.size() > 1) {
+    sourceObjectFifo->emitRemark() << "found more than one L3->L2 dma ops";
+    return failure();
```
jtuyls, line 199:
```suggestion
    return sourceObjectFifo->emitOpError() << "found more than one L3->L2 dma ops";
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,368 @@
...
+// 4. Now traverse each L2->L1 Dma op and perform the following :-
+//    a) Create a new L2 AllocOp based on the updated size (step 3 above) and
+//    create
+//       a logicalobjectfifo using the same.
+//    b) Split L3->L2 Dma op.
+//    c) SPlit L2->L1 Dma op.
+LogicalResult splitLogicalObjectFifos(
+    IRRewriter &rewriter, SmallVector<AMDAIE::DmaCpyNdOp> &l2ToL1DmaOps,
+    MLIRContext *context) {
+  if (l2ToL1DmaOps.size() == 0) return success();
```
jtuyls, line 138: Add insertion guard.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,368 @@
...
+        l3ToL2DmaOp.getSourceMixedOffsets();
+    for (auto &&[splitDim, nonSplitdim] :
+         llvm::zip_equal(splitDimsForL2, nonSplitDimsForL2)) {
+      std::optional<int64_t> constantOffset =
+          getConstantIntValue(staticL2AsSourceOffsets[splitDim]);
+      if (!constantOffset) {
+        l2ToL1DmaOp->emitRemark()
+            << "found a non-constant value for source offset at dim "
+            << splitDim;
+        return failure();
```
jtuyls, line 294: `emitOpError`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,368 @@
...
+      }
+      std::optional<int64_t> constantSize =
+          getConstantIntValue(staticL2AsTargetSizes[nonSplitdim]);
+      if (!constantSize) {
+        l3ToL2DmaOp->emitRemark()
+            << "found a non-constant value for target size at dim "
+            << nonSplitdim;
+        return failure();
```
jtuyls, line 302: `emitOpError`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,368 @@
...
+      int64_t offsetToAdd = constantOffset.value() * constantSize.value();
+      FailureOr<OpFoldResult> newOffset = updateL3SourceOffset(
+          rewriter, staticL3AsSourceOffsets[nonSplitdim], offsetToAdd, context);
+      if (failed(newOffset)) {
+        // TODO: Ideally we should be able to handle even +, -, *, /, etc.
+        //       But handle this later (if at all!) as such cases might not
+        //       arise.
+        l3ToL2DmaOp->emitRemark()
+            << "Unhandled expression for source offset at dim " << nonSplitdim;
+        return failure();
```
jtuyls, line 313: `emitOpError`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,368 @@
...
+    // Remove old dealloc.
+    memref::DeallocOp oldDeallocOp;
+    for (Operation *userOp : sourceAllocOp->getUsers()) {
+      if (auto deallocUser = dyn_cast<memref::DeallocOp>(userOp)) {
+        oldDeallocOp = deallocUser;
+      }
+    }
+    if (oldDeallocOp) {
+      toBeErased.insert(oldDeallocOp);
+    }
```
jtuyls, line 357:
```suggestion
    if (oldDeallocOp) toBeErased.insert(oldDeallocOp);
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -0,0 +1,368 @@
...
+        llvm::ArrayRef(staticL2AsSourceStrides));
+    rewriter.replaceOp(l2ToL1DmaOp, newL2ToL1DmaOp);
+
+    // Remove old dealloc.
+    memref::DeallocOp oldDeallocOp;
+    for (Operation *userOp : sourceAllocOp->getUsers()) {
+      if (auto deallocUser = dyn_cast<memref::DeallocOp>(userOp)) {
+        oldDeallocOp = deallocUser;
+      }
+    }
```
jtuyls, line 354:
```suggestion
      if (auto deallocUser = dyn_cast<memref::DeallocOp>(userOp))
        oldDeallocOp = deallocUser;
    }
```

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifosForConnectionReuse.cpp
```diff
@@ -0,0 +1,80 @@
...
+  IRRewriter rewriter(context);
+
+  SmallVector<AMDAIE::DmaCpyNdOp> l2ToL1DmaOps =
+      fetchDmaCpyNdOpsToSplit(moduleOp);
+
+  SplittingLogicalObjectFifoData splittingLogicalObjectFifoData;
+  splittingLogicalObjectFifoData.l2ToL1DmaOps = l2ToL1DmaOps;
+  if (failed(checkWhetherSplitIsPossible(splittingLogicalObjectFifoData))) {
+    LLVM_DEBUG(llvm::dbgs()
+               << "Cannot perform splitting of logicalobjectfifos");
```
jtuyls, line 62:
```suggestion
               << "Cannot perform splitting of logical objectFifos");
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifosForConnectionReuse.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+namespace {
+
+/// Utility to help fetch those input DmaCpyNd Ops which needs to be split.
+static SmallVector<AMDAIE::DmaCpyNdOp> fetchDmaCpyNdOpsToSplit(
+    ModuleOp moduleOp) {
+  SmallVector<AMDAIE::DmaCpyNdOp> l2ToL1DmaOps;
+  // We are currently walking through CoreOps gathering 3rd Input DmaOp (if
+  // applicable) from them.
+  // TODO: We will generalize this later.
```
jtuyls, line 25:
```suggestion
  // TODO(avarma): We will generalize this later.
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.h
```diff
@@ -0,0 +1,39 @@
...
+/// of logicalobjectfifos.
+struct SplittingLogicalObjectFifoData {
+  SmallVector<AMDAIE::DmaCpyNdOp> l2ToL1DmaOps;
+  SmallVector<size_t> splitDimsForL2;
+  SmallVector<size_t> nonSplitDimsForL2;
+  AMDAIE::DmaCpyNdOp l3ToL2DmaOp;
+};
+
+/// Utility to check whether splitting of logicalobjectfifos can be performed.
+/// If yes, it populates the struct `SplittingLogicalObjectFifoData` with the
```
jtuyls, line 24:
```suggestion
/// If possible, it populates the struct `SplittingLogicalObjectFifoData` with the
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifosForConnectionReuse.cpp
jtuyls, line 62: You could avoid exposing the `SplittingLogicalObjectFifoData` struct my moving this into the `splitLogicalObjectFifos` function and return `success()` if the check failed.

jtuyls (approved):

PR #660, opened 2024-08-09
Bump iree

yzhang93:
Following runtime changes in IREE
- Attaching pipeline layout to hal.interface.binding.subspan. https://github.com/iree-org/iree/pull/18098
- Adding flag placeholders to semaphores/events. https://github.com/iree-org/iree/pull/18122


nirvedhmeshram (approved):
Thanks, LGTM!

PR #664, opened 2024-08-11
Scripts to build aietools distros

makslevental:
Barebones but it works...

### DIFF
--- a/build_tools/build_aietools_distro/compile_test.sh
+++ b/build_tools/build_aietools_distro/compile_test.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+# Copyright 2024 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# This script tests whether the copy was successful by attempting to compile the adjacent main.cpp
+# using the copied Vitis distro. It also serves the dual purpose of illustrating explicitly how to use
+# xchesscc.
+
+set -euo pipefail
+
+AIETOOLS_DIR=${AIETOOLS_DIR:-Vitis/2024.1/aietools}
+AIEARCH=aie_ml
+
+export RDI_DATADIR=$AIETOOLS_DIR/data
+export me_DIR=${RDI_DATADIR}/$AIEARCH/lib
+export TCLLIBPATH=$AIETOOLS_DIR/tps/lnx64/target_$AIEARCH/chessdir/tcltk8.6/LNa64/lib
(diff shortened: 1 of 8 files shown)


build_tools/build_aietools_distro/compile_test.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
jtuyls, line 1: Maybe explain what this file is for as it's not used? Or can we put this file in CI as well to avoid it going stale?

build_tools/build_aietools_distro/copy.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+set -euo pipefail
```
jtuyls, line 3: Maybe explain what this file is for as it's not used? Or can we put this file in CI as well to avoid it going stale?

build_tools/build_aietools_distro/repair.py
```diff
@@ -0,0 +1,77 @@
...
+
+    return new_soname, dest_path
+
+
+wheel_policy = WheelPolicies()
+out_wheel = repair_wheel(
+    wheel_policy,
+    str(Path.cwd() / "dist" / "chess-0.0.0-py3-none-any.whl"),
+    abis=["manylinux_2_35_x86_64"],
+    lib_sdir="Vitis/2024.1/aietools/lib/lnx64.o",
```
jtuyls, line 74: We could use an environment variable to get the Vitis path? i.e. `SRC_AIETOOLS_DIR`?

jtuyls (changes requested):
Could you add headers to all the files?

Also, it might be useful to try putting this in CI to avoid files becoming stale?

makslevental:
> Also, it might be useful to try putting this in CI to avoid files becoming stale?

How? You want to dogfood? I.e., create the distro each time before use? Even if that's the case, from where? The only Vitis we have in CI is the copied version.

jtuyls:
>  The only Vitis we have in CI is the copied version.

Yeah, right.. Nvm.

build_tools/build_aietools_distro/repair.py
makslevental: That's not what this thing does - it's the dest dir in the wheel/zip where the libs are deposited.

build_tools/build_aietools_distro/compile_test.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# This script tests whether the copy was successful by attempting to compile the adjacent main.cpp
+# using the copied Vitis distro. IT also serves the dual purpose of illustrating explicitly how to use
```
jtuyls, line 4:
```suggestion
# using the copied Vitis distro. It also serves the dual purpose of illustrating explicitly how to use
```

jtuyls (commented):
Could you still add the headers?

jtuyls (approved):

PR #665, opened 2024-08-12
[Localize ObjectFifo] Don't crash when scf.forall has no mapping

newling:
While we probably don't expect an scf.forall without _either_ a thread or block mapping, it's better if this pass can handle the case.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELocalizeLogicalObjectFifo.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELocalizeLogicalObjectFifo.cpp
@@ -29,7 +29,7 @@ namespace mlir::iree_compiler::AMDAIE {
 namespace {
 
 template <class... T>
-scf::ForallOp getThreadMappedForallAncestor(Operation *op) {
+scf::ForallOp getMappedForallAncestor(Operation *op) {
   auto forall = op->getParentOfType<scf::ForallOp>();
   while (forall) {
     bool isMapped = [&forall]() {
@@ -49,12 +49,12 @@ scf::ForallOp getThreadMappedForallAncestor(Operation *op) {
 }
 
 scf::ForallOp getThreadMappedForallAncestor(Operation *op) {
-  return getThreadMappedForallAncestor<mlir::gpu::GPUThreadMappingAttr>(op);
+  return getMappedForallAncestor<mlir::gpu::GPUThreadMappingAttr>(op);
 }
 
 scf::ForallOp getThreadOrBlockMappedForallAncestor(Operation *op) {
-  return getThreadMappedForallAncestor<mlir::gpu::GPUThreadMappingAttr,
-                                       mlir::gpu::GPUBlockMappingAttr>(op);
+  return getMappedForallAncestor<mlir::gpu::GPUThreadMappingAttr,
+                                 mlir::gpu::GPUBlockMappingAttr>(op);
 }
 
 class AMDAIELocalizeLogicalObjectfifoPass
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELocalizeLogicalObjectFifo.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELocalizeLogicalObjectFifo.cpp
@@ -19,7 +19,7 @@
 #include "iree-amd-aie/IR/AMDAIEDialect.h"
 #include "iree-amd-aie/IR/AMDAIEOps.h"
 #include "iree-amd-aie/Transforms/Passes.h"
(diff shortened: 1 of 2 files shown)


newling:
The original code:
```
 if (isa<mlir::gpu::GPUThreadMappingAttr>(
          *forallOp.getMapping()->getValue().begin())) {
```

doesn't check if if `forallOp` even has a mapping, an optional attribute. So there is a segfault on getValue(). 

This PR is exactly equivalent to just adding a `has_value()` check in. The rest of the change is replacing recursion.

jtuyls:
> The original code:
> 
> ```
>  if (isa<mlir::gpu::GPUThreadMappingAttr>(
>           *forallOp.getMapping()->getValue().begin())) {
> ```
> 
> doesn't check if `forallOp` has a mapping, an optional attribute. So there is a segfault on getValue().
> 
> This PR is exactly equivalent to just adding a `has_value()` check in. The rest of the change is replacing recursion.

But it won't throw an error?

newling:
> > The original code:
> > ```
> >  if (isa<mlir::gpu::GPUThreadMappingAttr>(
> >           *forallOp.getMapping()->getValue().begin())) {
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > doesn't check if `forallOp` has a mapping, an optional attribute. So there is a segfault on getValue().
> > This PR is exactly equivalent to just adding a `has_value()` check in. The rest of the change is replacing recursion.
> 
> But it won't throw an error?

Not sure I understand question, are you suggesting I make a change to the logic? All this PR does is what was done in 

https://github.com/nod-ai/iree-amd-aie/commit/883ee07732ece0b3cccdcd871c1b7d4c3eeebed2 

w.r.t. mapping on scf.forall.

jtuyls:
> > > The original code:
> > > ```
> > >  if (isa<mlir::gpu::GPUThreadMappingAttr>(
> > >           *forallOp.getMapping()->getValue().begin())) {
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > doesn't check if `forallOp` has a mapping, an optional attribute. So there is a segfault on getValue().
> > > This PR is exactly equivalent to just adding a `has_value()` check in. The rest of the change is replacing recursion.
> > 
> > 
> > But it won't throw an error?
> 
> Not sure I understand question, are you suggesting I make a change to the logic? All this PR does is what was done in
> 
> [883ee07](https://github.com/nod-ai/iree-amd-aie/commit/883ee07732ece0b3cccdcd871c1b7d4c3eeebed2)
> 
> w.r.t. mapping on scf.forall.

Yeah, the PR says 'Don't crash when scf.forall has no mapping '. I agree there shouldn't be a segfault, but now the pass will just continue and then at some point there will be an error/crash again, at a point where it might not be clear that this is the root issue? I think the correct behaviour would be to throw an error in this pass if no scf.forall ops with thread mapping were found.

newling:
> > > > The original code:
> > > > ```
> > > >  if (isa<mlir::gpu::GPUThreadMappingAttr>(
> > > >           *forallOp.getMapping()->getValue().begin())) {
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > doesn't check if `forallOp` has a mapping, an optional attribute. So there is a segfault on getValue().
> > > > This PR is exactly equivalent to just adding a `has_value()` check in. The rest of the change is replacing recursion.
> > > 
> > > 
> > > But it won't throw an error?
> > 
> > 
> > Not sure I understand question, are you suggesting I make a change to the logic? All this PR does is what was done in
> > [883ee07](https://github.com/nod-ai/iree-amd-aie/commit/883ee07732ece0b3cccdcd871c1b7d4c3eeebed2)
> > w.r.t. mapping on scf.forall.
> 
> Yeah, the PR says 'Don't crash when scf.forall has no mapping '. I agree there shouldn't be a segfault, but now the pass will just continue and then at some point there will be an error/crash again, at a point where it might not be clear that this is the root issue? I think the correct behaviour would be to throw an error in this pass if no scf.forall ops with thread mapping were found.

I agree with your overall design idea, but this PR doesn't change the behavior. It is exactly equivalent to https://github.com/nod-ai/iree-amd-aie/pull/669/files

jtuyls:
>  This isn't about the case where there is no scf.forall with a thread mapping, it's about the case where there is an scf.forall with no mapping (thread or block).

Ah, I see, like a case with some `scf.forall` ops with mapping and others without?

jtuyls (approved):

newling:
> > This isn't about the case where there is no scf.forall with a thread mapping, it's about the case where there is an scf.forall with no mapping (thread or block).
> 
> Ah, I see, like a case with some `scf.forall` ops with mapping and others without?

Yes, that's the edge case being fixed

Issue #448, opened 2024-06-19
Privatize local memory for tensors


newling:
The pack-peel pipeline, matmul (m=n=1024, k=512) followed by addition of a bias (1-d vector with 1024 values) results in the final allocations (I've renamed the SSA values for clarity). 

```
%bias_local = memref.alloc() : memref<1x16x4xf32, 2 : i32>
%bias_shared = memref.alloc() : memref<2x64xf32, 1 : i32>
%B_local = memref.alloc() : memref<1x1x16x8x8x4xbf16, 2 : i32>
%A_local = memref.alloc() : memref<1x1x8x16x4x8xbf16, 2 : i32>
%B_shared = memref.alloc() : memref<1x2x64x64xbf16, 1 : i32>
%A_shared = memref.alloc() : memref<2x1x64x64xbf16, 1 : i32>
%C_local = memref.alloc() : memref<2x2x16x16x4x4xf32, 2 : i32>
%C_shared = memref.alloc() : memref<2x2x64x64xf32, 1 : i32> 
```

The above is for a design using a 2x2 array of AIE cores. The IR contains a loop over the 2x2 cores, indexing into arrays as follows

```
 scf.forall (%arg2, %arg3) in (2, 2) {

// Copy from slice of shared-memory A, to local-memory for A 
%subview_14 = memref.subview %A_shared[%arg2, 0, 0, 0] [1, 1, 64, 64] [1, 1, 1, 1]...
iree_linalg_ext.pack %subview_14 outer_dims_perm = [0, 1, 3, 2] inner_dims_pos = [2, 3] inner_tiles = [4, 8] into %A_local ...

%subview_15 = memref.subview %B_shared[0, %arg3, 0, 0] [1, 1, 64, 64] [1, 1, 1, 1]...
iree_linalg_ext.pack %subview_15 outer_dims_perm = [0, 1, 3, 2] inner_dims_pos = [2, 3] inner_tiles = [8, 4] into %B_local ...

%subview_16 = memref.subview %C_local[%arg2, %arg3, 0, 0, 0, 0] [1, 1, 16, 16, 4, 4] [1, 1, 1, 1, 1, 1]...
 
 ...
 }
```

For A and B, a view into shared memory is copied to the entire local buffer for A and B. For C, a slice of the local buffer is taken. 

I find this very confusing, and think it would be much better if C was already 'privatized' per core, so that instead of 

```
[...]

PR #667, opened 2024-08-12
[Distribute Cores] Do not assume all subviews are 'distributing'

newling:
One role of the distribute-cores-and-objectfifo pass is to private L2 memory allocations. 

That is, to replace IR like

```
  %alloc_3 = memref.alloc() : memref<2x2x8x8x4x4xi32, 2 : i32>
  ...
  %subview = memref.subview %alloc_3[%arg2, %arg3, 0, 0, 0, 0] [1, 1, 8, 8, 4, 4] 
                      [1, 1, 1, 1, 1, 1] : memref<2x2x8x8x4x4xi32, 2 : i32> to 
                      memref<1x1x8x8x4x4xi32, strided<[2048, 1024, 128, 16, 4, 1], offset: ?>, 2 : i32>
```

with 

```
  %[[ALLOC_3:.*]] = memref.alloc() : memref<1x1x8x8x4x4xi32, 2 : i32>
```

Before this PR, the pass assumed **all** memrefs on allocs where of this form. But this is not always the case. This PR tries to determine if the memref is indeed of this form, otherwise it does nothing. 

This is still not completely robust, we either do a proper privatization pass or get the root cause of the needing this in the first place: https://github.com/nod-ai/iree-amd-aie/issues/448

This PR also does a refactor of the code.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/distribute_cores_and_objectfifos.mlir
```diff
@@ -679,6 +679,31 @@ func.func @l1_temporary_buffer_for_matmul_elem() {
     return
 }
 
+// -----
+
+// We check that the case where an L2 distributed is handled correctly.
```
Abhishek-Varma, line 684: `where an L1 is not distributed` ?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -543,68 +607,77 @@ LogicalResult insertLogicalObjectFifoAccess(ModuleOp moduleOp) {
...
-        // inserted within the vectorized scf.for ops. We therefore would want
-        // to traverse to the outermost scf.for op in that case. Currently we
-        // bubble up this traversal till that operation whose parent is not a
-        // scf.for op. TODO: Generalize this later.
-        Operation *opToInsertRewriterPoint = op;
-        while (isa<scf::ForOp>(opToInsertRewriterPoint->getParentOp())) {
-          opToInsertRewriterPoint = opToInsertRewriterPoint->getParentOp();
+      bool hasAllocOperand = [op]() {
+        for (auto operand : op->getOperands()) {
+          auto definingOp = operand.getDefiningOp();
```
Abhishek-Varma, line 623: I'd prefer the following :-
```suggestion
        for (Value operand : op->getOperands()) {
          Operation* definingOp = operand.getDefiningOp();
```

Abhishek-Varma (commented):
I really liked how you cleaned up `distributeLocalMemory`.
Regarding the distributable concept itself - I'd like @jtuyls to take another look. Else the structure of the `distributeLocalMemory` looks good to me.

The rest of the code looks like clang-formats - thank you for that!

A few nit comments from my end.

jtuyls (changes requested):
> This is still not completely robust, we either do a proper privatization pass or get the root cause of the needing this in the first place: https://github.com/nod-ai/iree-amd-aie/issues/448

+1 for we need to make this more robust. This shared buffers on L1 complicate the logic quite a bit.

The root cause would the `scf.forall` ops operating in parallel to compute a different slice of the output tensor.  It doesn't seem that easy to handle that earlier, before we unroll.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -56,182 +61,238 @@ struct LocationMapInfo {
...
+                  [&](memref::ExtractStridedMetadataOp
+                          extractStridedMetadataOp) {
+                    rewriter.setInsertionPoint(extractStridedMetadataOp);
+                    auto newextractStridedMetadataOp =
+                        rewriter.create<memref::ExtractStridedMetadataOp>(
+                            extractStridedMetadataOp.getLoc(), newAlloc);
+                    rewriter.replaceAllUsesWith(
+                        extractStridedMetadataOp.getResults(),
+                        newextractStridedMetadataOp.getResults());
+                    toBeErased.push_back(extractStridedMetadataOp);
```
jtuyls, line 230: It seems like not much is happening here and this could be written as a more general handler for different (future) ops? (Just a thought, not necessarily for this PR)

newling: Ok. I'll refactor in a later PR. This part of the change was moving an if-else statement into an llvm switch statement.

newling:
> > This is still not completely robust, we either do a proper privatization pass or get the root cause of the needing this in the first place: #448
> 
> +1 for we need to make this more robust. This shared buffers on L1 complicate the logic quite a bit.
> 
> The root cause would the `scf.forall` ops operating in parallel to compute a different slice of the output tensor. It doesn't seem that easy to handle that earlier, before we unroll.

I haven't explored where would be easy/best to run this as a standalone pass, but it would ideally be it's own pass. Future work!

jtuyls (approved):
LGTM.

PR #670, opened 2024-08-12
[WIP] Towards stateful transform shake-up

newling:
Very minor changes. The test removed is basically a duplicate of the AIE2 equivalent test, doesn't add any coverage. 

Background: I'm grokking this pass to work out how to get around 

```
error: 'amdaie.circular_dma_cpy_nd' op access pattern of strided operation overlaps with next one, which is not supported for now
```

What would be very useful would be to reduce complexity so that we only have we need. For example, look at the arguments in 

```
def AIE_ObjectFifoCreateOp: AIE_Op<"objectfifo", [Symbol]> {
  let summary = "Create a circular buffer or channel between two tiles";
  let arguments = (
    ins SymbolNameAttr:$sym_name,
        Index:$producerTile,
        Variadic<Index>:$consumerTiles,
        AnyAttrOf<[ConfinedAttr<I32Attr, [IntMinValue<0>]>, ArrayAttr]>:$elemNumber,
        TypeAttrOf<AIE_ObjectFifoType>:$elemType,
        BDDimLayoutArrayAttr:$dimensionsToStream,
        BDDimLayoutArrayArrayAttr:$dimensionsFromStreamPerConsumer,
        DefaultValuedAttr<BoolAttr, "false">:$via_DMA,
        DefaultValuedAttr<BoolAttr, "false">:$plio
  );
```

how many of the arguments can we simplify/drop (@jtuyls) ? For example, I imagine `plio` isn't something we want.


compiler/plugins/target/AMD-AIE/aie/AIEOps.td
```diff
@@ -551,6 +550,20 @@ def AIE_ObjectFifoCreateOp: AIE_Op<"objectfifo", [Symbol]> {
...
+      auto elemNumber = getElemNumber();
+      auto intAttr = [&]() -> mlir::IntegerAttr {
+        if (auto arrAttr = llvm::dyn_cast<mlir::ArrayAttr>(elemNumber))
+          return llvm::dyn_cast<mlir::IntegerAttr>(arrAttr[index]);
+        else
+          return llvm::dyn_cast<mlir::IntegerAttr>(elemNumber);
+      }();
+      assert(intAttr && "expected integer attribute");
+      return intAttr.getInt();
+    }
```
makslevental, line 564: I deleted this thing and replaced it with `objFifoSize` because 

1) putting cpp into tablegen effectively hides it (eg completely confuses `clangd`); 
2) member methods also hide the side effects of the function;
3) single use functions like this should be kept as close to their use as possible.

This is roughly inline with google's and various other style guides that recommend private (anon namespace) by default.

makslevental (changes requested):

compiler/plugins/target/AMD-AIE/aie/AMDAIEObjectFifoStatefulTransform.cpp
```diff
@@ -176,9 +171,11 @@ class DMAChannelAnalysis {
       auto tile = memOp.getTile();
       for (auto &bl : r.getBlocks()) {
         for (auto op : bl.getOps<DMAStartOp>()) {
-          static_cast<DMAChannelDir>(op.getChannelDir()) == DMAChannelDir::MM2S
-              ? getProducerDMAChannel(tile)
-              : getConsumerDMAChannel(tile);
+          if (op.getChannelDir() == DMAChannelDir::MM2S) {
+            ++producerChannelsPerTile[tile];
```
makslevental, line 175: if you're going to do this, you might as well remove `getProducerDMAChannel` (and `getConsumerDMAChannel`).

compiler/plugins/target/AMD-AIE/aie/AMDAIEObjectFifoStatefulTransform.cpp
```diff
@@ -138,26 +128,31 @@ std::optional<Value> getOptionalSharedTile(ObjectFifoLinkOp op) {
...
-    AMDAIEDeviceModel deviceModel =
-        getDeviceModel(static_cast<AMDAIEDevice>(device.getDevice()));
-    for (int i = 0;
-         i < deviceModel.getNumLocks(tileOp.getCol(), tileOp.getRow()); i++) {
+  int getLockID(TileOp tileOp) {
+    auto maxLocks = deviceModel.getNumLocks(tileOp.getCol(), tileOp.getRow());
+    for (int i = 0; i < maxLocks; i++) {
       std::pair<Value, int> lockId = {tileOp, i};
-      if (int usageCnt = locksPerTile[lockId]; usageCnt == 0) {
-        locksPerTile[lockId] = 1;
```
makslevental, line 158: jeez wow :facepalm:

makslevental:
> how many of the arguments can we simplify/drop ? For example, I imagine plio isn't something we want.

certainly those last two can be dropped. `AIE_ObjectFifoType` (and thus `$elemType`) could entirely be dropped and just replaced with `memref` (trust me I've tried it...) but air uses it.... :shrug:

newling:
> > how many of the arguments can we simplify/drop ? For example, I imagine plio isn't something we want.
> 
> certainly those last two can be dropped. `AIE_ObjectFifoType` (and thus `$elemType`) could entirely be dropped and just replaced with `memref` (trust me I've tried it...) but air uses it.... 🤷

Ok I'll probably remove last 2 today at some point (in this PR or another if you accept this in which case I'll just land it and start afresh). Not sure what air problem is w.r.t. any of this code, it doesn't use any object fifo related stuff does it?

compiler/plugins/target/AMD-AIE/aie/AMDAIEObjectFifoStatefulTransform.cpp
```diff
@@ -79,20 +79,23 @@ std::vector<ObjectFifoCreateOp> getOutputObjectFifos(ObjectFifoLinkOp &op) {
...
 }
 
 int objFifoSize(ObjectFifoCreateOp op, int index = 0) {
-  if (llvm::isa<mlir::ArrayAttr>(op.getElemNumber()))
-    return llvm::dyn_cast<mlir::IntegerAttr>(
-               llvm::dyn_cast<mlir::ArrayAttr>(op.getElemNumber())[index])
-        .getInt();
-  else
-    return llvm::dyn_cast<mlir::IntegerAttr>(op.getElemNumber()).getInt();
+  auto elemNumber = op.getElemNumber();
```
jtuyls, line 90:
```suggestion
  Attribute elemNumber = op.getElemNumber();
```

newling:
> > > how many of the arguments can we simplify/drop ? For example, I imagine plio isn't something we want.
> > 
> > 
> > certainly those last two can be dropped. `AIE_ObjectFifoType` (and thus `$elemType`) could entirely be dropped and just replaced with `memref` (trust me I've tried it...) but air uses it.... 🤷
> 
> Ok I'll probably remove last 2 today at some point (in this PR or another if you accept this in which case I'll just land it and start afresh). Not sure what air problem is w.r.t. any of this code, it doesn't use any object fifo related stuff does it?

https://github.com/Xilinx/mlir-air/blob/b5a5116b21f70812e941fc2184be28980bb8eab1/mlir/lib/Conversion/AIRToAIEPass.cpp#L1102

compiler/plugins/target/AMD-AIE/aie/AMDAIEObjectFifoStatefulTransform.cpp
```diff
@@ -138,26 +141,31 @@ std::optional<Value> getOptionalSharedTile(ObjectFifoLinkOp op) {
...
-  DenseMap<std::pair<Value, int>, int> locksPerTile;
+ private:
+  DenseSet<std::pair<Value, int>> locks;
+  AMDAIEDeviceModel deviceModel;
+
+  void insert(LockOp l) {
+    auto id = l.getLockID().value();
+    auto tile = l.getTile();
+    locks.insert(std::pair<Value, int>(tile, id));
+  }
```
jtuyls, line 152:
```suggestion
  DenseSet<std::pair<Value, int8_t>> locks;
  AMDAIEDeviceModel deviceModel;

  void insert(LockOp l) {
    int8_t id = l.getLockID().value();
    Value tile = l.getTile();
    locks.insert(std::pair<Value, int8_t>(tile, id));
  }
```

compiler/plugins/target/AMD-AIE/aie/AMDAIEObjectFifoStatefulTransform.cpp
```diff
@@ -291,14 +301,15 @@ bool requiresDMAs(ObjectFifoCreateOp createOp,
 /// return 0.
 int findObjectFifoSize(DeviceOp &device, Value tile,
                        ObjectFifoCreateOp objFifo) {
-  if (objFifoSize(objFifo) == 0) return 0;
+  auto ofSize = objFifoSize(objFifo);
```
jtuyls, line 304:
```suggestion
  int ofSize = objFifoSize(objFifo);
```

jtuyls (commented):
> Background: I'm grokking this pass to work out how to get around

I would avoid doing too many rewrites **(inside this pass)** and rather delete and get things out and rewriting while doing that.

compiler/plugins/target/AMD-AIE/aie/AMDAIEObjectFifoStatefulTransform.cpp
makslevental: `uint8_t`

jtuyls: Yeah, that needs a change in the op, but would be even better.

makslevental: oh hmm I forgot to change that - the init value is legit `int8_t` but the id should be `uint8_t`.

PR #672, opened 2024-08-13
Add logical objectfifo buffer depth assignment pass

jtuyls:
Add pass to assign logical objectfifo buffer depths based on provided options. By default double buffering will be enabled on L2 and L1.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignLogicalObjectFifoDepth.cpp
```diff
@@ -0,0 +1,70 @@
...
+    registry.insert<AMDAIEDialect>();
+  }
+
+  void runOnOperation() override;
+};
+
+void AMDAIEAssignLogicalObjectFifoDepthPass::runOnOperation() {
+  Operation *parentOp = getOperation();
+  IRRewriter rewriter(parentOp->getContext());
+  // Only assign buffer depth for L2 and L1, not for L3 (main memory).
```
Abhishek-Varma, line 38: I believe the comment is incorrect? Since we're indeed assigning buffer depth for L3 as well via `l3BufferDepth`.

Abhishek-Varma (approved):
Just one nit to address. LGTM otherwise!

Abhishek-Varma:
Unrelated failure. The failure is in `air`.

PR #679, opened 2024-08-15
Add logical objectFifo hoisting pass

jtuyls:
Follow-up on https://github.com/nod-ai/iree-amd-aie/pull/676, which blocks subsumption when operands are in the same scope. Hoists `amdaie.logicalobjectfifo.from_memref` ops as much as possible, which enables more subsumption opportunities.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -213,6 +213,11 @@ def AMDAIEHoistForLoopAffineApply : Pass<"iree-amdaie-hoist-for-affine-apply"> {
   let constructor = "mlir::iree_compiler::AMDAIE::createAMDAIEHoistForLoopAffineApplyPass()";
 }
 
+def AMDAIEHoistLogicalObjFifo : Pass<"iree-amdaie-hoist-logical-objectfifo"> {
+  let summary = "Hoist logical objectFifo operations to the scope of its operands.";
```
newling, line 217:
```suggestion
  let summary = "Hoist logical objectFifo operations to the scope of a most nested operand, without hoisting through workgroup, controlcode, or func ops.";
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEHoistLogicalObjFifo.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+namespace mlir::iree_compiler::AMDAIE {
+
+/// Hoist logical objectFifo operations until one of the operands is located
+/// within the same scope.
+LogicalResult hoistLogicalObjFifoOp(RewriterBase &rewriter,
+                                    AMDAIE::LogicalObjectFifoFromMemrefOp op) {
+  Operation *parentOp = op;
+  while (parentOp) {
+    Operation *newParentOp = parentOp->getParentOp();
```
newling, line 22: I interpret this as
"If the new parent is a proper ancestor of any operand, keep hoisting"
which seems like not what you want?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEHoistLogicalObjFifo.cpp
```diff
@@ -0,0 +1,60 @@
...
+    : public impl::AMDAIEHoistLogicalObjFifoBase<
+          AMDAIEHoistLogicalObjFifoPass> {
+  void runOnOperation() override;
+};
+
+void AMDAIEHoistLogicalObjFifoPass::runOnOperation() {
+  Operation *parentOp = getOperation();
+  IRRewriter rewriter(parentOp->getContext());
+  parentOp->walk([&](AMDAIE::LogicalObjectFifoFromMemrefOp op) {
+    (void)hoistLogicalObjFifoOp(rewriter, op);
```
newling, line 50: I'm not sure if `walk` enumerates all the ops it will visit before hand, I don't think so, so I think you actually visit all the hoisted ops twice (once after hoisting) here. 

FWIW I find code that
1) walk to find targets
2) process targets

easier to grok.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
newling: I guess it makes sense to not hoist through these ops because they're isolated from above (could you check that instead?)

newling (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
newling: Scrap that idea, they're not isolated from above

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEHoistLogicalObjFifo.cpp
jtuyls: > "If the new parent is a proper ancestor of any operand, keep hoisting"

No, then it breaks,

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEHoistLogicalObjFifo.cpp
jtuyls: > so I think you actually visit all the hoisted ops twice (once after hoisting) 

The hoisted ops will be inserted before the ancestor op, so should not be visited again?

jtuyls:
@newling Could you help check the PR again?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEHoistLogicalObjFifo.cpp
newling: Oh yes, you're right.

newling (approved):
Looks good. Apologies for inserting noise with invalid concerns last time

PR #680, opened 2024-08-15
Canonicalizers for doubly strided ops such as npu.dma_cpy_nd

newling:
Doubly strided ops have, for both source and target, and for all of offsets/sizes/strides, 2 fields:

1) a vector of **dynamic** Values
2) a vector of **static** integers

For example it might be

1) dynamic = {v0}
2) static = {16, kDynamic, 32, 64}

The size of **dynamic** is always exactly equal to the number of appearances of kDynamic in the **static** vector. This pass does the following: it checks if any of the Values in **dynamic** are actually MLIR constants, and if they are then it removes the Value from the dynamic vector and updates the corresponding index in **static**. So for example if `v0` above is actually 

```
%v0 = arith.constant 6 : index
```

then this canonicalization updates dynamic/static to 

1) dynamic = {}
2) static = {16, 6, 32, 64}

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.cpp
@@ -7,11 +7,9 @@
 #include "iree-amd-aie/IR/AMDAIEOps.h"
 
 #include "iree-amd-aie/IR/AMDAIEDialect.h"
-#include "llvm/ADT/TypeSwitch.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
-#include "mlir/Dialect/Transform/IR/TransformOps.h"
-#include "mlir/IR/DialectImplementation.h"
+#include "mlir/Dialect/Utils/StaticValueUtils.h"
 #include "mlir/IR/OpDefinition.h"
 
 #define GET_OP_CLASSES
@@ -52,7 +50,6 @@ LogicalResult ControlCodeOp::verify() {
 // AMDAIE_CoreOp
 //===----------------------------------------------------------------------===//
 
-
 void CoreOp::build(OpBuilder &b, OperationState &result, AMDAIE::TileOp tileOp,
                    ValueRange inputDmas, ValueRange outputDmas) {
   build(b, result, b.getIndexType(), tileOp, inputDmas, outputDmas, nullptr);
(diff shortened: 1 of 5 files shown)


MaheshRavishankar (commented):
There is a canonicalizer upstream [here](https://github.com/llvm/llvm-project/blob/1900810b6fd7b913d04b23afd9cce7e83fff51d7/mlir/lib/Dialect/Tensor/IR/TensorOps.cpp#L2432) that does this, but it is a bit hard-coded for `tensor.extract_slice` like op that have a single `offset` , `size` and `stride`. It should be split up to handle each of the `offset` , `size` and `stride` separately. Then you could reuse that here. Maybe just use the same style here to be consistent

jtuyls (commented):
Nice, should help make IR more compact and readable! Would be great to do for all doubly strided ops in one go I think.

MaheshRavishankar:
Actually, if your wait for this PR (https://github.com/llvm/llvm-project/pull/104488) it might get even easier

newling:
@MaheshRavishankar thanks for the pointer, I've changed to use the same style for now. 
@jtuyls ok, will get this working for all the doubly strided ops

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.cpp
```diff
@@ -93,6 +90,70 @@ TileOp CoreOp::getTileOp() {
 // AMDAIE_DmaCpyNdBaseOp
 //===----------------------------------------------------------------------===//
 
+namespace {
+// Simplified from upstream MLIR's foldDynamicIndexList:
```
jtuyls, line 94:
```suggestion
/// Simplified from upstream MLIR's foldDynamicIndexList:
```
Nit, but IREE uses `///` for function/class comments.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.cpp
```diff
@@ -93,6 +90,70 @@ TileOp CoreOp::getTileOp() {
...
+      ofr = attr;
+      valuesChanged = true;
+    }
+  }
+  return success(valuesChanged);
+}
+
+template <typename OpType, typename ReplacementBuilder>
+// Based on upstream MLIR's
+// OpWithOffsetSizesAndStridesConstantArgumentFolder
```
jtuyls, line 110:
```suggestion
/// Based on upstream MLIR's
/// OpWithOffsetSizesAndStridesConstantArgumentFolder
```
Nit, but IREE uses /// for function/class comments.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.cpp
```diff
@@ -449,8 +522,8 @@ void LogicalObjectFifoRelease::build(OpBuilder &b, mlir::OperationState &result,
 // AMDAIE_NpuDmaCpyNdOp
 //===----------------------------------------------------------------------===//
 
-// Build a NpuDmaCpyNdOp with mixed static and dynamic entries and target and
-// source BD IDs.
+// Build a NpuDmaCpyNdOp with mixed static and dynamic entries and target
+// and source BD IDs.
```
jtuyls, line 526:
```suggestion
/// Build a NpuDmaCpyNdOp with mixed static and dynamic entries and target
/// and source BD IDs.
```

jtuyls (approved):
LGTM, just a few small nits

newling:
> LGTM, just a few small nits

Ok I'll use /// by default in future. As this particular file uses both /// and // quite a lot I'm not going to update this PR.

PR #682, opened 2024-08-15
[Testing] Better handling of verbose flag 

newling:
Also: don't fail if you can't get a peano hash, this isn't strictly needed to run the tests 
FYI @erwei-xilinx


erwei-xilinx (approved):
Thanks for the fix!

PR #685, opened 2024-08-17
Do not serialize-deserialize module before calling aie2xclbin

newling:
This PR does a few things, happy to split it up if requested. 

1) Before pulling mlir-aie into iree-amd-aie, it we required to serialize-deserialize to call the aie2xclbin program. But we now just use aie2xclbin as a function, no shell out anymore. So no serialization-deserialization needed. 

2) This PR moves dma-to-npu closer to lower-to-aie pass. I think we'd eventually like to change the lowering of npu instructions from

```
amdaie dialect -> aie dialect -> npu
```

to 

```
amdaie dialect  -> npu
```

, because the amdaie and aie dialects are very similar and this indirection doesn't provide us with anything afaict. Making that change is currently not possible (dma-to-npu must currently run after stateful transform pass), this change is a step in that direction though

Test changes: I removed some CHECKs for ` aiex.runtime_sequence` in tests/samples, because that's now sucked into the LX instructions (sequence of integers). In my mind the tests in tests/samples are only useful to check that compilation doesn't error/crash, so IMO removing CHECK lines there is fine. 

3) General clean-up, for example we don't need aiex-to-standard pass anymore.

PR #686, opened 2024-08-17
Do we assume that there is just 1 runtime_sequence per device?

newling:


makslevental:
what are you gonna do with multiple seqs?

newling:
> what are you gonna do with multiple seqs?

There were places in the code that were handling the case of multiple seqs, I wanted to make sure I could delete this code https://github.com/nod-ai/iree-amd-aie/blob/9109a1fcae55c8f2f5a2c934e9f57b3a6b1b9967/compiler/plugins/target/AMD-AIE/aie/AMDAIEDmaToNpu.cpp#L524 . Now in https://github.com/nod-ai/iree-amd-aie/pull/685

PR #689, opened 2024-08-19
Windows E2E

makslevental:
This PR adds Windows E2E support.


makslevental:
With current top commit here this works e2e with this (`pwsh`) command:

```
python.exe \
  C:\Users\maksim\dev_projects\iree-amd-aie\iree-amd-aie\build_tools\ci\cpu_comparison\run_test.py \
  C:\c-c \
  C:\bui \
  C:\Users\maksim\dev_projects\iree-amd-aie\iree-amd-aie\build_tools\llvm-aie \
  C:\Users\maksim\dev_projects\XRT-MCDM\build\WDebug\xilinx\xrt
```

_and_ the closed source build of the umddml shim; note that last arg in the above where the build artifacts of the MCDM driver are deposited, specifically `xrt_umddml.dll`.

Note for some reason this works with the build dir of iree (`C:\bui`) but not an install version 🤷 (probably something about symbols - will take a little investigation).

**Key takeaway**: we need `xrt_umddml.dll` distributed with the [publicly released version of the windows drivers](https://ryzenai.docs.amd.com/en/latest/inst.html#install-npu-drivers) (currently not happening). I'll make an internal post/email/something soon.

cc @MaheshRavishankar @kumardeepakamd @powderluv @stellaraccident

build_tools/ci/run_matmul_test.sh
```diff
@@ -88,7 +95,9 @@ fi
 if [ -z "${3-}" ]; then
   PEANO=/opt/llvm-aie
 else
-  PEANO=`realpath "$3"`
+  if [ -d "$PEANO" ]; then
```
newling:
```suggestion
  if [ "$PEANO" ]; then
```

Same for XRT_DIR and VITIS

Because if the user provided an invalid path, it would be good to fail informatively

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -7,11 +7,12 @@
 #include "XCLBinGen.h"
 
 #include <filesystem>
-#include <fstream>
 #include <functional>
 #include <random>
 #include <regex>
 #include <sstream>
+// ReSharper disable once CppUnusedIncludeDirective
```
newling, line 14: What does this mean?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -293,12 +297,30 @@ std::optional<std::string> dumpStrToDisk(const std::string &payload,
...
+  }
+  return false;
+}
+
 // Returns either:
 //  -- the output of running the tool, if run without failure, or
 //  -- an empty optional, if the tool fails to run.
-static std::optional<std::string> runTool(
-    const std::string &program, const std::vector<std::string> &args,
+LogicalResult runTool(
```
newling, line 311: The comment above isn't valid after this change

runtime/src/iree-amd-aie/aie_runtime/iree_aie_configure.cc
```diff
@@ -210,9 +210,12 @@ LogicalResult pushToBdQueueAndEnable(const AMDAIEDeviceModel &deviceModel,
 }
 
 LogicalResult addElfToTile(const AMDAIEDeviceModel &deviceModel,
-                           const TileLoc &tileLoc, const Path &elfPath,
-                           bool aieSim) {
+                           const TileLoc &tileLoc, Path &elfPath, bool aieSim) {
   auto devInst = const_cast<XAie_DevInst *>(&deviceModel.devInst);
+  if (!std::filesystem::exists(elfPath)) {
```
newling, line 215: `std::filesystem::exists` takes a const path (https://en.cppreference.com/w/cpp/filesystem/exists) so not sure why you de-consts this. But I like the error.

runtime/src/iree-amd-aie/driver/xrt/native_executable.cc
```diff
@@ -184,8 +184,19 @@ iree_status_t iree_hal_xrt_native_executable_create(
       return iree_make_status(IREE_STATUS_INTERNAL, "XCLBIN load error: %s",
                               e.what());
     }
-    device->register_xclbin(*xclbin);
-    xrt::hw_context context(*device, xclbin->get_uuid());
+    try {
+      device->register_xclbin(*xclbin);
+    } catch (std::runtime_error& e) {
```
newling, line 189:
```suggestion
    } catch (const std::exception & e) {
```
Super pedantic would be to then `catch (...)` after that https://en.cppreference.com/w/cpp/language/catch (not a request!)

runtime/src/iree-amd-aie/driver/xrt/native_executable.cc
```diff
@@ -267,8 +279,8 @@ static void iree_hal_xrt_native_executable_destroy(
 
   for (iree_host_size_t i = 0; i < executable->entry_point_count; ++i) {
     try {
-      delete executable->entry_points[i].kernel;
-      delete executable->entry_points[i].instr;
+      // delete executable->entry_points[i].kernel;
```
newling, line 282: Removable

tests/matmul/requirements.txt
```diff
@@ -1,3 +1,4 @@
 PyYAML>=5.4.1
 requests>=2.28.0
-enum_tools==0.6.4
\ No newline at end of file
+enum_tools==0.6.4
+numpy<2
```
newling, line 4: Why less than 2, that's a bit annoying, can we rather change our code to make it forward compatible?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -293,12 +297,30 @@ std::optional<std::string> dumpStrToDisk(const std::string &payload,
...
+    const std::string &program_, const std::vector<std::string> &args,
```
newling, line 312:
```suggestion
    std::string program, const std::vector<std::string> &args,
```
and then just
```
#if defined(_WIN32)
  if (!hasEnding(program, ".exe")) program += ".exe";
#endif
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -330,20 +349,34 @@ static std::optional<std::string> runTool(
     if (errorCode) {
       llvm::errs() << "Failed to create temporary file: " << errorCode.message()
                    << "\n";
-      return {};
+      return failure();
     }
   }
 
+  SmallVector<std::optional<StringRef>> redirects;
+#ifdef _WIN32
```
newling, line 357: Can envSmallVec not be the same type for windows and linux? Please leave a comment in code if not

newling (commented):
Looks good (well as good as it can maybe with preprocessor _WIN32 sprinkled around). I just left a few shallow/nit comments

build_tools/ci/run_matmul_test.sh
```diff
@@ -65,11 +65,18 @@ fi
 IREE_COMPILE_EXE=""
 TEST_RUNNER=""
 for dir in "${IREE_INSTALL_DIR}" "${IREE_INSTALL_DIR}/bin" "${IREE_INSTALL_DIR}/tools"; do
+  echo "Looking in $dir"
   if [ -f "${dir}/iree-compile" ]; then
     IREE_COMPILE_EXE="${dir}/iree-compile"
   fi
-  if [ -f "${dir}/testing/e2e/iree-e2e-matmul-test" ]; then
-    TEST_RUNNER="${dir}/testing/e2e/iree-e2e-matmul-test"
+  if [ -f "${dir}/iree-compile.exe" ]; then
```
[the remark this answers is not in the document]
makslevental: because if the linux (no exe) fails then the exe will succeed. Just very broadly, I'll say that I do not want to spend a lot of time fixing this script, which we agreed should either be ported to python or absorbed into `cpu_comparison/run_test.py`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
makslevental: it means my editor (and maybe yours if you use ReSharper) won't think `<fstream>` isn't used, which it's not if you're not on windows.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_configure.cc
makslevental: It takes a const reference which means copy ellision can happen. I de-consted because prior I was doing `make_preferred` in here which mutates.

runtime/src/iree-amd-aie/driver/xrt/native_executable.cc
makslevental: well no becaues i'm not sure if this works - @jtuyls added this when he was fighting the numerics/XRT issues. Currently I'm still figuring out if this passes on Linux (there's a new CI runner with an updated driver now).

makslevental: It passes on my machine but like I say: if it didn't happen on CI it didn't happen :)

build_tools/ci/build_test_cpp.sh
```diff
@@ -32,6 +32,9 @@ if [[ "$OSTYPE" == "linux-gnu"* ]]; then
   export CMAKE_TOOLCHAIN_FILE="$this_dir/linux_default_toolchain.cmake"
   export CC=clang
   export CXX=clang++
+else
```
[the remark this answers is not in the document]
kumardeepakamd: I believe for git-bash and MinGW it should be 'msys". Could test and assert on that.

kumardeepakamd (commented):
Great to see clang being used to build on Windows. We should make sure we can build with either MSVC cl.exe or clang?

makslevental:
> Great to see clang being used to build on Windows. We should make sure we can build with either MSVC cl.exe or clang?

We can - nothing changes in the scripts. It's what I was doing before I figured out I could use clang-cl

build_tools/ci/build_test_cpp.sh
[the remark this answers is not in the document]
makslevental: That's correct. I didn't put it because I wasn't sure we were gonna go with bash scripts for Windows users. After our convo I've decided I'm going to write the pwsh version of this same script and just use it in CI too.

build_tools/ci/run_matmul_test.sh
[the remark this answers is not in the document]
newling: Makes sense

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
newling: Got it, thanks

tests/matmul/requirements.txt
makslevental: numpy 2 introduced [breaking changes](https://numpy.org/devdocs/numpy_2_0_migration_guide.html) - it was their first version bump in like 9 years? I don't know I'm sure we upgrade/make changes but for right now I'm doing `numpy<2` everywhere (and then some years from now I'll figure out numpy 2).

runtime/src/iree-amd-aie/driver/xrt/native_executable.cc
makslevental: yea no chance i'm doing that lol

runtime/src/iree-amd-aie/aie_runtime/iree_aie_configure.cc
makslevental: restored

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
makslevental: removed

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
makslevental: you can't - all of the `(peanoDir / "bin" / "clang").string()` only work with const&

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
makslevental: you can't make a non-empty `ArrayRef<StringRef>` - it's an array _ref_. so you have to make a smallvector that then gets refed. but in the _WIN32 case it's empty/uninited/never called (because i'm constructing the optional from `nullopt`).

makslevental: added a comment

.github/workflows/ci-linux.yml
```diff
@@ -190,23 +183,19 @@ jobs:
...
-          # TODO(newling) make requirements.txt file
-          pip install numpy
           source /opt/xilinx/xrt/setup.sh
           python3 build_tools/ci/cpu_comparison/run_test.py \
             test_aie_vs_cpu \
-            iree-install \
+            $PWD/iree-install \
             $PWD/llvm-aie \
-            /opt/xilinx/xrt \
-            /opt/Xilinx/Vitis/2024.2
```
makslevental, line 202: TODO: restore

.github/workflows/ci-linux.yml
```diff
@@ -152,6 +150,7 @@ jobs:
           # on the guthub CI machine.
           sudo prlimit -lunlimited --pid $$
 
+          source /opt/xilinx/xrt/setup.sh
```
makslevental, line 153: TODO: respell using `export`

build_tools/ci/build_test_cpp.sh
makslevental, line 35: TODO: `else if`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIETargetCDODirect.cpp
```diff
@@ -168,8 +168,8 @@ LogicalResult addAieElfsToCDO(const AMDAIEDeviceModel &deviceModel,
       else
         fileName = "core_" + std::to_string(tileLoc.col) + "_" +
                    std::to_string(tileLoc.row) + ".elf";
-      if (failed(addElfToTile(deviceModel, tileLoc, workDirPath / fileName,
```
makslevental, line 171: TODO: restore

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -276,12 +280,27 @@ std::optional<std::string> dumpStrToDisk(const std::string &payload,
...
+                              ending.length(), ending) == 0;
+  }
+  return false;
+}
+
+LogicalResult runTool(
+    const std::string &program_, const std::vector<std::string> &args,
     bool verbose, std::optional<std::vector<std::string>> env = std::nullopt) {
+  std::string program;
+#if defined(_WIN32)
```
makslevental, line 295: TODO: use `!`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -276,12 +280,27 @@ std::optional<std::string> dumpStrToDisk(const std::string &payload,
...
+  if (hasEnding(program_, ".exe")) {
```
makslevental, line 296: TODO: use `!`

MaheshRavishankar (approved):

.github/workflows/ci-windows.yml
```diff
@@ -104,3 +102,59 @@ jobs:
...
+      - name: Create venv and install dependencies
+        run: |
+          python -m venv .venv
+          source .venv/Scripts/activate
+          pip install -r tests/matmul/requirements.txt
+
+      - name: E2E correctness matmul test
+        run: |
+          source .venv/Scripts/activate
+          export XILINX_XRT=/c/Xilinx/XRT
```
dezhiAmd, line 138: On windows OS, all applications are supposed to uses xrt_coreutil.dll from c:\windows\system32. But this line make the CI computer use the xrt_coreutil.dll from /c/Xilinx/XRT. 
This may cause all our tests get a different test result from a system deployed to general public

stsoe: Agree with Dezhi's statement.  On Windows XILINX_XRT should not be set when running tests.

makslevental: > On Windows XILINX_XRT should not be set when running tests.

As far as I am aware, XRT is open source and thus we are free to use it according to how it best suits our needs. So we have our usecase and implementation and you have yours. If you feel so strongly, you are free to refactor your code so that it doesn't expose these env variables (and I am free to fork).

dezhiAmd: @makslevental ,
What version of NPU KMD driver is installed on the windows computer used for CI?

makslevental: Please stop commenting on this PR. It is now 3 months since it was closed. If you have some current issue then create a new issue about it.

PR #690, opened 2024-08-20
Rerun tests on windows (flakes)

makslevental:


newling (approved):
I'd prefer targeting individual tests for repeating or disabling, but this PR is probably way simpler than targeting individual tests on windows so I'm fine with it

PR #692, opened 2024-08-21
[LowerToAIE] Don't force change memspace

newling:
This constraint the current convolution pipeline, where there are further memref.subviews which end up with different memspaces on src and dst. 

Removing this constraint doesn't seem to make anything fail, so I'm just removing it.


jtuyls (approved):
Interesting that you don't get into trouble with this as I remember issues with different memory space in `aie.objectfifo` (related) ops. But fine by me if this passes CI.

newling:
> Interesting that you don't get into trouble with this as I remember issues with different memory space in `aie.objectfifo` (related) ops. But fine by me if this passes CI.

Probably something waiting to reappear later, I'll keep an eye out for this.

PR #693, opened 2024-08-21
[LowerToAIE] Change handling of linalg operations (fix for conv lowering)

newling:
This PR removes the special case handling of linalg ops, which was causing convolution lowering to fail. 

The default switch case handles linalg ops better here, so this PR falls back to the default. In particular, ops with regions such as linalg conv 1d are handled correctly by the default switch case. With the cloning approach in the code this PR removes, the nested conv 1d op was causing problems in the `IRMapping mapper`, and we were ending up with a aie.core op who's operand was the result of linalg.conv op (lol). 

Please, let's stop using  `.Case<linalg::LinalgOp>([&](auto linalgOp)` when the only linalg op being tested is linalg.fill. Rather just use `Case<linalg::FillOp>` and maybe error out of other linalg ops are encountered -- this will make debugging much easier.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -396,9 +384,6 @@ LogicalResult coreToAIE(IRRewriter &rewriter, AMDAIE::CoreOp coreOp,
               return coreReleaseOpToAIE(rewriter, releaseOp, mapper,
                                         toBeErased);
             })
-            .Case<linalg::LinalgOp>([&](auto linalgOp) {
```
MaheshRavishankar, line 399: Just reading the description and PR, this seems like it is dropping general handling of `LinalgOp`. Is that to mean that we dont expect `LinalgOp`s to survive till here?

newling: Instead of handling them as a special case: https://github.com/nod-ai/iree-amd-aie/blob/57cd0bca89917e47639b9189683ac71b5d399f3c/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp#L399

They are now handled like other not-special ops: https://github.com/nod-ai/iree-amd-aie/blob/57cd0bca89917e47639b9189683ac71b5d399f3c/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp#L411

We (I) do expect LinalgOps to survive til here.

newling:
I thought this was a fix for the non-deterministic crash I'm seeing in this pass with convolution, but it's not. I think I've found the actual problem (IRMapping not being updated when operations are erased) I'll post a fix for this tomorrow

makslevental:
> I thought this was a fix for the non-deterministic crash I'm seeing in this pass with convolution, but it's not. I think I've found the actual problem (IRMapping not being updated when operations are erased) I'll post a fix for this tomorrow

really? where?

newling:
> Member

https://github.com/nod-ai/iree-amd-aie/pull/694

newling:
Bundled with https://github.com/nod-ai/iree-amd-aie/pull/705

PR #694, opened 2024-08-22
[LowerToAIE] Fix for non-deterministic error (observed with convolution)

newling:
What was happening was 

1) Value/Operation inserted into IRMapping
2) IRRewriter erased Value/Op
3) New Value/Op created with same address as erased Value/Op
4) Bad state: new Value/Op is a key in IRMapping, even though it was never added. 

That's my analysis, at least. Either way, this change makes my failure rate go from 10/20 to 0/20. 

An alternative to this PR might be to postpone erasing of ops even longer, so that only after all CoreOps have been processed are the ops erased.


makslevental:
Isn't the "correct" thing to do is to use `Conversion` instead of just an `OperationPass`? Like doesn't a conversion both handle the mapping for you _and_ actually defer the erases (that last part is definitely true).

makslevental:
I mean we should probably just merge this and be done with it because why rewrite a pass that we won't need soon.

makslevental (approved):

PR #696, opened 2024-08-23
[CreateLogicalObjectFifoLink] Fail if accesses not contiguous 

newling:
I was looking into this pass because convolution lowering hits 

```
error: 'amdaie.circular_dma_cpy_nd' op access pattern of strided operation overlaps with next one, which is not supported for now
```

But looking at the code now, it looks like it'll currently only work for the the access patterns are perfectly contiguous? If that's the case, I can look into fixing it (for both cases -- overlapping and non-contiguous). 

I left a comment in the code describing my thinking


jtuyls:
> But looking at the code now, it looks like it'll currently only work for the the access patterns are perfectly contiguous? 

I think 'contiguous access patterns' is ambiguous. The access pattern of a single DMA does not need be contiguous (e.g. `sizes: [2, 1024], strides: [2048, 1]` is not a contiguous access pattern), but the access patterns of multiple consumers/producers that need to be linked indeed need to be contiguous with respect to each other. I guess you're only referring to the latter, right?

newling:
> > But looking at the code now, it looks like it'll currently only work for the the access patterns are perfectly contiguous?
> 
> I think 'contiguous access patterns' is ambiguous. The access pattern of a single DMA does not need be contiguous (e.g. `sizes: [2, 1024], strides: [2048, 1]` is not a contiguous access pattern), but the access patterns of multiple consumers/producers that need to be linked indeed need to be contiguous with respect to each other. I guess you're only referring to the latter, right?

Yes, I'm referring to the multiple ops being linked. The code which sets all offsets to zero seems to be expecting something stricter than the code checks for (this PR makes the check stricter).

jtuyls (approved):
LGTM

PR #413, opened 2024-06-13
[DistributeCoresAndObjectFifos] Add local and shared memory distribution and rewrite

jtuyls:
Add `DistributeCoresAndObjectFifos` changes for cpp pipeline and e2e example, see all changes here: https://github.com/nod-ai/iree-amd-aie/compare/jornt_cpp_pipeline?expand=1#diff-888fdafbbd093dc98e4fae55a26e7fa56065fc7506fedc0107b8a897c19f816a

- Add local memory distribution: rewrite a larger memory across multiple cores into dedicated ones for each core
- Add shared memory distribution: allocate different memories for logical objectFifos on the same shared memory tile to ensure different buffers will be used
- Rewrite L1 tile assignment to use new `amdaie.logicalobjectfifo.access` operations
- Rewrite L2/L3 tile assignment into two parts: 1) finding all potential tiles and 2) selecting specific tiles

NOTE: In this PR, the `amdaie.logicalobjectfifo.access`  operation is not supported e2e and further lowering and usage will be handled in a follow-up. There are a few places were code is commented and temporary logic is inserted to keep all tests functional (see comments).

PR #697, opened 2024-08-23
[DistributeCoresAndObjectFifos][Windows flake] Don't erase op too early  

newling:
The call to replaceOp in the middle of this pass calls eraseOp, which is possibly the cause of the error observed on windows:

```
Error: <unknown>:0: error: 'arith.constant' op needs to have logic implemented for handling in distributeLocalMemory
```

In https://github.com/nod-ai/iree-amd-aie/pull/413 most of the replacements use the pattern replaceAllUsesWith, so I'm guessing it was just a mistake to use replaceOp here. 

Included in PR, a bit of code safety/sanitizing. 

Context: all this logic in distributeLocalMemory should be replaced with a robust 'privatization' pass. Basically, it's a stop gap to split the large L3 memory (describing all memories on all cores) into a local L3 memory.


makslevental (approved):
🙏

Issue #704, opened 2024-08-26
`pack_peel_pipeline_matmul` and `conv_fill_spec_pad` consistently fail (on Mac)


makslevental:
I don't know why only Mac picks this up but there are several tests that actually consistently fail with errors like:

```
/Users/runner/work/iree-amd-aie/iree-amd-aie/tests/transform_dialect/conv_fill_spec_pad.mlir:57:3: warning: op argument #0 is not consumed in the body but is marked as consumed
  transform.named_sequence @full_pipeline(%variant_op: !any {transform.consumed}) {
  ^
```

https://github.com/nod-ai/iree-amd-aie/actions/runs/10563277691/job/29263141735#step:9:11051

@jtuyls, @newling any idea?

yzhang93:
What is the error for `pack_peel_pipeline_matmul`?

makslevental:
I don't remember but if you remove `-E ...` here https://github.com/nod-ai/iree-amd-aie/blob/main/build_tools/ci/build_test_cpp.sh#L101 you'll see it fail in CI.

yzhang93:
It seems `conv_fill_spec_pad` fails flaky now, do you want me to merge the PR or not?

PR #706, opened 2024-08-26
Fix conv_fill_spec_pad.mlir

yzhang93:
This change will get rid of the warning `conv_fill_spec_pad.mlir:57:3: warning: op argument #0 is not consumed in the body but is marked as consumed`.

Also attempt to fix https://github.com/nod-ai/iree-amd-aie/issues/704.


newling (approved):
Thanks! Seems like the conv_fill_spec_pad problem is solved.

yzhang93:
> Thanks! Seems like the conv_fill_spec_pad problem is solved.

Well, I still see some failures for this test https://github.com/nod-ai/iree-amd-aie/actions/runs/10567875494/job/29277762727?pr=706#step:9:11037. But then it passed after repeating. So I guess now it doesn't consistently fail but only flaky.

newling:
Ah ok I thought this one failed consistently on OSX and didn't have 5 chances to pass

tests/transform_dialect/conv_fill_spec_pad.mlir
```diff
@@ -183,11 +183,11 @@ module attributes { transform.with_named_sequence } {
...
 
-    %conv_pre_contract =
-       transform.structured.match ops{["linalg.conv_1d_ncw_fcw"]}
-               in %fused_for_all : (!any) -> !any
+    // %conv_pre_contract =
+    //    transform.structured.match ops{["linalg.conv_1d_ncw_fcw"]}
+    //            in %fused_for_all : (!any) -> !any
 
-    transform.structured.vectorize %conv_pre_contract : !any
+    // transform.structured.vectorize %conv_pre_contract : !any
```
makslevental, line 190: Any reason to save this as a comment rather than just remove?

makslevental (approved):
Not familiar with these transform schedules but looks fine to me.

makslevental:
Why not just get rid of it? Personally I'm not a fan of transform schedules but that's just me :shrug:

tests/transform_dialect/conv_fill_spec_pad.mlir
yzhang93: Uh, I was just trying to see if this part of the codes caused flaky crash, but seems not. Without these, it still have some failures. So I'm going to keep blocking this test for MacOS Ci until figuring out the reason. 

Is there a MacOS machine that I can use to debug the CI failure?

PR #708, opened 2024-08-27
Fix LdScript emission

makslevental:
Without explicit locations for these sections, they get placed in arbitrary locations. The introduced linker flag will error out if new sections appear (and don't have specific locations/assignments).

See https://github.com/Xilinx/mlir-aie/pull/1717

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIETargetLdScript.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIETargetLdScript.cpp
@@ -23,43 +23,15 @@ static void writeLDScriptMap(raw_ostream &output, BufferOp buf, int offset) {
   output << ". += 0x" << llvm::utohexstr(numBytes) << ";\n";
 }
 
-///// ld.script format:
-//
-// MEMORY
-// {
-//    program (RX) : ORIGIN = 0, LENGTH = 0x0020000
-//    data (!RX) : ORIGIN = 0x20000, LENGTH = 0x0020000
-// }
-// ENTRY(_main_init)
-// INPUT(something.o)
-// SECTIONS
-// {
-//   . = 0x0;
-//   .text : {
-//      // the __start symbol from crt0.o has to come at address zero.
-//      *crt0.o(.text)
-//      . = 0x200;
-//      *(.text)
-//   } > program
-//   .data : { *(.data) } > data
-//   . = 0x20000;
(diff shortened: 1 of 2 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIETargetLdScript.cpp
```diff
@@ -95,9 +67,33 @@ SECTIONS
...
+  }
+  .rtstab : {
+     *(.rtstab)
+  }
+  .eoltab : {
+     *(.eoltab)
+  }
+  .chesstypeannotationtab : {
+     *(.chesstypeannotationtab)
+  }
```
jtuyls, line 93: Orthogonal to this PR specifically, but are all of these needed? I.e. aren't some of these metadata and wouldn't we be able to make the elf smaller by excluding those?

jtuyls (approved):
LGTM

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIETargetLdScript.cpp
makslevental: The way this works is with these sections specified explicitly, *if* any of the ELFs contain these sections, *then* those sections will be placed at these locations. E.g., this PR failed at first because I turned on `=error` and `mm.o` had a `.chesstypeannotationtab` section that currently gets placed at an arbitrary location. 

So the sections are already there and are already getting placed into the ELF. `-Wl,--gc-sections` pertains to your question and there I do not know what counts as "unused" and therefore "will be gced".

jtuyls: I see, thanks for the explanation! I am a rookie on ld scripts so will take your word for it :) At some point, it might be interesting to see what sections are actually copied into the elf.

makslevental: I ain't no pro either - 90% of what I'm saying is just stuff I'm absorbing by osmosis by reading PRs/discussions in llvm-aie.

makslevental: Actually that's not true - I did actually read some "docs" on it (shocking I know); I read about half of this thing https://mcyoung.xyz/2021/06/01/linker-script/.

PR #713, opened 2024-08-28
[StatefulTransform] Remove some unused logic

jtuyls:
Removing some unused logic as part of refactoring `AMDAIEStatefulTransform` and moving `AIE` transformations to `AMDAIE`:

1. DMAChannelAnalysis: we're not prefilling the device op and are not expecting to do that in the future as we're trying to get rid of this pass, so get rid of the constructor with device op parameter.
2. Remove topological sort of erased ops.
3. Simplify `getOptionalSharedTile`.


compiler/plugins/target/AMD-AIE/aie/AMDAIEObjectFifoStatefulTransform.cpp
```diff
@@ -1092,16 +1049,14 @@ struct AMDAIEObjectFifoStatefulTransformPass : mlir::OperationPass<DeviceOp> {
...
     device.walk([&](Operation *op) {
       if (isa<ObjectFifoCreateOp, ObjectFifoLinkOp, ObjectFifoAcquireOp,
-              ObjectFifoSubviewAccessOp, ObjectFifoReleaseOp>(op))
-        opsToErase.insert(op);
+              ObjectFifoSubviewAccessOp, ObjectFifoReleaseOp>(op)) {
+        op->dropAllUses();
+        rewriter.eraseOp(op);
+      }
     });
-    topologicalSort(opsToErase);
```
makslevental, line 1101: I added this actually (in mlir-aie) because if you don't do this in toposort order and it fails (and the pass doesn't complete) you can leave the IR in an invalid state thereby preventing debug dumps. But I guess `dropAllUses` takes care of the same problem.

makslevental (approved):
I see red I approve

compiler/plugins/target/AMD-AIE/aie/AMDAIEObjectFifoStatefulTransform.cpp
jtuyls: Yeah, this is not too important, I was just trying to get rid of some stuff. Soon none of this should be needed anyway.

PR #716, opened 2024-08-28
[canonicalization] Add AMDAIEDialectFoldInterface to keeps ops in amdaie.cores

newling:
See https://mlir.llvm.org/docs/Canonicalization/#globally-applied-rules
Copied from AIE dialect, which does the same thing for `aie.core`


newling:
@makslevental I don't think there's a way to do this in the .td file.

makslevental (approved):

PR #718, opened 2024-08-28
[AMDAIECoreToStandard] Improve error message when core not isolated 

newling:
In case we hit this error again, I'd like the message to be more informative. The new error is something like:
```
<unknown>:0: error: 'arith.constant' op is not in the core in which it is used. Cores must be `isolated` before this point.
<unknown>:0: note: see current operation: %0 = "arith.constant"() <{value = 0 : i32}> : () -> i32
<unknown>:0: error: 'aie.device' op Failed to lower to LLVM
<unknown>:0: note: see current operation:
"aie.device"() <{device = 8 : i32}> ({
```

compared to old error, something like: 
```
LLVM ERROR: operation destroyed but still has uses
Please report issues to https://github.com/iree-org/iree/issues and include the crash backtrace.
Stack dump without symbol names (ensure you have llvm-symbolizer in your PATH or set the environment var `LLVM_SYMBOLIZER_PATH` to point to it):
0  libIREECompiler.so 0x00007fa09d494f87 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) + 39
1  libIREECompiler.so 0x00007fa09d4931c0 llvm::sys::RunSignalHandlers() + 80
```

### DIFF
--- a/compiler/plugins/target/AMD-AIE/aie/AMDAIECoreToStandard.cpp
+++ b/compiler/plugins/target/AMD-AIE/aie/AMDAIECoreToStandard.cpp
@@ -126,28 +126,31 @@ struct AMDAIECoreToStandardFunc : OpConversionPattern<CoreOp> {
         tileRow(tileRow) {}
 
   LogicalResult matchAndRewrite(
-      CoreOp op, OpAdaptor adaptor,
+      CoreOp coreOp, OpAdaptor adaptor,
       ConversionPatternRewriter &rewriter) const override {
-    TileOp t = getTileOp(*op);
+    TileOp t = getTileOp(*coreOp);
     int col = t.getCol();
     int row = t.getRow();
 
     // Only pull code for the indicated function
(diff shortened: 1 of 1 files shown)


compiler/plugins/target/AMD-AIE/aie/AMDAIECoreToStandard.cpp
```diff
@@ -251,6 +257,28 @@ struct AMDAIECoreToStandardPass : mlir::OperationPass<ModuleOp> {
     if (failed(applyPartialConversion(m, target, std::move(patterns))))
       return signalPassFailure();
 
+    // Assert that cores are isolated
+    {
```
jtuyls, line 261: You could put it in a utility function inside the CoreOp itself so it can be reused in different places. But I wouldn't spend effort on that for AIE ops and wait for it until we solely rely on AMDAIE ops.

jtuyls (approved):
LGTM

PR #719, opened 2024-08-29
Fix GHA caches

makslevental:
GHA (github action) [caches](https://github.com/nod-ai/iree-amd-aie/actions/caches) aren't mutable so currently we're timestamping them:

```
windows-build-test-cpp-asserts-v1-f77db6670966b372878b4219599ba723feb41945-2024-08-28T19:33:14Z
```

 Thus PRs that run will upload new caches and push out old but useful caches, e.g., caches from main that could be used by PRs. 

~~This PR tries to mitigate that by using either PR number (in case the cache is produced by a PR) or `github.ref_name-N` where `N` always increments (in case the cache is produced by a branch run, like cache warming on `main`) to label the cache. The overwrite is done by first deleting any existing cache by that name and then uploading the new cache to the same label/key.~~

~~So PR caches look like this~~

```
linux-build-test-cpp-asserts-manylinux-v2-719
```

~~and will be fixed/reused for the duration of the PR.~~

~~And caches produced by `main` look like~~

```
linux-build-test-cpp-asserts-manylinux-v2-main-123
```

FYI for anyone wondering how this all works - the [`restore-keys`](https://github.com/nod-ai/iree-amd-aie/blob/main/.github/workflows/ci-linux.yml#L63) are ~~longest prefix matched~~ (see https://github.com/nod-ai/iree-amd-aie/pull/719#discussion_r1735941967); so

```
key: linux-build-test-cpp-asserts-manylinux-v2-${{ github.event.number || format('{0}-{1}', github.ref_name, github.run_number) }}
restore-keys: linux-build-test-cpp-
```

will create a cache with the aforementioned name but then restore the cache with ~~longest match~~ (see  https://github.com/nod-ai/iree-amd-aie/pull/719#discussion_r1735941967) starting from `linux-build-test-cpp-` (including caches created by main).~~

see https://github.com/nod-ai/iree-amd-aie/pull/719#issuecomment-2319190867

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -9,12 +9,23 @@ on:
     branches:
       - main
 
+  schedule:
+    # At minute 0 past every 6th hour. (see https://crontab.guru)
+    # this job is to keep the ccache cache warm
+    - cron: '0 */6 * * *'
+
 concurrency:
   # A PR number if a pull request and otherwise the commit hash. This cancels
   # queued and in-progress runs for the same PR (presubmit) or commit
   # (postsubmit).
   group: ci-build-test-cpp-linux-${{ github.event.number || github.sha }}
   cancel-in-progress: true
+
+permissions:
+  contents: read
+  # required to delete cache
+  # https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28#delete-github-actions-caches-for-a-repository-using-a-cache-key
+  actions: write
 
 jobs:
   build_and_ctest:
@@ -24,6 +35,7 @@ jobs:
       fail-fast: true
     env:
       CACHE_DIR: ${{ github.workspace }}/.container-cache
+      CACHE_KEY: linux-build-test-cpp-asserts-manylinux-v2-${{ github.event.number || github.ref_name }}
     steps:
       - name: Set unified TZ
         uses: szenius/set-timezone@v2.0
@@ -42,11 +54,16 @@ jobs:
         with:
           submodules: recursive
 
-      - name: Install static libs
+      - name: Install deps
         run: |
-          dnf install -y almalinux-release-devel
           yum remove -y openssl-devel zlib-devel || true
-          yum install -y protobuf-devel protobuf-compiler
+          
+          dnf install -y almalinux-release-devel
+          type -p yum-config-manager >/dev/null || yum -y install yum-utils
+          yum-config-manager -y --add-repo https://cli.github.com/packages/rpm/gh-cli.repo
+          
+          yum install -y protobuf-devel protobuf-compiler gh
+
       - name: Sync source deps
         run: |
           python ./sync_deps.py
@@ -59,7 +76,7 @@ jobs:
         uses: actions/cache/restore@v3
         with:
           path: ${{ env.CACHE_DIR }}
-          key: linux-build-test-cpp-asserts-manylinux-v2-${{ github.sha }}
+          key:  ${{ env.CACHE_KEY }}
           restore-keys: linux-build-test-cpp-
 
       - name: Build packages
@@ -80,12 +97,22 @@ jobs:
           path: iree-dist-linux.tar
           if-no-files-found: warn
 
+      - name: Overwrite cache
+        if: ${{ !cancelled() }}
+        continue-on-error: true
+        env:
+          # required to run gh with privileges
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh extension install actions/gh-actions-cache
+          gh actions-cache delete ${{ env.CACHE_KEY }} --confirm -R ${{ github.repository }}
+
       - name: Save cache
         uses: actions/cache/save@v3
         if: ${{ !cancelled() }}
         with:
           path: ${{ env.CACHE_DIR }}
-          key: linux-build-test-cpp-asserts-manylinux-v2-${{ github.sha }}
+          key: ${{ env.CACHE_KEY }}
 
   test_linux:
     name: E2E Test linux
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-linux.yml
```diff
@@ -59,7 +76,7 @@ jobs:
         uses: actions/cache/restore@v3
         with:
           path: ${{ env.CACHE_DIR }}
-          key: linux-build-test-cpp-asserts-manylinux-v2-${{ github.sha }}
+          key:  ${{ env.CACHE_KEY }}
           restore-keys: linux-build-test-cpp-
```
jtuyls, line 80: > FYI for anyone wondering how this all works - the [restore-keys](https://github.com/nod-ai/iree-amd-aie/blob/main/.github/workflows/ci-linux.yml#L63) are longest prefix matched; so

I might be missing something, but [here](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#matching-a-cache-key) it says:

> When a key doesn't match directly, the action searches for keys prefixed with the restore key. If there are multiple partial matches for a restore key, the action returns the most recently created cache.

Still, it would first look for an exact match with `key` and only search further if not found, so I think this behaves like you're intending.

jtuyls (approved):
LGTM

.github/workflows/ci-linux.yml
makslevental: Ctrl+f prefix here https://github.com/actions/cache?tab=readme-ov-file#inputs

newling:
Thanks for the explainer! Some noob questions:

What is the 123 in `linux-build-test-cpp-asserts-manylinux-v2-main-123`, is this the 123'rd cache of main ever created? Why do we need multiple caches of main? 

Do we even need caches of branches. Like, shouldn't they always be close enough to main to make the cache from main basically good enough?

What about branches from forks, same as branches from nod.ai?

makslevental:
> What is the 123 in `linux-build-test-cpp-asserts-manylinux-v2-main-123`, is this the 123'rd cache of main ever created? Why do we need multiple caches of main?

123 is [`github.run_number`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs). We need multiple caches to prevent a race condition when a cache warming run overlaps with a `main` build.

> Do we even need caches of branches. Like, shouldn't they always be close enough to main to make the cache from main basically good enough?

Now this is on point - we don't, in fact @ScottTodd suggested (obliquely) that we just do away with caches on PRs and only cache on push to main. I keep fiddling with this PR trying to figure out why ccache is still missing/hitting on and off and I'm about to give up and do exactly this (remove caches for PRs).

> What about branches from forks, same as branches from nod.ai?

I'm always confused about this - you're the one that pointed out to me that since the linux script/action has _always_ successfully cached things, even for forks. So according to that logic, everything here should work for forks. But reading the [docs](https://github.com/actions/cache/blob/main/tips-and-workarounds.md#use-cache-across-feature-branches), which talk about how feature branches that share a common ancestor can share caches, it doesn't make sense to me - `main` is a common ancestor but in another repo? I mean maybe it really does just find the common ancestor hash and gate on that but I would've assumed there'd be stricter checking.

newling:
> We need multiple caches to prevent a race condition when a cache warming run overlaps with a `main` build.
Makes sense

> I keep fiddling with this PR trying to figure out why ccache is still missing/hitting on and off and I'm about to give up and do exactly this (remove caches for PRs).
Good to know. Caching just main might also mean llvm (via iree) is built less frequently in CI. Situation I have in mind: I make a PR before iree is bumped, and then I'm stuck with the cache for that PR, even after main has a cache with the bumped llvm. 

> I'm always confused about this - you're the one that pointed out to me that since the linux script/action has _always_ successfully cached things, even for forks. So according to that logic, everything here should work for forks. But reading the [docs](https://github.com/actions/cache/blob/main/tips-and-workarounds.md#use-cache-across-feature-branches), which talk about how feature branches that share a common ancestor can share caches, it doesn't make sense to me - `main` is a common ancestor but in another repo? I mean maybe it really does just find the common ancestor hash and gate on that but I would've assumed there'd be stricter checking.

Maybe the strictness needs to be introduced by other means. The first time you make a PR to IREE/torch-mlir, you need to request someone in the IREE team starts CI. If that's what you mean by 'stricter checking'. For my part I'm glad branches 'just work'.

ScottTodd:
> Now this is on point - we don't, in fact @ScottTodd suggested (obliquely) that we just do away with caches on PRs and only cache on push to main. I keep fiddling with this PR trying to figure out why ccache is still missing/hitting on and off and I'm about to give up and do exactly this (remove caches for PRs).

We have a few different workflows doing different things:
* https://github.com/iree-org/iree/blob/main/.github/workflows/pkgci_build_packages.yml uses `actions/cache` without much customization. That goes through some layers of scripts and Docker to write into a local cache dir manually and have github save/restore that full directory. Pushes _and_ PRs both read from and write to that, so https://github.com/iree-org/iree/actions/caches is typically at 20GB/10GB used. I've been considering changing that to only write to the cache on pushes, since the "post enable cache" step also takes 2m30s to upload - which is a bottleneck for starting dependent jobs. Using a cache hosted off of GitHub (GCP/Azure/AWS) would be better for storage limits and transfer speeds but would cost some money and require occasional maintenance.
* https://github.com/iree-org/iree/blob/main/.github/workflows/ci.yml uses `hendrikmuhs/ccache-action` in a few places, with no Docker or scripts in the way. I have those set to only "save" to the cache  on `push` events

I have some ideas for other experiments to try tracked at https://github.com/iree-org/iree/issues/18185.

makslevental:
I'm changing this PR to just not create caches for PRs. Let's see how that goes for a while.

makslevental:
For the future: `CCACHE_COMPILERCHECK=content` will prevent mac from getting cache hits even though it's supposed to help https://github.com/hendrikmuhs/ccache-action/issues/146.

PR #720, opened 2024-08-29
Set ObjectFifo pipeline as default and some cleanup

yzhang93:
As discussed previously, we are now setting pack-peel and objectfifo pipeline as default.

Some cleanup including:
- Remove cl flag `numCores` which was added for historical task, but is no longer used.
- Rename sample tests to be more clear about the pipelines being used.
- Just keep a minimal number of ci tests for MLIR-AIR path.


erwei-xilinx (approved):

build_tools/ci/run_matmul_test.sh
```diff
@@ -527,72 +527,23 @@ run_matmul_test \
...
-# The later is what a more vectorization friendly packing looks like so we are expected failing the test here.
-# We should fix this failure.
-# run_matmul_test \
-#    --name_prefix "failure_0" \
-#    --lhs_rhs_type "i32" \
-#    --acc_type "i32" \
-#    --m "52"  --n "52" --k "63" \
-#    --expect_compile_failure "1"
-
-# Example of a run with a group of 2+ matmuls. Currently this test is passed
```
newling, line 590: Why is this comment removed?

newling (commented):
LGTM just the one comment about the comment removed

build_tools/ci/run_matmul_test.sh
yzhang93: Thanks for catching it!

PR #724, opened 2024-08-29
Simplify print_ir_aie2xclbin.sh

makslevental:
We don't actually need vitis and xrt should  be sourced outside of this script like in the ci.


newling (approved):
my favourite color

yzhang93 (approved):

PR #730, opened 2024-08-30
SSH into runners

makslevental:
This PR adds the ability to SSH into runners using https://github.com/mxschmitt/action-tmate. The way it works is like this:

1. You want to debug a runner either because it failed or for the heck of it;
2. You go to the `Actions` tab and find the list of actions in the left colum and select Mac or Windows (**but not Linux** - see Notes below): 
    Screenshot from 2024-08-30 17-28-05
3. You then go to the far right and click the `Run workflow` dropdown:

    Screenshot from 2024-08-30 17-28-45
    Note, the `Branch:` dropdown - currently (because this PR hasn't landed) I can only do it for this branch. After this PR lands you'll be able to do it for any branch;
5. You check either one of the options; hopefully the descriptions are sufficient - if not ask in this PR and I'll update the descriptions accordingly;
6. The runner will run to completion or failure and then, **at the end**, start spamming

    Screenshot from 2024-08-30 17-32-06 Note, not indefinitely but 600 spams;
7. Copy-paste and you'll arrive at a tmux session that looks like this:

    Screenshot from 2024-08-30 17-48-55

    Hit `ctrl-c` and you'll be dropped into `tmux`:

    Screenshot from 2024-08-30 17-50-15

9. The job (not the whole action) exits/ends when you quit the session.

A few things to call out:

1. You need an ssh key that's associated with your GH username (e.g., the one you're using to commit/push); Note you don't need to do the `-i` part, `ssh` will find it automatically if it's under `~/.ssh`;
2. This isn't enabled for the Linux runner. Why? I only added this to the **free GitHub** runners, **not any nod runners**, because this is clearly an attack vector. So don't add it to any of our runners (at least not until/unless I do [this](https://github.com/mxschmitt/action-tmate/issues/202));
3. Yes, this works on Windows too but you land in `git-bash`/`mingw`/`msys`. If you want powershell (or cmd) then you can do `powershell` (or `cmd`);
4. If you don't know how to use `tmux` just refer to this [cheatsheet](https://tmuxcheatsheet.com/).

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -9,11 +9,6 @@ on:
     branches:
       - main
 
-  schedule:
-    # At minute 0 past every 6th hour. (see https://crontab.guru)
-    # this job is to keep the ccache cache warm
-    - cron: '0 */6 * * *'
-
 concurrency:
   # A PR number if a pull request and otherwise the commit hash. This cancels
   # queued and in-progress runs for the same PR (presubmit) or commit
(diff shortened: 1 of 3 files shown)


jtuyls (approved):
Nice!

PR #735, opened 2024-09-02
Remove lock analysis from stateful transform 

newling:
Assignment of lock IDs is already performed later, in amdaie-assign-lock-ids. Deduplicating.


jtuyls (approved):
LGTM

PR #736, opened 2024-09-03
Build and cache llvm

makslevental:
This PR splits up build _only in CI_ into two steps

1. Build LLVM
2. Build iree 

Why do this if it only affects CI? Because now we can also cache/save/upload the LLVM distro for each successful build thereby giving everyone (not just CI) a distro of LLVM they can use to build iree locally. This cuts down the number of files you have to compile by like 60% (from 7500 -> 2900). 

In case you didn't know, the CI artifacts are at the bottom of the summary page for any given [run](https://github.com/nod-ai/iree-amd-aie/actions/runs/10696630365):

github com_nod-ai_iree-amd-aie_actions_runs_10713474448 (1)

**EDIT**: in fact we actually need this for HSA (because ROCR requires an up-to-date Clang)[^1].

**EDIT2**: This PR now also adds powershell scripts for building on windows

[^1]: https://github.com/nod-ai/ROCR-Runtime/blob/amd-staging/runtime/hsa-runtime/core/runtime/blit_shaders/CMakeLists.txt#L71


build_tools/build_llvm.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
```
jtuyls, line 1:
```suggestion
#!/bin/bash
#
# Copyright 2024 The IREE Authors
#
# Licensed under the Apache License v2.0 with LLVM Exceptions.
# See https://llvm.org/LICENSE.txt for license information.
# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```

compiler/plugins/target/AMD-AIE/aie/AIEDialect.h
```diff
@@ -74,6 +74,8 @@ using DMAChannelDirAttr = mlir::iree_compiler::AMDAIE::DMAChannelDirAttr;
 using AIEArch = mlir::iree_compiler::AMDAIE::AIEArch;
 using AIEDevice = mlir::iree_compiler::AMDAIE::AMDAIEDevice;
 using AIEDeviceAttr = mlir::iree_compiler::AMDAIE::AMDAIEDeviceAttr;
+using AIETargetModel = mlir::iree_compiler::AMDAIE::AMDAIEDeviceModel;
+using TileID = mlir::iree_compiler::AMDAIE::TileLoc;
```
jtuyls, line 78: Why this change?

makslevental: it simplifies the `sed`s that were/are necessary for AIR which were apparently producing malformed source (I can't find the exact failed CI right now). It's weird because it should have nothing to do with prebuilt LLVM or not but it's a good change nonetheless (because it eliminates/simplifies the `sed`s) so I'm not thinking too hard about it.

.github/workflows/ci-linux.yml
```diff
@@ -77,13 +77,26 @@ jobs:
...
-          bash build_tools/ci/build_test_cpp.sh
+          bash build_tools/build_llvm.sh
+          rm -rf llvm-build
+          export llvm_install_dir=$PWD/llvm-install
+          bash build_tools/build_test_cpp.sh
 
       - name: Create artifacts
         if: ${{ !cancelled() }}
         run: |
+          pushd third_party/iree/third_party/llvm-project && llvm_sha_short=$(git rev-parse --short HEAD) && popd
```
ScottTodd, line 88: Nice one liner for getting the llvm sha. I've been considering using the llvm sha in cache keys, since that represents what _mostly_ affects build time. Getting a cache hit from the same LLVM commit is probably more useful than a cache hit from a more recent timestamp, and cache uploads could be limited to only one per llvm sha.

.github/workflows/ci-windows.yml
```diff
@@ -49,18 +47,15 @@ jobs:
...
-            clone \
-            --recursive \
-            --shallow-submodules \
-            --depth 1 \
-            -b $BRANCH_NAME $REPO_ADDRESS .
+          git init
+          git remote add origin $REPO_ADDRESS
+          git -c protocol.version=2 fetch --depth 1 origin $BRANCH_NAME
+          git reset --hard FETCH_HEAD
+          git -c submodule."third_party/torch-mlir".update=none -c submodule."third_party/stablehlo".update=none -c submodule."src/runtime_src/core/common/aiebu".update=none submodule update --init --recursive --depth 1 --single-branch -j 10
```
ScottTodd, line 58: Another option for this style of git operations is to use a script like https://github.com/iree-org/iree/blob/main/build_tools/scripts/git/update_runtime_submodules.sh. Sample usage here: https://github.com/iree-org/iree/blob/45132f5b1f1697e1f4d3450f0d6371c2ec49700a/.github/workflows/ci.yml#L209-L213. That does require checking out the repo with the script (without submodules) though.

With such a script, developers and workflows can share the same steps without copy/pasting all this.

.github/workflows/ci-windows.yml
```diff
@@ -86,17 +81,37 @@ jobs:
           key: ${{ env.CACHE_KEY }}
           restore-keys: windows-build-test-cpp-
 
+      - name: Hack IREE
+        shell: bash
+        run: |
+          # TODO(max): send IREE a fix for this
+          # target_compile_definitions may only set INTERFACE properties on IMPORTED
+          sed -i '114s/PRIVATE/INTERFACE/g' third_party/iree/compiler/src/iree/compiler/API/CMakeLists.txt
```
ScottTodd, line 89: 👀 
Is 114 a line number, pointing here: https://github.com/iree-org/iree/blob/45132f5b1f1697e1f4d3450f0d6371c2ec49700a/compiler/src/iree/compiler/API/CMakeLists.txt#L114? At a minimum it would be nice to make that more stable and not reliant on the file remaining unchanged.

I haven't noticed that "target_compile_definitions may only set INTERFACE properties on IMPORTED" error/warning upstream. Seems reasonable enough to patch it if it is causing issues downstream.

build_tools/build_llvm.ps1
```diff
@@ -0,0 +1,91 @@
...
+& ccache -z
+
+& cmake `
+  -GNinja `
+  -DCMAKE_BUILD_TYPE=Release `
+  -DCMAKE_INSTALL_PREFIX=$install_dir `
+  -DCMAKE_OBJECT_PATH_MAX=4096 `
+  -DCMAKE_EXE_LINKER_FLAGS_INIT=-fuse-ld=lld `
+  -DCMAKE_SHARED_LINKER_FLAGS_INIT=-fuse-ld=lld `
+  -DCMAKE_MODULE_LINKER_FLAGS_INIT=-fuse-ld=lld `
```
ScottTodd, line 59: FYI CMake 3.29+ has access to `-DCMAKE_LINKER_TYPE=LLD`: https://cmake.org/cmake/help/latest/variable/CMAKE_LINKER_TYPE.html

(We plan on bumping the CMake minimum in IREE and replacing `-DIREE_ENABLE_LLD=ON` with that official option)

.github/workflows/ci-linux.yml
```diff
@@ -77,13 +77,26 @@ jobs:
         run: |
           export cache_dir="${{ env.CACHE_DIR }}"
           export CCACHE_COMPILERCHECK="string:$(clang --version)"
-          bash build_tools/ci/build_test_cpp.sh
+          bash build_tools/build_llvm.sh
+          rm -rf llvm-build
+          export llvm_install_dir=$PWD/llvm-install
```
ScottTodd, line 82: > **EDIT**: in fact we actually need this for HSA (because ROCR requires an up-to-date Clang)

Do you need to build that from source yourself here, or could that come from official/upstream LLVM [nightly] builds? I guess https://llvm.org/builds/ doesn't include Windows (any longer), but https://apt.llvm.org/ exists for Linux (Debian/Ubuntu) at least.

ScottTodd (approved):
The extra build feels rather complicated to me, and it would be nice if iree-amd-aie didn't need to build LLVM itself in this way, but the approach you've taken looks good on the surface at least.

.github/workflows/ci-linux.yml
makslevental: i've considered something like that before but i think it's not worth it - like how often do you have multiple shas of llvm being "asked for" by various PRs that would avail themselves of a cache hit? with how small the free cache limit is (10Gb) i think the answer is basically never?

.github/workflows/ci-windows.yml
makslevental: in general i'm not a fan of hiding stuff in scripts that get used in CI ymls - at least until github let's you jump to those scripts like it does for names in cpp. the exception (at least for iree-amd-aie) being cmake configure/build, we would put into a script.

here in particular, this incantation isn't useful for devs because it's overly complex because it needs to work for forks (which i guess you well know from that example repo you showed me).

.github/workflows/ci-windows.yml
makslevental: yea that's the line - i'm guessing no one ever exercises the combo of "windows" + "byo llvm".

.github/workflows/ci-linux.yml
makslevental: in ROCR i manged to get away with 

https://github.com/nod-ai/ROCR-Runtime/blob/amd-staging/.github/workflows/ci-linux.yml#L49-L54

```
      - name: Setup Cpp
        uses: aminya/setup-cpp@v1
        with:
          compiler: llvm-18
          cmake: true
          ninja: true
```

which jams a compatible ubuntu release of llvm into an almalinux runner. to make it work i had to install `sudo` on there. here, where we're building on nod-shared-builder-whatever, which is almalinux, it doesn't work because you get a complaint that "root isn't in the sudoers file" 🤷. I didn't try too hard because really ROCR is a cherry on-top and what i wanted here was to be able to reuse the LLVM builds that currently get thrown away.

makslevental:
> The extra build feels rather complicated to me, and it would be nice if iree-amd-aie didn't need to build LLVM itself in this way

yea i mean this is what we discussed on slack earlier this week - i would love to be able to either 

1) build iree distro
2) build iree-amd-aie using iree distro
3) load 2 -> 3 (somehow/some plugin mechanism)

or 

1) build iree distro
2) build `iree-amd-aie-opt`, `iree-amd-aie-run-module` using iree distro

both paths would save us _a lot_ of CI and dev. but it's spot time right (full rebuild once every ~2 weeks when we bump iree if you're not sloppy and the bump is major) and amortized it's not a lot of time so for now shaving 60% instead of 90% is good enough. revisit later 🤷

PR #737, opened 2024-09-03
[CanonicalizeDoublyStridedOp] Fix single dimension folder

newling:
I'm starting to look deeper into creating links when the circular_dma_cpy_nd's being linked are not mutually 'contiguous'. Noticed that Attributes become Constant ops unnecessarily in `create-logical-objectfifo-link`, so fixed this by removing ` getValueOrCreateConstantIndexOp`. Then noticed that test `canonicalize_doubly_strided_op.mlir` was failing, and this was because `.getTargetStrides` was being used where `getTargetMixedStrides` should've been, so fixed that. Also removed a redundant canonicalization call (print-ir-after-all showed 2 contiguous canonicalizations)


Abhishek-Varma (approved):
LGTM. The CI seems to be failing because of the new update to the Github workflow script perhaps ?

jtuyls (approved):
LGTM

PR #738, opened 2024-09-03
Fix CI for forks

makslevental:
♥️ @newling


jtuyls (approved):
LGTM, thanks for the quick fix!

makslevental:
Note the README doesn't need to be updated because a clone from this repo (as tested by the previous CI yml) works fine

PR #739, opened 2024-09-03
[CreateLogicalObjFifoLink] Refactor unsafe walk for non-zero offset removal

jtuyls:
In a larger change to start extracting `aie.flow` ops from `AMDAIEStatefulTransform`, I ran into a segfault issue inside the walk to remove non-zero offsets. I extracted the fix from that effort to reduce the scope of that PR.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECreateLogicalObjectFifoLink.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECreateLogicalObjectFifoLink.cpp
@@ -34,7 +34,6 @@ namespace mlir::iree_compiler::AMDAIE {
 template <CopyOpOperateOn OperateOn>
 LogicalResult checkForContiguousAccessPatterns(
     ArrayRef<std::pair<DoublyStridedCopyOpInterface, int64_t>> stridedOps) {
-
   for (auto &&[i, stridedOpAndOffset] : llvm::enumerate(stridedOps)) {
     DoublyStridedCopyOpInterface stridedOp = stridedOpAndOffset.first;
     std::optional<int64_t> extent;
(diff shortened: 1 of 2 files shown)


newling (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECreateLogicalObjectFifoLink.cpp
```diff
@@ -161,8 +159,30 @@ LogicalResult createLogicalObjectFifoLink(
...
+    newLinkOps.push_back(linkOp);
+  }
+  return success();
+}
+
+LogicalResult discardLinkNonZeroOffsets(RewriterBase &rewriter,
+                                        AMDAIE::LogicalObjectFifoLink linkOp) {
+  for (Value input : linkOp.getIns()) {
+    if (auto stridedOp = dyn_cast<AMDAIE::DoublyStridedCopyOpInterface>(
+            input.getDefiningOp())) {
```
Abhishek-Varma, line 173: Here and below this should've been `dyn_cast_or_null` because `getDefiningOp()` is being used.

Perhaps can be addressed in the `aie.flow` extraction PR? @jtuyls

jtuyls: Thanks for pointing this out. I went through the codebase for all cases and created a separate PR: https://github.com/nod-ai/iree-amd-aie/pull/744

Abhishek-Varma: Thank you for the amazing cleanup Jorn!

PR #744, opened 2024-09-04
Use dyn_cast_if_present and isa_and_present with getDefiningOp

jtuyls:
- Update `dyn_cast` to `dyn_cast_if_present` everywhere getDefiningOp is used
- Update `isa` to `isa_and_present ` everywhere getDefiningOp is used
- Update `dyn_cast_or_null` to `dyn_cast_if_present` because the former is deprecated (see below)
- Update `isa_and_nonnull` to `isa_and_present` because although not deprecated according to comments, the former calls the latter under the hood and to be consistent with `dyn_cast_if_present`

`dyn_cast_or_null` is deprecated (https://llvm.org/doxygen/Casting_8h_source.html):
> // Forwards to dyn_cast_if_present to avoid breaking current users. This is
// deprecated and will be removed in a future patch, use
// cast_if_present instead.
template <class X, class Y> auto dyn_cast_or_null(const Y &Val) {
  return dyn_cast_if_present<X>(Val);
}

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.cpp
@@ -83,7 +83,7 @@ LogicalResult CoreOp::verify() {
 }
 
 TileOp CoreOp::getTileOp() {
-  return dyn_cast<TileOp>(getTile().getDefiningOp());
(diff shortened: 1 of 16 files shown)


Abhishek-Varma (approved):
WOAH!

Thanks for the quick fix and cleaning up the entire codebase with this @jtuyls !

makslevental:
What does `if_present`actually mean?

PR #745, opened 2024-09-04
Refactor: lower DMAs to connection + controlcode DMA ops

jtuyls:
In the effort to enable control packets, a first step is to extract `aie.flow` logic from `AMDAIEStatefulTransform` to enable packet flow ops on `amdaie`. This PR adds `amdaie.connection` which resembles `aie.flow` and lowers circular DMA ops (`amdaie.circular_dma_cpy_nd`) to a connection (`amdaie.connection`) + control code (`amdaie.npu.circular_dma_cpy_nd`). Note that this also benefits the effort to move all/most DMA/PM/... configuration logic into the control code instead of the xclbin.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaCSE.cpp
```diff
@@ -0,0 +1,301 @@
...
+// sub-expression elimination on AMDAIE DMA operations within the same region.
+// This eliminates identical circular DMA operations until only one exists. This
+// is different from the main upstream  CSE behaviour on ops with the `Pure`
+// trait in that `Pure` ops with no users will be eliminated, while DMA ops
+// without users are kept around.
+//
+// The implementation is based on and largely copied from:
+// https://github.com/llvm/llvm-project/blob/main/mlir/lib/Transforms/CSE.cpp
+//
+// TODO(jornt): It might be worthwile to try using the main upstream CSE pass,
```
jtuyls, line 19: I spent some time trying to use upstream CSE instead (using Pure/MemoryEffects), but I wasn't able to get the same result. From what I am seeing, CSE would need changes to avoid accommodate this, but let me know if you see a way to achieve the same result with upstream CSE and/or a better way of accomplishing this.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIETraits.h
```diff
@@ -0,0 +1,22 @@
...
+
+#ifndef IREE_COMPILER_AMDAIE_TRAITS_H_
+#define IREE_COMPILER_AMDAIE_TRAITS_H_
+
+#include "mlir/IR/OpDefinition.h"
+
+namespace mlir::OpTrait::iree_compiler::AMDAIE {
+
+template <typename ConcreteType>
+class CircularDmaOp : public OpTrait::TraitBase<ConcreteType, CircularDmaOp> {
```
makslevental, line 15: Just curious: why a trait rather than an attribute?

jtuyls: Isn't this what `Traits` are meant for, to describe common properties of operations? I am trying to do something similar like `MemoryEffects`, but for our DMA operations. And ideally, I would like to find a good way to annotate the operands (with attributes?) as the resources the trait relates to, for example:

```
MyOp [CircularDmaOp] {
  ins Arg<Index, "", DmaRead>:$target,
       Arg<Index, "", DmaWrite>:$source
}
```

Then, it would be possible to create more general CSE logic to check whether another DMA uses the same resources and in what way.

Something like this seems to exist already for memory effects, but I don't how to use it, I don't see any code getting generated for it either when I tried using it and I don't see CSE leveraging it either..

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/dma_cse.mlir
```diff
@@ -0,0 +1,71 @@
...
+    %c1 = arith.constant 1 : index
+    %c2 = arith.constant 2 : index
+    %c6 = arith.constant 6 : index
+    amdaie.workgroup {
+      %0 = amdaie.connection(%arg0, %arg1) : (!amdaie.logicalobjectfifo<memref<1x1x8x16xi32>>, !amdaie.logicalobjectfifo<memref<8x16xi32, 1>>)
+      %1 = amdaie.connection(%arg0, %arg1) : (!amdaie.logicalobjectfifo<memref<1x1x8x16xi32>>, !amdaie.logicalobjectfifo<memref<8x16xi32, 1>>)
+      amdaie.controlcode {
+        scf.for %arg2 = %c1 to %c6 step %c2 {
+          %2 = amdaie.npu.circular_dma_cpy_nd %0([] [] [], [] [] [])
+          %3 = amdaie.npu.circular_dma_cpy_nd %0([] [] [], [] [] [])
```
newling, line 23: This doesn't look like cse to me. Is this like 

```
copy(dst, src)
copy(dst, src)
```

becoming 
```
copy(dst, src)
```
Because the first copy is dead? Seems more like a liveness analysis thing to me, if so. CSE is something in value semantics, no? 

Why is the scf.forall retained, is this a future optimization? i.e. %5 and %4 are the same aren't they (imagine unrolling).

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
```diff
@@ -374,7 +374,160 @@ def AMDAIE_NpuDmaCpyNdOp: AMDAIE_Op<"npu.dma_cpy_nd", [
...
+    }
+    
+    // Return the input circular dma copy operation.
+    ConnectionOp getConnectionOp() {
+      return dyn_cast<ConnectionOp>(getConnection().getDefiningOp());
+    }
+
+    // Return the source memref type. This is retrieved using information from
+    // the input DMA operation.
+    MemRefType getSourceMemrefType() {
```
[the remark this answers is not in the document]
jtuyls: We already have an interface similar to `Tensor_OpWithOffsetSizesAndStrides`: `DoublyStridedOpInterface`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/dma_cse.mlir
newling: > No, because (in our case) the copy is circular so the DMAs don't need to be programmed twice.
Makes sense. 

> Will be removed if you run cse after this.

Ok good. So maybe add extra lines

// CHECK-NOT: circular_dma_cpy_nd
// CHECK: amdaie.end

jtuyls: > Ok good. So maybe add extra lines
> // CHECK-NOT: circular_dma_cpy_nd
> // CHECK: amdaie.end

Done

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
```diff
@@ -374,7 +374,160 @@ def AMDAIE_NpuDmaCpyNdOp: AMDAIE_Op<"npu.dma_cpy_nd", [
   }];
 
   let hasCanonicalizer = 1;
+}
+
+def AMDAIE_NpuCircularDmaCpyNdOp: AMDAIE_Op<"npu.circular_dma_cpy_nd", [
+      AMDAIE_CircularDmaOp, AttrSizedOperandSegments, DoublyStridedOpInterface]>,
+    Results<(outs Index)> {
+  let summary = "The Npu uController's circular dma operator";
+  let description = [{TODO
```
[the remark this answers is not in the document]
jtuyls: Thanks, done

newling (approved):
I haven't gone over this very thoroughly, but overall looks good to me.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIETraits.h
makslevental: > Isn't this what Traits are meant for, to describe common properties of operations?

It's functionality rather than properties though - like it's cpp level stuff rather then IR/compiler/domain level stuff. Now yes, naturally sometimes those things go hand-in-hand e.g. TilingInterface/Traits naturally apply only to a certain kind of op (loopy things) but the Trait itself is for giving connections/hooks for how something happens in cpp land - how the IR is rewritten by cpp methods.

> similar to MemoryEffects

yea so I'm guessing the way this works is somewhere in the CSE system there's a check for like RAW/WAR/WAW or whatever so it's cpp level but I think it's a poor use, specifically because

> And ideally, I would like to find a good way to annotate the operands

traits aren't visible in the IR itself (they're visible in ODS)

> Something like this seems to exist already for memory effects, but I don't how to use it, I don't see any code getting generated for it when I tried using it and I don't see CSE leveraging it either..

I'll do some digging some time tomorrow

makslevental: btw as always this isn't biblical scripture/dogma - just my intuition

tests/samples/matmul_pack_peel_objectfifo.mlir
```diff
@@ -1,4 +1,4 @@
...
+// RUN: iree-opt --pass-pipeline="builtin.module(fold-memref-alias-ops,iree-amdaie-pack-to-dma,air-copy-to-dma,iree-amdaie-air-dma-to-amdaie-dma,iree-amdaie-insert-cores,cse,iree-amdaie-localize-logicalobjectfifo,iree-amdaie-distribute-cores-and-objectfifos,cse,canonicalize,iree-amdaie-dma-to-circular-dma,func.func(iree-amdaie-create-aie-workgroup),cse,iree-amdaie-dma-cse,iree-amdaie-canonicalize-doubly-strided-op,iree-amdaie-flatten-logicalobjectfifo,iree-amdaie-access-to-acquire-release,cse,canonicalize,iree-amdaie-dma-loop-subsumption,cse,canonicalize,iree-amdaie-assign-npu-dma-bd-ids,iree-amdaie-controlcode-loop-unroll,cse,canonicalize,iree-amdaie-dma-cse,iree-amdaie-create-logical-objectfifo-link,iree-amdaie-canonicalize-doubly-strided-op,canonicalize,iree-amdaie-canonicalize-npu-dma-cpy-nd,canonicalize,iree-amdaie-sink-into-core,canonicalize,iree-amdaie-lower-to-aie,canonicalize)" --split-input-file %s | FileCheck %s
```
Abhishek-Varma, line 1: Orthogonal to this PR: Do we still need this `RUN` command in the lit test ?

Since we're adding `DmaCSE` in `Passes.cpp` the following lit test should only suffice :-
```
iree-compile --iree-hal-target-backends=amd-aie --compile-to=executable-sources %s | iree-opt --pass-pipeline="builtin.module(hal.executable(hal.executable.variant(iree-hal-translate-target-executable-variants{target=amd-aie})))" --iree-amdaie-lower-to-aie-pipeline=objectFifo --iree-amdaie-tile-pipeline=pack-peel | FileCheck %s
```

Also, I see `DmaCSE` was included thrice in `Passes.cpp` but twice here - so the lit test `RUN` command above will also help avoid such discrepancies.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECoreLoopUnroll.cpp
```diff
@@ -25,20 +25,17 @@ LogicalResult coreLoopUnroll(RewriterBase &rewriter, AMDAIE::CoreOp coreOp) {
     llvm::SmallDenseSet<unsigned> depths;
     for (auto acqOp :
          forOp.getBody()->getOps<AMDAIE::LogicalObjectFifoAcquire>()) {
-      auto stridedOp = dyn_cast_if_present<DoublyStridedCopyOpInterface>(
-          acqOp.getDma().getDefiningOp());
-      if (!stridedOp) {
-        acqOp.emitOpError()
-            << "should operate on a doubly strided copy operation for core "
-               "loop unrolling to behave correctly";
+      auto copyOp = dyn_cast_if_present<CopyOpInterface>(acqOp.getDma().getDefiningOp());
```
Abhishek-Varma, line 28: Why we didn't use `DoublyStridedCopyOpInterface` here?

Is it because we now have `AMDAIE_NpuCircularDmaCpyNdOp` which is not copy-like ? (The only difference between `DoublyStridedCopyOpInterface` and `DoublyStridedOpInterface` is the presence of `CopyOpInterface` in the former)

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECreateAIEWorkgroup.h
```diff
@@ -217,6 +217,15 @@ class WorkgroupBuilder {
                       Block::iterator controlCodeEnd);
 
  private:
+  /// Build function that handles `amdaie.dma_cpy_nd` by converting it into a
+  /// workgroup DMA with potentially corresponding control code.
+  LogicalResult buildForCircularDmaCpyNdOp(AMDAIE::CircularDmaCpyNdOp dmaOp,
```
Abhishek-Varma, line 222: `amdaie.dma_cpy_nd` -> `amdaie.circular_dma_cpy_nd`

Abhishek-Varma (approved):
LGTM. Only a few nits/question(s) to address.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECoreLoopUnroll.cpp
jtuyls: Because the input is not a `CircularDmaCpyNdOp` anymore, but a `ConnectionOp`. Ìn general, this pass does not expect a strided op to be its input, just an op with a source and target (`CopyOpInterface`)

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIETraits.h
jtuyls: > traits aren't visible in the IR itself (they're visible in ODS)

That's fine, I don't think I need it to be

tests/samples/matmul_pack_peel_objectfifo.mlir
jtuyls: I just wanted to keep an example around with all (critical) transformations as a pass pipeline. We could discuss whether that's worthwhile. Added it three times here as well for now.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECreateAIEWorkgroup.h
jtuyls: Done

PR #752, opened 2024-09-05
Pytest and numerical failure diagnostics

newling:


makslevental:
I respect the hustle

build_tools/ci/cpu_comparison/output_comparer.py
```diff
@@ -2,65 +2,147 @@
...
+        return summary
+
+    def getSummaryString(self, max_discrepancies_to_show):
+        summary = self.getBaseString()
+        summary += self.getDiscrepenciesString(max_discrepancies_to_show)
+        summary += self.getSlicesString()
+        return summary
+
+
+def compare(npy_baseline1, npy_aie1, rtol, atol, max_discrepancies_to_show=50):
```
yzhang93, line 129: Why not use "npy_baseline" and "npy_aie"?

yzhang93 (approved):
Nice feature for debugging! Thanks!

build_tools/ci/cpu_comparison/output_comparer.py
newling: I can't remember the history here, but I've dropped the '1's.

Issue #762, opened 2024-09-10
[Convolution e2e] Numerical issue with i32 single AIE core


newling:
I will post more information but I have 1 branch where convolution is numerically compiles and runs and is correct (see title) in about 70% of the output indices. This task is just a stub to track this issue (I plan to get back to after packing works).

newling:
Fixed by https://github.com/nod-ai/iree-amd-aie/pull/765

PR #765, opened 2024-09-12
Fix `AIEX::NpuDmaMemcpyNdOp::getOffsetInBytes`

makslevental:
Current version doesn't take into account striding (effectively assumes unit stride for all dims). Not sure how this hasn't come up before...

Thanks @erwei-xilinx


jtuyls (approved):
LGTM. Is this/could this be causing the numerics issue you were seeing in batch matmul @yzhang93 ?

yzhang93:
Thanks @makslevental and @erwei-xilinx! This should fix the numeric issue we had for batch matmul size 2.

newling:
This also fixes https://github.com/nod-ai/iree-amd-aie/issues/762

PR #766, opened 2024-09-12
Don't download whole LLVM distro on runners

makslevental:
You only need `lld` not all of the LLVM distro to run our current tests[^1]. Prior this was being provided by `iree-lld` through `set(IREE_LLD_BINARY "$<TARGET_FILE:${IREE_LLD_TARGET}>")` in [iree_llvm.cmake](https://github.com/iree-org/iree/blob/c7351e1d62e92b711dfd582869dd640a3a9a9569/build_tools/cmake/iree_llvm.cmake#L84C3-L84C59) (inside of `macro(iree_llvm_configure_bundled)`) but now that we're building LLVM separately it is not. Maybe a bug in IREE's CMake (maybe not :shrug:) but we should be bundling everything a user needs to run our tests anyway.

This also resolves the annoying issue of downloading artifacts on Windows runners taking foreeeeeeeeeeeeeeeeeeeeeever.

Note, this was slightly annoying to debug because "smoke test" doesn't currently handle `-v` and without that `-v` you do not get any information about _why_ the smoke test failed if it failed in the shell out (cc @newling).

[^1]: This might change, at which point I'll revisit.


jtuyls (approved):
LGTM

ScottTodd (approved):

newling:
> Note, this was slightly annoying to debug because "smoke test" doesn't currently handle -v and without that -v you do not get any information about why the smoke test failed if it failed in the shell out (cc @newling).

Oh damn I should've thought of that. I'd like to move that to pytest eventually but not totally trivial so for I'll make a PR putting the smoke test last in the sequence so that this doesn't happen. Sometime today (unless someone beats me to it).

PR #770, opened 2024-09-12
[ObjectFifoStatefulTransform] Fix memory space. 

newling:
This patch (along with a modification to the tiling on main) enables int32, single core convolution to compile. It fixes the error

```
<unknown>:0: error: different memory spaces specified for source type 'memref<16xi32, 1 : i32>' and result memref type 'memref<1x1x4x4xi32, 2 : i32>'
<unknown>:0: note: see current operation: %52 = "memref.reinterpret_cast"(%7) <{operandSegmentSizes = array<i32: 1, 0, 0, 0>, static_offsets = array<i64: 0>, static_sizes = array<i64: 1, 1, 4, 4>, static_strides = array<i64: 16, 16, 4, 1>}> : (memref<16xi32, 1 : i32>) -> memref<1x1x4x4xi32, 2 : i32>
```

I guess this is another fix for what was probably undone in: https://github.com/nod-ai/iree-amd-aie/pull/692

If I understood the original logic better I could probably provide a less clunky fix. This PR gets the memspace from the row that the buffer is on. It assumes row 1 is memory tile, and rows 2+ are AIE cores. 

Last comment: the test updated makes more sense now, before it was doing math on memspace 1 memrefs.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/aie/AMDAIEObjectFifoStatefulTransform.cpp
+++ b/compiler/plugins/target/AMD-AIE/aie/AMDAIEObjectFifoStatefulTransform.cpp
@@ -330,18 +330,18 @@ SmallVector<BufferOp> createBuffers(OpBuilder &builder,
   SmallVector<BufferOp> buffers;
   if (deviceModel.isShimTile(tile.getCol(), tile.getRow())) return buffers;
   auto fifoType = cast<AIEObjectFifoType>(createOp.getElemType());
-  auto oldElemType = cast<MemRefType>(fifoType.getElementType());
-  auto shape = oldElemType.getShape();
-  auto elType = oldElemType.getElementType();
-  int memSpace;
-  if (tile.getRow() == 0)
-    memSpace = 0;
-  else if (tile.getRow() == 1)
-    memSpace = 1;
-  else
-    memSpace = 2;
-  Attribute memSpaceAttr = builder.getI32IntegerAttr(memSpace);
-  auto elemType = MemRefType::get(shape, elType, AffineMap{}, memSpaceAttr);
+  MemRefType fifoElemType = cast<MemRefType>(fifoType.getElementType());
+  ArrayRef<int64_t> shape = fifoElemType.getShape();
+  Type elType = fifoElemType.getElementType();
+
+  Attribute memSpace = [&]() -> Attribute {
+    if (!fifoElemType.getMemorySpace()) return Attribute{};
+    // row 0: memspace 0,
+    // row 1: memspace 1,
+    // else : memspace 2.
+    return builder.getI32IntegerAttr(std::min<int>(2, tile.getRow()));
+  }();
+  MemRefType elemType = MemRefType::get(shape, elType, AffineMap{}, memSpace);
 
   for (int ofElemIndex = 0; ofElemIndex < numBuffers; ofElemIndex++) {
     auto buff = builder.create<BufferOp>(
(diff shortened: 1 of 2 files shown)


compiler/plugins/target/AMD-AIE/aie/AMDAIEObjectFifoStatefulTransform.cpp
```diff
@@ -330,7 +330,19 @@ SmallVector<BufferOp> createBuffers(OpBuilder &builder,
...
+  ArrayRef<int64_t> shape = fifoElemType.getShape();
+  Type elType = fifoElemType.getElementType();
+
+  Attribute memSpace = [&]() -> Attribute {
+    if (!fifoElemType.getMemorySpace()) return Attribute{};
+    // row 0: memspace 0,
+    // row 1: memspace 1,
+    // else : memspace 2.
+    return builder.getI32IntegerAttr(std::min<int>(2, tile.getRow()));
+  }();
```
makslevental, line 343: just curious: why do it like this? `Attributes are reference types already so you can just do 

```cpp
Attribute memSpace;
if (fifoElemType.getMemorySpace())
  memSpace = builder.getI32IntegerAttr(std::min<int>(2, tile.getRow()));
```

and save yourself the stack allocation for the lambda...

Also shouldn't it be

```cpp
  Attribute memSpace;
  if (fifoElemType.getMemorySpace())
    memSpace = fifoElemType.getMemorySpace();
```

instead of hard-coded `2`?

newling: I could. No reason I've done it this way other than it started as a 10-liner lambda until I whittled it down to this. And my mental stack for tracking uninitialized variables across multiple lines is especially low...

> and save yourself the stack allocation for the lambda...
I hope 'the compiler' in lines this, but tbh I'm not sure

makslevental: i mean it does inline - you get an inlined struct allocation for the lambda. it's not big but it's probably the size of `std::funciton` or something.

makslevental: https://en.cppreference.com/w/cpp/language/lambda

> The lambda expression is a prvalue expression of unique unnamed non-union non-aggregate class type, known as closure type, which is declared (for the purposes of ADL) in the smallest block scope, class scope, or namespace scope that contains the lambda expression.

compiler/plugins/target/AMD-AIE/aie/AMDAIEObjectFifoStatefulTransform.cpp
```diff
@@ -330,7 +330,19 @@ SmallVector<BufferOp> createBuffers(OpBuilder &builder,
   SmallVector<BufferOp> buffers;
   if (deviceModel.isShimTile(tile.getCol(), tile.getRow())) return buffers;
   auto fifoType = cast<AIEObjectFifoType>(createOp.getElemType());
-  auto elemType = cast<MemRefType>(fifoType.getElementType());
+  MemRefType fifoElemType = cast<MemRefType>(fifoType.getElementType());
+  ArrayRef<int64_t> shape = fifoElemType.getShape();
+  Type elType = fifoElemType.getElementType();
+
+  Attribute memSpace = [&]() -> Attribute {
```
jtuyls, line 337: If I look at the tests, I don't see any memory space attributes ? I don't see how that happens, because here you do assign memory space attributes?

newling: The tests are not lowered through iree-amd-aie, our tiling strategy introduces these memory spaces. I haven't added a test for this, I guess I should. 

Here I'm only assigning a memory space attribute sometimes. It's just correcting the memory space if there is already one present.

jtuyls: I see. To avoid the hardcoding below based on row number, I think we should instead just remove all memory space attributes. That was the intention of the logic before (although that logic standardized on memSpace == 1).

newling: That would definitely be nicer. A bit more work, but I'll give it a try.

newling:
Closing while I work on a better solution

PR #772, opened 2024-09-12
[CI] reorder tests in linux CI to make debug easier

newling:
This PR makes the order of tests on linux

```
E2E comparison of AIE to llvm-cpu
E2E correctness matmul test
Printing IR from aie2xclbin
Smoke E2E comparison flag test
```

'smoke' test last because it's testing with verbose=0, not a good test to run first because diagnosing unrelated failures is impossible.


makslevental (approved):
Thanks

PR #773, opened 2024-09-12
Decouple packToDma pass from AIR pipeline

yzhang93:
It's not necessary to go from `LinalgExt::PackOp` -> `air::DmaMemcpyNdOp` -> `AMDAIE::DmaCpyNdOp`. With `packToDma` pass decoupled from AIR pipeline, we can refactor the codes to convert directly from PackOp to AMDAIE::DmaCpyNdOp (will address in the next PR).

For AIR pipeline, `DecomposeLinalgExtPackUnPackToAIR` can be used for both pack-peel and pad-pack pipelines. It doesn't have use `packToDma` pass. As a consequence, `AMDAIECanonicalizeDma` pass can also be eliminated, since similar logic already exists in MLIR-AIR and `AMDAIECanonicalizeDoublyStridedOp` for objectFifo.


erwei-xilinx:
Makes sense to me re AIR. Thanks!

MaheshRavishankar (approved):

PR #776, opened 2024-09-13
Refactor packToDma pass to work directly with objectFifo

yzhang93:
It's not necessary to go from `LinalgExt::PackOp -> air::DmaMemcpyNdOp -> AMDAIE::DmaCpyNdOp`. With packToDma pass decoupled from AIR pipeline in https://github.com/nod-ai/iree-amd-aie/pull/773, we can refactor the codes to convert directly from PackOp to AMDAIE::DmaCpyNdOp. As a consequence, `AIRDmaToAMDAIEDma` pass is eliminated.

I'm also thinking maybe it's good to rename `AMDAIEPackToDma` to `AMDAIEPackToLogicalObjectFifoDma`?


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -583,9 +583,7 @@ void addAMDAIEObjectFifoLoweringPasses(OpPassManager &passManager) {
   passManager.addPass(createEraseHALDescriptorTypeFromMemRefPass());
   passManager.addPass(memref::createFoldMemRefAliasOpsPass());
   passManager.addPass(createAMDAIEPackToDmaPass());
-  passManager.addPass(xilinx::air::createCopyToDmaPass());
```
yzhang93, line 586: In the current implementations, we don't have copy ops used for data movement, so there's no need to keep such pass. In the future, we can implement a new pass to convert copy ops to AMDAIE dma ops if needed.

jtuyls (approved):
LGTM

> I'm also thinking maybe it's good to rename AMDAIEPackToDma to AMDAIEPackToLogicalObjectFifoDma?

I am ok with the name as it is.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
yzhang93: Sorry... I thought you have switched to use pack based tiling strategy for convolution? I can write a pass to convert copy to dma if it's really needed. Let me know what you decide.

PR #779, opened 2024-09-16
[DmaOpInterface] Add static size retrieval method

jtuyls:
Add static size retrieval methods to the `AMDAIEDmaOpInterface`.

Needed as part of a larger effort to remove `StatefulTransform`: https://github.com/jtuyls/iree-amd-aie/tree/delete-stateful-transform

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDmaOpInterface.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDmaOpInterface.cpp
@@ -4,6 +4,8 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+#include <numeric>
+
 #include "iree-amd-aie/IR/AMDAIEDmaOpInterface.h"
 
 #include "iree-amd-aie/IR/AMDAIEAttrs.h"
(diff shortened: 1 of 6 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEDmaOpInterface.cpp
```diff
@@ -76,6 +78,23 @@ std::optional<int64_t> getStaticExtent(DoublyStridedOpInterface op) {
...
+  } else if constexpr (OperateOn == CopyOpOperateOn::Target) {
+    sizes = op.getTargetMixedSizes();
+  } else {
+    assert(false && "Function can only operate on Source or Target");
+  }
+  if (sizes.size() == 0) return 0;
+  std::optional<SmallVector<int64_t>> staticSizes = getConstantIntValues(sizes);
+  if (!staticSizes) return std::nullopt;
+  return std::accumulate(staticSizes->begin(), staticSizes->end(), 1,
+                         std::multiplies<>());
```
Abhishek-Varma, line 95: Why do it this way?

We can take the ShapedType from Source/Target and simply `getNumElements()` (can't recall the exact function name but I believe it's this).

Abhishek-Varma: Aah! Okay. Makes sense. Thanks.

Abhishek-Varma (approved):

PR #780, opened 2024-09-16
Lowering of linalg.copy to dmas (via iree.linalg_ext.pack) 

newling:
[UPDATED]

Support for lowering of linalg.copy was removed from the objectFifo pipeline in https://github.com/nod-ai/iree-amd-aie/pull/776 as there is nothing on the main branch objectFifo pipeline which needs this. I'm working on a new convolution pipeline which uses padding (L3 -> L2) and packing (L2 -> L1) for the objectFifo pipeline, so I'd like it back. This lazy PR goes linalg.copy -> iree.linalg_ext.pack -> 'whatever is next', and takes advantage of the existing lowering from iree.linalg_ext.pack


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPackToDma.cpp
```diff
@@ -321,6 +321,38 @@ void AMDAIEPackToDmaPass::runOnOperation() {
   MLIRContext *context = &getContext();
   IRRewriter rewriter(context);
 
+  // Convert all linalg.copy to iree_linalg_ext.pack ops.
+  //
+  // This is simple, the iree_linalg_ext.pack is just an identity pack
```
newling, line 326: (this comment makes no sense)

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPackToDma.cpp
```diff
@@ -321,6 +321,38 @@ void AMDAIEPackToDmaPass::runOnOperation() {
...
+  // operation.
+  //
+  // By doing this, we can bootstrap the work done for lowering the pack op to
+  // dmas: easier than writing a new pass to convert linalg.copy to dmas
+  // directly.
+  auto walkCopyToPackResult =
+      getOperation()->walk([&rewriter](linalg::CopyOp copyOp) -> WalkResult {
```
yzhang93, line 333: This pass happens after bufferization, so it has to be a memref::CopyOp?

yzhang93:
As discussed offline, I'm okay with the idea to convert memref.copy/linalg.copy to iree_linalg_ext.pack ops as an intermediate step and a stop-gap solution. I'd not add this as a separate pass though, because the IR after such pass could be misleading (all the copy ops are converted to the pack ops without any unpack ops).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEPackToDma.cpp
yzhang93: Also please add some tests.

newling: Before comprehensive bufferization:

```
 %16 = linalg.copy ins(%15 : tensor<1x1x4x4xi32>) outs(%extracted_slice_1 : tensor<1x1x4x4xi32>) -> tensor<1x1x4x4xi32>
```

After comprehensive bufferization:

```
linalg.copy ins(%alloc_3 : memref<1x1x4x4xi32, 1 : i32>) outs(%subview_1 : memref<1x1x4x4xi32, strided<[9216, 768, 64, 1], offset: ?>, #hal.descriptor_type<storage_buffer>>)
```

So it's not a memref.copy.

newling:
> As discussed offline, I'm okay with the idea to convert memref.copy/linalg.copy to iree_linalg_ext.pack ops as an intermediate step and a stop-gap solution. I'd not add this as a separate pass though, because the IR after such pass could be misleading (all the copy ops are converted to the pack ops without any unpack ops).

We already have pack ops without unpack ops, don't we? Only the result will have an unpack op.

yzhang93:
> We already have pack ops without unpack ops, don't we? Only the result will have an unpack op currently.

Yes, but what I really meant is the copy ops for L2->L3 should be something like unpack op instead of a pack op if we compare this to the fully pack-based pipeline.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertToDma.cpp
```diff
@@ -304,23 +305,49 @@ LogicalResult rewriteAsDma(PackOrUnpackOp op, IRRewriter &rewriter) {
...
 };
 
-void AMDAIEPackToDmaPass::runOnOperation() {
+void AMDAIEConvertToDmaPass::runOnOperation() {
   MLIRContext *context = &getContext();
   IRRewriter rewriter(context);
 
+  // Convert all linalg.copy to iree_linalg_ext.pack ops. We then bootstrap the
+  // work done for lowering the pack op to dmas as the next step.
+  auto walkCopyToPackResult =
```
yzhang93, line 327: I think it's better/cleaner to create a separate function for linalg.copy conversion.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertToDma.cpp
```diff
@@ -304,23 +305,49 @@ LogicalResult rewriteAsDma(PackOrUnpackOp op, IRRewriter &rewriter) {
...
+      getOperation()->walk([&rewriter](linalg::CopyOp copyOp) -> WalkResult {
+        (void)rewriter;
+        if (copyOp.getNumOperands() != 2 || copyOp.getNumResults() != 0) {
+          copyOp.emitOpError()
+              << "has " << copyOp.getNumOperands() << " operands and "
+              << copyOp.getNumResults()
+              << " results. It must have 2 operands and 0 results to convert "
+                 "to an iree.linalg_ext.pack op, here.";
```
yzhang93, line 335: Nit: don't have to use "here".

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -412,10 +412,44 @@ def AMDAIEPackAndTranspose :
...
 
-def AMDAIEPackToDma :
-  Pass<"iree-amdaie-pack-to-dma", ""> {
-  let summary = "Convert pack/unpack ops to AMDAIE DMA ops operating on logical objectFifos.";
-  let constructor = "mlir::iree_compiler::AMDAIE::createAMDAIEPackToDmaPass()";
+def AMDAIEConvertToDma :
+  Pass<"iree-amdaie-convert-to-dma", ""> {
+  let summary = "Convert linalg ops to AMDAIE DMA ops on logical objectFifos.";
+  let description = [{
+   Converts/lowers linalg.copy, linalg_ext.pack, and linalg_ext.unpack ops to
```
yzhang93, line 419: Nit: I would use the accurate name "iree_linalg_ext.pack", same for unpack.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -412,10 +412,44 @@ def AMDAIEPackAndTranspose :
...
+    %2 = amdaie.dma_cpy_nd(%0[0, 0, 0, 0] [1, 1, 8, 16] [128, 128, 16, 1],
+                           %1[0, 0, 0, 0] [1, 1, 8, 16] [128, 16, 16, 1]) :
+                          (!amdaie.logicalobjectfifo<memref<1x1x8x16xi32, 1>>,
+                           !amdaie.logicalobjectfifo<memref<8x16xi32, 1>>)
+    ```
+
+    The current approach taken with linalg.copy operations, is to lower them
+    to identity linalg_ext.pack operations, and then rely on the lowering of
+    linalg_ext.pack to dma operations to handle the actual lowering to DMAs.
+    We want want to reconsider this 'indirection' in the future.
```
yzhang93, line 448: Nit: use "iree_linalg_ext' and remove additional "want". In addition, I would delete the last sentence in this file and move it to where it is implemented in cpp.

yzhang93 (changes requested):

newling:
> > We already have pack ops without unpack ops, don't we? Only the result will have an unpack op currently.
> 
> Yes, but what I really meant is the copy ops for L2->L3 should be something like unpack op instead of a pack op if we compare this to the fully pack-based pipeline.

I've added some simple logic to ensure copies from L1->L2 and L2->L3 become unpacks instead of packs.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertToDma.cpp
```diff
@@ -302,25 +303,84 @@ LogicalResult rewriteAsDma(PackOrUnpackOp op, IRRewriter &rewriter) {
   return rewriteAsDma(rewriter, op, input, output, innerTiles);
 }
 
+/// Convert a linalg.copy operation on 2 memrefs to an equivalent pack/unpack
+/// operation. If the linalg.copy operation is to a memory close closer to the
+/// core it is converted to a pack operation, otherwise an unpack operation.
+LogicalResult copyToPack(IRRewriter &rewriter, linalg::CopyOp copyOp) {
+  (void)rewriter;
```
yzhang93, line 310: Why is this needed?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertToDma.cpp
```diff
@@ -302,25 +303,84 @@ LogicalResult rewriteAsDma(PackOrUnpackOp op, IRRewriter &rewriter) {
...
+  // MemRefTypes with no memory space attribute return 0 here, so this is safe.
+  uint32_t srcMemspace = cast<MemRefType>(src.getType()).getMemorySpaceAsInt();
+  uint32_t dstMemspace = cast<MemRefType>(dst.getType()).getMemorySpaceAsInt();
+  // Memory space 0 : L3
+  // Memory space 1 : L2
+  // Memory space 2 : L1
+  // So the copy is towards the core if the source memory space is less than
+  // the destination memory space. We'll convert copies towards the core to
+  // packs and all other copies to unpacks, just so that we don't ever have have
+  // pack ops which move data away from the core -- while valid this will be
```
yzhang93, line 335: Nit: Remove additional "have".

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertToDma.cpp
```diff
@@ -302,25 +303,84 @@ LogicalResult rewriteAsDma(PackOrUnpackOp op, IRRewriter &rewriter) {
...
+  // potentially confusing IR as packs usually go towards the core, and it might
+  // break implicit developer assumptions if packs are used to move data away
```
yzhang93, line 337: I don't think it's necessary to add these explanations, the comments above the function are concise and enough.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertToDma.cpp
```diff
@@ -302,25 +303,84 @@ LogicalResult rewriteAsDma(PackOrUnpackOp op, IRRewriter &rewriter) {
...
+  // bootstrap the work done for lowering the pack/unpack op to dmas as the next
+  // step. This is easy to implement, but not the most direct lowering, so
+  // we might want to revisit this.
+  WalkResult convertCopiesWalkResult =
+      getOperation()->walk([&rewriter](linalg::CopyOp copyOp) {
+        if (failed(copyToPack(rewriter, copyOp)))
+          return WalkResult::interrupt();
+        return WalkResult::advance();
+      });
+  if (convertCopiesWalkResult.wasInterrupted()) return signalPassFailure();
```
yzhang93, line 382: It's good to emit the error message.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.h
```diff
@@ -213,9 +213,9 @@ std::unique_ptr<Pass> createAMDAIEObjFifoBufferizationPass();
 std::unique_ptr<Pass> createAMDAIEPackAndTransposePass(
     AMDAIEPackAndTransposeOptions options = {});
 
-/// Create pass to lower pack/unpack ops to AMDAIE DMA ops operating on 
+/// Create pass to lower pack/unpack ops to AMDAIE DMA ops operating on
```
yzhang93, line 216: Also update this comment.

yzhang93 (commented):
Thanks for the revision. Some additional comments to address.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertToDma.cpp
newling: It's already emitted. There is no additional information to add, all failures from copyToPack already emit

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.h
newling: To what?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertToDma.cpp
newling: Oops, good spot

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.h
yzhang93: lower copy/pack/unpack ops...

yzhang93:
Not sure why this test didn't fail, but you have to update the command here https://github.com/nod-ai/iree-amd-aie/blob/main/tests/samples/matmul_pack_peel_objectfifo.mlir#L1.

newling:
> One last request, otherwise LGTM.
> 
> Not sure why this test didn't fail, but you have to update the command here https://github.com/nod-ai/iree-amd-aie/blob/main/tests/samples/matmul_pack_peel_objectfifo.mlir#L1.

Well spotted. I've updated this. I have this PR to check if these tests run: https://github.com/nod-ai/iree-amd-aie/pull/790

yzhang93 (approved):

Issue #763, opened 2024-09-10
[Convolution] Multiple AIE cores objectfifo 'link' issue. 


newling:
When targeting multiple AIE cores for convolution, the slices of the input image which go the cores overlaps (such is the nature of convolution!). This doesn't work with the objectfifo lowering through AIE, because aie.link ops expect the data that goes to each tile to be mutually exclusive (more specifically, it expects the data to be 'contiguous'). 

The plan is to do some major refactoring of the lowering pipeline with a goal to remove the aie.link operation entirely. Once this is done, I will return to trying to lowering convolution to multiple array tiles.

jtuyls:
@newling This PR takes care of removing the usage of `aie.objectfifo.link` altogether: https://github.com/nod-ai/iree-amd-aie/pull/784

PR #781, opened 2024-09-17
Add objectFifo bufferization pass

jtuyls:
Introduces new buffer related ops and a pass to bufferize logical objectFifos. This is part of the effort to move `StatefulTransform` logic into `AMDAIE` logic: https://github.com/nod-ai/iree-amd-aie/compare/main...jtuyls:iree-amd-aie:delete-stateful-transform?expand=1.

The newly introduced ops are:

- `amdaie.buffer`: buffer op which will get translated into `aie.buffer`
- `amdaie.lock`: lock op which will get translated into `aie.lock`
- `amdaie.logicalobjectfifo.from_buffers`: logical objectFifo op created from a set of buffers. Will help with lowering to synchronized DMA patterns (`aie.mem`, `aie.memtile_dma`), without the need for AIE objectFifos (`aie.objectfifo`, `aie.objectfifo.link`).

NOTE: this pass is intentionally not added to the pass pipeline yet  as that requires many more changes: https://github.com/nod-ai/iree-amd-aie/compare/main...jtuyls:iree-amd-aie:delete-stateful-transform?expand=1


compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
```diff
@@ -166,6 +166,66 @@ def AMDAIE_WorkgroupOp : AMDAIE_Op<"workgroup",
...
+    This operation represents a lock on an AIE tile. The operation is fully 
+    specified through a tile and lock ID value, designating the exact physical 
+    lock to be used. This op helps with guaranteeing/verifying correct reuse of
+    the same lock. The op accepts an optional initialization value.
+
+    Example:
+
+    ```mlir
+    %tile = amdaie.tile(%c0, %c1)
+    %lock = amdaie.lock(%tile(0), 2)
```
yzhang93, line 216: What does `%tile(0)` and `2` represent? It's good to add more descriptions.

yzhang93 (approved):
LGTM. One naive question, why do we need to move StatefulTransform logic into AMDAIE dialect and introduce these AIE ops in AMDAIE?

jtuyls:
> LGTM. One naive question, why do we need to move StatefulTransform logic into AMDAIE dialect and introduce these AIE ops in AMDAIE?

This is part of my effort to start introducing control packets and instead of trying to extend `StatefulTransform`, because:

- It's hard to keep extending the pass without changing AIE ops, and I can't easily do that because MLIR-AIR depends on them, and to avoid deviating from MLIR-AIE ops. In addition to the changes I need for control packets, there is also this `aie.objectfifo.link` issue for conv (https://github.com/nod-ai/iree-amd-aie/issues/763), which should get solved by this rewrite as we get rid of the  `aie.objectfifo.link` altogether.
- The pass is doing way too much and is overly complicated, so, therefore the logic of the pass is put in multiple different transformations, like `AMDAIEObjFifoBufferization`.
- The pass operates on `aie.objectfifo` related ops and we can simplify the flow by just skipping that layer by going from logical objectFifos directly to `aie.buffer` etc. The result is that we get rid of a whole layer of logic and reduce the overall amount of code/complexity needed to get to the same result: https://github.com/nod-ai/iree-amd-aie/compare/main...jtuyls:iree-amd-aie:delete-stateful-transform?expand=1

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIELogicalObjFifoOpInterface.cpp
```diff
@@ -8,3 +8,37 @@
...
+  return copyLikOps;
+}
+
+SmallVector<mlir::CopyOpInterface> getCopyLikeProducers(
+    LogicalObjFifoOpInterface op) {
+  SmallVector<mlir::CopyOpInterface> copyLikOps;
+  for (Operation *userOp : op->getUsers()) {
+    if (auto copyOp = dyn_cast<CopyOpInterface>(userOp);
+        dyn_cast_if_present<LogicalObjFifoOpInterface>(
+            copyOp.getTarget().getDefiningOp()) == op) {
```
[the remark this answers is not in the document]
jtuyls: The copy op reads from a source and writes to a target. From the target's perspective the copy op is the producer.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
jtuyls: Added it to the description.

PR #783, opened 2024-09-18
[ObjectFifo] Create a pass to convert temporary alloc to amdaie.buffer

Abhishek-Varma:
-- This commit creates a pass `iree-amdaie-temporary-alloc-bufferization`
   to convert temporary alloc/buffers to amdaie.buffer ops.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/temporary_alloc_bufferization.mlir
```diff
@@ -0,0 +1,87 @@
...
+      memref.dealloc %alloc : memref<1024xf32, 2 : i32>
+      memref.dealloc %alloc_1 : memref<1024xf32, 2 : i32>
+      amdaie.end
+    }
+    amdaie.controlcode {
+      amdaie.end
+    }
+  }
+  return
+}
```
jtuyls, line 87: Add empty line.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETemporaryAllocBufferization.cpp
```diff
@@ -0,0 +1,114 @@
...
+    IRRewriter &rewriter, WorkgroupOp workgroupOp, CoreOp coreOp,
+    SmallVector<Operation *> &toBeErased) {
+  // Step 1. Get all buffers within a CoreOp.
+  // TODO(avarma): All temporary buffers, if present, are at the beginning of
+  // the
+  //               e2e computation within each coreOp and their corresponding
+  //               dealloc ops mark the end of the e2e computation. Therefore we
+  //               are, in order to ensure reuse of the buffer space,
+  //               considering all memref.allocs before the first dealloc op as
+  //               unique set of temporary buffers.
```
jtuyls, line 41: This comment block needs to be formatted I think.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETemporaryAllocBufferization.cpp
```diff
@@ -0,0 +1,114 @@
...
+  auto bufferOp = rewriter.create<AMDAIE::BufferOp>(
+      rewriter.getUnknownLoc(), bufferType, tileOp, nullptr);
+  return bufferOp;
+}
+
+static LogicalResult bufferizeTemporaryAllocInCoreOp(
+    IRRewriter &rewriter, WorkgroupOp workgroupOp, CoreOp coreOp,
+    SmallVector<Operation *> &toBeErased) {
+  // Step 1. Get all buffers within a CoreOp.
+  // TODO(avarma): All temporary buffers, if present, are at the beginning of
```
jtuyls, line 35: What's the TODO?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETemporaryAllocBufferization.cpp
```diff
@@ -0,0 +1,114 @@
...
+  unsigned numOfUniqueAllocOps = 0;
+  bool foundDeallocOp = false;
+  coreOp.walk([&](Operation *op) {
+    if (auto allocOp = dyn_cast<memref::AllocOp>(op)) {
+      allocOps.push_back(allocOp);
+      toBeErased.push_back(allocOp);
+      if (!foundDeallocOp) numOfUniqueAllocOps++;
+    } else if (auto deallocOp = dyn_cast<memref::DeallocOp>(op)) {
+      toBeErased.push_back(deallocOp);
+      foundDeallocOp = true;
```
jtuyls, line 52: Can't you just look at the users of the `AllocOp` to find the `Dealloc`? And shouldn't we just expect every alloc to have a dealloc? This would simplify the logic below (with `numOfUniqueAllocOps`) as well I think.

Abhishek-Varma: Every alloc is indeed having a dealloc op.
But, the main intention of `numOfUniqueAllocOps` was to reuse the temporary buffers. Example :

aie.core {
     %a = memref.alloc
     <SOME USE>
     dealloc %a
     %b = memref.alloc
     <SAME USE LIKE ABOVE>
     dealloc %b
}
So, we shouldn't need to re-create `amdaie.buffer` for `%b` . We can reuse the buffer created for `%a`. So, the logic I added with `foundDeallocOp` is just to get hold of the unique set of AllocOps (the test case I added highlights this well).

I don't think simply fetching Dealloc user of Alloc is solving the `numOfUniqueAllocOps`  logic for reuse.

jtuyls: > So, we shouldn't need to re-create amdaie.buffer for %b . We can reuse the buffer created for %a. So, the logic I added with foundDeallocOp is just to get hold of the unique set of AllocOps (the test case I added highlights this well).

I see. In that case, you need to check the sizes of the allocs as well and keep track of which alloc the dealloc belongs to and I don't see that logic?

jtuyls: That logic is fairly more involved than what we need so I am not sure why we're adding it? And it might already exist on `memref.alloc` somewhere else?

Abhishek-Varma: As discussed offline, I've removed the optimization on temporary buffer that was being done with certain assumptions.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETemporaryAllocBufferization.cpp
```diff
@@ -0,0 +1,98 @@
...
+    } else if (auto deallocOp = dyn_cast<memref::DeallocOp>(op)) {
+      toBeErased.push_back(deallocOp);
+    }
+  });
+  // Bail out early in case of no temporary buffers.
+  if (allocOps.size() == 0) return success();
+  // Step 2. Traverse unique allocOps and create an aie.buffer for them.
+  SmallVector<BufferOp> temporaryBuffers;
+  unsigned tempBufferIndex = 0;
+  for (unsigned i = 0, n = allocOps.size(); i < n; i++) {
```
jtuyls, line 49:
```suggestion
  for (memref::AllocOp allocOp : allocOps) {
```
No need for interating on index.

jtuyls (approved):
LGTM, just one nit.

PR #785, opened 2024-09-18
[DmaToNpu] Add packet support for NpuDmaMemcpyNdOp

jtuyls:
From: https://github.com/Xilinx/mlir-aie/pull/1671/files#diff-2dc661023141402b83d465f580b2aa2e72aad010c1b19da829d1e014fd2d324cR373

Adds packet routing support for `NpuDmaMemcpyNdOp`.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/aie/AMDAIEDmaToNpu.cpp
+++ b/compiler/plugins/target/AMD-AIE/aie/AMDAIEDmaToNpu.cpp
@@ -376,6 +376,13 @@ struct DmaToNpuPattern : OpConversionPattern<NpuDmaMemcpyNdOp> {
     // repeat_count
     repeat_count = IntegerAttr::get(i32ty, sizes[3] - 1);
 
+    // enable_packet
+    if (auto packetInfo = op.getPacket()) {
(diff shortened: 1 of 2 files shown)


compiler/plugins/target/AMD-AIE/aie/AMDAIEDmaToNpu.cpp
```diff
@@ -376,6 +376,13 @@ struct DmaToNpuPattern : OpConversionPattern<NpuDmaMemcpyNdOp> {
     // repeat_count
     repeat_count = IntegerAttr::get(i32ty, sizes[3] - 1);
 
+    // enable_packet
+    if (auto packetInfo = op.getPacket()) {
```
yzhang93, line 380: Where is the first place that packet control is turned on?

yzhang93 (approved):

compiler/plugins/target/AMD-AIE/aie/AMDAIEDmaToNpu.cpp
jtuyls: E2E you mean? I have one hacked example working locally that enables it for L3 -> L2 data movement, but I am working on generalizing it.

PR #790, opened 2024-09-19
[CI] Reenable tests, fixes from last 3 weeks

newling:
Escapees recaptured:

image


yzhang93:
Nice and glad to see these tests added back to ci! So I think the reason is it's not picking up tests outside compiler / runtime directories?

yzhang93 (approved):

newling:
> Nice and glad to see these tests added back to ci! So I think the reason is it's not picking up tests outside compiler / runtime directories?

Yes, that's what I what I understand. For some reason, cmake was handling that test directory differently.

PR #794, opened 2024-09-23
[ConvertToUnpack] Apply correct permutation when lowering unpack to DMA

newling:
Inverse permutation needs to be applied when converting unpack to DMA, not permutation. PR adds test with permutation P != P^{-1}, without which this bug is hidden. I've left a comment with that test explaining why this fix is required. Also, `--cse`  is removed from the test lit command, to reduce noise.


jtuyls (approved):
LGTM

Issue #701, opened 2024-08-24
`medium_i32_i32_m1536_n1536_k2048` consistently fails on Windows


makslevental:
Haven't a clue. XFAILed in https://github.com/nod-ai/iree-amd-aie/pull/689/commits/906dd3b38de5d0fbd66b365c2f69316668386ffb.

cc @newling

PR #795, opened 2024-09-23
Use python bindings for tests

makslevental:
This PR begins the migration to `pytest` for all of our `e2e` tests using IREE's python bindings for both the [compiler and the runtime](https://iree.dev/reference/bindings/python/). I've tested it thoroughly (by running `run_matmul.sh` and `cpu_comparison/run.py` ~100~50 times).

Two major changes were required to make this happen:

1. Move all CLI args to `AMDAIEOptions` so that they're registered at plugin load time. This was necessary for enabling `Session.set_flags` in Python. Going forward we should make sure to put all CLI args in the same place;
2. The XRT HAL had to be drastically refactored because in its current state (prior to this PR), it doesn't support multiple dispatches in a single process - I had to track down all kinds of resource leaks (why this PR took so much longer than I thought);
     - There's a test `1536x2048x1536` that was prior disabled for Windows but is now disabled entirely. I don't know what the connection is (I wasn't able to track it down) but since every other test passes, including the 1000 run test, and even after repeating overall ~100~50 runs, I think it's safe to conclude there's something wrong with that shape rather than the HAL.

The migration isn't complete because the [bindings](https://github.com/iree-org/iree/blob/9e091155b8b4147a523352e954e161b655399d80/runtime/bindings/python/invoke.cc#L65) and the [VM itself](https://github.com/iree-org/iree/blob/d55785dc6affb9c0ffe787e0dc0e8f6f32a73da2/runtime/src/iree/vm/value.h#L56) don't support bf16 yet; I will send them a PR soon.

Note, in its current state the test scripts run ~100~50 times. I will remove this before landing.

**TODO** (for the next iteration): right now `repeat_runs` doesn't work because of how the XRT command buffer works (it's free'ed after every dispatch and with it the `xrt::kernel` and `xrt::hw_context`). Likely I'm doing something wrong (possibly using the wrong command buffer model) so I think it should be fixable.


runtime/src/iree-amd-aie/driver/xrt/xrt_device.cc
```diff
@@ -104,7 +117,9 @@ static void iree_hal_xrt_device_destroy(iree_hal_device_t* base_device) {
 
   iree_hal_allocator_release(device->device_allocator);
   iree_arena_block_pool_deinitialize(&device->block_pool);
+  xrtDeviceHandle device_hdl = device->device_hdl;
   iree_allocator_free(host_allocator, device);
+  (void)xrtDeviceClose(device_hdl);
```
makslevental, line 122: The way `xrtDeviceOpen` and `xrtDeviceClose` work is there's a `static std::map<...>` in XRT that maintains "references" to the [device objects](https://github.com/nod-ai/XRT/blob/a01266e20dff0ec1f59059a3377102a12e2c7d89/src/runtime_src/core/common/api/xrt_device.cpp#L46).

runtime/src/iree-amd-aie/driver/xrt/xrt_driver.cc
```diff
@@ -131,11 +110,11 @@ static iree_status_t iree_hal_xrt_driver_dump_device_info(
     iree_hal_driver_t* base_driver, iree_hal_device_id_t device_id,
     iree_string_builder_t* builder) {
   iree_hal_xrt_driver_t* driver = iree_hal_xrt_driver_cast(base_driver);
-  xrt::device* device = driver->device;
+  xrtDeviceHandle device_hdl = driver->device_hdl;
   IREE_RETURN_IF_ERROR(
       iree_string_builder_append_cstring(builder, "\n- Platform:"));
-
-  std::string platform_info = device->get_info<xrt::info::device::platform>();
+  xrt::device device(xrtDeviceToXclDevice(device_hdl));
```
makslevental, line 116: This makes exactly zero sense but it's correct - `xrtDeviceHandle` is **not** the handle that corresponds to the underlying kernel resource; the device handle that actually corresponds to the underlying kernel driver resource (the one that's deposited into the `static map<...>` mentioned below) is fetched through [this helper](https://github.com/nod-ai/XRT/blob/a01266e20dff0ec1f59059a3377102a12e2c7d89/src/runtime_src/core/common/api/xrt_device.cpp#L670).

tests/test_matmul.py
```diff
@@ -0,0 +1,155 @@
...
+
+    @func(T.tensor(32, 16, T.i8()), T.tensor(16, 32, T.i8()))
+    def matmul_i8_i32(lhs, rhs):
+        cst = arith.constant(T.i32(), 0)
+        v0 = tensor.empty([32, 32], T.i32())
+        v1 = linalg.fill(cst, outs=[v0])
+        return linalg.matmul(lhs, rhs, outs=[v1])
+
+    arg0 = np.ones((32, 16), dtype=np.int8)
+    arg1 = np.ones((16, 32), dtype=np.int8)
```
jtuyls, line 25: Can we not use `ones` as input? If the data is rearranged incorrectly (for example through the DMAs), this will still lead to  correct results because all inputs are `1`.

build_tools/download_peano.ps1
```diff
@@ -0,0 +1,11 @@
+# Copyright 2024 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+$ErrorActionPreference = 'Stop'
+
+$RELEASE = "19.0.0.2024082221+90abe71b"
```
jtuyls, line 9: We should probably use the same version as on Linux? https://github.com/nod-ai/iree-amd-aie/blob/a56844ce13786150c8198da688453a62c4e8b97d/build_tools/download_peano.sh#L9

build_tools/ci/run_matmul_test.sh
```diff
@@ -659,9 +654,9 @@ run_matmul_test_on_shapes ${i32_shapes_small[@]} \
 i32_shapes_medium=(
   '1024x1024x1024'
 )
-if [ "$OSTYPE" != "msys" ]; then
-  i32_shapes_medium+=('1536x2048x1536')
-fi
+#if [ "$OSTYPE" != "msys" ]; then
```
jtuyls, line 657:
```suggestion
# TODO(jornt): re-enable `1536x2048x1536`
#if [ "$OSTYPE" != "msys" ]; then
```

jtuyls: Or maybe better remove and create an issue for it (which I think you already had)?

cmake/iree_aie_xrt.cmake
```diff
@@ -176,7 +175,10 @@ foreach(_core_lib IN LISTS _core_libs)
                              ${IREE_XRT_SOURCE_DIR}/runtime_src)
   target_include_directories(${_core_lib} SYSTEM PUBLIC
                              ${IREE_XRT_SOURCE_DIR}/runtime_src/core/common/elf)
-  target_compile_definitions(${_core_lib} PUBLIC -DBOOST_BIND_GLOBAL_PLACEHOLDERS)
+  target_compile_definitions(${_core_lib}
+                             PUBLIC
+                             -DBOOST_BIND_GLOBAL_PLACEHOLDERS
+                             $<$<CONFIG:Debug>:-DXRT_VERBOSE>)
```
jtuyls, line 181: Is this for debugging purposes?

makslevental: Yes

tests/conftest.py
```diff
@@ -0,0 +1,154 @@
...
+    "ui16",
+    "ui32",
+    "ui64",
+    "f16",
+    "f32",
+    "f64",
+    "bf16",
+]:
+    tf = getattr(T, t)
+    tf.__name__ = t
```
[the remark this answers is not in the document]
makslevental: To get the tests to print legible labels. These names are actually [lambdas](https://github.com/llvm/llvm-project/blob/7b8f7beadcf1814b1f1aa985d344ca17747531a7/mlir/python/mlir/extras/types.py#L54) (not my smartest design decision but 🤷).

build_tools/download_peano.ps1
makslevental: Whoops didn't notice the bump

makslevental: Whoops didn't notice the bump

build_tools/ci/run_matmul_test.sh
makslevental: Ya https://github.com/nod-ai/iree-amd-aie/issues/701 (I know the shape there is a permutation but it's gotta be the same malfunction...)

cmake/iree_aie_xrt.cmake
jtuyls: Should we enable it by default?

makslevental: Nah it's too "loud".

jtuyls (approved):
LGTM

cmake/iree_aie_xrt.cmake
jtuyls: Sorry, I was assuming it was enabled by default now. If it's not, that's good

newling:
Off the top of my head things which we need to reach parity, hopefully we can reuse what's already done in cpu_comparison/run.py:

- When the numerics fail, give good feedback 
- Random seed control 
- Only sample for very large matmuls, don't perform full CPU computation (maybe not needed)

Also, can you/someone please at some point:
- document in the README how to run the pytests in the test dir
- create an issue describing what work still needs to be done
- keep track of migration of tests from old script to pytest, ideally they shouldn't be run in both places. Or maybe they've all been moved already?

makslevental:
> keep track of migration of tests from old script to pytest, ideally they shouldn't be run in both places. Or maybe they've all been moved already?

None have been moved, some have been duplicated.

PR #799, opened 2024-09-25
Delete obviated tests

makslevental:
This functionality is now better tested in `compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/test` and `runtime/src/iree-amd-aie/aie_runtime/test`.


makslevental:
Hey @newling can I delete "smoke test" as well? I think we have a good handle on those flags now?

newling:
This test was initially created check that flags like `aie2xclbin-print-ir-before-all` were working, but those flags are now basically not relevant so I'm ok with the test being removed (although it seems to be testing more now than it did when I initially created it). 

> Hey @newling can I delete "smoke test" as well? I think we have a good handle on those flags now?

The [test](https://github.com/nod-ai/iree-amd-aie/blob/2384626d3e19f511f33356aa776b0e2635f64357/.github/workflows/ci-linux.yml#L216) is there to prevent drift, to make sure those flags keep working. Why do you want to delete it?

makslevental:
> The [test](https://github.com/nod-ai/iree-amd-aie/blob/2384626d3e19f511f33356aa776b0e2635f64357/.github/workflows/ci-linux.yml#L216) is there to prevent drift, to make sure those flags keep working. Why do you want to delete it?

Because I'm assuming that now we understand the caveats of how argparser parses flags so why keep a test around that just tests that understanding? It's not like the implementation of that parsing is going to drift?

jtuyls (approved):
LGTM

newling:
> > The [test](https://github.com/nod-ai/iree-amd-aie/blob/2384626d3e19f511f33356aa776b0e2635f64357/.github/workflows/ci-linux.yml#L216) is there to prevent drift, to make sure those flags keep working. Why do you want to delete it?
> 
> Because I'm assuming that now we understand the caveats of how argparser parses flags so why keep a test around that just tests that understanding? It's not like the implementation of that parsing is going to drift?
> 
> EDIT: and it's not free - every time we _change_ the flags/args to `cpu_comparison/run.py` we have to update the smoke test and since the logic is different from the other tests it takes some amount of squinting to change.

Fair points, you can remove it. I just hope the flag `--do-not-run-aie` keeps working, that's the one I care most about in that test.

PR #801, opened 2024-09-25
[Utils] Remove unused utils

newling:
@jtuyls I'm ok keeping these if you think they might be useful in the future. I'm making this PR in case it's something you overlooked when changing to not use StatefulTransform


makslevental (approved):
I've said it before: 

I'm a simple man: I see red, I approve.

:smile:

jtuyls (approved):

PR #802, opened 2024-09-25
[Router] Consider channel ID demand and congestion in Dijkstra

jtuyls:
Pulls in changes from: https://github.com/Xilinx/mlir-aie/pull/1643:

> Changes on router behaviours: Previously, the router used a two-step process: first finding paths with Dijkstra's algorithm and then allocating channel IDs based on heuristics. Now, channel ID allocation is integrated into Dijkstra's algorithm to consider demand and congestion on individual channels directly.

NOTE: I had to bump the `DEMAND_COEFF` from 1.1 to 1.5 to find a solution for the `unit_vecmul_4x4.mlir` testcase. Not sure whether this makes sense or there might be an underlying issue? cc @Yu-Zhewen

### DIFF
--- a/runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
+++ b/runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
@@ -15,9 +15,15 @@
 
 #define DEBUG_TYPE "iree-aie-runtime-router"
 
-#define OVER_CAPACITY_COEFF 0.02
-#define USED_CAPACITY_COEFF 0.02
-#define DEMAND_COEFF 1.1
+#define OVER_CAPACITY_COEFF 0.1
+#define USED_CAPACITY_COEFF 0.1
+// NOTE(jornt): Set demand coeff high enough to find solutions. A coeff of 1.1
+// - 1.3 failed to find solutions in some tests.
+#define DEMAND_COEFF 1.5
+#define DEMAND_BASE 1.0
+#define MAX_CIRCUIT_STREAM_CAPACITY 1
+#define MAX_PACKET_STREAM_CAPACITY 32
+
 static constexpr double INF = std::numeric_limits<double>::max();
 
 namespace MLIRAIELegacy {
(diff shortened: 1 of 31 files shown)


runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
```diff
@@ -43,295 +49,194 @@ int columns(const mlir::iree_compiler::AMDAIE::AMDAIEDeviceModel &deviceModel);
...
+  // capacity
+  void bumpDemand(size_t i, size_t j) {
+    if (usedCapacity[i][j] >= MAX_CIRCUIT_STREAM_CAPACITY) {
+      demand[i][j] *= DEMAND_COEFF;
+    }
+  }
 };
 
-struct FlowNode {
+using Flow = struct Flow {
```
makslevental, line 108: nit: I'm not a fan of typedefs for structs - I don't remember what it was but aie-rt makes things difficult for us becaues they do this everywhere.

makslevental:
> Not sure whether this makes sense or there might be an underlying issue?

I'm trying to think how these behave differently - looking at the code closely I don't actually see how this PR handles congestion differently? It's still basically 

```
      double history = 1.0 + OVER_CAPACITY_COEFF * overCapacity[&ch];
      double congestion = 1.0 + USED_CAPACITY_COEFF * usedCapacity[&ch];
      demand[&ch] = history * congestion;
```

right?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
jtuyls: Yeah, I will change it.

jtuyls:
> > Not sure whether this makes sense or there might be an underlying issue?
> 
> I'm trying to think how these behave differently - looking at the code closely I don't actually see how this PR handles congestion differently? It's still basically
> 
> ```
>       double history = 1.0 + OVER_CAPACITY_COEFF * overCapacity[&ch];
>       double congestion = 1.0 + USED_CAPACITY_COEFF * usedCapacity[&ch];
>       demand[&ch] = history * congestion;
> ```
> 
> right?

It might just be the coefficients? I do see the load being spread over more connections.

makslevental (approved):
LGTM

PR #803, opened 2024-09-27
Remove `chesshack`

makslevental:
As of 2024.2, Chess has caught up to LLVM. This will unblock https://github.com/nod-ai/iree-amd-aie/pull/732 (i.e., aiesimulator support).


newling (approved):
Can't resist with a PR title like this

PR #805, opened 2024-09-27
Retrieve numArbiters and numMSels from device model

jtuyls:
Removes the hardcoded `numArbiters` and `numMSels` and retrieves it from the device model instead. I wasn't able though to find it in a retrievable way inside `aie-rt`, so I created a separate `AMDAIEDeviceConfig` struct for such cases.


runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
```diff
@@ -866,9 +868,19 @@ FailureOr<std::tuple<MasterSetsT, SlaveAMSelsT>> emitPacketRoutingConfiguration(
   for (const auto &[physPortAndID, packetFlowports] : sortedPacketFlows) {
     // The Source Tile of the flow
     TileLoc tileLoc = physPortAndID.physPort.tileLoc;
+    // Make sure the generator for the tile is initialized with the correct
+    // number of arbiters and msels.
+    uint8_t numArbiters =
+        1 + deviceModel.getStreamSwitchArbiterMax(tileLoc.col, tileLoc.row);
```
makslevental, line 874: why `+1`?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -200,6 +200,20 @@ inline ::XAie_TxnOpcode txnToTxn(XAie_TxnOpcode t) {
...
+struct AMDAIEDeviceConfig {
+  /// Currently, the max arbiter/msel is hidden inside aie-rt.
+  uint8_t streamSwitchCoreArbiterMax{0};
+  uint8_t streamSwitchCoreMSelMax{0};
+  uint8_t streamSwitchMemTileArbiterMax{0};
+  uint8_t streamSwitchMemTileMSelMax{0};
+  uint8_t streamSwitchShimArbiterMax{0};
+  uint8_t streamSwitchShimMSelMax{0};
+  AMDAIEDeviceConfig() = default;
+};
```
makslevental, line 215: This is basically a nit/style thing but if there's no plan for anyone elsewhere to use this struct _as a struct_ (like grouped together) then it should be "flattened" into fields on `AMDAIEDeviceModel `. Or (if you think expanding the constructor args is tedious) then it should be at least moved into the definition of `AMDAIEDeviceModel ` like this

```
struct AMDAIEDeviceModel {
  ...
struct StreamSwitchPacketSwitchConfig {
  /// Currently, the max arbiter/msel is hidden inside aie-rt.
  uint8_t streamSwitchCoreArbiterMax{0};
  uint8_t streamSwitchCoreMSelMax{0};
  uint8_t streamSwitchMemTileArbiterMax{0};
  uint8_t streamSwitchMemTileMSelMax{0};
  uint8_t streamSwitchShimArbiterMax{0};
  uint8_t streamSwitchShimMSelMax{0};
  AMDAIEDeviceConfig() = default;
  } packetSwitchConfig;
```

Also if you do `const StreamSwitchPacketSwitchConfig&& packetSwitchConfig` as the arg you'll get the `std::move`s for free I think.

makslevental (approved):
LGTM small things

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
jtuyls: Because `getStreamSwitchArbiterMax` returns the max value, and the first arbiter is `0`, so the number of arbiters is max + 1. This is to be consistent with how it's internally defined in `aie-rt`, and it's consistent with `std::numeric_limits::max` as well.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
jtuyls: > Or (if you think expanding the constructor args is tedious)

Yes, this is exactly the reason. A separate struct will be more versatile towards possibly future parameters.

> then it should be at least moved into the definition of AMDAIEDeviceModel

Sure, I can do that.

> Also if you do const StreamSwitchPacketSwitchConfig&& packetSwitchConfig as the arg you'll get the std::moves for free I think.

That will incur a copy then. I could make it `StreamSwitchPacketSwitchConfig&&`, but that would still need the move on the caller/user side and inside the constructor to avoid a copy. I prefer the pass by value here as it is more versatile towards move by value/move and towards adding a default value without needing multiple overloads.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
makslevental: oh duh - whoops sorry poor reading comprehension (or if I blame someone else 😄, `getStreamSwitchArbiterMax`  should be `getStreamSwitchArbiterMaxIdx`)

PR #806, opened 2024-09-27
[Router] Add packet flow groups for channel sharing

jtuyls:
Pulls in changes from: https://github.com/Xilinx/mlir-aie/pull/1694

This fixes deadlock issues with larger matmuls when all connections are made packet connections and therefore enables optional packet routing on all connections.

### DIFF
--- a/runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
+++ b/runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
@@ -71,6 +71,8 @@ struct SwitchboxConnect {
   std::vector<std::vector<int>> usedCapacity;
   // how many packet streams are actually using this Channel
   std::vector<std::vector<int>> packetFlowCount;
+  // only sharing the channel with the same packet group id
+  std::vector<std::vector<int>> packetGroupId;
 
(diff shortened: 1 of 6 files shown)


runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
```diff
@@ -394,6 +424,15 @@ std::optional<std::map<PathEndPoint, SwitchSettings>> Router::findPaths(
...
   }
 
+  // group flows based on packetGroupId
+  std::map<int, std::vector<Flow>> groupedFlows;
+  for (auto &f : impl->flows) {
+    if (groupedFlows.count(f.packetGroupId) == 0) {
+      groupedFlows[f.packetGroupId] = std::vector<Flow>();
+    }
+    groupedFlows[f.packetGroupId].push_back(f);
+  }
```
makslevental, line 434:
```suggestion
  std::map<int, std::vector<Flow>> groupedFlows;
  for (auto &f : impl->flows) {
    groupedFlows[f.packetGroupId].push_back(f);
  }
```

just accessing std::map with a key creates the value

https://godbolt.org/z/bTPv1e61G

makslevental (approved):
LGTM

PR #811, opened 2024-10-01
[DistributeCoresAndObjectFifos] Fix for case of mixed operands (subviews and allocs). 

newling:
Fix for a regression in depthwise convolution with packing, when we use the upstream pass to remove unit extent dimensions `useRankReducingSlices = true`. Enabling the unit extent dimension removal results in a op:

```
linalg.generic {indexing_maps = [#map, #map], iterator_types = ["parallel", "parallel"]} ins(%alloc : memref<4x4xi32, 2 : i32>) outs(%subview : memref<4x4xi32, strided<[4, 1]>, 2 : i32>) { ... }
```

appearing in the IR, which is new because it's the first time we see an op with one operand directly from a `memref.alloc`, and another from a `memref.subview` operation. This PR adds a fix for this case (by simplifying the logic...). Before this PR, the assumption was that all/none of the operands were from `memref.alloc` (I think).

PR #813, opened 2024-10-02
[WIP] Lower memref.alloc to amdaie.buffer 

newling:
This along with https://github.com/nod-ai/iree-amd-aie/pull/811 gets depthwise convolution with packing working again

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEObjFifoBufferization.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEObjFifoBufferization.cpp
@@ -131,6 +131,78 @@ class AMDAIEObjFifoBufferizationPass
   void runOnOperation() override;
 };
 
+/// 
+/// Some memref allocations are accessed directly, as a result of the pass 
+///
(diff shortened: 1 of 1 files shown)


Abhishek-Varma:
Hi @newling 

Could you use [this](https://github.com/nod-ai/iree-amd-aie/pull/783) pass instead after AMDAIEObjectfifoBufferization ?

I don't think we would be needing this PR.

Also, I'll be raising a PR in a while enabling the above (merged) pass by default in the main pipeline.

newling:
> Hi @newling
> 
> Could you use [this](https://github.com/nod-ai/iree-amd-aie/pull/783) pass instead after AMDAIEObjectfifoBufferization ?
> 
> I don't think we would be needing this PR.
> 
> Also, I'll be raising a PR in a while enabling the above (merged) pass by default in the main pipeline.

Ok, cool, I'll try this tomorrow. Thanks for keeping an eye on the PRs and seeing this one appear!

newling:
@Abhishek-Varma your pass does the trick for me. 

One caveat: if someone adds `HoistStaticallyBoundAllocationsPass` before `AMDAIEObjectfifoBufferization` then the pass doesn't work, because the pass assumes the temporary allocations are inside the CoreOps. I might make a PR to address this later...

PR #816, opened 2024-10-02
Direct Driver HAL

makslevental:
quote-you-know-crazy-straws-they-go-all-over-the-place-these-straws-are-sane-they-never-lost-mitch-hedberg-129-5-0527

Let's try this again: this is a HAL that skips/avoids/elides XRT *completely*. It does so by (essentially) going straight to the [XDNA driver's (kmq) shim](https://github.com/amd/xdna-driver/tree/main/src/shim) which talks directly to the kernel module. Note that XDNA's shim actually couples tightly to XRT itself so the bulk of the actual "work" here is/was decoupling. Thus, we vendor the decoupled shim. Just to be explicit/clear: after this PR, on Linux, the only thing needed in the environment is the [kernel module itself](https://github.com/amd/xdna-driver/tree/main/src/driver/amdxdna) - not XRT, nor XDNA's shim. 

Call outs:

1. Only Linux for now (Windows soon);
2. On Linux,  we no longer even need to clone XRT (completely really does mean completely);
    - We no longer use `.xclbin` as the artifact and now just load the `.pdi` directly;
3. This uses the so-called kernel-managed queue (`kmq`) path; user-managed queues (`umq`) aren't supported until aie4;
4. The current implementation here is still sync and still uses direct (rather than indirect) command buffers; this PR is already large/complex enough without changing the paradigms. The next PR will add/implement those;
5. Nothing about how drivers themselves are bumped needs to change - you/anyone can continue to install the XDNA drivers using their full instructions there and everything will work the same, i.e., the XRT pieces will still be skipped even if you have them installed. But I will probably put together a script that builds/installs _only_ their kernel module for those that want to opt-out of XRT in their environment;
6. For now, the XRT HAL will still work on Linux but it should be considered deprecated (hence no longer tested in CI).

TODOs before landing:

- [x] remove instructional comments leftover from the `null` HAL;
- [x] undo "OO"-ness of the HAL
- [x] add missing trace zones
- [x] remove smart pointers where possible
- [ ] ~~move `hwq` into `kernel`~~
- [x] remove exceptions from the shim.
- [x] parameterize number of cols/rows (currently hardcoded 4x4 for Phoenix);

shout-out to @benvanik for putting together the [`null` HAL](https://github.com/iree-org/iree/tree/main/runtime/src/iree/hal/drivers/null) lickety-split for me to study while working on this PR.


.github/workflows/ci-linux.yml
```diff
@@ -49,7 +49,10 @@ jobs:
           git remote add origin $REPO_ADDRESS
           git -c protocol.version=2 fetch --depth 1 origin $BRANCH_NAME
           git reset --hard FETCH_HEAD
-          git -c submodule."third_party/torch-mlir".update=none -c submodule."third_party/stablehlo".update=none -c submodule."src/runtime_src/core/common/aiebu".update=none submodule update --init --recursive --depth 1 --single-branch -j 10
+          git -c submodule."third_party/torch-mlir".update=none \
+              -c submodule."third_party/stablehlo".update=none \
+              -c submodule."third_party/XRT".update=none \
```
makslevental, line 54: look ma no XRT

build_tools/ci/cpu_comparison/run.py
```diff
@@ -1023,6 +1028,15 @@ def all_tests(
         default="",
     )
 
+    parser.add_argument(
+        "--device-hal",
+        default="xrt",
```
dezhiAmd, line 1033: Default HAL is xrt or xrt-lite?

makslevental: done

build_tools/ci/cpu_comparison/run.py
```diff
@@ -1023,6 +1028,15 @@ def all_tests(
         default="",
     )
 
+    parser.add_argument(
+        "--device-hal",
+        default="xrt-lite",
+        const="xrt-lite",
+        nargs="?",
```
yzhang93, line 1035: What are "const" and "nargs" used for? Also what's the difference between "xrt" and "xrt-lite"? When should we choose "xrt"/"xrt-lite"?

build_tools/ci/run_matmul_test.sh
```diff
@@ -28,16 +28,16 @@ if [ "$#" -lt 2 ] || [ "$#" -gt 5 ]; then
...
-   #    6) <vitis-install-dir>     (optional)
     echo -e "Illegal number of parameters: $#, expected 2-5 parameters." \
             "\n The parameters are as follows:" \
             "\n     1) <output-dir>               (required)" \
             "\n     2) <iree-install-dir>         (required)" \
             "\n     3) <peano-install-dir>        (optional)" \
-            "\n     4) <xrt-dir>                  (optional)" \
-            "\n     5) <vitis-install-dir>        (optional)" \
+            "\n     4) <vitis-install-dir>        (optional)" \
+            "\n     5) <xrt-dir>                  (optional)" \
```
yzhang93, line 40: Why do you change the order of 4 and 5? Just want to make sure the parameters are passing in the same order between `run_matmul_test.sh` and `run.py`.

runtime/src/iree-amd-aie/driver/xrt-lite/buffer.cc
```diff
@@ -0,0 +1,156 @@
...
+namespace {
+const iree_hal_buffer_vtable_t iree_hal_xrt_lite_buffer_vtable = {
+    .recycle = iree_hal_buffer_recycle,
+    .destroy = iree_hal_xrt_lite_buffer_destroy,
+    .map_range = map_range,
+    .unmap_range = unmap_range,
+    .invalidate_range = invalidate_range,
+    .flush_range = flush_range,
+};
+}
```
yzhang93, line 156: Add empty line.

runtime/src/iree-amd-aie/driver/xrt-lite/cts/executable_cache_test.mlir
```diff
@@ -0,0 +1,33 @@
...
+      %3 = flow.dispatch.tensor.load %0, offsets = [0, 0], sizes = [32, 32], strides = [1, 1] : !flow.dispatch.tensor<readonly:tensor<32x32xf32>> -> tensor<32x32xf32>
+      %4 = flow.dispatch.tensor.load %1, offsets = [0, 0], sizes = [32, 32], strides = [1, 1] : !flow.dispatch.tensor<readonly:tensor<32x32xf32>> -> tensor<32x32xf32>
+      %5 = tensor.empty() : tensor<32x32xf32>
+      %6 = linalg.fill ins(%c0_f32 : f32) outs(%5 : tensor<32x32xf32>) -> tensor<32x32xf32>
+      %7 = linalg.matmul ins(%3, %4 : tensor<32x32xf32>, tensor<32x32xf32>) outs(%6 : tensor<32x32xf32>) -> tensor<32x32xf32>
+      flow.dispatch.tensor.store %7, %2, offsets = [0, 0], sizes = [32, 32], strides = [1, 1] : tensor<32x32xf32> -> !flow.dispatch.tensor<writeonly:tensor<32x32xf32>>
+      return
+    }
+  }
+}
```
yzhang93, line 33: Also add empty line here and please check other files as well.

build_tools/ci/cpu_comparison/run.py
makslevental: re `const` and `nargs`: https://stackoverflow.com/a/40324463

re `xrt` vs. `xrt-lite`: this PR adds the `xrt-lite` HAL (the previous/current/original HAL is the `xrt` HAL). If you're on Linux you should choose `xrt-lite` and if you're on Windows, for now, you should choose `xrt`.

build_tools/ci/run_matmul_test.sh
makslevental: Because after this PR XRT becomes optional (i.e., unused on Linux) but Vitis remains so placing `xrt-dir` fifth means it can be omitted. For `run.py`, the directory is already optionally passed using `--xrt-dir`.

runtime/src/iree-amd-aie/driver/xrt-lite/device.cc
```diff
@@ -0,0 +1,296 @@
...
+    IREE_ASSERT(iree_status_is_ok(status));
+    iree_arena_block_pool_initialize(ARENA_BLOCK_SIZE, host_allocator,
+                                     &block_pool);
+
+    IREE_TRACE_ZONE_END(z0);
+  }
+};
+
+static iree_status_t iree_hal_xrt_lite_device_create_executable_cache(
+    iree_hal_device_t* base_value, iree_string_view_t identifier,
```
nirvedhmeshram, line 56: nit: for here and everywhere else base_value -> base_device

runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc
```diff
@@ -0,0 +1,216 @@
...
+        iree_hal_buffer_allocated_buffer(bindings.values[j].buffer));
+    ebuf.add_arg_bo(*bo);
+    bo->sync(shim_xdna::direction::host2device);
+  }
+
+  shim_xdna::hw_q* hwq = context->get_hw_queue();
+  hwq->issue_command(ebuf.get_exec_buf_bo());
+  hwq->wait_command(ebuf.get_exec_buf_bo(), 0);
+
+  for (iree_host_size_t j = 0; j < bindings.count; ++j) {
```
nirvedhmeshram, line 193: I think this should not be needed as we are doing this when we map/unmap the buffers, but I know the xrt hal is doing the same thing, so maybe add a TODO about this?

runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc
```diff
@@ -0,0 +1,216 @@
...
+  unsigned int opcode = 3;
+  ebuf.add_arg_64(opcode);
+  ebuf.add_arg_bo(*bo_ctrl_code);
+  ebuf.add_arg_32(kernel_params.asm_inst.size());
+
+  for (iree_host_size_t j = 0; j < bindings.count; ++j) {
+    shim_xdna::bo* bo = iree_hal_xrt_lite_buffer_handle(
+        iree_hal_buffer_allocated_buffer(bindings.values[j].buffer));
+    ebuf.add_arg_bo(*bo);
+    bo->sync(shim_xdna::direction::host2device);
```
nirvedhmeshram, line 186: I think this should not be needed as we are doing this when we map/unmap the buffers, but I know the xrt hal is doing the same thing, so maybe add a TODO about this?

runtime/src/iree-amd-aie/driver/xrt-lite/device.cc
```diff
@@ -0,0 +1,296 @@
...
+  IREE_ASSERT_ARGUMENT(out_device);
+  *out_device = nullptr;
+
+  IREE_TRACE_ZONE_BEGIN(z0);
+
+  iree_hal_xrt_lite_device* device = nullptr;
+  iree_host_size_t total_size = iree_sizeof_struct(*device) + identifier.size;
+  IREE_RETURN_IF_ERROR(
+      iree_allocator_malloc(host_allocator, total_size, (void**)&device));
+  device = new (device) iree_hal_xrt_lite_device(options, host_allocator);
```
nirvedhmeshram, line 267: can you add a comment that this is calling the constructor of `iree_hal_xrt_lite_device` . Took me a while to figure out as we normally dont see structs with contructors in a hal implementation.

nirvedhmeshram (commented):
Nice! I am wondering though, one of the issues with XRT HAL was the heavy dependence on C++ which was leaking into the HAL C structs that cant have C++ objects so we had pointers to them and ran into ownership issues of those pointers. XRT-lite is using even more C++, however we perhaps have more control over it as the API is lower level. So lets see how this scales with time. Great work!

makslevental:
> Nice! I am wondering though, one of the issues with XRT HAL was the heavy dependence on C++ which was leaking into the HAL C structs that cant have C++ objects so we had pointers to them and ran into ownership issues of those pointers. 

Interesting - do you remember which objects specifically? I have to admit that one thing that isn't clear to me (in HAL land) is what is allowed-to-be/should-be an "inlined struct" and what needs to be a pointer. Currently I believe I'm managing lifetimes correctly but I haven't actually rigorously checked (I keep failing to successfully get an [ASAN build working for us](https://github.com/nod-ai/iree-amd-aie/pull/808)).

runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc
makslevental: I'll try removing and see what happens.

runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc
makslevental: Same

runtime/src/iree-amd-aie/driver/xrt-lite/device.cc
makslevental: If enough people find this super weird I'll just make it conventional. Vote here if you don't like it!

nirvedhmeshram:
> > Nice! I am wondering though, one of the issues with XRT HAL was the heavy dependence on C++ which was leaking into the HAL C structs that cant have C++ objects so we had pointers to them and ran into ownership issues of those pointers.
> 
> Interesting - do you remember which objects specifically? I have to admit that one thing that isn't clear to me (in HAL land) is what is allowed-to-be/should-be an "inlined struct" and what needs to be a pointer. Currently I believe I'm managing lifetimes correctly but I haven't actually rigorously checked (I keep failing to successfully get an [ASAN build working for us](https://github.com/nod-ai/iree-amd-aie/pull/808)).

I think all the smart pointer patterns you see in the XRT HAL are a symptom of that issue. We essentially used the .release() on smart pointers to take ownership of the objects as you couldnt just new/delete them in the XRT API, and an example of the ownership issues I am referring to are shown by the need of PRs like this
https://github.com/nod-ai/iree-amd-aie/pull/641
although looks like that was later fixed (as I don't see the global device in the xrt_device.h in the code anymore)
We also had trouble with the kernel being deleted because we were deleting the xclbin, but we technically had ownership of the kernel.

dezhiAmd:
Currently "xrt" does provide a path forward regarding to 

- reusable command buffers

- indirect dispatch or runtime modification of previously submitted packet

Please check this [confluence page](https://confluence.amd.com/display/AIE/AIE+Compiler+Artifacts)
shows a proposal of adding load_pdi as a IPU instruction. Once this proposal is landed, there is no explicit LoadPdi() function in c++ code. Instead, load PDI is specified in IPU instruction. 

What is the path forward if we use "xrt-lite"?

runtime/src/iree-amd-aie/driver/xrt-lite/device.cc
```diff
@@ -0,0 +1,296 @@
...
+    IREE_TRACE_ZONE_END(z0);
+  }
+};
+
+static iree_status_t iree_hal_xrt_lite_device_create_executable_cache(
+    iree_hal_device_t* base_value, iree_string_view_t identifier,
+    iree_loop_t loop, iree_hal_executable_cache_t** out_executable_cache) {
+  IREE_TRACE_ZONE_BEGIN(z0);
+
+  iree_hal_xrt_lite_device* device =
```
AWoloszyn, line 60: For type safety we often use a wrapper for these types of casts in iree. 
e.g.:
```

static iree_hal_hip_device_t* iree_hal_hip_device_cast(
    iree_hal_device_t* base_value) {
  IREE_HAL_ASSERT_TYPE(base_value, &iree_hal_hip_device_vtable);
  return (iree_hal_hip_device_t*)base_value;
}
```

runtime/src/iree-amd-aie/driver/xrt-lite/device.cc
AWoloszyn: The placement new doesn't really bother me here, what does trip me up is the dangling memory after the new-ed block that we fill with the identifier. This is what we do in C but feels weirder in C++.

runtime/src/iree-amd-aie/driver/xrt-lite/nop_executable_cache.cc
```diff
@@ -0,0 +1,106 @@
...
+
+iree_status_t iree_hal_xrt_lite_nop_executable_cache_create(
+    shim_xdna::device* shim_device, iree_string_view_t identifier,
+    iree_allocator_t host_allocator,
+    iree_hal_executable_cache_t** out_executable_cache) {
+  IREE_ASSERT_ARGUMENT(out_executable_cache);
+  *out_executable_cache = nullptr;
+  IREE_TRACE_ZONE_BEGIN(z0);
+
+  iree_hal_xrt_lite_nop_executable_cache* executable_cache = nullptr;
```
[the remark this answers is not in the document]
makslevental: hey i like that!

runtime/src/iree-amd-aie/driver/xrt-lite/device.cc
makslevental: yea ofc i saw this everywhere but elided it assuming it was an artifact all the other HALs being written in C. but now investigating `reinterpret_cast` I see that it is indeed not type safe. so i'll change this back.

runtime/src/iree-amd-aie/driver/xrt-lite/device.cc
makslevental: any alternative come to mind? in fact in this case (`iree_hal_xrt_lite_device`) i actually have `iree_string_view_t identifier;` in the wrong order/position.

runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc
makslevental: So this one wasn't removeable - I left a todo to figure out where this is supposed to be happening.

runtime/src/iree-amd-aie/driver/xrt-lite/util.h
```diff
@@ -20,6 +20,14 @@ iree_status_t unimplemented_ok_status(Params...) {
 }
 
 template <typename... Params>
-void unimplemented_ok_void(Params...) {}
+void unimplemented_ok_void(Params...){}
+#ifndef NDEBUG
+#define IREE_HAL_XRT_LITE_CHECKED_VTABLE_CAST(base_value, vtable, subvalue_t) \
+  (IREE_HAL_ASSERT_TYPE(base_value, &vtable),                                 \
```
AWoloszyn, line 26: FWIW `IREE_HAL_ASSERT_TYPE` should already be getting elided in a non-debug build. It would turn into 
```
while (false && (iree_hal_resource_is(....)) {
}
```
which the compiler "SHOULD" just be removed entirely.

runtime/src/iree-amd-aie/driver/xrt-lite/device.cc
```diff
@@ -0,0 +1,298 @@
...
+      base_device, iree_hal_xrt_lite_device_vtable, iree_hal_xrt_lite_device);
+
+  IREE_TRACE_ZONE_END(z0);
+  return iree_hal_xrt_lite_semaphore_create(device->host_allocator,
+                                            initial_value, out_semaphore);
+}
+
+static iree_status_t iree_hal_xrt_lite_device_queue_execute(
+    iree_hal_device_t* base_device, iree_hal_queue_affinity_t queue_affinity,
+    const iree_hal_semaphore_list_t wait_semaphore_list,
```
dezhiAmd, line 108: Are we supposed to use the "wait_semphore_list" or "signal_semaphore_list" for telling the host when the execution result is available?

makslevental: Please consult consult [iree/runtime/src/iree/hal/semaphore.c](https://github.com/iree-org/iree/blob/a28f76fdf3661196e30f4720c29ee4e2427ad415/runtime/src/iree/hal/semaphore.c).

runtime/src/iree-amd-aie/driver/xrt-lite/registration/driver_module.c
```diff
@@ -0,0 +1,144 @@
...
+#include "iree-amd-aie/driver/xrt-lite/registration/driver_module.h"
+
+#include "iree-amd-aie/driver/xrt-lite/api.h"
+#include "iree/base/api.h"
+#include "iree/base/internal/flags.h"
+
+IREE_FLAG(int32_t, xrt_lite_n_core_rows, 4,
+          "Number of core rows to use on NPU.");
+IREE_FLAG(int32_t, xrt_lite_n_core_cols, 4,
+          "Number of core cols to use on NPU.");
```
jtuyls, line 16: `4` is the default here? Shouldn't we just fail if not provided, as this might create issues in the future?

runtime/src/iree-amd-aie/driver/xrt-lite/shim/linux/kmq/kernel.cpp
```diff
@@ -0,0 +1,124 @@
+//
+// Created by mlevental on 10/11/24.
+//
```
jtuyls, line 3: Add a copyright header?

runtime/src/iree-amd-aie/driver/xrt-lite/shim/linux/kmq/kernel.h
```diff
@@ -0,0 +1,35 @@
+//
+// Created by mlevental on 10/11/24.
+//
```
jtuyls, line 3: Add a copyright header?

runtime/src/iree-amd-aie/driver/xrt-lite/shim/linux/kmq/shim_debug.cpp
```diff
@@ -0,0 +1,35 @@
+//
+// Created by mlevental on 10/3/24.
+//
```
jtuyls, line 3: Add a copyright header?

build_tools/ci/amdxdna_driver_utils/amdxdna_ioctl.py
```diff
@@ -0,0 +1,217 @@
...
+        ctypes.sizeof(struct_amdxdna_drm_query_aie_metadata),
+        get_void_ptr_to_struct(metadata).value,
+    )
+
+    fcntl.ioctl(drv_fd, ioctls.DRM_IOCTL_AMDXDNA_GET_INFO, info_params)
+
+    return format_struct(metadata)
+
+
+def get_core_n_rows(drv_fd):
```
jtuyls, line 146: Why can't we retrieve this in the HAL driver itself?

runtime/src/iree-amd-aie/driver/xrt-lite/driver.cc
```diff
@@ -0,0 +1,133 @@
+// Copyright 2024 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "iree-amd-aie/driver/xrt-lite/api.h"
+#include "iree-amd-aie/driver/xrt-lite/util.h"
+
+#define IREE_HAL_XRT_LITE_DEVICE_ID_DEFAULT 0
```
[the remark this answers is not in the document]
makslevental: I don't know; this is used to populate 

```c++
  static const iree_hal_device_info_t device_infos[1] = {
      {
          .device_id = IREE_HAL_XRT_LITE_DEVICE_ID_DEFAULT,
          .name = iree_string_view_literal("default"),
      },
  };
```

and I'm not sure what/where `device_id` is used - I mean I can grep and it's used for retrieval/association of various things. 

What I don't have clear in my head right this moment is how this will look when we/if we try to partition - my guess is these IDs capture/correspond to something like multiple GPUs, which would be one way to implement/model multiple partitions running concurrently on one AIE device. But then again maybe not :shrug:.

runtime/src/iree-amd-aie/driver/xrt-lite/device.cc
```diff
@@ -0,0 +1,285 @@
...
+#include "iree-amd-aie/driver/xrt-lite/api.h"
+#include "iree-amd-aie/driver/xrt-lite/device.h"
+#include "iree-amd-aie/driver/xrt-lite/direct_command_buffer.h"
+#include "iree-amd-aie/driver/xrt-lite/nop_executable_cache.h"
+#include "iree-amd-aie/driver/xrt-lite/nop_semaphore.h"
+#include "iree-amd-aie/driver/xrt-lite/util.h"
+#include "iree/hal/utils/deferred_command_buffer.h"
+#include "iree/hal/utils/deferred_work_queue.h"
+
+#define ARENA_BLOCK_SIZE (32 * 1024)
```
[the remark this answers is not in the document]
makslevental: Can't say - it is exactly the same for all HALs. Some of the HALs have this comment repasted

```
  // Total size of each block in the device shared block pool.
  // Larger sizes will lower overhead and ensure the heap isn't hit for
  // transient allocations while also increasing memory consumption.
  iree_host_size_t arena_block_size;
```

Note, the "block pool" this pertains is that used to allocate the [command buffer](https://github.com/iree-org/iree/blob/7dc8c26bbd50ca08b13e041e3ac510612a094e57/runtime/src/iree/hal/utils/deferred_command_buffer.c#L184-L186).

runtime/src/iree-amd-aie/driver/xrt-lite/registration/driver_module.c
makslevental: Is there a config where 4x4 isn't correct? I would rather have a poor, but not failing, experience in the case someone forgets the flag (it's quite annoying to me that these flags are globals/CLI rather than somehow passed/tunnelled through the IR).

build_tools/ci/amdxdna_driver_utils/amdxdna_ioctl.py
makslevental: You can - I have helper methods in [shim/linux/kmq/device.cpp](https://github.com/nod-ai/iree-amd-aie/blob/fc364b727c72050c20aceaaaaff3550a94a3c4e9/runtime/src/iree-amd-aie/driver/xrt-lite/shim/linux/kmq/device.cpp#L252-L275) that do the exact same things but I had the feeling that, since this is core config data, it should be passed in. I.e., you might want to override at some point (for some reason).

runtime/src/iree-amd-aie/driver/xrt-lite/registration/driver_module.c
makslevental: Maybe you're right tho

jtuyls: > it's quite annoying to me that these flags are globals/CLI rather than somehow passed/tunnelled through the IR

Yeah, we should be able to do that in the future.

runtime/src/iree-amd-aie/driver/xrt-lite/driver.cc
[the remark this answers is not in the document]
makslevental: yea https://github.com/nod-ai/iree-amd-aie/blob/977edb6f9dbbb2ca8d97d7aaeb3ce96c2288d254/runtime/src/iree-amd-aie/driver/xrt-lite/shim/linux/kmq/amdxdna_accel.h#L413 - that's the "query" struct and I can't find right now where/how the setting of the `start_col` works but it's in there somewhere (or possibly it's one of their hidden features - i.e., only visible in the TurinAI/amd-aie repo).

runtime/src/iree-amd-aie/driver/xrt-lite/registration/driver_module.c
makslevental: Okay I've made them required and added checks/fails if they're not provided.

runtime/src/iree-amd-aie/driver/xrt-lite/driver.cc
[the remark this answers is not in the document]
makslevental: a little experiment: I added another entry to `device_infos` and I got this

```
$ iree-run-module --xrt_lite_n_core_rows=4 --xrt_lite_n_core_cols=4 --dump_devices

# ============================================================================
# Enumerated devices for driver 'xrt-lite'
# ============================================================================

# ===----------------------------------------------------------------------===
# --device=xrt-lite://
#   default
# ===----------------------------------------------------------------------===
# ===----------------------------------------------------------------------===
# --device=xrt-lite://
#   default-2
# ===----------------------------------------------------------------------===
```

**useful note**: `--xrt_lite_n_core_rows --xrt_lite_n_core_cols` have have to be passed first because of how [flag parsing works](https://github.com/iree-org/iree/blob/dc6f0cd5f4a784531bde3dcac48f2effc28c4224/runtime/src/iree/base/internal/flags.c#L526)

jtuyls (approved):
LGTM

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
```diff
@@ -54,6 +54,9 @@ struct AMDAIEOptions {
   std::string enableAMDAIEUkernels{"none"};
   bool enablePacketFlow{false};
 
+  enum class DeviceHAL { XRT, XRT_LITE };
+  DeviceHAL deviceHal{DeviceHAL::XRT};
```
makslevental, line 58:
```suggestion
  DeviceHAL deviceHal{DeviceHAL::XRT_LITE};
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -269,56 +346,69 @@ LogicalResult AIETargetBackend::serializeExecutable(
...
+    if (ordinalCount > 1) {
       llvm::sys::path::append(entryPointWorkDir, entryPointNamesFb[ordinal]);
-    auto err = llvm::sys::fs::create_directories(entryPointWorkDir);
-    if (err)
+    }
+
+    if (auto err = llvm::sys::fs::create_directories(entryPointWorkDir)) {
       return moduleOp.emitOpError()
-             << "failed to create working directory for xclbin generation: "
+             << "failed to create working directory for pdi generation: "
```
makslevental, line 400:
```suggestion
             << "failed to create working directory for artifact generation: "
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -269,56 +346,69 @@ LogicalResult AIETargetBackend::serializeExecutable(
...
              << err.message();
+    }
     llvm::outs().flush();
-    SmallString<128> xclbinPath(entryPointWorkDir);
-    llvm::sys::path::append(xclbinPath, entryPointNamesFb[ordinal] + ".xclbin");
+
+    SmallString<128> artifactPath(entryPointWorkDir);
+    llvm::sys::path::append(artifactPath, entryPointNamesFb[ordinal] + ".pdi");
```
makslevental, line 406:
```suggestion
    switch (options.deviceHal) {
      case AMDAIEOptions::DeviceHAL::XRT:
        llvm::sys::path::append(artifactPath, entryPointNamesFb[ordinal] + ".xclbin");
        break;
      case AMDAIEOptions::DeviceHAL::XRT_LITE:
        llvm::sys::path::append(artifactPath, entryPointNamesFb[ordinal] + ".pdi");
        break;
      default:
        llvm::errs() << "Unsupported device HAL\n";
        return failure();
    }
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -269,56 +346,69 @@ LogicalResult AIETargetBackend::serializeExecutable(
...
     SmallString<128> npuInstPath(entryPointWorkDir);
     llvm::sys::path::append(npuInstPath,
                             entryPointNamesFb[ordinal] + ".npu.txt");
 
-    // Convert ordinal to hexadecimal string for xclbin kernel id.
+    // Convert ordinal to hexadecimal string for pdi kernel id.
```
makslevental, line 411:
```suggestion
    // Convert ordinal to hexadecimal string for kernel id.
```

build_tools/ci/run_matmul_test.sh
```diff
@@ -147,6 +148,8 @@ cd ${OUTPUT_DIR}
 export MATMUL_TESTS_RUN=0
 export MATMUL_TESTS_FAILS=0
 
+DEVICE_HAL="${DEVICE_HAL:-xrt-lite}"
```
dezhiAmd, line 151: In order to run this test on windows, do we need to change this line so DEVICE_HAL can be xrt?

makslevental: https://github.com/nod-ai/iree-amd-aie/blob/main/.github/workflows/ci-windows.yml#L180

dezhiAmd: On the CI computer, this link of course is a good solution. But can we make it easier for developers to run this script on a local machine?

runtime/src/iree-amd-aie/driver/xrt/cts/CMakeLists.txt
```diff
@@ -0,0 +1,114 @@
+# Copyright 2024 The IREE Authors
```
dezhiAmd, line 1: Why we need to add a CTS test for XRT HAL driver in this PR?

This not mentioned in PR description. What does the CTS testdo?

Refer to the PR description below:

```
Let's try this again: this is a HAL that skips/avoids/elides XRT completely. It does so by (essentially) going straight to the [XDNA driver's (kmq) shim](https://github.com/amd/xdna-driver/tree/main/src/shim) which talks directly to the kernel module. Note that XDNA's shim actually couples tightly to XRT itself so the bulk of the actual "work" here is/was decoupling. Thus, we vendor the decoupled shim. Just to be explicit/clear: after this PR, on Linux, the only thing needed in the environment is the [kernel module itself](https://github.com/amd/xdna-driver/tree/main/src/driver/amdxdna) - not XRT, nor XDNA's shim.

Call outs:

Only Linux for now (Windows soon);
On Linux, we no longer even need to clone XRT (completely really does mean completely);
We no longer use .xclbin as the artifact and now just load the .pdi directly;
This uses the so-called kernel-managed queue (kmq) path; user-managed queues (umq) aren't supported until aie4;
The current implementation here is still sync and still uses direct (rather than indirect) command buffers; this PR is already large/complex enough without changing the paradigms. The next PR will add/implement those;
Nothing about how drivers themselves are bumped needs to change - you/anyone can continue to install the XDNA drivers using their full instructions there and everything will work the same, i.e., the XRT pieces will still be skipped even if you have them installed. But I will probably put together a script that builds/installs only their kernel module for those that want to opt-out of XRT in their environment;
For now, the XRT HAL will still work on Linux but it should be considered deprecated (hence no longer tested in CI).
TODOs before landing:

 remove instructional comments leftover from the null HAL;
 undo "OO"-ness of the HAL
 add missing trace zones
 remove smart pointers where possible
 move hwq into kernel
 remove exceptions from the shim.
 parameterize number of cols/rows (currently hardcoded 4x4 for Phoenix);
```

build_tools/ci/run_matmul_test.sh
makslevental: For the final time: XRT is deprecated so I am not spending absolutely any time documenting/fixing/describing/explaining anything about it.

> So not all developers need to read the CI script to duplicate the effort?

Developers are responsible for reading code relevant to their jobs.

PR #615, opened 2024-07-31
[ObjectFifo] Add InsertTemporaryBuffer Pass

yzhang93:
This PR adds a fix for matmul-elementwise in objectFifo pipeline. 

In particular, the access op with `None` type (shown below) was not handled by the current pipeline, and thus caused failure in LowerToAIEPass.
```
%52 = amdaie.logicalobjectfifo.access(%21, None) : !amdaie.logicalobjectfifo<memref<1x1x8x8x4x4xi32, strided<[2048, 1024, 128, 16, 4, 1], offset: ?>, 2 : i32>> -> memref<1x1x8x8x4x4xi32, strided<[2048, 1024, 128, 16, 4, 1], offset: ?>, 2 : i32>
linalg.fill ins(%c0_i32 : i32) outs(%52 : memref<1x1x8x8x4x4xi32, strided<[2048, 1024, 128, 16, 4, 1], offset: ?>, 2 : i32>)
```

This PR adds a temporary buffer and replaces the users of such access ops inside each core.

PR #817, opened 2024-10-02
Do not insert None accesses (rather than create and then destroy them later)

newling:
A pass was introduced in https://github.com/nod-ai/iree-amd-aie/pull/615 to eliminate None access patterns. This PR simplifies the stack, by instead just not inserting None accesses in the first place. 

Error in one of the e2e tests is 

```ld.lld: error: undefined symbol: malloc```

Weird. I'll try this again once https://github.com/nod-ai/iree-amd-aie/pull/814 lands

### CODE REVIEW

File: compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.td [hunk 1]
```diff
@@ -59,6 +59,7 @@ def LogicalObjectFifoPort: I32EnumAttr<"LogicalObjectFifoPort",
 def MemoryAccess: I32EnumAttr<"MemoryAccess",
   "The memory access type",
   [
+    // TODO(newling) remove None (requires test factoring)
```
jtuyls, line 62: There is a use for None accesses in peeled matmul (in prolog/main), where it will be used for an access which is nor read nor write, but is later for example consolidated with a write after the peeled computation is put together inside the cores. You can have a loop at the `mlir-after-all` to see that for a standalone matmul, those none accesses will be removed before `AMDAIENoneAccessToTemporaryBuffer` pass. So, I don't think removing `None` access will work as is and I also don't see why we should remove it?

jtuyls: It would be easier to combine the `AMDAIENoneAccessToTemporaryBuffer` and `AMDAIETemporaryAllocBufferizationPass` into a single one. Then you would go `None` -> `amdaie.buffer` directly.

newling: Ok. My goal here is just to simplify, to rather go straight to the final state if possible. i.e. why insert complex layering around a memref, if the layering is just going to be removed, rather just don't put the layering on it the first place...

Your explanation sounds a bit dodgy or I'm not following (probably the latter :)). So... if the None layering is not inserted, then the final IR is incorrect? So None access layering cannot be removed at any point?  TBH I have an extremely poor understanding of what the 4 access patterns actually mean (and I can't find any documentation for it). 

Comparing the IR I get with this PR vs that on main, I can see some locks are missing with this PR. But why should None access result in locks? They seem to be needed, as the numerics with this PR are off.

jtuyls: > i.e. why insert complex layering around a memref, if the layering is just going to be removed, rather just don't put the layering on it the first place...

Yeah, this layering is not just inserted and then removed. The logical objectfifo layering around the memrefs, describes a synchronized memref with producers and consumers. An access to an objectfifo is is not like a an access to a memref, in that the necessary synchronization stubs need to be inserted (to for example wait for the producer, or to release the data to a consumer).

> Your explanation sounds a bit dodgy or I'm not following (probably the latter :)). So... if the None layering is not inserted, then the final IR is incorrect?

Yes, then the IR is incorrect as the synchronization ops are missing. Note that this is not case for accesses hat end up being fully internal (temporary buffers), and they are converted into locally allocated memref and then into `amdaie.buffers`s.

> So None access layering cannot be removed at any point? TBH I have an extremely poor understanding of what the 4 access patterns actually mean (and I can't find any documentation for it).

The access patterns describe an access with respect to a logical objectFifo, i.e. is the access a producer (write) or consumer (read), or undetermined (None) at that point.

newling: So the pass https://github.com/nod-ai/iree-amd-aie/pull/615 is making an assumption that all None accesses are, at the point it is run, of a special type ("fully internal") ? That should really be tested for, or at least documented. 

>  undetermined (None)
That's confusing, 'None' suggests no read and no write, why is not called 'Undetermined' ?

jtuyls: > So the pass https://github.com/nod-ai/iree-amd-aie/pull/615 is making an assumption that all None accesses are, at the point it is run, of a special type ("fully internal") ? That should really be tested for, or at least documented.

Yes.

> That's confusing, 'None' suggests no read and no write, why is not called 'Undetermined' ?

It's also used for temporary buffers as discussed here, which are `None` with respect to synchronization. `None` is kind of the sentinel value for this. Maybe `Undetermined` is a slightly better name though because a temporary buffer could be `Undetermined` initially as well..

newling: Got it thanks. I suppose everything can pass for `Undetermined` if you squint hard enough! But then what's the difference between `Undetermined` and `Any`. Shouldn't everything be Read and Write until proven otherwise (has a traditional fixed point algo feel to it?) i.e. 4 categories. [ReadWrite, Read, Write, None] where:
ReadWrite : might be read, might be write
Read : definitely not write, might be read
Write : definitely not read, might be write
None : definitely not read, definitely not write.

jtuyls: > ReadWrite : might be read, might be write

I think I prefer `Any` for this. `ReadWrite` makes me think about both read and write, which while I don't see an actual use case for it now, can in principal be implemented.

PR #818, opened 2024-10-02
Don't capture this unnecessarily 

newling:
Introduced and unaddressed in https://github.com/nod-ai/iree-amd-aie/pull/812


makslevental (approved):
PR title sounds like a lyric from a 2000s boyband song lol

yzhang93:
I added "this" because it didn't pass the options correctly to the rewritter function in my first attempt with the change. But I guess some other changes fixed it and now it works without passing "this"! Thanks for making the change.

PR #819, opened 2024-10-02
[DmaLoopSubsumption] Add checks for all dims after inserting new access pattern dimension

yzhang93:
This PR adds checks for all dimensions of new sizes/strides after insertion to make sure they don't exceed the max value. Without this PR, it only checks the inserted dimension, and ignores the case that inter size becomes intra size after insertion which could exceed the max value.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/dma_loop_subsumption.mlir
```diff
@@ -428,6 +428,62 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
 
 // -----
 
+// Don't subsume if original inter size becomes new intra size and is too large.
```
jtuyls, line 431:
```suggestion
// Don't subsume if a dimension's index would change (by adding new dimensions) and the size becomes too large for the new dimension's max value.
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/dma_loop_subsumption.mlir
```diff
@@ -428,6 +428,62 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
...
+        amdaie.end
+      }
+    }
+    return
+  }
+}
+
+// -----
+
+// Don't subsume if original inter size becomes new intra size and is too large.
```
jtuyls, line 459:
```suggestion
// Don't subsume if a dimension's index would change (by adding new dimensions) and the size becomes too large for the new dimension's max value.
```

jtuyls (changes requested):
Could you add the following as well to the `DmaDimConfig` to model that the last/outermost intra dimension can have any size, and add tests for it: 

```
stepSizes[1] = std::numeric_limits<uint32_t>::max();
```

Here:

- https://github.com/nod-ai/iree-amd-aie/blob/cd5cdea472bd768da2321e22fb60604f48a642c3/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaUtils.h#L345
- https://github.com/nod-ai/iree-amd-aie/blob/cd5cdea472bd768da2321e22fb60604f48a642c3/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaUtils.h#L353

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaLoopSubsumption.cpp
```diff
@@ -289,17 +289,36 @@ struct SubsumeLoopIntoDMA
...
+        // they are not out of the range.
+        std::optional<SmallVector<int64_t>> maybeNewSourceSizes =
+            getConstantIntValues(newSourceSizes);
+        std::optional<SmallVector<int64_t>> maybeNewSourceStrides =
+            getConstantIntValues(newSourceStrides);
+        assert(maybeNewSourceSizes.has_value() &&
+               "expect constant source size values");
+        assert(maybeNewSourceStrides.has_value() &&
+               "expect constant source stride values");
+        SmallVector<int64_t> newSourceSizesInt = maybeNewSourceSizes.value();
```
newling, line 303: nit: value() returns a reference, but here you make a copy, consider  `const &`?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaLoopSubsumption.cpp
```diff
@@ -289,17 +289,36 @@ struct SubsumeLoopIntoDMA
                 insertSourceStrides, updateSourceOffsets))) {
           return failure();
         }
+
```
newling, line 292: What about a 

```
auto getAppended = [](ArrayRef<OpFoldResult> prefix, ArrayRef<int64_t> suffix) -> SmallVector<int64_t> {
  // ...
};
```

to reduce dup? Then

```
auto newSourceSizesInt = getAppended(newSourceSizes, insertSourceSizes); 
auto newSourcesStridesInt = getAppended(...);
...

File: compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/dma_loop_subsumption.mlir [hunk 5]
```diff
@@ -514,12 +556,6 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
         scf.for %arg5 = %c0 to %c64 step %c1 {
           %2 = amdaie.npu.dma_cpy_nd %0([0, 0, 0] [1, 8, 16] [128, 16, 1], [] [] [])
         }
-        scf.for %arg6 = %c0 to %c1023 step %c1 {
-          %3 = amdaie.npu.dma_cpy_nd %0([0, 0] [8, 16] [16, 1], [] [] [])
-        }
-        scf.for %arg7 = %c0 to %c1024 step %c1 {
```
jtuyls, line 520: I think it's useful to keep this one (or create a new test with a large loop extent) as the subsumption should happen in this case.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/dma_loop_subsumption.mlir
```diff
@@ -463,12 +516,6 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
         scf.for %arg5 = %c0 to %c64 step %c1 {
           %2 = amdaie.npu.dma_cpy_nd %0([] [] [], [0, 0, 0] [1, 8, 16] [128, 16, 1])
         }
-        scf.for %arg6 = %c0 to %c1023 step %c1 {
-          %3 = amdaie.npu.dma_cpy_nd %0([] [] [], [0, 0] [8, 16] [16, 1])
-        }
-        scf.for %arg7 = %c0 to %c1024 step %c1 {
```
jtuyls, line 469: I think it's useful to keep this one (or create a new test with a large loop extent) as the subsumption should happen in this case.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaLoopSubsumption.cpp
yzhang93: Very good point! I've made the change.

jtuyls (approved):
LGTM

PR #822, opened 2024-10-03
[Matmul+Truncf] Enable Matmul+Truncf for shorter shape on Pack-Peel + Objectfifo

Abhishek-Varma:
-- This commit includes `arith.truncf`, `vector.transfer_read` and `vector.transfer_write` into `amdaie.core` op.
-- This is required to make "Matmul + truncf" work with vectorization enabled for `arith.truncf` op.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_cores.mlir
```diff
@@ -300,3 +300,28 @@ module {
...
+// -----
+
+// CHECK-LABEL: @insert_truncf_within_core
+//       CHECK:     scf.forall
+//       CHECK:       amdaie.tile
+//       CHECK:       amdaie.core
+//       CHECK:         vector.transfer_read
+//       CHECK:         arith.truncf
+//       CHECK:         vector.transfer_write
+//       CHECK:         amdaie.end
```
jtuyls, line 313: I prefer that we don't indent the `CHECK` further than one space from `//` to be consistent with the rest of the codebase (and this file).

jtuyls (commented):
Could you add an e2e test?

Abhishek-Varma:
> Could you add an e2e test?

This is the first PR in the series of PRs that need to go in before I can add an e2e test for "Matmul + truncf" with the vectorization enabled for elementwise.

Abhishek-Varma:
As discussed offline with @jtuyls - I have push most of the changes required in the same PR.

Will be raising a separate PR for flattening `arith.truncf` and once that goes in I'll add an e2e test in this PR.

Will rename the PR title and description accordingly - so marking this as a draft.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_loops_for_vectorization.mlir
```diff
@@ -246,4 +227,30 @@ module {
   }
 }
 
+// -----
 
+// CHECK-LABEL: @element_wise
+module {
+  func.func @element_wise(%arg0: tensor<1x1x4x4x4x4xf32>, %arg1: tensor<1x1x4x4x4x4xbf16>) -> tensor<1x1x4x4x4x4xbf16>{
```
newling, line 234: Can you go over this one more time an simplify as much as possible? (3 dimensions suffice here?). 

If you have have shape 4x6x8 then you can also check that the loop is on the dimension of size 4 (I think that's what your code is supposed to do).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -66,12 +66,29 @@ class AMDAIEInsertLoopsForVectorizationPass
...
                              IRRewriter &rewriter) {
+    if (isa<linalg::CopyOp, linalg::FillOp>(genericOp)) return failure();
+
     auto iteratorTypes = genericOp.getIteratorTypesArray();
     auto numIterators = iteratorTypes.size();
 
     // No outer dimensions to tile if fewer than 4 iterators.
     if (numIterators < 4) return failure();
 
+    // Enable generating loops for vectorization in case of element-wise ops.
```
newling, line 77: Can you add a comment motivating the decision to tile all but the final 2 dimensions? 

Is this because the final 2 dimensions correspond to the m and n dimensions which are correspondingly tiled in this pass for matmul?

Abhishek-Varma:
Here's the short shaped [Matmul + truncf e2e IR log](https://gist.github.com/Abhishek-Varma/5519d5f61864ac5e7130c9db2d66236a) that this PR currently enables (the bigger shapes need to be addressed incrementally who e2e IR, in case someone wants to take a look, is [here](https://gist.github.com/Abhishek-Varma/966ccadcaa838520b3d7224e582561f4)) and the numerics were verified locally.

But for the e2e test via `cpu_comparisons/run.py` I get the following for the CPU run itself let alone AIE :-

```
iree/runtime/src/iree/tooling/numpy_io.c:419:
   UNIMPLEMENTED; unsupported data encoding; outputting results; processing function outputs;
   `sync func @matmul_truncf(%input0: tensor<32x32xbf16>, %input1: tensor<32x32xbf16>) -> (%output0: tensor<32x32xbf16>)`
```

The above is not a verification issue at the MLIR level, else it'd have bailed out earlier.

I checked the input MLIR file that's getting generated locally but see no issue with it :-
```
// input 32x32xbf16
// input 32x32xbf16

func.func @matmul_truncf(%arg0: tensor<32x32xbf16>, %arg1: tensor<32x32xbf16>) -> tensor<32x32xbf16>
{
  %cst = arith.constant 0.0 : f32
  %0 = tensor.empty() : tensor<32x32xf32>
  %1 = linalg.fill ins(%cst : f32) outs(%0 : tensor<32x32xf32>) -> tensor<32x32xf32>
  %2 = linalg.matmul ins(%arg0, %arg1 : tensor<32x32xbf16>, tensor<32x32xbf16>)
    outs(%1: tensor<32x32xf32>) -> tensor<32x32xf32>
  %3 = arith.truncf %2 : tensor<32x32xf32> to tensor<32x32xbf16>
  return %3: tensor<32x32xbf16>
}
```

Am I missing some other template to include?

newling:
> Here's the short shaped [Matmul + truncf e2e IR log](https://gist.github.com/Abhishek-Varma/5519d5f61864ac5e7130c9db2d66236a) that this PR currently enables (the bigger shapes need to be addressed incrementally who e2e IR, in case someone wants to take a look, is [here](https://gist.github.com/Abhishek-Varma/966ccadcaa838520b3d7224e582561f4)) and the numerics were verified locally.
> 
> But for the e2e test via `cpu_comparisons/run.py` I get the following for the CPU run itself let alone AIE :-
> 
> ```
> iree/runtime/src/iree/tooling/numpy_io.c:419:
>    UNIMPLEMENTED; unsupported data encoding; outputting results; processing function outputs;
>    `sync func @matmul_truncf(%input0: tensor<32x32xbf16>, %input1: tensor<32x32xbf16>) -> (%output0: tensor<32x32xbf16>)`
> ```
> 
> The above is not a verification issue at the MLIR level, else it'd have bailed out earlier.
> 
> I checked the input MLIR file that's getting generated locally but see no issue with it :-
> 
> ```
> // input 32x32xbf16
> // input 32x32xbf16
> 
> func.func @matmul_truncf(%arg0: tensor<32x32xbf16>, %arg1: tensor<32x32xbf16>) -> tensor<32x32xbf16>
> {
>   %cst = arith.constant 0.0 : f32
>   %0 = tensor.empty() : tensor<32x32xf32>
>   %1 = linalg.fill ins(%cst : f32) outs(%0 : tensor<32x32xf32>) -> tensor<32x32xf32>
>   %2 = linalg.matmul ins(%arg0, %arg1 : tensor<32x32xbf16>, tensor<32x32xbf16>)
>     outs(%1: tensor<32x32xf32>) -> tensor<32x32xf32>
>   %3 = arith.truncf %2 : tensor<32x32xf32> to tensor<32x32xbf16>
>   return %3: tensor<32x32xbf16>
> }
> ```
> 
> Am I missing some other template to include?

No this is all correct. I think the issue is that iree-run-module doesn't support writing bfloat16 values. I'll investigate further.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -758,6 +758,30 @@ def run(self, config):
                 lower_to_aie_pipeline="objectFifo",
             )
 
+        # Test(s) of the form matmul(A,B) + truncf(C) where A:MxK, B:KxN and C:MxN
+        template_name = test_files_dir / "matmul_truncf_32x32x32_bf16_bf16.mlir"
+        identity_mat = np.eye(32, dtype=np.float32)
+        ones = np.ones(32 * 32, dtype=np.float32).reshape([32, 32])
+        lhs = ones * 192
+        input_args = generate_inputs(
+            template_name, output_dir, 1, {1: lhs, 2: identity_mat}
```
jtuyls, line 767: Could we do `rhs == identity_matrix * 2` for example to make the expected output different from the input? We should also think whether these inputs are not too straightforward, resulting in only testing a very specific case. With `rhs == identity_matrix * 2` it looks reasonable though as an initial test.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_loops_for_vectorization.mlir
```diff
@@ -137,25 +137,6 @@ module {
     return %0 : !t3_f32
   }
 
-
-  // A check that a linalg.generic where the number of operands is not 3, does
```
jtuyls, line 141: Why is this removed?

build_tools/ci/cpu_comparison/run.py
Abhishek-Varma: Sure. I've added the change.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_loops_for_vectorization.mlir
Abhishek-Varma: I have just replaced it with another test below.
Both the tests turned out to be similar anyway so removed the former one and kept the one that mimics `arith.truncf` IR behavior instead.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -758,6 +758,31 @@ def run(self, config):
                 lower_to_aie_pipeline="objectFifo",
             )
 
+        # Test(s) of the form matmul(A,B) + truncf(C) where A:MxK, B:KxN and C:MxN
+        template_name = test_files_dir / "matmul_truncf_32x32x32_bf16_bf16.mlir"
+        identity_mat = np.eye(32, dtype=np.float32)
+        ones = np.ones(32 * 32, dtype=np.float32).reshape([32, 32])
```
newling, line 764: Because truncf(192 * 2) = 192 * 2, this isn't testing that truncf is doing the right thing (the compiler might just be ignoring it!). Can you choose a matmul where truncf is required, maybe

lhs = ones * 101
rhs = identity_mat * 3 

truncf(101 * 3) = 302  (because 101*3 = 303, and 303 is in range [256, 512), the truncf round it down to the nearest power of 2. 

 nit: can you make the tensor have sizes m=n=k=8 (just to keep it as small and simple as possible)

build_tools/ci/cpu_comparison/matmul_template/matmul_truncf_MxK_KxN.mlir
```diff
@@ -0,0 +1,13 @@
+// input ${M}x${K}x${TYPE1}
+// input ${K}x${N}x${TYPE1}
+
```
newling, line 3: You have a template version of the test and a non-template version -- you can remove one of them I think?

newling (changes requested):
Looks good. My only comments are wrt the e2e testing.

build_tools/ci/cpu_comparison/run.py
Abhishek-Varma: Thanks! I've made the changes. 8x8x8 was failing compilation so added the change required for that too. Please take a look.

build_tools/ci/cpu_comparison/matmul_template/matmul_truncf_MxK_KxN.mlir
Abhishek-Varma: Have removed the non-template version.

newling (approved):
Looks good, thanks!

PR #824, opened 2024-10-03
[AMDAIETemporaryAllocBufferization] Bufferize temporary allocs that aren't defined in core

newling:
Update the pass to handle `memref.alloc` in any scope, not just in `amdaie.core`

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETemporaryAllocBufferization.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETemporaryAllocBufferization.cpp
@@ -13,47 +13,82 @@ namespace mlir::iree_compiler::AMDAIE {
 
 namespace {
 
-static std::optional<BufferOp> createBufferForTemporaryAllocOp(
-    IRRewriter &rewriter, WorkgroupOp workgroupOp, memref::AllocOp allocOp,
-    CoreOp coreOp, unsigned index) {
-  OpBuilder::InsertionGuard g(rewriter);
-  TileOp tileOp = coreOp.getTileOp();
-  // Reset rewriter's location to after last tile's declaration.
-  auto tiles = workgroupOp.getBody()->getOps<TileOp>();
-  assert(!tiles.empty() && "no tiles in workgroupOp");
-  rewriter.setInsertionPointAfter(*std::prev(tiles.end(), 1));
-  auto bufferType = cast<MemRefType>(allocOp.getType());
-  auto bufferOp = rewriter.create<AMDAIE::BufferOp>(
-      rewriter.getUnknownLoc(), bufferType, tileOp, nullptr);
-  return bufferOp;
-}
+///
+/// Some memref allocations are accessed directly, as a result of the pass
+///
+/// `iree-amdaie-none-access-to-temporary-buffer`
+///
+/// Thie memref.allocs need to be converted to amdaie.buffer ops
+///
+///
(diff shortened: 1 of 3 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/temporary_alloc_bufferization.mlir
```diff
@@ -86,3 +76,34 @@ func.func @temp_buffer() {
...
+//   CHECK-DAG:   %[[C0:.*]] = arith.constant 0 : index
+//   CHECK-DAG:   %[[C1:.*]] = arith.constant 1 : index
+//   CHECK-DAG:   %[[TILE_0:.*]] = amdaie.tile(%[[C0]], %[[C0]])
+//   CHECK-DAG:   %[[BUFFER_0:.*]] = amdaie.buffer(%[[TILE_0]]) : memref<4xi32>
+//   CHECK-DAG:   %[[TILE_1:.*]] = amdaie.tile(%[[C0]], %[[C1]])
+//   CHECK-DAG:   %[[BUFFER_1:.*]] = amdaie.buffer(%[[TILE_1]]) : memref<4xi32>
+//       CHECK:   amdaie.core(%[[TILE_0]]
+//       CHECK:     %[[CAST:.*]] = memref.reinterpret_cast %[[BUFFER_0]]
+//       CHECK:   amdaie.core(%[[TILE_1]]
+//       CHECK:     %[[CAST:.*]] = memref.reinterpret_cast %[[BUFFER_1]]
```
jtuyls, line 92: I would prefer for style consistency that we don't indent the `CHECK` and place them one space after `//` as we do in the rest of the codebase (except for this file apparently :P).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETemporaryAllocBufferization.cpp
```diff
@@ -13,47 +13,42 @@ namespace mlir::iree_compiler::AMDAIE {
...
-    SmallVector<Operation *> &toBeErased) {
-  // Step 1. Get all buffers within a CoreOp.
-  SmallVector<memref::AllocOp> allocOps;
-  coreOp.walk([&](Operation *op) {
-    if (auto allocOp = dyn_cast<memref::AllocOp>(op)) {
-      allocOps.push_back(allocOp);
-      toBeErased.push_back(allocOp);
-    } else if (auto deallocOp = dyn_cast<memref::DeallocOp>(op)) {
-      toBeErased.push_back(deallocOp);
+LogicalResult bufferizeTemporaryMemrefs(Operation *parentOp) {
```
jtuyls, line 16: While this is an improvement, I think we should go from `access` directly to `amdaie.buffer` and avoid the alloc intermediary.

jtuyls (approved):
LGTM

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETemporaryAllocBufferization.cpp
newling: Yes ok I think that makes sense, I'll pick that up next.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/temporary_alloc_bufferization.mlir
newling: Noted. :P

Issue #820, opened 2024-10-03
Numerics issue with vectorized conv2d


newling:
Enabling vectorization (see https://github.com/nod-ai/iree-amd-aie/pull/789) for convolution results in numerical failure. The values are off only slightly (although they are definitely not correct, there is no floating point rounding issue here). Experiments with the input values suggests that the problem is that the input image data (i.e. not the kernel data) is being read incorrectly, with an incorrect offset inside the scf.for loop (not confirmed). 

Some things I've tried: 

Setting optimization flags in LLVM to 0 (the default is 2) has no effect. 
Inverting the scf.for loop order has no effect.
Using different versions of peano has no effect. 

This task is to find the source of the error, and if it's peano create a reproducer for the team. 

Attached are the `ll` and `opt.ll` files (vectorized and unvectorized)  : [ll_files.zip](https://github.com/user-attachments/files/17248940/ll_files.zip) (they're quite small, vectorized_input.opt.ll is 94 lines only) 

The MLIR IR looks basically identical except for the inner-loop. 

// With vectorization:

```
scf.for %arg1 = %c0 to %c3 step %c1 {
  scf.for %arg2 = %c0 to %c3 step %c1 {
    scf.for %arg3 = %c0 to %c4 step %c1 {
      %0 = vector.transfer_read %reinterpret_cast_21[%c0, %arg1, %arg3, %arg2, %c0], %cst {in_bounds = [true, true]} : memref<1x3x4x6x8xbf16>, vector<4x8xbf16>
      %1 = vector.transfer_read %reinterpret_cast_22[%arg1, %arg2, %arg3, %c0, %c0, %c0], %cst {in_bounds = [true, true]} : memref<3x3x4x1x8x4xbf16>, vector<8x4xbf16>
      %2 = vector.transfer_read %reinterpret_cast[%c0, %c0, %c0, %c0, %c0], %cst_20 {in_bounds = [true, true], permutation_map = affine_map<(d0, d1, d2, d3, d4) -> (d2, d4)>} : memref<1x1x4x1x4xf32>, vector<4x4xf32>
      %3 = arith.extf %0 : vector<4x8xbf16> to vector<4x8xf32>
      %4 = arith.extf %1 : vector<8x4xbf16> to vector<8x4xf32>
[...]

PR #825, opened 2024-10-03
Remove load alignment removal pass 

newling:
If I remove this pass the optimized IR (see https://github.com/nod-ai/iree-amd-aie/issues/820 file vectorized_input.opt.ll) changes from 

```
  %9 = load <32 x bfloat>, ptr getelementptr inbounds ([144 x bfloat], ptr @buff_6, i20 0, i20 8), align 64
```

to 

```
  %9 = load <32 x bfloat>, ptr getelementptr inbounds ([144 x bfloat], ptr @buff_6, i20 0, i20 8), align 2
```

which to llvm-novice brain looks more sensible (how you the aligment be 64 when the offset from the pointer is 8??)

Also, this makes conv numerics pass :D


newling:
Seems like the pass is still needed for matmul.

PR #826, opened 2024-10-04
[DmaLoopSubsumption] Relax circular dma loop subsumption condition

yzhang93:
1. This PR relaxes the condition for circular dma ops loop subsumption, so that npu.circular_dma_cpy_nd ops can be hoisted out of the loop even if there is other npu.dma_cpy_nd user of the same connection op after it.
2. With this change, we can further subsume loops and hoist npu.dma_cpy_nd ops out of the loop. This PR makes use of https://github.com/nod-ai/iree-amd-aie/pull/812 and brings the dma optimizations in Passes.cpp.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -588,6 +591,8 @@ void addAMDAIEObjectFifoLoweringPasses(OpPassManager &passManager,
   passManager.addPass(createCanonicalizerPass());
 
   passManager.addPass(createAMDAIEDmaCompositionPass());
+  passManager.addPass(createAMDAIECanonicalizeDoublyStridedOpPass());
+  passManager.addPass(createAMDAIEDmaLoopSubsumptionPass());
```
jtuyls, line 595: Can't we put the canonicalization inside the `DmaComposition` pass to avoid `DmaComposition` -> `Canonicalization` -> `DmaLoopSubsumption`?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/dma_loop_subsumption_circular.mlir
```diff
@@ -58,3 +58,95 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
...
+          %2 = amdaie.npu.circular_dma_cpy_nd %0([0] [2048] [1], [] [] [])
+          %3 = amdaie.npu.dma_cpy_nd %0([] [] [], [0, %1] [32, 64] [128, 1])
+          %4 = amdaie.npu.circular_dma_cpy_nd %0([0] [2048] [1], [] [] [])
+        }
+        amdaie.end
+      }
+    }
+    return
+  }
+}
```
jtuyls, line 152: Ad new line

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaLoopSubsumption.cpp
```diff
@@ -501,6 +501,25 @@ struct SubsumeLoopIntoDMA
       return false;
     };
 
+    auto hasCircularUsersInSameScope =
```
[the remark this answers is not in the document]
yzhang93: I don't know how to make it work with what you suggested. I think it is a little confusing to use `currentUser`, what I mean is the current dma op that is being checked. I should probably use `currentCircularDma`.

The logic is checking the users in the same scope in a reverse order. There are three situations for each user in the loop:
1. If the current user from the connection op is a NpuCircularDmaCpyNdOp, and the loop has not reached the circular dma op we are checking (currentCircularDma = false) then return true;
2. If the loop reaches the circular dma op we are checking, then set `currentCircularDma = true`, and continue to the next user. 
3. If there is a next user, and it is before what we are checking (now currentCircularDma = true), then return true.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/dma_loop_subsumption_circular.mlir
```diff
@@ -58,3 +58,95 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
...
+
+// -----
+
+// Ensure no subsumption happens, if there are more than one npu.circular_dma_cpy_nd users
```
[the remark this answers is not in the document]
yzhang93: It won't change if I remove `%3 = amdaie.npu.dma_cpy_nd %0([] [] [], [0, %1] [32, 64] [128, 1])`. I've simplified this test.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/dma_loop_subsumption_circular.mlir
```diff
@@ -58,3 +58,95 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
...
+// CHECK:         amdaie.npu.circular_dma_cpy_nd %[[CONNECTION]]([0] [2048] [1], [] [] [])
+// CHECK:         amdaie.npu.dma_cpy_nd %[[CONNECTION]]([] [] [], [0, 0, 0, 0] [2, 2, 32, 64] [0, 64, 128, 1])
+#map = affine_map<(d0) -> (d0 * 64)>
+#executable_target_amdaie_xclbin_fb = #hal.executable.target<"amd-aie", "amdaie-xclbin-fb", {target_device = "npu1_4col", ukernels = "none"}>
+module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb} {
+  func.func @circular_and_other_dma_subsume(%arg0: !amdaie.logicalobjectfifo<memref<2048xi32, 1 : i32>, 2>, %arg1: !amdaie.logicalobjectfifo<memref<256x128xi32>>) {
+    amdaie.workgroup {
+      %0 = amdaie.connection(%arg0, %arg1) : (!amdaie.logicalobjectfifo<memref<2048xi32, 1 : i32>, 2>, !amdaie.logicalobjectfifo<memref<256x128xi32>>)
+      amdaie.controlcode {
+        scf.forall (%arg2, %arg3) in (2, 2) {
```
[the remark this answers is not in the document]
yzhang93: I'd keep it as is, since this is the IR mimic the real generated IR and we want to see if both npu ops are hoisted out of the loop.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaLoopSubsumption.cpp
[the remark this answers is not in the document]
newling: I see, sorry I misread the code the first time. Thanks for the explanation

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/dma_loop_subsumption_circular.mlir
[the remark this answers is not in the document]
newling: Ok. I'm a bit surprised, I thought these ops were basically just register writes, and so the second write can be eliminated.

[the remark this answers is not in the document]
yzhang93: Yeah, this is a made-up test. Actually the second op would go away even before this pass if they are the same.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
yzhang93: Done.

newling (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaComposition.cpp
```diff
@@ -38,27 +38,54 @@ class AMDAIEDmaCompositionPass
...
                                       onlyZeroStrideOnOuterDim);
-  }
-  populateStridedOpCombinationPattern(patterns);
-  if (failed(applyPatternsAndFoldGreedily(parentOp, std::move(patterns)))) {
-    parentOp->emitOpError("failed to compose strided operations");
-    return signalPassFailure();
+    if (failed(applyPatternsAndFoldGreedily(parentOp, std::move(patterns)))) {
+      parentOp->emitOpError("failed to compose strided operations");
+      return signalPassFailure();
+    }
```
jtuyls, line 88: Applying the patterns one by one will always miss some composition opportunities. We should be populating all patterns and applying them together. This will be the most versatile and keep rewriting until all DMA composition opportunities are exhausted:

```
RewritePatternSet patterns(context);
populateDmaLoopSubsumptionPattern(patterns, std::move(deviceModel),
                                      onlyZeroStrideOnOuterDim);
populateStridedOpCombinationPattern(patterns);
populateCanonicalizeDoublyStridedOpPatterns(patterns, false);
```

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaComposition.cpp
newling: We've got a couple of passes which look like they're doing canonicalization
of DMA ops. Can they be put into a single AMDAIECanonicalizeDma.cpp, which 
we then just run like vanilla canonicalization? 

AMDAIECanonicalizeDoublyStridedOp.cpp
AMDAIECanonicalizeNpuDmaCpyNd.cpp
AMDAIECombineStridedOps.cpp
AMDAIEDmaComposition.cpp
AMDAIEDmaCSE.cpp
AMDAIEDmaLoopSubsumption.cpp
AMDAIENormalizeLoopBounds.cpp

newling: (obviously not for this PR, but maybe some refactoring for later)

yzhang93: > Applying the patterns one by one will always miss some composition opportunities. We should be populating all patterns and applying them together. This will be the most versatile and keep rewriting until all DMA composition opportunities are exhausted:
> 
> ```
> RewritePatternSet patterns(context);
> populateDmaLoopSubsumptionPattern(patterns, std::move(deviceModel),
>                                       onlyZeroStrideOnOuterDim);
> populateStridedOpCombinationPattern(patterns);
> populateCanonicalizeDoublyStridedOpPatterns(patterns, false);
> ```

I still couldn't make it work as expected if populate all patterns and make them under the same RewritePatternSet. @jtuyls Could you try it and give some suggestions?

jtuyls: > %48 = amdaie.npu.dma_cpy_nd %8([] [] [], %31[0, 0, %45] [4, 32, 64] [4096, 128, 1]) : source_type = !amdaie.logicalobjectfifo<memref<16384xi32>>

Isn't this the same as:

```
%48 = amdaie.npu.dma_cpy_nd %8([] [] [], %31[0, %45] [128, 64] [128, 1]) : source_type = !amdaie.logicalobjectfifo<memref<16384xi32>>
```

I think you can adjust `CombineStridedOps`, so the following is combined into the result above:

```
48 = amdaie.npu.dma_cpy_nd %8([] [] [], %31[0, %45] [32, 64] [128, 1]) : source_type = !amdaie.logicalobjectfifo<memref<16384xi32>>
%49 = amdaie.npu.dma_cpy_nd %8([] [] [], %31[32, %45] [96, 64] [128, 1]) : source_type = !amdaie.logicalobjectfifo<memref<16384xi32>>
```

newling: I meant canonicalization in a  general sense. Canonicalization in the strict sense should be done directly with op class methods. I don't want to spend time arguing abstract points for which I have no evidence but my intuition says it's good to have lots of _lowering_ passes, but few, general purpose canonicalization/cse/normalization passes which can be run anywhere.

yzhang93: > I think you can adjust `CombineStridedOps`, so the following is combined into the result above:
> 
> ```
> 48 = amdaie.npu.dma_cpy_nd %8([] [] [], %31[0, %45] [32, 64] [128, 1]) : source_type = !amdaie.logicalobjectfifo<memref<16384xi32>>
> %49 = amdaie.npu.dma_cpy_nd %8([] [] [], %31[32, %45] [96, 64] [128, 1]) : source_type = !amdaie.logicalobjectfifo<memref<16384xi32>>
> ```

Are you sure this is combinable for a more general case regardless of what we already know from this example? It is not consistent with the current logic that inserting a new dimension if two npu has the same number of dimensions of addressing.

jtuyls: > > I think you can adjust `CombineStridedOps`, so the following is combined into the result above:
> > ```
> > 48 = amdaie.npu.dma_cpy_nd %8([] [] [], %31[0, %45] [32, 64] [128, 1]) : source_type = !amdaie.logicalobjectfifo<memref<16384xi32>>
> > %49 = amdaie.npu.dma_cpy_nd %8([] [] [], %31[32, %45] [96, 64] [128, 1]) : source_type = !amdaie.logicalobjectfifo<memref<16384xi32>>
> > ```
> 
> Are you sure this is combinable for a more general case regardless of what we already know from this example?

Yes, because the above two DMA ops are just an extension of each other, so no new dimension is needed. This is an edge case which is just not handled currently.

jtuyls: > I meant canonicalization in a general sense. Canonicalization in the strict sense should be done directly with op class methods.

In that sense, you could argue that every lowering pass is a performing a canonicalization. so that the `canonical` form is the most efficiently lowered form for the given hardware, and therefore you could put the whole codegen inside a single canonicalization pass. That's against the point of having lowering passes, so I think we shouldn't look at it like that, and unless we have a very good reason to deviate, we should stay with the definition and use of canonicalization within MLIR: https://mlir.llvm.org/docs/Canonicalization/.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -541,11 +542,22 @@ void buildAMDAIETransformPassPipeline(
...
-                                       bool enablePacketFlow) {
+                                       bool enablePacketFlow,
+                                       TilePassPipeline useTilePipeline) {
   passManager.addPass(createEraseHALDescriptorTypeFromMemRefPass());
   passManager.addPass(memref::createFoldMemRefAliasOpsPass());
   passManager.addPass(createCanonicalizerPass());
-  passManager.addPass(createAMDAIEConvertToDmaPass());
+  // For matmul pipelines, we do transpose on target side for pack ops to get
+  // better performance. While for convolution pipelines, the same settings
+  // cause 'aie.dma_bd' error, so for now keep using transpose on source for
```
jtuyls, line 552: What's the exact error?

jtuyls (approved):
LGTM

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
yzhang93: The error is  `'aie.dma_bd' op Cannot give more than 3 dimensions for step sizes and wraps in this  tile (got 4 dimensions).` The full dump IR is https://gist.github.com/yzhang93/fd34425cc223eccbd233c7d1051f2d63

PR #827, opened 2024-10-04
[NpuDmaCpyNdOp/NpuDmaWaitOp] Return optional async token and wait for multiple

jtuyls:
This PR makes the return type of `AMDAIE::NpuDmaCpyNdOp` optional and more explicit by introducing an `amdaie.async_token` type. The `AMDAIE::NpuDmaWaitOp` will now operate on operands of the async token type . This is consistent with how `gpu.async_token` is used in the GPU dialect.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIETypes.td
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIETypes.td
@@ -15,6 +15,16 @@ class AMDAIEDialect_Type<string name, string typeMnemonic, list<Trait> traits = 
   let mnemonic = typeMnemonic;
 }
 
+// The types for async tokens which can be returned from async operations. The
+// dedicated types for `source` and `target` can be used to specify on which
+// side of (for example a copy/DMA) operation should be synchronized if there
+// are multiple.
+def AMDAIE_AsyncTokenType : AMDAIEDialect_Type<"AsyncToken", "async_token">;
+def AMDAIE_AsyncSourceTokenType : AMDAIEDialect_Type<"AsyncSourceToken", "async_source_token">;
+def AMDAIE_AsyncTargetTokenType : AMDAIEDialect_Type<"AsyncTargetToken", "async_target_token">;
+def AMDAIE_AnyAsyncTokenType 
+  : AnyTypeOf<[AMDAIE_AsyncTokenType, AMDAIE_AsyncSourceTokenType, AMDAIE_AsyncTargetTokenType]>;
+
 def AMDAIE_LogicalObjectFifoType :
     AMDAIEDialect_Type<"LogicalObjectFifo", "logicalobjectfifo"> {
   let summary = "The logical objectfifo type encapsulating a memref";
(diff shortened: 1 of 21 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIETypes.td
```diff
@@ -15,6 +15,12 @@ class AMDAIEDialect_Type<string name, string typeMnemonic, list<Trait> traits =
   let mnemonic = typeMnemonic;
 }
 
+def AMDAIE_AsyncTokenType : AMDAIEDialect_Type<"AsyncToken", "async_token">;
+def AMDAIE_AsyncSourceTokenType : AMDAIEDialect_Type<"AsyncSourceToken", "async_source_token">;
+def AMDAIE_AsyncTargetTokenType : AMDAIEDialect_Type<"AsyncTargetToken", "async_target_token">;
```
Abhishek-Varma, line 20: Is there no way to add a brief description about them in the `.td` file? I understand `async_token` has correspondence with GPU dialect, so I had to read about it from the corresponding `.td` file to understand - hence asking.

And from what I understood :-
Presence of only `async_source_token` means the dma transfer execution from source side has begun and shouldn't be blocking.
Presence of only `async_target_token` means the dma transfer execution from target side has begun and shouldn't be blocking.

Presence of `async_source_token` and `async_target_token` means the same but now the dma transfer is bi-directional in nature ? I am trying to understand with an example.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECreateAIEWorkgroup.cpp
```diff
@@ -185,16 +185,19 @@ LogicalResult WorkgroupBuilder::buildForDmaCpyNdOp(
       circularDmaTargetOffsets, circularDmaTargetSizes,
       circularDmaTargetStrides, circularDmaSourceOffsets,
       circularDmaSourceSizes, circularDmaSourceStrides);
+  Type ty =
+      !sourceMemSpace
+          ? static_cast<Type>(
+                controlCodeRewriter.getType<AMDAIE::AsyncSourceTokenType>())
+          : static_cast<Type>(
+                controlCodeRewriter.getType<AMDAIE::AsyncTargetTokenType>());
   auto npuDmaCpy = controlCodeRewriter.createAndLookup<AMDAIE::NpuDmaCpyNdOp>(
```
Abhishek-Varma, line 194: From what I infer in this part of the code - we add `async_source` for L3->L2 and `async_target` for other DMA transfers ? If that's too trivialized, please correct me.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIETypes.td
jtuyls: >Presence of only async_source_token means the dma transfer execution from source side has begun and shouldn't be blocking.

No, it means that the DMA transfer operation isn't be blocking, but a token is returned when the source DMA is done executing. Similarly, `async_target_token `, means that the DMA operation isn't blocking, but a token is returned when the target DMA is done executing. Having both `async_source_token` and `async_target_token` means that both the source and target DMAs will be returning a token that can be waited for.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECreateAIEWorkgroup.cpp
jtuyls: Yes

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIETypes.td
jtuyls: > Is there no way to add a brief description about them in the .td file?

I added more information to the descriptions and comments in the `AMDAIEOps.td` and `AMDAIETypes.td` files.

PR #829, opened 2024-10-04
[CI] Support bf16 output in end-to-end numerical testing 

newling:
See comments in https://github.com/nod-ai/iree-amd-aie/pull/822

@Abhishek-Varma this should unblock you (although I haven't tested your PR with this yet...)

### DIFF
--- a/build_tools/ci/cpu_comparison/input_generator.py
+++ b/build_tools/ci/cpu_comparison/input_generator.py
@@ -36,7 +36,7 @@ import re
 from numpy.random import Generator, MT19937, SeedSequence
 
 
-def convert_f32_to_bf16(float32_value):
+def convert_f32_to_bf16(float32_array):
     """
     IEEE float32 to bfloat16
 
@@ -59,19 +59,17 @@ def convert_f32_to_bf16(float32_value):
     to:   [SEEEEEEEEMMMMMMM]
                            ================= remove 16 bits of mantissa
     """
-    int32_repr = float32_value.view(np.int32)
-    bf16_int_repr = int32_repr >> 16
-    return np.uint16(bf16_int_repr)
+    v0 = float32_array.view(np.uint32) >> 16
+    return v0.astype(np.uint16)
 
 
-def convert_bf16_to_f32(bfloat16_value):
+def convert_bf16_to_f32(bfloat16_array):
     """
     IEEE bfloat16 to float32. See docstring of convert_f32_to_bf16 for a
     bit of info on the mantissa/exponent manipulation.
     """
-    v0 = np.uint32(bfloat16_value) << 16
-    float32_value = np.frombuffer(v0.tobytes(), dtype=np.float32)[0]
-    return float32_value
+    v0 = bfloat16_array.astype(np.uint32) << 16
+    return np.frombuffer(v0.tobytes(), dtype=np.float32)
 
 
 def generate_bfloat16_data(num_values, lower_bound, upper_bound, rng):
@@ -79,7 +77,7 @@ def generate_bfloat16_data(num_values, lower_bound, upper_bound, rng):
     float_data = rng.integers(lower_bound, upper_bound, num_values).astype(np.float32)
 
     # Convert float32 data to bfloat16
-    bf16_data = [convert_f32_to_bf16(f) for f in float_data]
+    bf16_data = convert_f32_to_bf16(float_data)
 
     # Pack bfloat16 data into binary format
     binary_data = struct.pack(f"{len(bf16_data)}H", *bf16_data)
@@ -228,7 +226,6 @@ def np_from_binfile(bin_file, type_str):
 
     element_type_str = type_str.strip().split("x")[-1]
 
-
     # Get a numpy type from the string.
     np_type = None
     if element_type_str == "bf16":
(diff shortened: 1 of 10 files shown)


build_tools/ci/cpu_comparison/input_generator.py
```diff
@@ -64,6 +64,16 @@ def convert_f32_to_bf16(float32_value):
...
 
 
+def convert_bf16_to_f32(bfloat16_value):
+    """
+    IEEE bfloat16 to float32. See docstring of convert_f32_to_bf16 for a
+    bit of info on the mantissa/exponent manipulation.
+    """
+    v0 = np.uint32(bfloat16_value) << 16
+    float32_value = np.frombuffer(v0.tobytes(), dtype=np.float32)[0]
+    return float32_value
```
makslevental, line 74: wouldn't it make more sense to cast to uint16? won't this foil packing?

newling: So, bfloat16_value _is_ a uint16. 

Are you suggesting 

v0 = np.uint16(bfloat16_value) << 16 

? That basically wipes clean all the interesting 16 bits. 

I speak confidently because there's a test in this PR which passes ..

newling: Oh, I didn't add the test file lol

newling: Test file added

makslevental: i'm saying i don't get why you're using 32 bits for a 16 bit value - i'm assuming that "native" bf16 is packed two (or 4) to a machine word (on the host) where stuff bf16 into fp32 spaces out the values. 

>I speak confidently because there's a test in this PR which passes ..

I don't know any better ... i'm just asking for personal edification

newling: From bf16 to fp32:


from:   [SEEEEEEEEMMMMMMM]
to:       [SEEEEEEEEMMMMMMMMMMMMMMMMMMMMMMM]

where the extra mantissa bits are zeros, so we want:
to:       [SEEEEEEEEMMMMMMM0000000000000000]


We start with the following  (which is stored as uint16, bfloat16_value **_is_** uint16, maybe that's what's confusing here)  
[SEEEEEEEEMMMMMMM]

and cast it to uint32:
[0000000000000000SEEEEEEEEMMMMMMM]

and then bitshift it to
[SEEEEEEEEMMMMMMM0000000000000000]

which is what we want.

Abhishek-Varma:
> See comments in #822
> 
> @Abhishek-Varma this should unblock you (although I haven't tested your PR with this yet...)

Thank you so much @newling for this prompt fix - I tried this and it works!

build_tools/ci/cpu_comparison/input_generator.py
Abhishek-Varma: Changing `uint32` to `uint16` seems to have solved the issue with [numerics comparison reported here](https://github.com/nod-ai/iree-amd-aie/pull/829#issuecomment-2395994235).

build_tools/ci/cpu_comparison/input_generator.py
```diff
@@ -165,6 +175,80 @@ def write_input(bin_filename, num_elements, element_type, input_number, input_se
...
+
+    shape = [int(x) for x in type_str.strip().split("x")[0:-1]]
+
+    # Load data with the numpy type specified.
+    array = np.fromfile(bin_file, dtype=np_type)
+    array = array.reshape(shape)
+
+    # If the numpy type was just a proxy, do some extra processing.
+    if element_type_str == "bf16":
+        array = np.array([convert_bf16_to_f32(x) for x in array])
```
jtuyls, line 247:
```suggestion
        array = convert_bf16_to_f32(array)
```

newling: Are you sure this works, do I need to test it before choosing 'Commit suggestion' ?

newling:
Also, what's the smallest integer that cannot be represented as bfloat16, is it 128? I'm wondering now if the discrepency is just that CPU backend doesn't actually do anything in bf16 (maybe?)

newling:
AIE is rounding down, llvm-cpu is rounding up. I did a test where the fp32 output is 527, llvm-cpu rounds to 528 and aie rounds to 524. 

This is true for any setting of rounding_mode!

```
%3 = arith.truncf %2 to_nearest_even 
%3 = arith.truncf %2 downward
%3 = arith.truncf %2 upward
```

etc llvm-cpu always rounds up AIE (peano) always rounds down. 

What I think we should do: write a test which doesn't rely on llvm-cpu. 

I'll try and set this up.

newling:
I would like to land this, before helping on the truncf PR with @Abhishek-Varma . Can I please get a complete review?

Abhishek-Varma (approved):
Thank you @newling!

makslevental:
randomly found this just now https://github.com/iree-org/iree/blob/e2a2b2b52de077db5daf63bd8c9255d6f3be2036/compiler/src/iree/compiler/Codegen/Common/ConvertBf16ToUInt16Buffers.cpp - not sure if it's useful :shrug:

newling:
Cool, maybe. We don't do any casting in C++ atm. I came across this this morning https://github.com/jax-ml/ml_dtypes so I should probably use that instead of manually bit slicing.

makslevental:
> Cool, maybe. We don't do any casting in C++ atm. I came across this this morning https://github.com/jax-ml/ml_dtypes so I should probably use that instead of manually bit slicing.

I already added that https://github.com/nod-ai/iree-amd-aie/blob/main/tests/requirements.txt#L6

PR #833, opened 2024-10-07
Run aievec lowering passes in 'main' pipeline

newling:
This PR puts the aievec passes in the main 'IREE' pipeline, rather than running them in aie2xclbin. 

I found the trickiest was figuring out why it didn't "just work" without adding Affine and AIEVec as legal dialects in the pattern target in AMDAIECoreToStandard, So I'd like to refactor that pass soon -- there are too many dialects listed there. I'd also like to throw the final lowering to LLVM over the wall from aie2xclbin too, but that was causing other issues which I'd rather address in another PR.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
+++ b/compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
@@ -642,6 +642,7 @@ struct DetectNonCanonicalOpsPass
 };
 
 void buildCanonicalizeVectorForAIEVec(OpPassManager &pm) {
+
   // TODO: Add passes to split vectors that won't fit in registers
   pm.addPass(createCanonicalizeVectorForAIEVecPass());
   pm.addPass(mlir::createCanonicalizerPass());
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Test/samples/matmul_pack_peel_air_e2e.mlir
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Test/samples/matmul_pack_peel_air_e2e.mlir
@@ -18,14 +18,14 @@ func.func @matmul_i8_i32(%lhs: tensor<32x16xi8>, %rhs: tensor<16x32xi8>) -> tens
 
 // -----
 
-func.func @matmul_bf16(%lhs: tensor<16x32xbf16>, %rhs: tensor<32x16xbf16>) -> tensor<16x16xbf16>
+func.func @matmul_bf16(%lhs: tensor<16x32xbf16>, %rhs: tensor<32x16xbf16>) -> tensor<16x16xf32>
 {
-  %cst = arith.constant 0.000000e+00 : bf16
-  %0 = tensor.empty() : tensor<16x16xbf16>
-  %1 = linalg.fill ins(%cst : bf16) outs(%0 : tensor<16x16xbf16>) -> tensor<16x16xbf16>
+  %cst = arith.constant 0.000000e+00 : f32
+  %0 = tensor.empty() : tensor<16x16xf32>
+  %1 = linalg.fill ins(%cst : f32) outs(%0 : tensor<16x16xf32>) -> tensor<16x16xf32>
   %res = linalg.matmul ins(%lhs, %rhs: tensor<16x32xbf16>, tensor<32x16xbf16>)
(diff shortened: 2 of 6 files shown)


makslevental:
My name is Max and I support this message (thanks for continuing to push on this because I concur aie2xclbin shouldn't have passes in it at all).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Test/samples/matmul_pack_peel_air_e2e.mlir
```diff
@@ -18,14 +18,14 @@ func.func @matmul_i8_i32(%lhs: tensor<32x16xi8>, %rhs: tensor<16x32xi8>) -> tens
 
 // -----
 
-func.func @matmul_bf16(%lhs: tensor<16x32xbf16>, %rhs: tensor<32x16xbf16>) -> tensor<16x16xbf16>
```
jtuyls, line 21: Why is this output data type change needed?

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
```diff
@@ -642,6 +642,7 @@ struct DetectNonCanonicalOpsPass
 };
 
 void buildCanonicalizeVectorForAIEVec(OpPassManager &pm) {
+
```
jtuyls, line 645: Is this change on purpose?
```suggestion
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Test/samples/matmul_pack_peel_air_e2e.mlir
newling: We don't support matmul with output type bf16. The lowering of the vector.contract to aievec.matmul fails (there is a compilation error, I can't remember exactly what it is). Because this PR adds the lowering from vector to aievec, we only hit this error with this PR.

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
newling: I don't know if this is a clang-format change or me being careless -- I'll commit the suggestion.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Test/samples/matmul_pack_peel_air_e2e.mlir
jtuyls: I see, thanks!

jtuyls (approved):
LGTM

newling:
> Turns out that this prepones a few other bugs that wasn't captured by the lit tests.
> 
> [Matmul + Truncf PR](https://github.com/nod-ai/iree-amd-aie/pull/822) tries to enable vectorization for all element-wise operations - no issue was caught earlier. But now, it causes an issue because `VectorToAIEVecConversions.cpp` places lot of `addDynamicallyLegalOp` constraints on various arith ops. Specifically an error is thrown for `arith.addf` op for [this lit test](https://github.com/nod-ai/iree-amd-aie/blob/c84cca03034b61c97a081ebce47cf31a06255356/compiler/plugins/target/AMD-AIE/iree-amd-aie/Test/samples/matmul_elementwise_pack_peel_air_e2e.mlir#L60).
> 
> I guess it'd be nice if we could bridge the gap between `AMDAIEVectorization.cpp` and the expectations laid out by `VectorToAIEVecConversions.cpp`. One way is clearly to enforce vectorization based on the expectations of `VectorToAIEVecConversions.cpp`.
> 
> For now, to not at least get in the first step toward supporting Matmul + Truncf in, I've added a stop-gap solution.
> 
> Please let me know your thoughts on the overall plan.

I like the plan. I guess in  [this lit test](https://github.com/nod-ai/iree-amd-aie/blob/c84cca03034b61c97a081ebce47cf31a06255356/compiler/plugins/target/AMD-AIE/iree-amd-aie/Test/samples/matmul_elementwise_pack_peel_air_e2e.mlir#L60) aievec shouldn't be interested in f32 addf. 

Yes -- it'd be nice to consolidate AMDAIEVectorization with the later aievec passes. I looked into merging the 2 a few weeks ago, but the main problem I faced was the AMDAIEVectorization runs before bufferization, aievec runs after bufferization.

PR #834, opened 2024-10-08
Temporarily disable double buffering for better performance

yzhang93:
Temporarily set default buffer depth as 1 for l1/l2 space, and use a larger tile size. This will give much better performance. Will enable it after optimizing double buffering and pipelining.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
@@ -174,8 +174,7 @@ FailureOr<ParameterSetting> ParameterSetting::create(linalg::LinalgOp linalgOp,
   // the element types of all tensors which need to be allocated in memory
   // simultaneously.
   FailureOr<unsigned> maybeScaleFactor =
-      isObjectFifo ? getTilingScaleFactor(initType.getElementType())
-                   : getTilingScaleFactor(lhsType.getElementType());
+      getTilingScaleFactor(lhsType.getElementType());
   if (failed(maybeScaleFactor)) {
     return linalgOp.emitOpError(
         "does not have the expected bitwidth (64, 32, 16, or 8), could not "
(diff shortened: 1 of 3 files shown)


jtuyls (changes requested):
I don't think we should be disabling double buffering by default because in principal double buffering should result in better performance, so either:

- There is some overhead trumping double buffering at the moment. For example, disabling double buffering and increasing tile size will result in less control code, which might be the bottleneck at the moment.
- There is a bug/blocker in double buffering that causing no double buffering to actually happen.

Additionally, most tests should keep using double buffering by default as this is better stress testing the codegen as it leads to more complex code. NOTE: I don't think we have any single buffer tests, which would be good to add to make sure that's also tested.

Because the better results you're seeing shouldn't be there in principal, I think we should instead get to the bottom of this instead of merging for marginal gains (with respect to the total gain we still need).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -174,8 +174,7 @@ FailureOr<ParameterSetting> ParameterSetting::create(linalg::LinalgOp linalgOp,
   // the element types of all tensors which need to be allocated in memory
   // simultaneously.
   FailureOr<unsigned> maybeScaleFactor =
-      isObjectFifo ? getTilingScaleFactor(initType.getElementType())
```
jtuyls, line 177: We should still take the initType into account to be safe. To improve the tile sizes safely, while also making them as large as possible, we should compute it based on all three (`lhs`, `rhs`, `init`) and whether or not double buffering is enabled for L1.

yzhang93:
> Additionally, most tests should keep using double buffering by default as this is better stress testing the codegen as it leads to more complex code. NOTE: I don't think we have any single buffer tests, which would be good to add to make sure that's also tested.

What tests are you referring to? Do you think it's worth to make single/double buffering a flag and add some e2e tests?

> Because the better results you're seeing shouldn't be there in principal, I think we should instead get to the bottom of this instead of merging for marginal gains (with respect to the total gain we still need).

Okay, agreed. I'm going to close this PR.

jtuyls:
> What tests are you referring to? Do you think it's worth to make single/double buffering a flag and add some e2e tests?

The main e2e tests in `run_matmul_tests.sh`. Yes, I think it would be useful to have a flag for that and add some tests for it.

Also, this would still be useful as well to get optimal tile sizes for both double buffer and single buffer:

> We should still take the initType into account to be safe. To improve the tile sizes safely, while also making them as large as possible, we should compute it based on all three (lhs, rhs, init) and whether or not double buffering is enabled for L1.

PR #837, opened 2024-10-09
Bump IREE 10/09/2024

makslevental:
This is ahead of schedule I guess but I'd like to get this PR locally https://github.com/iree-org/iree/pull/18675 (the PR built locally with no changes).

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
@@ -682,12 +682,11 @@ void addMLIRAIRLoweringPasses(OpPassManager &passManager, AMDAIEDevice device,
   passManager.addPass(createCSEPass());
   {
     xilinx::air::AIRFuseChannelsOptions options;
-    std::vector<std::string> mode;
     if (useTilePipeline == TilePassPipeline::PackPeelPipeline &&
         matmulElementwiseFusion) {
-      mode.push_back("L1");
+      const static llvm::SmallVector<std::string> mode = {"L1"};
+      options.clAggressiveMode = mode;
     }
-    options.clAggressiveMode = ArrayRef(mode);
     passManager.addPass(xilinx::air::createAIRFuseChannels(options));
   }
   passManager.addPass(createCanonicalizerPass());
@@ -753,14 +752,13 @@ void addMLIRAIRLoweringPasses(OpPassManager &passManager, AMDAIEDevice device,
     // with given factors, and subsequently unrolled in
     // AIRUnrollOuterPerfectlyNestedLoopsPass, to enforce SHIM DMA BD count
     // within the hardware limit.
-    std::vector<unsigned> tile_sizes;
     if (useTilePipeline == TilePassPipeline::PackPeelPipeline) {
-      tile_sizes = {2, 2};
+      const static llvm::SmallVector<unsigned> tile_sizes = {2, 2};
+      options.clTileSizes = tile_sizes;
     } else if (useTilePipeline == TilePassPipeline::PadPackPipeline) {
-      tile_sizes = {4, 4};
-    } else
-      tile_sizes = {};
-    options.clTileSizes = ArrayRef(tile_sizes);
+      const static llvm::SmallVector<unsigned> tile_sizes = {4, 4};
+      options.clTileSizes = tile_sizes;
+    }
     passManager.addNestedPass<func::FuncOp>(
         xilinx::air::createAffineLoopOptPass(options));
   }


newling (approved):
Approving assuming CI passing

newling:
```
  690 |     options.clAggressiveMode = ArrayRef(mode);
      |     ~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~
```

My interpretation of this code is that it has UB, and it'll be a heavy lift to remove the UB, but if you don't mind leaving the UB then it shouldn't be too hard.

makslevental:
> My interpretation of this code is that it has UB, and it'll be a heavy lift to remove the UB, but if you don't mind leaving the UB then it shouldn't be too hard.

Why? there's an [arrayref ctor from std::vec](https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/ADT/ArrayRef.h#L144-L146). Not sure why it wasn't being found but let's see if assignment ctor works.

newling:
> > My interpretation of this code is that it has UB, and it'll be a heavy lift to remove the UB, but if you don't mind leaving the UB then it shouldn't be too hard.
> 
> Why? there's an [arrayref ctor from std::vec](https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/ADT/ArrayRef.h#L144-L146). Not sure why it wasn't being found but let's see if assignment ctor works.
> 
> Oh you're saying because the std::vec gets dropped almost immediately? That's not UB just use-after-free? yea that's probably true - soln is just to do `static std::vector<std::string> mode;`.

 Maybe. Something like

 ```
 static const std::vector<std::string> modeA {"L1"}; 
 static const std::vector<std::string> modeB {};
 const std::vector<std::string> & mode = x ? modeA : modeB;
 ```
 
 to make it less scary when it's not just a single threads, single call, single etc.

makslevental:
> static const std::vector<std::string> modeA {"L1"}; 
> static const std::vector<std::string> modeB {};
> const std::vector<std::string> & mode = x ? modeA : modeB;
> ```
> 
> to make it less scary when it's not just a single threads, single call, single etc.

you don't need the empty vec - that's just "CLI arg not passed" right?

newling:
> > static const std::vectorstd::string modeA {"L1"};
> > static const std::vectorstd::string modeB {};
> > const std::vectorstd::string & mode = x ? modeA : modeB;
> > ```
> > 
> > to make it less scary when it's not just a single threads, single call, single etc.
> > ```
> 
> you don't need the empty vec - that's just "CLI arg not passed" right? anyway assignment didn't work from vec but there's this `std::initializer_list` [ctor](https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/ADT/ArrayRef.h#L275) that should work

Makes sense, so no need for this indirection, good point

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -682,12 +682,11 @@ void addMLIRAIRLoweringPasses(OpPassManager &passManager, AMDAIEDevice device,
   passManager.addPass(createCSEPass());
   {
     xilinx::air::AIRFuseChannelsOptions options;
-    std::vector<std::string> mode;
+    // static because clAggressiveMode is an ArrayRef
     if (useTilePipeline == TilePassPipeline::PackPeelPipeline &&
         matmulElementwiseFusion) {
-      mode.push_back("L1");
+      options.clAggressiveMode = {"L1"};
```
newling, line 688: Still UB me thinks. This is like 

ArrayRef<std::string> x = {"L1"}; 

which can't be good, where does the string actually live? 

But feel free to land because I don't know if caching works nicely for IREE bumps, and I promise to clean this later.

makslevental: I don't understand why you think is undefined behavior? What about this do you think is compiler/impl dependent? 

> where does the string actually live?

It's a local so I'm almost 100% it's stack allocation.  Regardless, the ctor I'm using here moves the string:

```
    /// The declaration here is extra complicated so that "arrayRef = {}"
    /// continues to select the move assignment operator.
```

If you're very worried I can link a godbolt to double-check.

makslevental: > List-initializing std::initializer_list
> The backing array has the same lifetime as any other temporary object, except that initializing an std::initializer_list object from the backing array extends the lifetime of the array exactly like binding a reference to a temporary.

https://en.cppreference.com/w/cpp/language/list_initialization

newling: I am still very worried. 

Here's what happens I think: 
1) You create your local ` xilinx::air::AIRFuseChannelsOptions options;` which is basically an ArrayRef<std::string>.
2) You pass your local `options` into `xilinx::air::createAIRFuseChannels(options)`. 
3) Inside that pass creator function, the ArrayRef is copied. This is just a copy of 2 stack pointers, the string is not copied!
4) So inside your pass object, you end storing pointers to stack. 
5) By the time you run the pass, that stack pointer is probably long since overwritten.

makslevental: Ok I concede I've been bamboozled by whoever implemented that ctor:

> Anyone using ArrayRef should already be aware of the fact that it does not do lifetime extension

https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/ADT/ArrayRef.h#L111

because in my understanding avoiding the kind of UB you're worried about is exactly what `std::initializer_list` was invented for. Ok then the solution is a `static llvm::SmallVector` because there's no other way for the reference to outlive the pass (since it's the passman that's passed in in this function). I'll patch it up before I land.

newling: Thanks. Yes I think using `static` is the only easy solution.

PR #838, opened 2024-10-09
Move most of LLVM lowering out of aie2xclbin

newling:
Builds on top of https://github.com/nod-ai/iree-amd-aie/pull/836 (that PR included in this PR)
Is a continuation of https://github.com/nod-ai/iree-amd-aie/commit/be42d1918f43ed04595a1384cd9274305e42f961 -- it that PR we moved vector->aievec into the main pipeline, this PR moves aievec->llvm into the main pipeline (and a few other dialect lowerings to llvm). Moving the lowering of func too llvm is tricky, because of the position of the pass `AMDAIECoreToStandardFunc` which lowers from CoreOp to func.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -308,6 +308,17 @@ def AMDAIELinkExecutables :
   let constructor = "mlir::iree_compiler::AMDAIE::createAMDAIELinkExecutablesPass()";
 }
 
+def AMDAIELoadAlignmentReset :
+    Pass<"iree-amdaie-load-alignment-reset", ""> {
+    let summary = "Reset the alignment of the LLVM load operations.";
+    let constructor = "mlir::iree_compiler::AMDAIE::createAMDAIELoadAlignmentResetPass()";
+    let description = [{
+      Reset the alignment of the LLVM load operations to the 'unset'
+      optional value. This is a workaround for an issue in peano, and
```
jtuyls, line 317: Have you checked with peano folks on this?

jtuyls (approved):
LGTM

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
newling: Yes, there's a ticket on JIRA (AIECC-589) which they're aware of, and we've discussed a bit on the peano slack channel. Just to note: this pass isn't new/changed -- i've just transplanted it from aie2xclbin.cpp.

jtuyls: Yeah, I am aware this existed already, but I find it very strange. Will check the discussions and JIRA to understand more.

PR #850, opened 2024-10-18
[SplitLogicalObjectFifos] Add support for dma tranposed on the target side

yzhang93:
Previous PR https://github.com/nod-ai/iree-amd-aie/pull/812 added support to transpose dma dimensions on the target side for control code optimization. However, this new dma addressing wasn't supported in `SplitLogicalObjectFifosForConnectionReuse` pass. 

This PR keeps the original logic and most of the original codes while adding the support for the new dma format.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos_for_connection_reuse_transpose.mlir
```diff
@@ -0,0 +1,135 @@
...
+    %alloc_2 = memref.alloc() : memref<128x128xi32>
+    %alloc_3 = memref.alloc() : memref<1x1x8x8x4x4xi32, 2 : i32>
+    %tile = amdaie.tile(%c1, %c3)
+    %tile_4 = amdaie.tile(%c0, %c2)
+    %0 = amdaie.logicalobjectfifo.from_memref %alloc_1, {%tile} : memref<2x2x32x32xi32, 1 : i32> -> !amdaie.logicalobjectfifo<memref<2x2x32x32xi32, 1 : i32>>
+    %1 = amdaie.logicalobjectfifo.from_memref %alloc_2, {%tile} : memref<128x128xi32> -> !amdaie.logicalobjectfifo<memref<128x128xi32>>
+    scf.forall (%arg4, %arg5) in (2, 2) {
+      %2 = affine.apply #map(%arg5)
+      %3 = affine.apply #map(%arg4)
+      %4 = amdaie.dma_cpy_nd(%0[0, 0, 0, 0] [2, 32, 2, 32] [2048, 32, 1024, 1], %1[%3, %2] [64, 64] [128, 1]) : (!amdaie.logicalobjectfifo<memref<2x2x32x32xi32, 1 : i32>>, !amdaie.logicalobjectfifo<memref<128x128xi32>>)
```
Abhishek-Varma, line 78: I'd prefer adding an analysis between L2 ObjectFifo's shape `<2x2x32x32>` and the sizes involved in L2 as target `[2, 32, 2, 32]` to infer whether a transpose has taken place instead of using a pass flag for the same.

Abhishek-Varma (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos_for_connection_reuse_transpose.mlir
yzhang93: That's a good point. Changes have made.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -404,15 +434,19 @@ LogicalResult splitLogicalObjectFifos(
...
+          getConstantIntValue(newL2Sizes[nonSplitdim]);
       if (!constantSize) {
         return l3ToL2DmaOp->emitOpError()
                << "found a non-constant value for target size at dim "
                << nonSplitdim;
       }
       int64_t offsetToAdd = constantOffset.value() * constantSize.value();
+
+      // If the dma transpose is on the target side, L3 source side data are
+      // continuous and doesn't have "nonSplitDim".
```
Abhishek-Varma, line 446:
```suggestion
      // continuous and don't have `nonSplitDim`.
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -353,39 +342,80 @@ LogicalResult splitLogicalObjectFifos(
...
+    SmallVector<OpFoldResult> staticL2AsSourceSizes =
         l2ToL1DmaOp.getSourceMixedSizes();
 
     // Now we'll create a new L2 buffer based on the new shape inferred earlier
     // via `staticL2AsTargetSizes`.
     LogicalObjectFifoFromMemrefOp oldL2ObjectFifo =
         l2ToL1DmaOp.getSourceObjectFifo();
-    AMDAIE::LogicalObjectFifoFromMemrefOp source = createNewLogicalObjectFifo(
-        rewriter, oldL2ObjectFifo, staticL2AsTargetSizes);
+
```
Abhishek-Varma, line 412:
```suggestion
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -353,39 +342,80 @@ LogicalResult splitLogicalObjectFifos(
...
+    for (size_t dim : splitDimsForL2) {
+      staticL2AsTargetOffsets[transposeDim[dim]] = zeroVal;
+      staticL2AsTargetSizes[transposeDim[dim]] = oneVal;
+    }
+    // Modify the L3 source sizes to match the new L2 target sizes.
+    for (auto &&[splitDim, nonSplitdim] :
+         llvm::zip_equal(splitDimsForL2, nonSplitDimsForL2)) {
+      staticL3AsSourceSizes[splitDim] =
+          staticL2AsTargetSizes[transposeDim[nonSplitdim]];
+    }
```
Abhishek-Varma, line 398: Can we try to fuse these two into a single loop with :- 
```
 for (auto &&[splitDim, nonSplitdim] : llvm::zip_equal(splitDimsForL2, nonSplitDimsForL2))
```
Or is there a dependency which gets invalidated ?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -353,39 +342,80 @@ LogicalResult splitLogicalObjectFifos(
...
+      staticL2AsTargetOffsets[dim] = zeroVal;
+      staticL2AsTargetSizes[dim] = oneVal;
+      staticL3AsSourceOffsets[dim] = zeroVal;
+      staticL3AsSourceSizes[dim] = oneVal;
+    }
+  } else {
+    // The L2 target side has transposed dimensions, while the L3 source side
+    // data are continuous and doesn't have "nonSplitDim".
+    // Hardcoded the transposed dimensions for now.
+    const SmallVector<size_t> transposeDim = {0, 2, 1, 3};
```
Abhishek-Varma, line 388: Since we have added a check earlier : `if (l2TargetShape.size() != staticL2AsTargetSizes.size()) `, I'd rather we infer even the dimensions where transpose has taken place inside the loop `for (auto [s1, s2] : llvm::zip_equal(l2TargetShape, staticL2AsTargetSizes))`  itself and get rid of this hardcoding.

yzhang93: Well, now I think there could be issues if we blindly compare values. There could be dimensions that have same values, e.g, [2, 32, 32, 32] -> [2, 32, 32, 32] after transposing dim 1 and 2. Then there's no way to tell if the dma is transposed or not.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
yzhang93: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -353,39 +342,81 @@ LogicalResult splitLogicalObjectFifos(
...
+
+  // Check if the L3->L2 dma is transposed on the target side.
+  bool dmaTransposeOnSource = true;
+  for (auto [s1, s2] : llvm::zip_equal(l2TargetShape, staticL2AsTargetSizes)) {
+    if (s1 != getConstantIntValue(s2)) {
+      dmaTransposeOnSource = false;
+      break;
+    }
+  }
+  if (staticL3AsSourceSizes.size() != staticL2AsTargetSizes.size()) {
```
yzhang93, line 371: I also added this check to deal with the case I mentioned `[2, 32, 32, 32] transposed to [2, 32, 32, 32]`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
Abhishek-Varma: Oh okay. That makes sense. Thanks for the explanation.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
yzhang93: If the dma is tranposed on the target side then the source side data is continuous. The dma ops are converted from pack ops, and for matmul cases we know the dimensions of source and target memrefs are not the same. For example, we have [128, 128] on the source side, and [1, 2, 32, 32] on the target side, if we transpose on the target side, then the dma addressing has 2 dimensions on source side and 4 dimensions on target side. While on the side, if we transpose on source side, then both dma addressing have 4 dimensions.

Abhishek-Varma: Makes sense. Thanks for the explanation!

Abhishek-Varma (approved):
LGTM!

PR #853, opened 2024-10-21
Bump mlir-air to 24cb14e6d2233e819a5455928e4237ef319e6fc8

erwei-xilinx:
- Convolution with `air` pipeline now works once again.
- In MLIR-AIR, we now have a ukernel for vectorized i8 conv2d (https://github.com/Xilinx/mlir-air/blob/main/test/xrt/14_conv2d_i8_extern_vec/aie.py), following the new convolution tiling recipe in iree-amd-aie.


Abhishek-Varma (approved):
Thanks @erwei-xilinx !

PR #856, opened 2024-10-23
[Vectorization][ObjectFifo] Enable larger Matmul + Truncf

Abhishek-Varma:
-- This PR adds the following pieces to enable working of larger Matmul + Truncf.
1. Updates `insert-loops-for-vectorization` to work on bufferized inputs, collapse unit dimensions and coalesce the generated loops for tiling.
2. Makes the pipeline as `IREEComprehensiveBufferization` -> `insert-cores` -> `insert-loops-for-vectorization` -> `vectorization`.
3. Creates and adds a new pass `function-outlining` - which is invoked right after `insert-cores` pass.
4. Adds two canonicalization patterns that canonicalizes away trivial read/write access patterns to enable other dependent canonicalizations in `aievec`.
5. Adds e2e CI test for 128x128x256 Matmul + Truncf bf16 -> bf16.

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -795,6 +795,29 @@ class MatmulSet(TestSet):
             output_type=get_output_type(test_name),
         )
 
+        # Large shape Matmul + Truncf
+        generate_matmul_test(test_name, template_name, 128, 128, 256, "bf16", "f32")
+        identity_mat = np.eye(128, dtype=np.float32)
+        ones = np.ones(128 * 128, dtype=np.float32).reshape([128, 128])
+        lhs = ones * 101
+        rhs = identity_mat * 3
+        input_args = generate_inputs(test_name, output_dir, 1, {1: lhs, 2: rhs})
+        aie_vs_baseline(
+            config,
+            test_name,
+            input_args,
+            ones * 302,  # exected output
+            use_ukernel=False,
+            tile_pipeline="pack-peel",
+            lower_to_aie_pipeline="objectFifo",
+            function_name=None,
+            seed=1,
+            rtol=0,
+            atol=0,
+            n_repeats=1,
+            output_type=get_output_type(test_name),
+        )
+
 
 class SmokeSet(TestSet):
     def __init__(self):
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -795,7 +795,7 @@ class MatmulSet(TestSet):
             output_type=get_output_type(test_name),
         )
 
-        # Large shape Matmul + Truncf
+        # 128x128x256 shape test of Matmul + Truncf
         generate_matmul_test(test_name, template_name, 128, 128, 256, "bf16", "f32")
         identity_mat = np.eye(128, dtype=np.float32)
         lhs_ones = np.ones(128 * 256, dtype=np.float32).reshape([128, 256])
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -798,15 +798,17 @@ class MatmulSet(TestSet):
         # Large shape Matmul + Truncf
         generate_matmul_test(test_name, template_name, 128, 128, 256, "bf16", "f32")
         identity_mat = np.eye(128, dtype=np.float32)
-        ones = np.ones(128 * 128, dtype=np.float32).reshape([128, 128])
-        lhs = ones * 101
-        rhs = identity_mat * 3
+        lhs_ones = np.ones(128 * 256, dtype=np.float32).reshape([128, 256])
+        rhs_ones = np.ones(256 * 128, dtype=np.float32).reshape([256, 128])
+        out_ones = np.ones(128 * 128, dtype=np.float32).reshape([128, 128])
+        lhs = lhs_ones * 2
+        rhs = rhs_ones * 3
         input_args = generate_inputs(test_name, output_dir, 1, {1: lhs, 2: rhs})
         aie_vs_baseline(
             config,
             test_name,
             input_args,
-            ones * 302,  # exected output
+            out_ones * 1536,  # exected output
             use_ukernel=False,
             tile_pipeline="pack-peel",
             lower_to_aie_pipeline="objectFifo",
--- a/compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
+++ b/compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
@@ -320,13 +320,13 @@ struct CanonicalizeTrivialReadAccessSubviewOpPattern
         }))
       return failure();
     SmallVector<Value> newIndices;
-    for (OpFoldResult x : subViewOp.getMixedOffsets()) {
+    for (OpFoldResult offset : subViewOp.getMixedOffsets()) {
       Value indexVal;
-      if (auto attr = dyn_cast<Attribute>(x)) {
+      if (auto attr = dyn_cast<Attribute>(offset)) {
         indexVal = rewriter.create<arith::ConstantIndexOp>(
             readOp.getLoc(), cast<IntegerAttr>(attr).getInt());
       } else {
-        indexVal = cast<Value>(x);
+        indexVal = cast<Value>(offset);
       }
       newIndices.push_back(indexVal);
     }
@@ -353,13 +353,13 @@ struct CanonicalizeTrivialWriteAccessSubviewOpPattern
         }))
       return failure();
     SmallVector<Value> newIndices;
-    for (OpFoldResult x : subViewOp.getMixedOffsets()) {
+    for (OpFoldResult offset : subViewOp.getMixedOffsets()) {
       Value indexVal;
-      if (auto attr = dyn_cast<Attribute>(x)) {
+      if (auto attr = dyn_cast<Attribute>(offset)) {
         indexVal = rewriter.create<arith::ConstantIndexOp>(
             writeOp.getLoc(), cast<IntegerAttr>(attr).getInt());
       } else {
-        indexVal = cast<Value>(x);
+        indexVal = cast<Value>(offset);
       }
       newIndices.push_back(indexVal);
     }
--- a/compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
+++ b/compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
@@ -316,11 +316,13 @@ static LogicalResult isAllZeroOffsetAccess(mlir::OperandRange indices) {
   return success();
 }
 
-/// Utility to fetch indices of Subview op which would be used by a new vector
-/// transfer_read/transfer_write op with trivial access pattern.
-static SmallVector<Value> fetchNewIndices(PatternRewriter &rewriter,
-                                          Location loc,
-                                          memref::SubViewOp subViewOp) {
+/// Utility to convert OpFoldResult vector of offsets of a Subview op to
+/// a vector of values.
+static SmallVector<Value> opFoldResultsToValues(PatternRewriter &rewriter,
+                                                Location loc,
+                                                memref::SubViewOp subViewOp) {
+  OpBuilder::InsertionGuard g(rewriter);
+  rewriter.setInsertionPoint(subViewOp);
   SmallVector<Value> newIndices;
   for (OpFoldResult offset : subViewOp.getMixedOffsets()) {
     Value indexVal;
@@ -354,7 +356,7 @@ struct CanonicalizeTrivialReadAccessSubviewOpPattern
     if (!subViewOp) return failure();
     if (failed(isAllZeroOffsetAccess(readOp.getIndices()))) return failure();
     SmallVector<Value> newIndices =
-        fetchNewIndices(rewriter, readOp.getLoc(), subViewOp);
+        opFoldResultsToValues(rewriter, readOp.getLoc(), subViewOp);
     rewriter.replaceOpWithNewOp<vector::TransferReadOp>(
         readOp, readOp.getType(), subViewOp.getSource(), newIndices,
         readOp.getPadding(), readOp.getInBoundsValues());
@@ -381,7 +383,7 @@ struct CanonicalizeTrivialWriteAccessSubviewOpPattern
     if (!subViewOp) return failure();
     if (failed(isAllZeroOffsetAccess(writeOp.getIndices()))) return failure();
     SmallVector<Value> newIndices =
-        fetchNewIndices(rewriter, writeOp.getLoc(), subViewOp);
+        opFoldResultsToValues(rewriter, writeOp.getLoc(), subViewOp);
     rewriter.create<vector::TransferWriteOp>(
         writeOp.getLoc(), writeOp.getVector(), subViewOp.getSource(),
         newIndices, writeOp.getInBoundsValues());
--- a/compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
+++ b/compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
@@ -304,6 +304,45 @@ class FlattenContiguousRowMajorTransferWritePattern
 
 }  // namespace copied_from_mlir
 
+/// Utility to check if the indices provided are all 0.
+static LogicalResult isAllZeroOffsetAccess(mlir::OperandRange indices) {
+  if (!llvm::all_of(indices, [](Value val) {
+        IntegerAttr attr;
+        if (!matchPattern(val, m_Constant(&attr))) return false;
+        return attr.getInt() == 0;
+      })) {
+    return failure();
+  }
+  return success();
+}
+
+/// Utility to fetch indices of Subview op which would be used by a new vector
+/// transfer_read/transfer_write op with trivial access pattern.
+static SmallVector<Value> fetchNewIndices(PatternRewriter &rewriter,
+                                          Location loc,
+                                          memref::SubViewOp subViewOp) {
+  SmallVector<Value> newIndices;
+  for (OpFoldResult offset : subViewOp.getMixedOffsets()) {
+    Value indexVal;
+    if (auto attr = dyn_cast<Attribute>(offset)) {
+      indexVal = rewriter.create<arith::ConstantIndexOp>(
+          loc, cast<IntegerAttr>(attr).getInt());
+    } else {
+      indexVal = cast<Value>(offset);
+    }
+    newIndices.push_back(indexVal);
+  }
+  return newIndices;
+}
+
+/// A rewriter function to canonicalize the following :-
+/// INPUT:
+///       %b = memref.subview %a [offset0, offset1, ...]
+///       %c = vector.transfer_read %b[0, 0, ...]
+/// OUTPUT:
+///       %c = vector.transfer_read %a[offset0, offset1, ...]
+///
+/// This is needed to enable other set of staged canonicalizations in this pass.
 struct CanonicalizeTrivialReadAccessSubviewOpPattern
     : public OpRewritePattern<vector::TransferReadOp> {
   using OpRewritePattern<vector::TransferReadOp>::OpRewritePattern;
@@ -313,23 +352,9 @@ struct CanonicalizeTrivialReadAccessSubviewOpPattern
     auto subViewOp = dyn_cast_if_present<memref::SubViewOp>(
         readOp.getSource().getDefiningOp());
     if (!subViewOp) return failure();
-    if (!llvm::all_of(readOp.getIndices(), [](Value val) {
-          IntegerAttr attr;
-          if (!matchPattern(val, m_Constant(&attr))) return false;
-          return attr.getInt() == 0;
-        }))
-      return failure();
-    SmallVector<Value> newIndices;
-    for (OpFoldResult offset : subViewOp.getMixedOffsets()) {
-      Value indexVal;
-      if (auto attr = dyn_cast<Attribute>(offset)) {
-        indexVal = rewriter.create<arith::ConstantIndexOp>(
-            readOp.getLoc(), cast<IntegerAttr>(attr).getInt());
-      } else {
-        indexVal = cast<Value>(offset);
-      }
-      newIndices.push_back(indexVal);
-    }
+    if (failed(isAllZeroOffsetAccess(readOp.getIndices()))) return failure();
+    SmallVector<Value> newIndices =
+        fetchNewIndices(rewriter, readOp.getLoc(), subViewOp);
     rewriter.replaceOpWithNewOp<vector::TransferReadOp>(
         readOp, readOp.getType(), subViewOp.getSource(), newIndices,
         readOp.getPadding(), readOp.getInBoundsValues());
@@ -337,6 +362,14 @@ struct CanonicalizeTrivialReadAccessSubviewOpPattern
   }
 };
 
+/// A rewriter function to canonicalize the following :-
+/// INPUT:
+///       %b = memref.subview %a [offset0, offset1, ...]
+///       vector.transfer_write %val, %b[0, 0, ...]
+/// OUTPUT:
+///       vector.transfer_write %val, %a[offset0, offset1, ...]
+///
+/// This is needed to enable other set of staged canonicalizations in this pass.
 struct CanonicalizeTrivialWriteAccessSubviewOpPattern
     : public OpRewritePattern<vector::TransferWriteOp> {
   using OpRewritePattern<vector::TransferWriteOp>::OpRewritePattern;
@@ -346,23 +379,9 @@ struct CanonicalizeTrivialWriteAccessSubviewOpPattern
     auto subViewOp = dyn_cast_if_present<memref::SubViewOp>(
         writeOp.getSource().getDefiningOp());
     if (!subViewOp) return failure();
-    if (!llvm::all_of(writeOp.getIndices(), [](Value val) {
-          IntegerAttr attr;
-          if (!matchPattern(val, m_Constant(&attr))) return false;
-          return attr.getInt() == 0;
-        }))
-      return failure();
-    SmallVector<Value> newIndices;
-    for (OpFoldResult offset : subViewOp.getMixedOffsets()) {
-      Value indexVal;
-      if (auto attr = dyn_cast<Attribute>(offset)) {
-        indexVal = rewriter.create<arith::ConstantIndexOp>(
-            writeOp.getLoc(), cast<IntegerAttr>(attr).getInt());
-      } else {
-        indexVal = cast<Value>(offset);
-      }
-      newIndices.push_back(indexVal);
-    }
+    if (failed(isAllZeroOffsetAccess(writeOp.getIndices()))) return failure();
+    SmallVector<Value> newIndices =
+        fetchNewIndices(rewriter, writeOp.getLoc(), subViewOp);
     rewriter.create<vector::TransferWriteOp>(
         writeOp.getLoc(), writeOp.getVector(), subViewOp.getSource(),
         newIndices, writeOp.getInBoundsValues());
--- a/compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
+++ b/compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
@@ -30,6 +30,7 @@
 #include "mlir/Dialect/Vector/Transforms/LoweringPatterns.h"
 #include "mlir/Dialect/Vector/Transforms/VectorRewritePatterns.h"
 #include "mlir/Dialect/Vector/Utils/VectorUtils.h"
+#include "mlir/IR/Matchers.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Pass/PassManager.h"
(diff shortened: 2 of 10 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -102,7 +156,7 @@ class AMDAIEInsertLoopsForVectorizationPass
       };
       auto lhsType = elType(genericOp->getOperand(0));
       auto rhsType = elType(genericOp->getOperand(1));
-      auto resType = elType(genericOp->getResult(0));
+      auto resType = elType(genericOp->getOperand(2));
```
jtuyls, line 159: Was the former wrong? `auto resType = elType(genericOp->getResult(0));`. Why?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -74,21 +125,24 @@ class AMDAIEInsertLoopsForVectorizationPass
...
-      auto tiled = linalg::tileLinalgOp(rewriter, genericOp, opts);
-      const auto &loops = tiled.value().loops;
-      assert(!loops.empty() && "expected at least one loop here");
-      rewriter.replaceOp(genericOp, loops[0]->getResult(0));
+      collapseUnitDims(rewriter, genericOp, isTensorType);
+      std::optional<SmallVector<int64_t>> tileSizes =
+          formTileSizesForElementwise(genericOp);
+      if (!tileSizes) {
+        return failure();
+      }
```
jtuyls, line 141: Use `emitOpError` with an error message.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -74,21 +125,24 @@ class AMDAIEInsertLoopsForVectorizationPass
...
+      if (failed(
+              performTiling(rewriter, genericOp, *tileSizes, isTensorType))) {
+        return failure();
+      }
```
jtuyls, line 145: Use `emitOpError` with an error message.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -151,7 +205,15 @@ class AMDAIEInsertLoopsForVectorizationPass
...
-    rewrite(rewriter, genericOp);
+    collapseUnitDims(rewriter, genericOp, isTensorType);
+    std::optional<SmallVector<int64_t>> tileSizes =
+        formTileSizesForMatmul(genericOp);
+    if (!tileSizes) {
+      return failure();
+    }
+    if (failed(performTiling(rewriter, genericOp, *tileSizes, isTensorType))) {
+      return failure();
+    }
```
jtuyls, line 216: Use `emitOpError` with an error message.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -43,11 +44,31 @@ class AMDAIEInsertLoopsForVectorizationPass
     return nDims - firstDim;
   };
 
+  /// Tile the generic op using `tileSizes` and coalesce the generated tiling
+  /// loops.
+  static LogicalResult performTiling(IRRewriter &rewriter,
```
jtuyls, line 49: This function seems to never return a failure?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -243,6 +243,12 @@ def AMDAIEFlattenLogicalObjectFifo :
   let constructor = "mlir::iree_compiler::AMDAIE::createAMDAIEFlattenLogicalObjectFifoPass()";
 }
 
+def AMDAIEFunctionOutlining :
```
jtuyls, line 246: Add more info like what the conditions are for outlining, i.e. it happens on linalg ops.
Also, I think `AMDAIELinalgFunctionOutlining` would be a more descriptive name.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_loops_for_vectorization.mlir
```diff
@@ -257,3 +267,84 @@ module {
     return %0 : tensor<4x6x8xbf16>
   }
 }
+
+// -----
+
+// CHECK-LABEL: @element_wise_bufferized
```
jtuyls, line 273: What are you specifically checking for here that wasn't covered earlier?

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/function_outlining.mlir
```diff
@@ -0,0 +1,91 @@
...
+// CHECK-SAME:    %[[B:.*]]: memref<1x1x8x4x8x4xbf16>,
+// CHECK-SAME:    %[[C:.*]]: memref<1x1x8x8x4x4xf32>) {
+// CHECK:            amdaie.core
+// CHECK:               func.call @generic_matmul_outlined(%[[A]], %[[B]], %[[C]])
+// CHECK-NOT:           linalg.generic
+// CHECK:               amdaie.end
+// CHECK:            }
+// CHECK:            return
+// CHECK:        }
+func.func @matmul_example(%A: memref<1x1x4x8x4x8xbf16>, %B: memref<1x1x8x4x8x4xbf16>, %C: memref<1x1x8x8x4x4xf32>) {
```
newling, line 23: Is the idea that if you outline the same function twice, then both calls will be to the same function? i.e. code size reduction. If so, can we have a lit test for this?  I assume that it works, because it sounds like you've got an e2e test passing now that was not passing before. But a lit test would still be useful IMO

newling (changes requested):
I think this should be at divided into multiple PRs, one containing just the function outlining. 

Nice to see this approach working!

erwei-xilinx:
> Nice progress! This PR needs a bit more polish. Also please talk with @erwei-xilinx and see if he would help to make AIR tests work with the changes.

Thanks for the heads up, @yzhang93! WiP on AIR support: https://github.com/nod-ai/iree-amd-aie/pull/857.

build_tools/ci/run_matmul_test.sh
```diff
@@ -593,20 +593,20 @@ run_matmul_test \
...
-    --m "512"  --n "512" --k "512"
-
-run_matmul_test \
-  --name_prefix "packPeel_t_bf16" \
-  --tile_pipeline "pack-peel" \
-  --lhs_rhs_type "bf16" \
-  --acc_type "f32" \
-  --m "128" --n "256" --k "512" \
-  --do_transpose_rhs "1"
+# run_matmul_test \
```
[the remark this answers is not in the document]
jtuyls: Yeah, I am asking to add a comment in the code so someone reading it a few months from now (potentially), knows why this was disabled.

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
```diff
@@ -39,6 +40,73 @@ namespace mlir::iree_compiler::aievec {
 
 using namespace mlir;
 
+struct CanonicalizeTrivialReadAccessSubviewOpPattern
```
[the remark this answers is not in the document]
Abhishek-Varma: Canonicalizations in general are supposed to be implemented with rewrite pattern.

Let me know if I understood "standard rewrite function" correctly : do you mean `funcName(IRRewriter &rewriter, vector::TransferReadOp readOp)` ?

One case where that function wouldn't work is the following case :-
```
%b = subview %a
%c = subview %b
%d = transfer_read %c
```
This needs to go through stages :-
```
%b = subview %a
%d = transfer_read %b
```
and then :-
```
%d = transfer_read %a
```

These and mostly other things are better off handled via a Rewrite pattern which runs in a loop till no change in the DAG takes place.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -56,11 +77,41 @@ class AMDAIEInsertLoopsForVectorizationPass
...
+                               linalg::GenericOp &genericOp,
+                               bool isTensorType) {
+    linalg::ControlDropUnitDims options;
+    options.rankReductionStrategy =
+        linalg::ControlDropUnitDims::RankReductionStrategy::ExtractInsertSlice;
+    FailureOr<linalg::DropUnitDimsResult> result =
+        linalg::dropUnitDims(rewriter, genericOp, options);
+    if (failed(result)) {
+      return;
+    }
```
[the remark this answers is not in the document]
Abhishek-Varma: No, because collapsing of unit dims won't take place if the input is `2x3x4`, but that shouldn't stop us from moving ahead to tile this like before.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
Abhishek-Varma: Since this pass has been updated to work for `memref`, using `getResult(...)` doesn't work as the ops work only on operand basis. And this change I made would work for both `tensor` as well as `vector`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_loops_for_vectorization.mlir
Abhishek-Varma: Collapsing of unit dims as well as `tensor` vs `memref` args.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -783,6 +783,31 @@ def run(self, config):
             output_type=get_output_type(test_name),
         )
 
+        # Large shape Matmul + Truncf
```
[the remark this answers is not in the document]
Abhishek-Varma: So, larger shapes like 1024x1024x1024 and 512x2048x4096 (MxNxK) all work.

But shapes like 1536x1536x2048 and 4096x4096x2048 (MxNxK) suffer from PM issue.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFunctionOutlining.cpp
```diff
@@ -0,0 +1,112 @@
...
+  ModuleOp moduleOp = getOperation();
+  MLIRContext *context = &getContext();
+  IRRewriter rewriter(context);
+
+  auto outlinedToAFunction = [&](linalg::LinalgOp &computeOp) -> func::FuncOp {
+    // Form outlined FuncName.
+    std::string computeName = "";
+    if (isMatmul(computeOp)) {
+      computeName = "_matmul";
+    } else {
```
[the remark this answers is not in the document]
Abhishek-Varma: For now the check is only on it being Matmul and Elemwise.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFunctionOutlining.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+    // Create terminator op returning the cloned compute op's results.
+    rewriter.setInsertionPointToEnd(outlinedFuncBody);
+    rewriter.create<func::ReturnOp>(clonedComputeOp->getLoc(), ValueRange({}));
+
+    return outlinedFunc;
+  };
+
+  moduleOp.walk([&](linalg::LinalgOp computeOp) {
+    if (isa<linalg::FillOp, linalg::CopyOp>(computeOp))
```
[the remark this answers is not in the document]
Abhishek-Varma: Yes, these are the only linalg ops that I'm not outlining. Because either they're not really compute ops or they're just invoked once.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -56,11 +77,41 @@ class AMDAIEInsertLoopsForVectorizationPass
...
+    if (isTensorType) {
+      rewriter.replaceOp(genericOp, result->replacements);
+    } else {
+      rewriter.eraseOp(genericOp);
+    }
```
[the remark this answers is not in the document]
Abhishek-Varma: Because this pass is now made to work on bufferized linalg ops, it won't have a return value, only operands.
Hence I need to remove this too.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -793,8 +818,8 @@ def run(self, config):
         output_dir = config.output_dir
 
         # The most basic test, direct from .mlir file using all defaults
-        test_files_dir = file_dir / "test_files"
-        aie_vs_llvm_cpu(config, test_files_dir / "matmul_int32.mlir")
+        # test_files_dir = file_dir / "test_files"
+        # aie_vs_llvm_cpu(config, test_files_dir / "matmul_int32.mlir")
```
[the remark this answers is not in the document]
Abhishek-Varma: I've made `objectFifo` default and have changed these tests to use `air` pipeline explicitly, albeit commented.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/function_outlining.mlir
Abhishek-Varma: Hi @newling @jtuyls @yzhang93 - I have raised a separate PR for creating a standalone pass for outlining https://github.com/nod-ai/iree-amd-aie/pull/862 as this pass' creation in itself won't cause any issue - the review comments I've received for this pass has therefore been addressed and carried over to this new PR. Please take a look.

I'm keeping rest of the infra within this PR itself and would rebase this once #862 gets merged.

@newling - I have also added a lit test in this PR as rest of the infra is needed to make the e2e work. Please take a look.

Abhishek-Varma:
> > Nice progress! This PR needs a bit more polish. Also please talk with @erwei-xilinx and see if he would help to make AIR tests work with the changes.
> 
> Thanks for the heads up, @yzhang93! WiP on AIR support: #857.

Thank you so much @erwei-xilinx for the fix! ❤️ 

I took the changes from https://github.com/Xilinx/mlir-air/pull/752 (created by Erwei) and applied to local `mlir-air` submodule.

I confirm all the tests I commented in this PR works with Erwei's fix.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/function_outlining.mlir
newling: Thanks @Abhishek-Varma. I'll take a look at https://github.com/nod-ai/iree-amd-aie/pull/862 asap

newling: > @newling - I have also added a lit test in this PR as rest of the infra is needed to make the e2e work. Please take a look.

I was thinking a test of just the outlining pass, something like 
```
func.func @main() {
  %0 = linalg.matmul ...
  %1 = linalg.matmul ...
}
```
being transformed to 
```
func.func @outlined_matmul() {
   ...
}

func.func @main() {
   %0 = func.call @outlined_matmul ... 
   %1 = func.call @outlined_matmul ... 
}
```

build_tools/ci/cpu_comparison/run.py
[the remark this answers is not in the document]
yzhang93: Okay, maybe good to remove the above comment of "large shape".

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFunctionOutlining.cpp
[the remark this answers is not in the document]
yzhang93: I'd suggest to check the compute ops that need to be outlined instead of the excluded ones.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFunctionOutlining.cpp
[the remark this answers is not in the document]
yzhang93: I suggest to check if the linalg ops are matmul/elementwise outside the function. And emit a debug message here if it fails.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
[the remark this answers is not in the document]
Abhishek-Varma: `tensor`:
```
%a = linalg.generic .... ins(%b : tensor<1x2x3xf32>) outs(%c : tensor<1x2x3xf32>)
```

`bufferized`:
```
linalg.generic .... ins(%b : memref<1x2x3xf32>) outs(%c : memref<1x2x3xf32>)
```

So, if we're forming new a linalg op at the same spot in the IR, we need to replace the old value with the new value which then gets removed by canonicalization. But we don't have the notion of this replacement in case of bufferized inputs, hence we need to remove it explicitly.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFunctionOutlining.cpp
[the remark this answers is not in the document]
Abhishek-Varma: Addressed in https://github.com/nod-ai/iree-amd-aie/pull/862

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFunctionOutlining.cpp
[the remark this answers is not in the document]
Abhishek-Varma: Addressed in https://github.com/nod-ai/iree-amd-aie/pull/862

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -43,11 +44,31 @@ class AMDAIEInsertLoopsForVectorizationPass
...
+    auto opts = linalg::LinalgTilingOptions().setTileSizes(tileSizes);
+    auto tiled = linalg::tileLinalgOp(rewriter, genericOp, opts);
+    const auto &tileLoops = tiled.value().loops;
+    SmallVector<scf::ForOp> loops = llvm::map_to_vector(
+        tileLoops, [](Operation *loop) { return cast<scf::ForOp>(loop); });
+    (void)mlir::coalesceLoops(rewriter, loops);
+    if (isTensorType) {
+      rewriter.replaceOp(genericOp, loops[0]->getResult(0));
+    } else {
+      rewriter.eraseOp(genericOp);
```
jtuyls, line 62: Could you document that this function can work with both tensors and memrefs above?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -43,11 +44,31 @@ class AMDAIEInsertLoopsForVectorizationPass
     return nDims - firstDim;
   };
 
+  /// Tile the generic op using `tileSizes` and coalesce the generated tiling
+  /// loops in order to minimize the overhead of loop control/branch statements.
+  static void performTiling(IRRewriter &rewriter,
+                                     linalg::GenericOp genericOp,
```
jtuyls, line 50: This looks like the file is not formatted.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -43,11 +44,31 @@ class AMDAIEInsertLoopsForVectorizationPass
...
+                                     SmallVector<int64_t> &tileSizes,
+                                     bool isTensorType) {
```
jtuyls, line 52: Instead of passing this bool, I think you can derive it from the generic and simplify the function signature.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -56,11 +77,42 @@ class AMDAIEInsertLoopsForVectorizationPass
...
+    return tileSizes;
+  }
+
+  /// Collapse unit dims of the generic op before tiling for vectorization. Since
+  /// this is optinal we need not return failure if the collapsing cannot take
+  /// place. Eg: For <2x3x4> since there aren't any unit dimensions, it'd return
+  /// failure, hence we can simply return.
+  static void collapseUnitDims(IRRewriter &rewriter,
+                               linalg::GenericOp &genericOp,
+                               bool isTensorType) {
```
jtuyls, line 102: Instead of passing this bool, I think you can derive it from the generic and simplify the function signature.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -74,21 +126,21 @@ class AMDAIEInsertLoopsForVectorizationPass
     // No outer dimensions to tile if fewer than 3 iterators.
     if (numIterators < 3) return failure();
 
+    bool isTensorType = isa<TensorType>(genericOp->getOperand(0).getType());
```
jtuyls, line 129: Is this how tensor types are always derived? Wouldn't it be better to check the result?

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
Abhishek-Varma: Good point. I'm now checking if at all there exists any result or not. Please take a look now.

jtuyls (approved):
LGTM

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
```diff
@@ -303,6 +304,92 @@ class FlattenContiguousRowMajorTransferWritePattern
...
+        if (!matchPattern(val, m_Constant(&attr))) return false;
+        return attr.getInt() == 0;
+      })) {
+    return failure();
+  }
+  return success();
+}
+
+/// Utility to fetch indices of Subview op which would be used by a new vector
+/// transfer_read/transfer_write op with trivial access pattern.
```
newling, line 320: This function seems to just convert from OpFoldResults to Values, I think the comment should be about what it does not how it is used. Similarly I think it should be named `opFoldResultsToValues`

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
```diff
@@ -303,6 +304,92 @@ class FlattenContiguousRowMajorTransferWritePattern
...
+static SmallVector<Value> fetchNewIndices(PatternRewriter &rewriter,
+                                          Location loc,
+                                          memref::SubViewOp subViewOp) {
+  SmallVector<Value> newIndices;
+  for (OpFoldResult offset : subViewOp.getMixedOffsets()) {
+    Value indexVal;
+    if (auto attr = dyn_cast<Attribute>(offset)) {
+      indexVal = rewriter.create<arith::ConstantIndexOp>(
```
newling, line 328: I think the rewriter's insertion point should be set at the start of this function (to subViewOp), to make the function reusable/modular

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
```diff
@@ -303,6 +304,92 @@ class FlattenContiguousRowMajorTransferWritePattern
...
+/// This is needed to enable other set of staged canonicalizations in this pass.
+struct CanonicalizeTrivialReadAccessSubviewOpPattern
+    : public OpRewritePattern<vector::TransferReadOp> {
+  using OpRewritePattern<vector::TransferReadOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(vector::TransferReadOp readOp,
+                                PatternRewriter &rewriter) const override {
+    auto subViewOp = dyn_cast_if_present<memref::SubViewOp>(
+        readOp.getSource().getDefiningOp());
+    if (!subViewOp) return failure();
```
newling, line 354: Unfortunately you can't assume that memref is rank-preserving 

https://mlir.llvm.org/docs/Dialects/MemRef/#memrefsubview-memrefsubviewop

"A subview operation may additionally reduce the rank of the resulting view by removing dimensions that are statically known to be of size 1."

compiler/plugins/target/AMD-AIE/aievec/test/precanonicalization-aieml-llvmir.mlir
```diff
@@ -166,3 +166,42 @@ func.func @arith_truncf(%inp: vector<2x3xf32>) -> vector<2x3xbf16> {
...
+// CHECK:         %[[COLLAPSE_SHAPE:.*]] = memref.collapse_shape %[[ARG0]]
+// CHECK-SAME:        into memref<1024xbf16, strided<[1]>>
+// CHECK:         %[[READ:.*]] = vector.transfer_read %[[COLLAPSE_SHAPE]]
+// CHECK:         %[[SHAPE_CAST:.*]] = vector.shape_cast %[[READ]]
+// CHECK-SAME:        vector<32xbf16> to vector<1x1x4x8xbf16>
+// CHECK:         return %[[SHAPE_CAST]]
+func.func @trivial_read_access(%arg0: memref<4x8x4x8xbf16, strided<[256, 32, 8, 1]>>) -> vector<1x1x4x8xbf16> {
+    %c0 = arith.constant 0 : index
+    %cst = arith.constant 0.000000e+00 : bf16
+    %subview = memref.subview %arg0[2, 3, 0, 0] [1, 1, 4, 8] [1, 1, 1, 1] : memref<4x8x4x8xbf16, strided<[256, 32, 8, 1]>> to memref<1x1x4x8xbf16, strided<[256, 32, 8, 1], offset: 608>>
```
newling, line 184: Maybe add a tests where the subview offset (which is an OpFoldResult) is a value, to cover that path through the code (make an index value func.func argument for example)

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -672,10 +683,21 @@ void addMLIRAIELoweringPasses(OpPassManager &pm) {
...
-                              bool matmulElementwiseFusion) {
+                              bool matmulElementwiseFusion,
+                              bool enableVectorizationPasses) {
   // Add passes for preparing for lowering to MLIR-AIR
   passManager.addPass(createEraseHALDescriptorTypeFromMemRefPass());
   passManager.addPass(memref::createFoldMemRefAliasOpsPass());
+  {
+    // Vectorization passes
+    OpPassManager &funcPassManager = passManager.nest<func::FuncOp>();
+    // FIXME(newling) https://github.com/nod-ai/iree-amd-aie/issues/820
```
newling, line 694: Hmm, have you rebased on the newest master? I think I've fixed this issue and vectorization can always be enabled.

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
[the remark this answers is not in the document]
newling: What is the initial motivation for this pattern, why can't lowering handle the IR with a subview followed by a transfer_read, and will we still fall over if the transfer_read is not "trivial" (all zero offsets) ? 

It would be useful to have a short summary in the codebase IMO (apologies if it's already there, I haven't looked at the full PR yet).

newling (changes requested):

newling:
@Abhishek-Varma  PR looks good other than my small remaining comments. My main concern is that vectorization is being disabled for convolution, hopefully that's not necessary after rebasing?

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
Abhishek-Varma: Yes, and the pattern would still work fine. I have added a test case too.

Abhishek-Varma:
> @Abhishek-Varma PR looks good other than my small remaining comments. My main concern is that vectorization is being disabled for convolution, hopefully that's not necessary after rebasing?

Hi @newling - yes, this was an err because of rebasing - I've removed the disabling of vectorization for conv now.

PR #862, opened 2024-10-29
[LinalgFunctionOutlining] Create a pass to outline linalg compute ops

Abhishek-Varma:
This PR creates a standalone `iree-amdaie-linalg-function-outlining` pass which will be used to later enable the [larger Matmul + Truncf e2e](https://github.com/nod-ai/iree-amd-aie/pull/856).


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+#define DEBUG_TYPE "iree-amdaie-linalg-function-outlining"
+
+namespace mlir::iree_compiler::AMDAIE {
+
+namespace {
+
+/// Utility to outline the linalg compute op.
+static FailureOr<func::FuncOp> outlinedToAFunction(
+    IRRewriter &rewriter, ModuleOp &moduleOp, linalg::LinalgOp &computeOp) {
```
newling, line 25: I think it's unusual to pass ModuleOp/LinalgOp by reference, they're just glamorous pointers. Is there a reason not to pass them by value? Maybe IRRewriter too (although I'm not sure about this one).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -243,6 +243,12 @@ def AMDAIEFlattenLogicalObjectFifo :
   let constructor = "mlir::iree_compiler::AMDAIE::createAMDAIEFlattenLogicalObjectFifoPass()";
 }
 
+def AMDAIELinalgFunctionOutlining :
+  Pass<"iree-amdaie-linalg-function-outlining", "ModuleOp"> {
+  let summary = "Outlining of linalg compute ops";
```
newling, line 248: Can you add some additional information in a description = [{ }] block about which linalg ops specifically are outlined (i.e. linalg.fill is not) and what the motivation for this pass is, and what some assumptions are? Sorry, I know that most of our passes in this file don't contain descriptions, but I think it'd be a good thing if we all started adding high-level information here for people without context.

yzhang93 (changes requested):
Some of my comments are still in the other PR that need to be addressed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
Abhishek-Varma: I have added that. Please check if the wordings are correct.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
```diff
@@ -0,0 +1,115 @@
...
+    computeName = "_matmul";
+  } else if (isElementwise(computeOp)) {
+    computeName = "_elementwise";
+  } else {
+    return failure();
+  }
+  std::string outlinedFuncName =
+      computeOp->getName().stripDialect().str() + computeName + "_outlined";
+  if (auto outlinedFuncOp = dyn_cast_if_present<func::FuncOp>(
+          moduleOp.lookupSymbol(outlinedFuncName))) {
```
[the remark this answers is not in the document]
Abhishek-Varma: Hi @newling - thank you so much for the nudge in this case!

I initially thought that adding just a clear demarcation using `isMatmul` and `isElementwise` would be general enough for a particular dispatch, but missed this point in the generalisation.

I've therefore addressed this comment by indeed going ahead with `Option 1` as you suggested - involved playing around with IREE's `isStructurallyEquivalentTo` but I couldn't make that to work, therefore went ahead with upstream's `isEquivalentTo`.

Please take a look. I've also added corresponding comments in the codebase.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
```diff
@@ -0,0 +1,115 @@
...
+static FailureOr<func::FuncOp> outlinedToAFunction(
+    IRRewriter &rewriter, ModuleOp &moduleOp, linalg::LinalgOp &computeOp) {
+  // Form outlined FuncName.
+  std::string computeName = "";
+  if (isMatmul(computeOp)) {
+    computeName = "_matmul";
+  } else if (isElementwise(computeOp)) {
+    computeName = "_elementwise";
+  } else {
+    return failure();
```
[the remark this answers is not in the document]
Abhishek-Varma: I've removed `failure()` altogether and am now simply skipping those compute ops from the `walk` before invoking this function to outline. Let me know if that sounds okay to you.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
Abhishek-Varma: I've removed the references from these ops, except `rewriter` itself because that uses `context` and I've seen it being referenced in the codebase. If there's a stronger opinion from anyone to have it removed, please let me know and I shall do that.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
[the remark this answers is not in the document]
newling: Nice, I didn't know about `isEquivalentTo`. This seems to do exactly what we want. 

I probably would have tried a slightly different design -- I would have created a `func.func` for every compute operation, and then compared func.funcs for equivalence (using `isReqionEquivalentTo`) before returning from outlinedToAFunc. This would require keeping a `DenseSet<FuncOp>` instead of the DenseMap, which would mean we could erase compute ops immediately rather than waiting till the end of the pass (smaller mental stack while reading code!) 

This is just a minor design preference though -- I will make a PR with this change after this lands.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
newling: Thanks. 

Is 

```
One assumption this pass currently makes: All elementwise/matmul linalg ops within a
    dispatch have same body content.
```

still true? Otherwise looks good.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include "mlir/Dialect/Linalg/Transforms/Transforms.h"
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
+
+#define DEBUG_TYPE "iree-amdaie-linalg-function-outlining"
+
+namespace mlir::iree_compiler::AMDAIE {
+
+namespace {
+
+unsigned uniqueOutlinedMatmul = 0;
```
newling, line 23: Design nit: I would make these (private) members of the Pass, and make outlinedToAFunction a member function of the Pass too. Just because I'm a bit unsure about these non-static global variables, not seen this C++ design before.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
```diff
@@ -0,0 +1,149 @@
...
+unsigned uniqueOutlinedElementwise = 0;
+
+/// Utility to outline the linalg compute op.
+static FailureOr<func::FuncOp> outlinedToAFunction(
+    IRRewriter &rewriter, ModuleOp moduleOp, linalg::LinalgOp computeOp,
+    std::string outlineFuncName,
+    DenseMap<Operation *, std::string> &computeOpToOutlinedFuncMap) {
+  // Check if the compute op is equivalent to a previously outlined compute op.
```
newling, line 31: At this point we expect moduleOp's symbol table to not contain `outlineFuncName`. I think this is worth asserting, to avoid to unlikely case that there is already a function with the name you've just created. I suppose you could increment the  uniqueOutlined index in a while loop until an unused name is found.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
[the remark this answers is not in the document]
newling: It's ok. I think it'd still be good to improve the possible situation in N months time when someone is trying to understand why matmul transpose is running out of PM, whereas matmul is not -- if rather than silently not outlining matmul transpose, you made the pass fail for unconsidered ops, the person might be able to avoid the situation. 

I guess it's a trade off between failing too often when you shouldn't and letting issues silently slip through.

newling (approved):
Functionally looks good to me. My only remaining comments are mostly minor design things which I'm happy to bikeshed, so accepting as is. But @Abhishek-Varma please ping me if you'd like me to take another look.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
MaheshRavishankar, line 23: Global values like this arent a good idea. This can cause race conditions when run in multi-threaded context.

I think LLVM already has a way to deduplicate symbols. I dont know how to call it though.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
MaheshRavishankar: Same question. Doesnt operation equivalence check the region as well?

MaheshRavishankar (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
```diff
@@ -0,0 +1,149 @@
...
+  // same as the previous equivalent outlined compute op in order to lookup the
+  // Symbol table.
+  for (auto &[op, funcName] : computeOpToOutlinedFuncMap) {
+    if (!OperationEquivalence::isEquivalentTo(
+            computeOp.getOperation(), op,
+            OperationEquivalence::ignoreValueEquivalence, /*flags=*/nullptr,
+            OperationEquivalence::IgnoreLocations))
+      continue;
+    outlineFuncName = funcName;
+    break;
```
yzhang93, line 42: Why don't you directly check if the `isEquivalentTo` condition holds, and then update outlineFuncName and break?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+void AMDAIELinalgFunctionOutliningPass::runOnOperation() {
+  ModuleOp moduleOp = getOperation();
+  MLIRContext *context = &getContext();
+  IRRewriter rewriter(context);
+
+  DenseMap<Operation *, std::string> computeOpToOutlinedFuncMap;
+  SmallVector<Operation *> toBeErased;
+  moduleOp.walk([&](linalg::LinalgOp computeOp) {
+    // Form outlined FuncName.
```
yzhang93, line 115: Also add comments of what linalg ops are currently outlined.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -243,6 +243,19 @@ def AMDAIEFlattenLogicalObjectFifo :
...
+  let description = [{
+    Outlines matmul/elementwise linalg compute ops only. This pass essentially minimises
+    the code footprint overall as the unrolling in the later passes lead to lot repeated
```
yzhang93, line 251: Nit: 1) "loop unrolling"; 2) "lead to lots of repeated codes".

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/linalg_function_outlining.mlir
```diff
@@ -0,0 +1,177 @@
...
+      indexing_maps = [affine_map<(d0) -> (d0)>, affine_map<(d0) -> (d0)>],
+      iterator_types = ["parallel"]
+    } ins(%A : memref<4xf32>)
+      outs(%C : memref<4xbf16>) {
+    ^bb0(%in: f32, %out: bf16):
+      %8 = arith.truncf %in : f32 to bf16
+      linalg.yield %8 : bf16
+    }
+    amdaie.end
+  }
```
yzhang93, line 175: I think the above three elementwise cores are good enough for test purpose. Why don't you delete this or change this to a matmul.

yzhang93 (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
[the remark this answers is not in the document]
Abhishek-Varma: Makes sense.

Instead of failing I'm emitting a remark.

Failing this pass shouldn't be an apt way - because outlining is supposed to be an improvement in the performance (time/space) and shouldn't really prevent the pipeline to move further in compilation.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
Abhishek-Varma: I forgot to remove that from the `.td` after adding the Equivalence check. I've removed it now. Operation equivalence indeed checks the region as well.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
Abhishek-Varma: Hi @MaheshRavishankar @newling - so I tried looking into upstream and found a pass called `duplicate-function-elimination` whose description can be found [here](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/Func/Transforms/Passes.td#L43-L52).

I agree we shouldn't use global variables like this - I have made the changes that incorporates the same logic without having to use a global variable and without using the above pass `duplicate-function-elimination`.

I did try to see how to use the above de-duplication pass and I'm attaching the patch in [this gist](https://gist.github.com/Abhishek-Varma/7bdad538e0080cde16442da6c133c0a4).

The main reason why I implemented the logic (for creating unique outlined function) the way I did with counter variables for corresponding compute op was :-
1. We aren't creating function definitions for every occurrence - and can reuse the existing one from the get-go - time complexity-wise, I believe what's being done in this patch is better.
2. Every unique instance from the same category of compute ops would have an incremental number associated with it which we can just inspect from the IR and figure the total unique instances too just by looking at the count.

Now, by using the above patch (`deduplicate-function-elimination`), both of the aforementioned points doesn't hold true. Here's why :-
Reg 1. Clearly we're creating N instances of functions for all N instances of LinalgOps - time complexity wise this is not good.
Reg 2. Now, the pass `duplicate-function-elimination` in order to perform deduplication, requires that the symbol name should be different even if the body is same. And the pass picks up the first instance and uses that to replace the other instances. What therefore happens now is :-
```
INPUT:
Assume we have five outlined functions for elemwise linalg.generic.
Suppose :-
linalg generic 1, linalg generic 2 and linalg generic 3 are equivalent.
linalg generic 4 and linalg generic 5 are equivalent.

OUTPUT:
func.func @generic_elemwise_0_outlined
func.func @generic_elemwise_3_outlined
```

Whereas with the logic I implemented we'd get :-
```
OUTPUT:
func.func @generic_elemwise_0_outlined
func.func @generic_elemwise_1_outlined
```

Apart from being able to reuse the function definition from the Symbol Table when we can, I'm being particular about this unique instance numbering too because I feel it helps (as it's incremental) and doesn't appear as a random count/number in the function name in the IR.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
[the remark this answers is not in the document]
Abhishek-Varma: > I would have created a func.func for every compute operation

Hi @newling - please refer to [this](https://github.com/nod-ai/iree-amd-aie/pull/862#discussion_r1827565532) thread. I strongly feel that we shouldn't be creating `func.func` for every compute ops if we could just reuse it by looking up the Symbol table.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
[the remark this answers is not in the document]
newling: I don't like remarks, I think I'd prefer complete silence than a remark, because there are ops for which we don't want outlining (fill). Matmul and matmul transpose should be handled in the same way here, eventually.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+  void runOnOperation() override;
+};
+
+void AMDAIELinalgFunctionOutliningPass::runOnOperation() {
+  ModuleOp moduleOp = getOperation();
+  MLIRContext *context = &getContext();
+  IRRewriter rewriter(context);
+
+  unsigned uniqueOutlinedMatmul = 0;
```
newling, line 86: Problem if this pass is run twice, as you'll create same names

Abhishek-Varma: I don't think so, the `linalg.generic` for all Matmul won't even exist after the first run and the same applies for all Elementwise ops.

So, even if this pass is run more than once it won't cause any issue.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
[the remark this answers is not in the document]
Abhishek-Varma: Sure, I've removed the remark and going ahead with skipping the unsupported ops for now. Please take a look.

yzhang93 (approved):

MaheshRavishankar (approved):

PR #866, opened 2024-10-30
Towards vectorized convolution (second PR)

newling:
This is just a copy of MLIR op definitions from mlir-aie. These ops are needed to lower `transfer_read` with insufficient alignment correctly. In a future PR, the ops added in this PR will be used to lower `transfer_read`.


compiler/plugins/target/AMD-AIE/aievec/AIEVecOps.cpp
```diff
@@ -605,6 +605,178 @@ ParseResult FMAElemOp::parse(OpAsmParser &parser, OperationState &result) {
...
+  Type stype = sourceType.getElementType();
+  Type rtype = resultType.getElementType();
+  if (stype != rtype)
+    return emitError("source and result element type must be same");
+
+  return success();
+}
+
+// Parse Ext op.
+ParseResult ExtOp::parse(OpAsmParser &parser, OperationState &result) {
```
jtuyls, line 660: Does this really need cpp parse/print functions? It would be nice to just use tablegen. Same for `ShiftOp` below. No hard objections though as it's copy/paste from mlir-aie `aievec`.

jtuyls (approved):
LGTM

compiler/plugins/target/AMD-AIE/aievec/AIEVecOps.cpp
newling: ( @jsetoain if you have an opinion )

jsetoain: Jorn is correct, none of these need custom parsers/printers, and they've ended up the way they are for non-technical reasons. It's a matter of finding the time to clean up. I'll create an issue to do keep track of this.

PR #867, opened 2024-10-30
Pass to align transfer_reads

newling:
This PR adds the necessary pass to align `transfer_read`. It is based on mlir-aie's aievec, see:

https://github.com/Xilinx/mlir-aie/blob/d3da586305ebc22e5ecdf1d3e682b44853436e91/lib/Dialect/AIEVec/Transforms/VectorToVectorConversions.cpp#L123

Some changes were needed for our use case, however. The main one is that the lowering in this PR skips the `vector.extract_strided_slice` operation, because we have an offset which is not constant. i.e. the offsets in https://mlir.llvm.org/docs/Dialects/Vector/#vectorextract_strided_slice-vectorextractstridedsliceop cannot be integers for us, because they are determined from loop induction variables. The pass implemented here goes straight to aievec extract and shift operations, where mlir Values are used for offsets. 

Also included in this PR: an aievec.shift folder. I can make this a separate PR if preferred. 

This PR enables vectorization for convolution and resolves https://github.com/nod-ai/iree-amd-aie/issues/820


jtuyls:
Could you rename your PRs to be more descriptive of what they actually add? I was looking at the commit history earlier and `Towards vectorized convolution (second PR)` isn't a great commit message.

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
```diff
@@ -925,6 +927,156 @@ struct CanonicalizeVectorForAIEVecPass
...
+/// This is currently hard-wired for AIE2 (phoenix) HW constraints.
+FailureOr<Value> getAlignedTransferRead(vector::TransferReadOp readOp,
+                                        IRRewriter &rewriter) {
+  //`alignBits`: transfer_reads must be aligned to this number of bits,
+  //             otherwise the reads will be scalarized.
+  //
+  //`maxVectorSizeBits`: the maximum size of a vector that can be read in a
+  //                     single transfer_read operation.
+  int alignBits = 256;
+  int maxVectorSizeBits = 1024;
```
jtuyls, line 946: We should retrieve this from the `AMDAIEDeviceModel`. Ideally, you check whether it can be retrieved from aie-rt and otherwise add it here: https://github.com/nod-ai/iree-amd-aie/blob/c6156e3b203401c25d348ae995fd2f1ffc138787/runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h#L224

compiler/plugins/target/AMD-AIE/aievec/Passes.h
```diff
@@ -39,6 +41,54 @@ std::unique_ptr<mlir::Pass> createCanonicalizeVectorForAIEVecPass();
...
+ *
+ * If we want to read 32 bytes starting from 0x01, we can use a similar
+ * approach. The only consideration is that the shift operation requires 64-byte
+ * inputs, so the order of the of the shift and extracts is reversed.
+ *
+ * We do not currently support unaligned reads of vectors which are not 32-bytes
+ * or 64-bytes in length.
+ *
+ * TODO(newling) use this same approach to align writes to unaligned memory.
+ *  */
```
jtuyls, line 86: IREE uses `///` everywhere:
```suggestion
/// This pass ensures that reads from AIE tile memory are aligned according to
/// hardware constraints. For example, suppose we have 128 bytes in tile memory,
/// represented in hex as:
///
///    0x00 0x01 ... 0x7E 0x7F
///
/// On AIE-2, the (vector) read instructions from the tile memory into registers
/// must be aligned to 256-bits (32-bytes). So if we want to read 64 bytes
/// starting from 0x00 that is fine, but if we want to read 64 bytes starting
/// from 0x01, then we cannot use a vector read instruction directly. To work
/// around this constraint, we do the following:
///
/// 1. Perform a wider read, that loads 128 bytes (2x as many as we want)
///    starting from 0x00 into a larger register. That is, bytes 0x00-0x7F are
///    loaded, so we have 1 'junk' byte at the beginning and 63 'junk' bytes at
///    the end.
///
/// 2. Extract the target bytes 0x01 ... 0x40 from the larger register into a
///    smaller register in 2 steps, using 2 AIE specific instructions:
///
///   a) Extract:
///      https://www.xilinx.com/htmldocs/xilinx2023_2/aiengine_ml_intrinsics/intrinsics/group__intr__gpvectorconv__elem.html
///
///   b) Shift:
///      https://www.xilinx.com/htmldocs/xilinx2023_2/aiengine_ml_intrinsics/intrinsics/group__intr__gpvectorop__shift.html
///
///   First, we use the extract instruction to split the read 128-bytes into two
///   halves, 0x00-0x3F and 0x40-0x7F, each in its own 64-byte register. Then, we
///   use a shift operation to combine the upper 31 bytes from the first half
///   and the lower 33 bytes from the second half into a new 64-byte register.
///   This new register contains exactly the 64 bytes we want to read, starting
///   from 0x01.
///
/// If we want to read 32 bytes starting from 0x01, we can use a similar
/// approach. The only consideration is that the shift operation requires 64-byte
/// inputs, so the order of the of the shift and extracts is reversed.
///
/// We do not currently support unaligned reads of vectors which are not 32-bytes
/// or 64-bytes in length.
///
/// TODO(newling) use this same approach to align writes to unaligned memory.
```

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
```diff
@@ -925,6 +927,156 @@ struct CanonicalizeVectorForAIEVecPass
...
+    return readOp.getVector();
+  }
+
+  Value newIndex = rewriter.createOrFold<arith::SubIOp>(loc, oldIndex, offset);
+
+  SmallVector<bool> inBounds = readOp.getInBoundsValues();
+  bool allInBounds =
+      std::all_of(inBounds.begin(), inBounds.end(), [](bool b) { return b; });
+
+  if (shortBytes != 64 && shortBytes != 32) {
```
jtuyls, line 1014: These constants are architecture dependent as well, right?

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
```diff
@@ -925,6 +927,156 @@ struct CanonicalizeVectorForAIEVecPass
...
+    return readOp.getVector();
+  }
+
+  // Create the aligned transfer read for a vector 2x as long that covers the
+  // elements of the unaligned vector.
+  Value longVec = rewriter.create<vector::TransferReadOp>(
+      loc, longType, readOp.getSource(), SmallVector<Value>{newIndex}, padding,
+      SmallVector<bool>{allInBounds});
+
+  Value replacement = [&]() -> Value {
```
jtuyls, line 1029: Why the lambda? You could just do:
```
ShiftOp shiftOp;
if (shortBytes == 64) {
  ...
} else if (shotrBytes == 32) {
  ...
} else {
  return failure()...;
}
```

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
```diff
@@ -925,6 +927,156 @@ struct CanonicalizeVectorForAIEVecPass
...
+      // - Extract upper 64 bytes
+      // - Apply shift to obtain new 64 bytes
+      Value low = rewriter.create<ExtOp>(loc, shortType, longVec,
+                                         rewriter.getI8IntegerAttr(0));
+      Value upp = rewriter.create<ExtOp>(loc, shortType, longVec,
+                                         rewriter.getI8IntegerAttr(1));
+      return rewriter.createOrFold<ShiftOp>(loc, shortType, low, upp,
+                                            offsetBytes_i32);
+    }
+    assert(shortBytes == 32 && "Expect nb bytes 64 or 32.");
```
jtuyls, line 1041: I think it would be better to throw an error as `shortBytes` depends on data type + shape, so this might actually be hit in a use case from what I can tell?

jtuyls (changes requested):

jtuyls:
> This PR enables vectorization for convolution and resolves #820

Are you planning to add an e2e testcase?

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
newling: There's nothing there related to the core, are you sure that's a good place to add it? We already have vector size related stuff elsewhere like matmul instruction size:  https://github.com/nod-ai/iree-amd-aie/blob/c6156e3b203401c25d348ae995fd2f1ffc138787/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp#L50

compiler/plugins/target/AMD-AIE/aievec/Passes.h
newling: For now I'd prefer to keep this consistent with the style of the file (which use /** */ commenting). I intend to align the style of aievec code that has been copied and pasted from mlir-aie to iree-amd-aie in a future PR. Specifically I'd like to create a Passes.td file for aievec.

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
newling: Yes, well actually it's just 64 that is -- it's the number of bytes that the shift instruction operands must be. Good spot -- let me clean up the archictecture specific parts, and rerequest a review.

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
newling: Ok. I think lambdas are great but ok, ok. 

Note that I've already checked that shortBytes is either 32 or 64 at this point -- I do this _before_ creating longVec (because I like to make all IR checks before modifying IR). So this assertion should never be hit (in general this is true whenever I use `assert`. If one of my asserts is hit, it's because I made a mistake). 

But let me rejiggle this to not use the lambda nonetheless.

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
jtuyls: Yeah, at least all hardware related information is in one place then. Doing this will make it really easy to support multiple architectures. I know we currently have hardcoded information all over the place, but that will make it really cumbersome/buggy to support multiple/new architectures. All that information, like the map you're referencing, should be moved to the device model.

compiler/plugins/target/AMD-AIE/aievec/Passes.h
jtuyls: I see, sure. At some point we should align it.

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
jtuyls: > Note that I've already checked that shortBytes is either 32 or 64 at this point -- I do this before creating longVec (because I like to make all IR checks before modifying IR). So this assertion should never be hit (in general this is true whenever I use assert. If one of my asserts is hit, it's because I made a mistake).

Ok, that's the same way I try to use error vs assert. `assert` makes sense then.

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
```diff
@@ -925,6 +929,176 @@ struct CanonicalizeVectorForAIEVecPass
...
+      op->emitOpError()
+          << "has no AMDAIEDevice in the target attribute configuration. This "
+             "device-specific information is required to determine what vector "
+             "sizes and alignments are supported.";
+      return signalPassFailure();
+    }
+    AMDAIE::AMDAIEDeviceModel deviceModel =
+        AMDAIE::getDeviceModel(maybeDevice.value());
+
+    (void)deviceModel;
```
jtuyls, line 1089:
```suggestion
```

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
```diff
@@ -925,6 +929,176 @@ struct CanonicalizeVectorForAIEVecPass
...
+  }
+
+  void runOnOperation() override {
+    Operation *op = getOperation();
+
+    auto targetAttr = IREE::HAL::ExecutableTargetAttr::lookup(op);
+    std::optional<AMDAIE::AMDAIEDevice> maybeDevice =
+        mlir::iree_compiler::AMDAIE::getConfigAMDAIEDevice(targetAttr);
+    if (!maybeDevice) {
+      maybeDevice = AMDAIE::AMDAIEDevice::npu1_4col;
```
jtuyls, line 1079:
```suggestion
```

compiler/plugins/target/AMD-AIE/aievec/VectorToVectorConversions.cpp
```diff
@@ -925,6 +929,176 @@ struct CanonicalizeVectorForAIEVecPass
...
+  } else if (shortBits == shiftOperandBits / 2) {
+    // - Apply shift to obtain new 64 bytes, bottom 32 being the required ones
+    // - Extract lower 32 bytes
+    Value shift = rewriter.createOrFold<ShiftOp>(loc, longType, longVec,
+                                                 longVec, offsetBytes_i32);
+    replacement = rewriter.create<ExtOp>(loc, shortType, shift,
+                                         rewriter.getI8IntegerAttr(0));
+  }
+
+  else {
```
jtuyls, line 1047:
```suggestion
  } else {
```

jtuyls (changes requested):

newling:
> Are you planning to add an e2e testcase?

The e2e numerical tests are now vectorized because of the change here in Passes.cpp

jtuyls (approved):
LGTM, but maybe we could wait for https://github.com/nod-ai/iree-amd-aie/pull/865 to merge as those new uncached builds (MacOS 13) take a lot of time. Also, I think we can merge without the Windows E2E check. Hopefully, that one comes back online soon.

PR #868, opened 2024-11-01
Bump mlir-air to 3d1a4e19ff748897a37e0eb88b59999197dbc0f8

erwei-xilinx:


build_tools/ci/cpu_comparison/run.py
```diff
@@ -537,8 +537,8 @@ def aie_vs_llvm_cpu(
     config,
     test_file,
     use_ukernel=False,
-    tile_pipeline="pack-peel",
-    lower_to_aie_pipeline="objectFifo",
+    tile_pipeline="pad-pack",
+    lower_to_aie_pipeline="air",
```
yzhang93, line 541: Can you leave the default as "pack-peel" and "objectFifo"? You can add a test for pad-pack pipeline if you like, but I think the current tests are enough.

yzhang93 (approved):

build_tools/ci/run_matmul_test.sh
```diff
@@ -556,16 +556,16 @@ run_matmul_test \
 ###################################################################
 
 # TODO: re-enable after fixing in AIR
```
yzhang93, line 558: Nit: You can delete this line.

PR #870, opened 2024-11-01
Add HW specific vector sizes to device model

newling:
Initial commit. Future work: move all other HW specific constants here, for example https://github.com/nod-ai/iree-amd-aie/blob/9aeeeb8055cd17798d858f3a0daf795cd637548b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp#L50

This PR is depended on by https://github.com/nod-ai/iree-amd-aie/pull/867

### DIFF
--- a/runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
+++ b/runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
@@ -8,23 +8,16 @@
 #define IREE_AIE_RUNTIME_H
 
 #include <optional>
-#include <ostream>
-#include <sstream>
 #include <tuple>
 #include <type_traits>
 
+#include "iree-amd-aie/aie_runtime/AMDAIEEnums.h"
 #include "llvm/ADT/Twine.h"
 #include "llvm/Support/Debug.h"
-#include "llvm/Support/FormattedStream.h"
 #include "macros.h"
-// clang-format off
-#include "iree-amd-aie/aie_runtime/AMDAIEEnums.h"
-// clang-format on
 
 extern "C" {
-#include "xaie_hwcfg.h"
 #include "xaiengine.h"
-#include "xaiengine/xaie_device_aieml.h"
 #undef s8
 #undef u8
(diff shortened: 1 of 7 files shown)


runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -328,6 +351,14 @@ struct AMDAIEDeviceModel {
   uint8_t getStreamSwitchArbiterMax(uint8_t col, uint8_t row) const;
   uint8_t getStreamSwitchMSelMax(uint8_t col, uint8_t row) const;
 
+  uint32_t getVectorLoadStoreAlignmentBits() const {
```
makslevental, line 354: I strongly dislike the whole OOP getters pattern thing - that's why this object is a struct and not a class in the first place - so that no one could make anything private/protected in the future (which these aren't btw so why would I use these getters ever ...). What you want to do is make these `const` so no one can change them

jtuyls (changes requested):
Isn't this already included in: https://github.com/nod-ai/iree-amd-aie/pull/867 ? I don't think it should be a separate PR.

newling:
> Isn't this already included in: #867 ? I don't think it should be a separate PR.

"This struct is meant to be a thin wrapper around aie-rt" is no longer true. It seems like a significant decision to add information about the core instructions to this struct, that is why I made it a separate PR. But I'm fine just including it in #867

makslevental:
> "This struct is meant to be a thin wrapper around aie-rt" 

i mean this isn't/wasn't a hard rule - nothing i ever do is like that lol. i think putting arch specific stuff in `AMDAIEDeviceModel` is a perfectly fine place. i would prefer it wasn't a nested struct (I had a comment to that effect earlier but removed it since I'm now not qualified to nitpick) but that's fine.

newling:
@makslevental thanks for keeping an eye on this, and you have my permission to nitpick my PRs as much as you like

PR #871, opened 2024-11-01
[aievec] ShiftOp folder 

newling:
ShiftOp has 2 64 byte vector operands, and 1 64 byte vector result. The result is a concatenation of the N upper bytes from the first operand, and the 64 - N lower bytes from the second operand. This folder considers the special cases where N = 0 and N = 64, when it is possible to just return one of the operands directly. 

https://www.xilinx.com/htmldocs/xilinx2024_1/aiengine_ml_intrinsics/intrinsics/group__intr__gpvectorop__shift.html#gaa83afc81114a5e0d3aebea1e973248bc


jtuyls (approved):
I guess you want to get this in separately?

PR #874, opened 2024-11-05
Add hardware-aware canonicalization

jtuyls:
Canonicalization of the offsets/strides/sizes in doubly-strided operations, like DMA ops, can lead to overflow of the number of available bits in the hardware buffer descriptor fields. This PR adds logic to the `AMDAIECanonicalizeDoublyStridedOpPass` to not canonicalize if it would lead to such an overflow.

Note that there is more work to be done on avoiding overflow of available hardware bits and making it more robust, which is not addressed in this PR:
- Using aie-rt's transaction data structure and APIs to generate the control code transaction instead of the manual transaction creation we're currently doing: https://github.com/nod-ai/iree-amd-aie/blob/20867183c610ec870344d074c4be78e0aaac1515/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEControlCodeToTransaction.cpp#L33. As aie-rt has a lot of checks and will throw errors if hardware field bits are exceeded, this will make the flow way more robust, in the sense that errors will be thrown instead of hard-to-debug silent overflows, leading to hangs or numerical errors.
- A pass that can decanonicalize offset/strides/sizes access patterns, for example if available bits are already exceeded before canonicalization, for example when strides/sizes are derived from the `pack` operations. Note that this should typically not happen, but in general DMA ops could be create that exceed hardware limits and a decanonicalization transformation would let the flow handle more cases.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECanonicalizeDoublyStridedOp.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECanonicalizeDoublyStridedOp.cpp
@@ -6,12 +6,14 @@
 
 #include "iree-amd-aie/IR/AMDAIEDialect.h"
 #include "iree-amd-aie/Transforms/AMDAIEDmaUtils.h"
+#include "iree-amd-aie/Transforms/AMDAIEUtils.h"
 #include "iree-amd-aie/Transforms/Passes.h"
 #include "iree-amd-aie/Transforms/Transforms.h"
+#include "iree-amd-aie/aie_runtime/iree_aie_runtime.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
-#define DEBUG_TYPE "iree-amdaie-canonicalize-doubly-strided-dma"
+#define DEBUG_TYPE "iree-amdaie-canonicalize-doubly-strided-op"
 
 namespace mlir::iree_compiler::AMDAIE {
 
@@ -19,9 +21,17 @@ namespace {
 
 /// Recognize linear accesses across multiple DMA access dimensions and fold
 /// them.
-struct FoldDmaOpLinearDims
+class FoldDmaOpLinearDims
     : public OpInterfaceRewritePattern<AMDAIE::DoublyStridedOpInterface> {
   using OpInterfaceRewritePattern::OpInterfaceRewritePattern;
+
+ public:
+  FoldDmaOpLinearDims(
+      MLIRContext *ctx,
+      std::optional<std::reference_wrapper<const AMDAIEDeviceModel>>
+          deviceModel = std::nullopt)
+      : OpInterfaceRewritePattern<AMDAIE::DoublyStridedOpInterface>(ctx),
+        deviceModel(deviceModel) {}
 
   LogicalResult matchAndRewrite(AMDAIE::DoublyStridedOpInterface op,
                                 PatternRewriter &rewriter) const override {
@@ -34,14 +44,30 @@ struct FoldDmaOpLinearDims
     SmallVector<OpFoldResult> targetStrides = op.getTargetMixedStrides();
     SmallVector<OpFoldResult> newSourceOffsets, newSourceSizes,
         newSourceStrides, newTargetOffsets, newTargetSizes, newTargetStrides;
+    SmallVector<int64_t> maxSourceSizes, maxTargetSizes;
+    if (deviceModel) {
+      std::optional<uint8_t> sourceMemSpace = op.getSourceMemorySpaceAsUInt();
+      std::optional<uint8_t> targetMemSpace = op.getTargetMemorySpaceAsUInt();
+      if (!sourceMemSpace || !targetMemSpace) {
+        return rewriter.notifyMatchFailure(
+            op,
+            "expected a source and target memory space for hardware aware "
+            "linear dimension folding");
+      }
+      AMDAIE::DmaDimConfig dmaDimConfig(
+          deviceModel.value(), sourceMemSpace.value(), targetMemSpace.value());
+      maxSourceSizes = dmaDimConfig.getMaxSizes<CopyOpOperateOn::Source>();
+      maxTargetSizes = dmaDimConfig.getMaxSizes<CopyOpOperateOn::Target>();
+    }
     LogicalResult sourceRes = foldLinearDims(
         op.getContext(), sourceOffsets, sourceSizes, sourceStrides,
-        newSourceOffsets, newSourceSizes, newSourceStrides);
+        newSourceOffsets, newSourceSizes, newSourceStrides, maxSourceSizes);
     LogicalResult targetRes = foldLinearDims(
         op.getContext(), targetOffsets, targetSizes, targetStrides,
-        newTargetOffsets, newTargetSizes, newTargetStrides);
+        newTargetOffsets, newTargetSizes, newTargetStrides, maxTargetSizes);
     if (failed(sourceRes) && failed(targetRes)) {
-      return failure();
+      return rewriter.notifyMatchFailure(
+          op, "neither a source nor a target change");
     }
 
     rewriter.setInsertionPointAfter(op);
@@ -51,6 +77,11 @@ struct FoldDmaOpLinearDims
     rewriter.replaceOp(op, newDoublyStridedOp.getOperation());
     return success();
   }
+
+ private:
+  /// If the device model is set, this rewriter will use the device model to
+  /// keep hardware stride/size limitations into account.
+  std::optional<std::reference_wrapper<const AMDAIEDeviceModel>> deviceModel;
 };
 
 /// Fold single dimension linear accesses and make them implicit.
@@ -140,8 +171,21 @@ void AMDAIECanonicalizeDoublyStridedOpPass::runOnOperation() {
   Operation *parentOp = getOperation();
   MLIRContext *context = &getContext();
   RewritePatternSet patterns(context);
-
-  populateCanonicalizeDoublyStridedOpPatterns(patterns, foldSingleDims);
+  std::optional<AMDAIEDeviceModel> deviceModel;
+  if (hardwareAware) {
+    auto targetAttr = IREE::HAL::ExecutableTargetAttr::lookup(parentOp);
+    std::optional<AMDAIEDevice> maybeDevice = getConfigAMDAIEDevice(targetAttr);
+    if (!maybeDevice.has_value()) {
+      parentOp->emitOpError(
+          "hardware-aware canonicalization is enabled, but op has no "
+          "AMDAIEDevice "
+          "in the target attribute configuration");
+      return signalPassFailure();
+    }
+    deviceModel = AMDAIE::getDeviceModel(maybeDevice.value());
+  }
+  populateCanonicalizeDoublyStridedOpPatterns(patterns, foldSingleDims,
+                                              deviceModel);
   if (failed(applyPatternsAndFoldGreedily(parentOp, std::move(patterns)))) {
     parentOp->emitOpError(
         "failed to canonicalize doubly strided DMA operations");
@@ -151,10 +195,13 @@ void AMDAIECanonicalizeDoublyStridedOpPass::runOnOperation() {
 
 }  // namespace
 
-void populateCanonicalizeDoublyStridedOpPatterns(RewritePatternSet &patterns,
-                                                 bool foldSingleDims) {
+void populateCanonicalizeDoublyStridedOpPatterns(
+    RewritePatternSet &patterns, bool foldSingleDims,
+    std::optional<std::reference_wrapper<const AMDAIEDeviceModel>>
+        deviceModel) {
   patterns.add<FoldDmaOpUnitDims>(patterns.getContext());
-  patterns.add<FoldDmaOpLinearDims>(patterns.getContext());
+  patterns.add<FoldDmaOpLinearDims>(patterns.getContext(),
+                                    std::move(deviceModel));
   if (foldSingleDims) {
     patterns.add<FoldDmaOpSingleDims>(patterns.getContext());
   }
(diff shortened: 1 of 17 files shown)


makslevental:
>  transaction based control code generation means moving to using aie-rt.

We're already always using aie-rt (which I know you know). Moving to "transactions" just means not leaving aie-rt to the last step and thus aie-rt will error right at the point where the width overflow happens (right after the "illegal" canonicalization, or thereabouts).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECanonicalizeDoublyStridedOp.cpp
```diff
@@ -51,6 +77,14 @@ struct FoldDmaOpLinearDims
     rewriter.replaceOp(op, newDoublyStridedOp.getOperation());
     return success();
   }
+
+ private:
+  /// If hardware aware folding is enabled, and the device model is set, this
```
newling, line 82: The logic could be simplified I think:

don't have hardwareAwareFolding as a member of this class, just deviceModel. The decision of whether to be hardware aware is based on whether deviceModel is nullopt or not.

The logic in the pass can then be

 std::optional<AMDAIEDevice> maybeDevice; 
 if (hardware aware){
   // try to set maybeDevice. If fail to do so, signalPassFailure. 
 }
 
  populateCanonicalizeDoublyStridedOpPatterns(maybeDevice).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECombineStridedOps.cpp
```diff
@@ -71,10 +72,16 @@ struct CombineStridedOps
...
-      uint8_t sourceMemspaceInt = npuDmaOp.getSourceMemorySpaceAsUInt();
-      uint8_t targetMemspaceInt = npuDmaOp.getTargetMemorySpaceAsUInt();
-      AMDAIE::DmaDimConfig dmaDimConfig(deviceModel, sourceMemspaceInt,
-                                        targetMemspaceInt);
+      std::optional<uint8_t> sourceMemspaceInt =
+          npuDmaOp.getSourceMemorySpaceAsUInt();
+      std::optional<uint8_t> targetMemspaceInt =
+          npuDmaOp.getTargetMemorySpaceAsUInt();
+      if (!sourceMemspaceInt || !targetMemspaceInt) {
+        return npuDmaOp.emitOpError()
```
newling, line 80: emitting errors from patterns, is there an alternative?

jtuyls:
> Can you please clarify these points in the summary:
> 
> > Moving to transaction based control code generation, as aie-rt will throw errors if hardware field bits are exceeded.
> 
> Throwing errors if hardware resources are exceeded sounds like a good thing, so I'm interpreting this as: moving to transaction based control code generation means moving to using aie-rt. Is that correct?

Yes, using aie-rt APIs and transaction data structure to generate the control code transaction instead of the manual way we're doing currently, which doesn't perform (m)any checks: https://github.com/nod-ai/iree-amd-aie/blob/20867183c610ec870344d074c4be78e0aaac1515/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEControlCodeToTransaction.cpp#L33 

> > A pass that can decanonicalize offset/strides/sizes access patterns, for example if available bits are already before canonicalization.
> 
> Is this about undoing preexisting overflow?

Yes, if the initial strides/sizes generated from the pack would overflow for example. Typically, this shouldn't happen, but in principal it could.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/AMDAIEDmaUtilsTest.cpp
```diff
@@ -271,6 +271,97 @@ TEST_F(AccessPatternCombinationTest, FailCombineAccessPatterns) {
...
+class FoldLinearDimsTest : public ::testing::Test {
+ protected:
+  FoldLinearDimsTest() : rewriter(&context), loc(UnknownLoc::get(&context)) {
+    context.loadDialect<arith::ArithDialect>();
+  }
+
+  SmallVector<OpFoldResult> toOpFoldResult(const SmallVector<int64_t> &values) {
+    return llvm::map_to_vector(values, [&](int64_t v) -> OpFoldResult {
+      return getAsIndexOpFoldResult(&context,
+                                    v);  // rewriter.getI64IntegerAttr(v);
```
newling, line 283:
```suggestion
                                    v);  
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/AMDAIEDmaUtilsTest.cpp
```diff
@@ -271,6 +271,97 @@ TEST_F(AccessPatternCombinationTest, FailCombineAccessPatterns) {
                              {128, 1}, {32, 0}, {96, 64}, {128, 1}, 4, false);
 }
 
+class FoldLinearDimsTest : public ::testing::Test {
+ protected:
+  FoldLinearDimsTest() : rewriter(&context), loc(UnknownLoc::get(&context)) {
+    context.loadDialect<arith::ArithDialect>();
+  }
+
+  SmallVector<OpFoldResult> toOpFoldResult(const SmallVector<int64_t> &values) {
```
newling, line 280:
```suggestion
  SmallVector<OpFoldResult> toOpFoldResult(ArrayRef<int64_t> values) {
```

and elsewhere in this file (big nit)

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/AMDAIEDmaUtilsTest.cpp
```diff
@@ -271,6 +271,97 @@ TEST_F(AccessPatternCombinationTest, FailCombineAccessPatterns) {
...
+    SmallVector<OpFoldResult> expectedStridesValues =
+        toOpFoldResult(expectedStrides);
+    SmallVector<OpFoldResult> newOffsets;
+    SmallVector<OpFoldResult> newSizes;
+    SmallVector<OpFoldResult> newStrides;
+    if (shouldSucceed) {
+      EXPECT_TRUE(succeeded(foldLinearDims(&context, offsetsValues, sizesValues,
+                                           stridesValues, newOffsets, newSizes,
+                                           newStrides, maxSizes)));
+      EXPECT_EQ(newOffsets, expectedOffsetsValues);
```
newling, line 311: Common trailing code in if- and else- can be moved after.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/canonicalize_doubly_strided_op_hardware_aware.mlir
```diff
@@ -0,0 +1,53 @@
+// RUN: iree-opt --pass-pipeline="builtin.module(func.func(iree-amdaie-canonicalize-doubly-strided-op{hardware-aware=true},canonicalize))" --split-input-file -allow-unregistered-dialect --verify-diagnostics %s | FileCheck %s
```
newling, line 1: Do you need to have `canonicalize` in the test pass pipeline?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaUtils.h
```diff
@@ -131,13 +131,20 @@ LogicalResult combineAccessPatterns(RewriterBase &rewriter,
 
 /// Fold subsequent dimensions within a strided access pattern that describe a
 /// single linear access. Returns `success` if folding took place.
+/// Accepts optional maximum size constraints as an array of integers. Note that
+/// maxSizes` is expected to be provided in the same order as `sizes` and they
```
newling, line 135:
```suggestion
/// `maxSizes` is expected to be provided in the same order as `sizes` and they
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaUtils.h
```diff
@@ -131,13 +131,20 @@ LogicalResult combineAccessPatterns(RewriterBase &rewriter,
...
+/// are compared from right to left/innermost to outermost. Also note that the
```
newling, line 136:
```suggestion
/// are compared from right to left (innermost to outermost). Also note that the
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDmaUtils.h
```diff
@@ -131,13 +131,20 @@ LogicalResult combineAccessPatterns(RewriterBase &rewriter,
...
+/// number of max sizes might exceed the number of sizes and the the other way
```
newling, line 137:
```suggestion
/// number of max sizes might exceed the number of sizes and the other way
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -244,17 +250,23 @@ class AMDAIEUnrollLocalLoops : public OpRewritePattern<scf::ForOp> {
...
         if (auto dmaOp = dyn_cast<AMDAIE::DmaCpyNdOp>(*it)) {
-          uint8_t sourceMemSpaceInt = dmaOp.getSourceMemorySpaceAsUInt();
-          uint8_t targetMemSpaceInt = dmaOp.getTargetMemorySpaceAsUInt();
-          if (targetMemSpaceInt > sourceMemSpaceInt) {
+          std::optional<uint8_t> sourceMemSpaceInt =
+              dmaOp.getSourceMemorySpaceAsUInt();
+          std::optional<uint8_t> targetMemSpaceInt =
+              dmaOp.getTargetMemorySpaceAsUInt();
+          if (!sourceMemSpaceInt || !targetMemSpaceInt) {
+            return dmaOp.emitOpError()
```
newling, line 258: Emiting error from pattern. Assert, or notifyMatchFailure?

newling (changes requested):
Looks good, I have no blocking comments, just a few small requests/suggestions.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECanonicalizeDoublyStridedOp.cpp
```diff
@@ -34,12 +46,26 @@ struct FoldDmaOpLinearDims
     SmallVector<OpFoldResult> targetStrides = op.getTargetMixedStrides();
     SmallVector<OpFoldResult> newSourceOffsets, newSourceSizes,
         newSourceStrides, newTargetOffsets, newTargetSizes, newTargetStrides;
+    SmallVector<int64_t> maxSourceSizes, maxTargetSizes;
+    if (hardwareAwareFolding && deviceModel) {
+      std::optional<uint8_t> sourceMemSpace = op.getSourceMemorySpaceAsUInt();
+      std::optional<uint8_t> targetMemSpace = op.getTargetMemorySpaceAsUInt();
+      if (!sourceMemSpace || !targetMemSpace) {
+        return op.emitOpError()
```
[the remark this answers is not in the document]
jtuyls: It should be `notifyMatchFailure`. I will update it here and in other places I touched.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
jtuyls: Updated to `notifyMatchFailure`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/canonicalize_doubly_strided_op_hardware_aware.mlir
jtuyls: Not needed, so I removed it.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/AMDAIEDmaUtilsTest.cpp
jtuyls: Thanks for the catch, updated it.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECombineStridedOps.cpp
jtuyls: Updated to notifyMatchFailure

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECanonicalizeDoublyStridedOp.cpp
jtuyls: Yeah, I think that makes sense for the pattern class, so updated it,

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIECanonicalizeDoublyStridedOp.cpp
```diff
@@ -140,8 +174,19 @@ void AMDAIECanonicalizeDoublyStridedOpPass::runOnOperation() {
   Operation *parentOp = getOperation();
   MLIRContext *context = &getContext();
   RewritePatternSet patterns(context);
-
-  populateCanonicalizeDoublyStridedOpPatterns(patterns, foldSingleDims);
+  auto targetAttr = IREE::HAL::ExecutableTargetAttr::lookup(parentOp);
+  std::optional<AMDAIE::AMDAIEDeviceModel> deviceModel;
```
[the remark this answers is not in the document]
jtuyls: I used `AMDAIE::` for consistency, but I saw you were just using `AMDAIEDeviceModel` in your PR and I like the brevity, so I think we could update it in all places.

jtuyls:
@newling I addressed the comments, could you check again?

newling (approved):
Thanks for addressing my comments @jtuyls

PR #875, opened 2024-11-05
[Strix][KernelDispatch][Ukernel] Enable Matmul intrinsics for Strix

Abhishek-Varma:
-- This commit enables Matmul intrinsics for Strix bf16xbf16xf32 ukernel.
-- This is an initial PR whose scope is only to generate correct target intrinsics
    based on whether we target 'phoenix' (default) or 'strix'.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
@@ -103,9 +103,8 @@ FailureOr<uint32_t> getAIEMacNumElements(Type inputType, Type outputType) {
   return failure();
 }
 
-FailureOr<std::array<uint32_t, 3>> getAIEMatmulInstructionSize(Type elTypeLhs,
-                                                               Type elTypeRhs,
-                                                               Type elTypeAcc) {
+FailureOr<std::array<uint32_t, 3>> getAIEMatmulInstructionSize(
+    Type elTypeLhs, Type elTypeRhs, Type elTypeAcc, AMDAIEDevice targetDevice) {
   bool allFloatingPoint = isa<FloatType>(elTypeLhs) &&
                           isa<FloatType>(elTypeRhs) &&
                           isa<FloatType>(elTypeAcc);
@@ -123,7 +122,13 @@ FailureOr<std::array<uint32_t, 3>> getAIEMatmulInstructionSize(Type elTypeLhs,
 
   if (allFloatingPoint) {
     if (nBitsLhs == 16 && nBitsRhs == 16 && nBitsAcc == 32) {
-      return std::array<uint32_t, 3>{4, 4, 8};
+      if (targetDevice == AMDAIEDevice::npu4) {
+        // Strix intrinsics.
+        return std::array<uint32_t, 3>{8, 8, 8};
+      } else {
+        // Phoenix intrinsics.
+        return std::array<uint32_t, 3>{4, 4, 8};
+      }
     }
     // There is only 1 floating point case in the table (handled above).
     return failure();
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
@@ -27,7 +27,7 @@ using detail::findLargestFactor;
 namespace {
 
 FailureOr<std::array<uint32_t, 3>> getMatmulInstructionSize(
-    linalg::LinalgOp op) {
+    linalg::LinalgOp op, AMDAIEDevice targetDevice) {
   auto getElementType = [](Value v) {
     return cast<ShapedType>(v.getType()).getElementType();
   };
@@ -39,17 +39,19 @@ FailureOr<std::array<uint32_t, 3>> getMatmulInstructionSize(
   auto elTypeRhs = getElementType(op->getOperand(1));
   auto elTypeAcc = getElementType(op->getResult(0));
 
-  return getAIEMatmulInstructionSize(elTypeLhs, elTypeRhs, elTypeAcc);
+  return getAIEMatmulInstructionSize(elTypeLhs, elTypeRhs, elTypeAcc,
+                                     targetDevice);
 }
 
 FailureOr<std::array<uint32_t, 3>> getPackedSize(linalg::LinalgOp linalgOp,
                                                  uint64_t M, uint64_t N,
-                                                 uint64_t K) {
+                                                 uint64_t K,
+                                                 AMDAIEDevice targetDevice) {
   // Depending on the operand/result element types, there might be a specific
   // vector instruction size that must be used on AIE. Some types do not have
   // vector instructions, for example if operands are 32-bit types.
   FailureOr<std::array<uint32_t, 3>> maybeInstructionSize =
-      getMatmulInstructionSize(linalgOp);
+      getMatmulInstructionSize(linalgOp, targetDevice);
 
   // Operand/result element types do not have vector instructions. In this case,
   // try for a packing of 4x4x8, but if the tensor dimensions M, N, and K are
@@ -95,7 +97,8 @@ class ParameterSetting {
   }
 
   static FailureOr<ParameterSetting> create(linalg::LinalgOp linalgOp,
-                                            bool isPackPeel, bool isObjectFifo);
+                                            bool isPackPeel, bool isObjectFifo,
+                                            AMDAIEDevice targetDevice);
 
   uint32_t getM0() const { return M0; }
   uint32_t getN0() const { return N0; }
@@ -142,9 +145,9 @@ class ParameterSetting {
   uint32_t k1Pack;
 };
 
-FailureOr<ParameterSetting> ParameterSetting::create(linalg::LinalgOp linalgOp,
-                                                     bool isPackPeel,
-                                                     bool isObjectFifo) {
+FailureOr<ParameterSetting> ParameterSetting::create(
+    linalg::LinalgOp linalgOp, bool isPackPeel, bool isObjectFifo,
+    AMDAIEDevice targetDevice) {
   auto initType =
       llvm::cast<ShapedType>(linalgOp.getDpsInitOperand(0)->get().getType());
   ArrayRef<int64_t> initShape = initType.getShape();
@@ -184,7 +187,7 @@ FailureOr<ParameterSetting> ParameterSetting::create(linalg::LinalgOp linalgOp,
 
   unsigned scaleFactor = maybeScaleFactor.value();
 
-  auto maybePackedSize = getPackedSize(linalgOp, M, N, K);
+  auto maybePackedSize = getPackedSize(linalgOp, M, N, K, targetDevice);
   if (failed(maybePackedSize)) return failure();
   auto [m1Pack, n1Pack, k1Pack] = maybePackedSize.value();
 
@@ -315,11 +318,12 @@ static SmallVector<int64_t> setInnerPermB(bool isMatmulTransposeB) {
 
 static LogicalResult setRootConfigForPackPeelPipeline(
     mlir::FunctionOpInterface entryPointFn, linalg::LinalgOp linalgOp,
-    LowerToAIEPassPipeline useLowerToAIEPipeline, bool isMatmulTransposeB) {
+    LowerToAIEPassPipeline useLowerToAIEPipeline, bool isMatmulTransposeB,
+    AMDAIEDevice targetDevice) {
   bool isObjectFifo =
       useLowerToAIEPipeline == LowerToAIEPassPipeline::ObjectFifo;
-  auto maybePackPeelTiling =
-      ParameterSetting::create(linalgOp, /*isPackPeel=*/true, isObjectFifo);
+  auto maybePackPeelTiling = ParameterSetting::create(
+      linalgOp, /*isPackPeel=*/true, isObjectFifo, targetDevice);
   if (failed(maybePackPeelTiling)) return failure();
   auto packPeelTiling = maybePackPeelTiling.value();
 
@@ -415,9 +419,9 @@ static LogicalResult setRootConfigForPackPeelPipeline(
 
 static LogicalResult setRootConfigForPadPackPipeline(
     mlir::FunctionOpInterface entryPointFn, linalg::LinalgOp linalgOp,
-    bool isMatmulTransposeB) {
+    bool isMatmulTransposeB, AMDAIEDevice targetDevice) {
   auto maybePadPackTiling = ParameterSetting::create(
-      linalgOp, /*isPackPeel=*/false, /*isObjectFifo=*/false);
+      linalgOp, /*isPackPeel=*/false, /*isObjectFifo=*/false, targetDevice);
   if (failed(maybePadPackTiling)) return failure();
   auto padPackTiling = maybePadPackTiling.value();
 
@@ -471,11 +475,12 @@ static LogicalResult setRootConfigForPadPackPipeline(
 //===----------------------------------------------------------------------===//
 
 static LogicalResult setRootConfigForConvDecomposePipeline(
-    mlir::FunctionOpInterface entryPointFn, linalg::LinalgOp linalgOp) {
+    mlir::FunctionOpInterface entryPointFn, linalg::LinalgOp linalgOp,
+    AMDAIEDevice targetDevice) {
   MLIRContext *context = entryPointFn.getContext();
 
   FailureOr<std::array<uint32_t, 3>> maybeInstructionSize =
-      getMatmulInstructionSize(linalgOp);
+      getMatmulInstructionSize(linalgOp, targetDevice);
   int64_t OW = 4;
   int64_t OC = 4;
   int64_t IC = 8;
@@ -683,13 +688,14 @@ static bool isMatmulTransposeB(linalg::GenericOp genericOp) {
 /// transposition.
 static LogicalResult setTransposeLikeOpRootConfig(
     mlir::FunctionOpInterface entryPointFn, linalg::LinalgOp linalgOp,
-    TilePassPipeline passPipeline,
-    LowerToAIEPassPipeline useLowerToAIEPipeline) {
+    TilePassPipeline passPipeline, LowerToAIEPassPipeline useLowerToAIEPipeline,
+    AMDAIEDevice targetDevice) {
   if (passPipeline == TilePassPipeline::PackPeelPipeline)
-    return setRootConfigForPackPeelPipeline(entryPointFn, linalgOp,
-                                            useLowerToAIEPipeline, true);
+    return setRootConfigForPackPeelPipeline(
+        entryPointFn, linalgOp, useLowerToAIEPipeline, true, targetDevice);
   else if (passPipeline == TilePassPipeline::PadPackPipeline)
-    return setRootConfigForPadPackPipeline(entryPointFn, linalgOp, true);
+    return setRootConfigForPadPackPipeline(entryPointFn, linalgOp, true,
+                                           targetDevice);
   return linalgOp.emitError(
       "Unhandled pass pipeline in setTransposeLikeOpRootConfig.");
 }
@@ -698,16 +704,18 @@ static LogicalResult setTransposeLikeOpRootConfig(
 // Root Configurations
 //===----------------------------------------------------------------------===//
 
-static LogicalResult setRootConfig(
-    mlir::FunctionOpInterface entryPointFn, linalg::GenericOp genericOp,
-    TilePassPipeline passPipeline,
-    LowerToAIEPassPipeline useLowerToAIEPipeline) {
+static LogicalResult setRootConfig(mlir::FunctionOpInterface entryPointFn,
+                                   linalg::GenericOp genericOp,
+                                   TilePassPipeline passPipeline,
+                                   LowerToAIEPassPipeline useLowerToAIEPipeline,
+                                   AMDAIEDevice targetDevice) {
   assert(!getLoweringConfig<IREE::Codegen::LoweringConfigAttr>(genericOp) &&
          "expected lowering_config is not set");
 
   if (isMatmulTransposeB(genericOp) &&
-      succeeded(setTransposeLikeOpRootConfig(
-          entryPointFn, genericOp, passPipeline, useLowerToAIEPipeline))) {
+      succeeded(
+          setTransposeLikeOpRootConfig(entryPointFn, genericOp, passPipeline,
+                                       useLowerToAIEPipeline, targetDevice))) {
     return success();
   }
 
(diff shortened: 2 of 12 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -417,6 +417,16 @@ def AMDAIELoweringStrategy :
...
+      )}]>,
+    Option<"targetDevice", "target-device",
+      "mlir::iree_compiler::AMDAIE::AMDAIEDevice",
+      /*default=*/"mlir::iree_compiler::AMDAIE::AMDAIEDevice::npu1_4col",
+      "AIE device to target",
+      [{::llvm::cl::values(
+        clEnumValN(mlir::iree_compiler::AMDAIE::AMDAIEDevice::npu1_4col, "phoenix",
+                   "Compile for Phoenix"),
+        clEnumValN(mlir::iree_compiler::AMDAIE::AMDAIEDevice::npu4, "strix",
+                   "Compile for Strix")
```
jtuyls, line 429: I think we should  use `NPU` naming: `npu1_4col`, `npu4`, instead of phoenix/strix, for consistency and to support multiple versions of a device generation. For example, while we're not targeting the 4 column version, there exists a strix 4 column version next to the 8 column one.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.h
```diff
@@ -29,6 +30,8 @@ enum class TilePassPipeline {
   None
 };
 
+enum class AMDAIEDevice : uint32_t;
```
jtuyls, line 33: This should not be needed?

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
makslevental: https://github.com/amd/xdna-driver/blob/ca8ecbf43bb62ff6e44185e5d6a34846e3ea1a7e/src/driver/amdxdna/amdxdna_pci_drv.c#L47

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.h
Abhishek-Varma: Correct. Removed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
```diff
@@ -123,7 +122,13 @@ FailureOr<std::array<uint32_t, 3>> getAIEMatmulInstructionSize(Type elTypeLhs,
 
   if (allFloatingPoint) {
     if (nBitsLhs == 16 && nBitsRhs == 16 && nBitsAcc == 32) {
-      return std::array<uint32_t, 3>{4, 4, 8};
+      if (targetDevice == AMDAIEDevice::npu4) {
```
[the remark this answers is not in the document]
Abhishek-Varma: The integer sizes specific to `npu4` is not known to me and I tried looking into https://github.com/Xilinx/mlir-aie/blob/main/include/aie/Dialect/AIEVec/IR/AIEVecOps.td but I believe since npu4 hasn't been made public into `mlir-aie` yet, this will need to wait.

I've added an assert - let me know if I should remove it.

Also, I've now moved the API to AMDAIEDeviceModel class in `iree_aie_runtime.cc`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -27,7 +27,7 @@ using detail::findLargestFactor;
 namespace {
 
 FailureOr<std::array<uint32_t, 3>> getMatmulInstructionSize(
```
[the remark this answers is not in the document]
Abhishek-Varma: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_ukernel.mlir
```diff
@@ -295,3 +295,48 @@ func.func @zero_fill_matmul_elmwise(%arg0 : tensor<8x16x4x8xbf16>, %arg1 : tenso
 // CHECK: linalg.generic
 // CHECK-SAME: iterator_types = ["parallel", "parallel", "parallel", "parallel"]
 // CHECK: return
+
+// -----
+
+#executable_target_amdaie_xclbin_fb = #hal.executable.target<"amd-aie", "amdaie-xclbin-fb", {target_arch = "chip-tbd", ukernels = "all"}>
```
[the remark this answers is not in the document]
Abhishek-Varma: This is testing that the ukernel name/call being generated has identified the intrinsics correctly and because of this a rank-2 matmul wouldn't work. The rank needs to be either 4 or 6 (the latter for pack-peel pipeline that we want to focus and ensure the intrinsics for).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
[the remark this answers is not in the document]
jtuyls: > The integer sizes specific to npu4 is not known to me 

i32 cases shouldn't matter, anything should work as it's scalarized.
i8 is `8x8x8` as well.
Note bfloat16 is only 8x8x8 here as bfp16 is used under the hood for the ukernel we're targeting

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.h
```diff
@@ -7,6 +7,8 @@
 #ifndef IREE_AMD_AIE_TRANSFORMS_KERNELDISPATCH_H_
 #define IREE_AMD_AIE_TRANSFORMS_KERNELDISPATCH_H_
 
+#include "iree-amd-aie/aie_runtime/AMDAIEEnums.h"
+#include "iree-amd-aie/aie_runtime/iree_aie_runtime.h"
```
jtuyls, line 11: I don't think that this include is needed in the header? It should be put in the cpp file.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/CMakeLists.txt
```diff
@@ -54,7 +54,8 @@ iree_lit_test_suite(
     "lowering_strategy.mlir"
     "lowering_strategy_conv.mlir"
     "lowering_strategy_failures.mlir"
-    "lowering_strategy_objectfifo.mlir"
+    "lowering_strategy_objectfifo_phoenix.mlir"
+    "lowering_strategy_objectfifo_strix.mlir"
```
jtuyls, line 58: I think it's better to rename these files to npu1, npu4 as well for clarity.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_loops_for_vectorization.mlir
```diff
@@ -9,7 +9,8 @@
 !t4_f32  = tensor<64x64x64x64xf32>
 
 
-module {
+#executable_target_amdaie_xclbin_fb = #hal.executable.target<"amd-aie", "amdaie-xclbin-fb", {target_device = "npu1_4col", ukernels = "none"}>
```
jtuyls, line 12: It's useful to add a test without device as well and check that the device not found error is thrown.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lowering_strategy_objectfifo_phoenix.mlir
```diff
@@ -1,7 +1,7 @@
-// RUN: iree-opt --split-input-file --pass-pipeline='builtin.module(iree-amdaie-lowering-strategy{use-pass-pipeline=pack-peel use-lower-to-aie-pipeline=objectFifo})' %s | FileCheck %s
+// RUN: iree-opt --split-input-file --pass-pipeline='builtin.module(iree-amdaie-lowering-strategy{use-pass-pipeline=pack-peel use-lower-to-aie-pipeline=objectFifo target-device=npu1_4col})' %s | FileCheck %s --check-prefix=PHOENIX
```
jtuyls, line 1: If you're splitting up the tests in two files, you don't need the prefix.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lowering_strategy_objectfifo_strix.mlir
```diff
@@ -0,0 +1,26 @@
+// RUN: iree-opt --split-input-file --pass-pipeline='builtin.module(iree-amdaie-lowering-strategy{use-pass-pipeline=pack-peel use-lower-to-aie-pipeline=objectFifo target-device=npu4})' %s | FileCheck %s --check-prefix=STRIX
```
jtuyls, line 1: If you're splitting up the tests in two files, you don't need the prefix.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc
```diff
@@ -629,6 +629,135 @@ struct AMDAIEDeviceModel getDeviceModel(AMDAIEDevice device) {
...
+/// table return failure.
+///
+/// Example: consider the first line of the table:
+///   `vector<4x16xi8>`  | `vector<16x8xi4>`  | `vector<4x8xi32>`
+///
+/// This first line says that if 'lhs' is an i8 tensor, 'rhs' is an i4 tensor
+/// and 'accumulator' is an i32 tensor, then there is an AIE instruction for
+/// matmul with m = 4, n = 8, k = 16.
+static llvm::DenseMap<uint32_t, std::array<uint32_t, 3>>
+    &getIntegerMatmulInstructionSizeMap() {
```
jtuyls, line 670: This would be a good time to rename this to `getNpu1IntegerMatmulInstructionSizeMap`

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc
```diff
@@ -629,6 +629,135 @@ struct AMDAIEDeviceModel getDeviceModel(AMDAIEDevice device) {
...
+      // `vector<4x2xi32>`  | `vector<2x4xi16>`  | `vector<4x4xi64>`
+      {getElementTypeKey(32, 16, 64), {4, 4, 2}},
+  };
+  return matmulIntSizes;
+}
+
+/// Return the AIE instruction size (m, n, k) for the integer types with
+/// bitwidths nBitsLhs, nBitsRhs, and nBitsAcc. Based on the table above.
+static llvm::FailureOr<std::array<uint32_t, 3>>
+getAIEIntegerMatmulInstructionSize(uint32_t nBitsLhs, uint32_t nBitsRhs,
```
jtuyls, line 707:
```suggestion
getNpu1IntegerMatmulInstructionSize(uint32_t nBitsLhs, uint32_t nBitsRhs,
```

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc
```diff
@@ -629,6 +629,135 @@ struct AMDAIEDeviceModel getDeviceModel(AMDAIEDevice device) {
...
+    }
+    // There is only 1 floating point case in the table (handled above).
+    return failure();
+  }
+
+  assert(allInteger &&
+         "expected all element types to either be all float types or all "
+         "integer types");
+  assert((device != AMDAIEDevice::npu4) &&
+         "currently not compiling for all integer element types targetting "
```
jtuyls, line 755:
```suggestion
         "currently not compiling for all integer element types targeting "
```

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc
```diff
@@ -629,6 +629,135 @@ struct AMDAIEDeviceModel getDeviceModel(AMDAIEDevice device) {
...
+      }
+    }
+    // There is only 1 floating point case in the table (handled above).
+    return failure();
+  }
+
+  assert(allInteger &&
+         "expected all element types to either be all float types or all "
+         "integer types");
+  assert((device != AMDAIEDevice::npu4) &&
```
jtuyls, line 754: Shouldn't this be: `device == AMDAIEDevice::npu1_4col ...` as integer types are only supported for npu1 for now.

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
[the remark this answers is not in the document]
Abhishek-Varma: Yeah, so I see that we maintain a [map of intrinsics for all integer types](https://github.com/nod-ai/iree-amd-aie/blob/fc0480187f9dabb2cbf077a3ed643a8fd5cf6fca/runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc#L674-L700) - it is those specific instrinsic sizes for all types that I'm not sure of. currently for `npu4`. How to derive the same (and what is the source where I can take a look at for `npu4`) ?

One thing which I thought was to just change 4 -> 8 and 2 -> 4 and have a new API `getNpu4IntegerMatmulInstructionSize`.

I could add just an `i8` instance for `npu4` but that'd be too specific for the integer types as currently we do that only for float types whereas we have a generic intrinsic map for integer bit-widths for `npu1_4col`.

Hence currently added an assert.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.h
Abhishek-Varma: No, these are required here because of the following reasons :-

1. I use `AMDAIEDevice` [here](https://github.com/nod-ai/iree-amd-aie/blob/fc0480187f9dabb2cbf077a3ed643a8fd5cf6fca/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.h#L43) within the same file.

2. [PassDetail.h](https://github.com/nod-ai/iree-amd-aie/blob/e25d3484be8f53b89aff1d51a9180c2c3768c14d/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/PassDetail.h#L10) includes KernelDispatch.h -> and since this is how the [`.td` derives the types/attr too](https://github.com/nod-ai/iree-amd-aie/blob/e25d3484be8f53b89aff1d51a9180c2c3768c14d/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/PassDetail.h#L86).

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc
Abhishek-Varma: Yes, I have changed it now.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.h
jtuyls: AMDAIEDevice is defined in `AMDAIEEnums.h.inc`, so I don't think you need `iree_aie_runtime.h`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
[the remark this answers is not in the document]
jtuyls: Yeah, I don't think we should add them in this PR. We should throw an error though instead of an assert

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.h
Abhishek-Varma: Sure, I've moved `iree_aie_runtime.h` from this file and included it in `KernelDispatch.cpp` and `InsertLoopsForVectorization.cpp`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp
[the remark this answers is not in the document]
Abhishek-Varma: Done.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc
jtuyls: I don't see this having been renamed

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lowering_strategy_objectfifo_npu1.mlir
```diff
@@ -1,7 +1,7 @@
...
+// RUN: iree-opt --split-input-file --pass-pipeline='builtin.module(iree-amdaie-lowering-strategy{use-pass-pipeline=pack-peel use-lower-to-aie-pipeline=objectFifo target-device=npu1_4col})' %s | FileCheck %s
 
 // CHECK-{LITERAL}: #config = #iree_codegen.lowering_config<tile_sizes = [[64, 64], [0, 0, 1], [1, 1, 0, 0, 0, 0]]>
-// CHECK-{LITERAL}: #packingConfig = #amdaie.packing_config<packing_config = [{packedSizes = [32, 32, 32], transposePackIndices = [1], unpackEmpty = [false], innerPerm = [[1, 0]], outerPerm = [[0, 1]]}, {packedSizes = [0, 0, 0, 4, 4, 8], transposePackIndices = [0, 1, 2], unpackEmpty = [false, false, true], innerPerm = [[0, 1], [1, 0], [0, 1]], outerPerm = [[0, 1, 3, 2], [0, 1, 3, 2], [0, 1, 3, 2]]}]>
+// CHECK-{LITERAL}: #packingConfig = #amdaie.packing_config<packing_config = [{packedSizes = [32, 32, 32], transposePackIndices = [1], unpackEmpty = [false], innerPerm = [[1, 0]], outerPerm = [[0, 1]]}, {packedSizes = [0, 0, 0, 8, 8, 8], transposePackIndices = [0, 1, 2], unpackEmpty = [false, false, true], innerPerm = [[0, 1], [1, 0], [0, 1]], outerPerm = [[0, 1, 3, 2], [0, 1, 3, 2], [0, 1, 3, 2]]}]>
```
jtuyls, line 4: Why did this change for phoenix?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc
Abhishek-Varma: Oops. My bad. Did it now.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lowering_strategy_objectfifo_npu1.mlir
Abhishek-Varma: Correct. Taking a look.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_ukernel.mlir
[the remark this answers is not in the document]
newling: Is it strix specific? Is it testing functionality added in this PR?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc
```diff
@@ -629,6 +629,135 @@ struct AMDAIEDeviceModel getDeviceModel(AMDAIEDevice device) {
...
+  }
+
+  auto nBitsLhs = elTypeLhs.getIntOrFloatBitWidth();
+  auto nBitsRhs = elTypeRhs.getIntOrFloatBitWidth();
+  auto nBitsAcc = elTypeAcc.getIntOrFloatBitWidth();
+
+  if (allFloatingPoint) {
+    if (nBitsLhs == 16 && nBitsRhs == 16 && nBitsAcc == 32) {
+      if (device == AMDAIEDevice::npu4) {
+        // Strix nup4 intrinsics.
```
newling, line 740:
```suggestion
        // Strix npu4 intrinsics.
```

Isn't `strip` npu2+ ?  (@jtuyls)

jtuyls: `npu2` might have been the 4 column strix version, not 100% sure

newling: Ok, that's pretty confusing.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lowering_strategy_objectfifo_npu1.mlir
Abhishek-Varma: So this turned out to be an issue with `CHECK-LITERAL`.

I even did a `[hello, 8, 8]` and it still passed. 🫨 

Seems like this bug just carried itself since so long! Have made the changes.

This therefore means that other lit tests `lowering_strategy_*` would also need their `CHECK` statements fixed. I don't want to include them as part of this PR though.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_ukernel.mlir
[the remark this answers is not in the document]
Abhishek-Varma: Yes, this is Strix specific and basically adds tests in other passes of the pipeline which would transitively ingest the output by the main functionality (`KernelDispatch` related) that's added in this PR.

I've even done a similar check for the e2e lit test.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc
jtuyls: Yes..

jtuyls (approved):
LGTM

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lowering_strategy_objectfifo_npu1.mlir
jtuyls: Ok, thanks. We could fix that in another PR.

newling (approved):
LGTM

PR #876, opened 2024-11-06
[Testing] Make it possible to run individual tests 

newling:
After this PR it is possible to run individual tests. 

```
./run.py --help 
```

Prints
```
...
  --tests TESTS         A comma-separated list of test names or sets to run. Available test sets: BatchMatmul, Convolution, ConvolutionNHWCQ, Matmul, MatmulFullBias, MatmulThinBias, MatmulTruncf, MultipleDispatches, UKernel, VanillaMatmul, All. Available individual tests: batch_matmul_1_128_128_256_bf16_f32, batch_matmul_1_128_128_256_i32_i32,
                        batch_matmul_2_64_64_64_bf16_f32, batch_matmul_2_64_64_64_i32_i32, conv_2d_nhwc_hwcf_2_14_bf16_f32, conv_2d_nhwc_hwcf_2_14_i32_i32, conv_2d_nhwc_hwcf_2_14_i8_i32, convolution_nhwc_q, depthwise_conv_2d_nhwc_hwc_1_14_i32_i32, matmul_f32_8_4_8, matmul_f32_8_8_4, matmul_full_bias_128_128_256_i32_i32,
                        matmul_thin_bias_1024_1024_512_bf16_f32_ukernel0, matmul_thin_bias_1024_1024_512_bf16_f32_ukernel1, matmul_truncf_128_256_bf16_f32, matmul_truncf_8_8_bf16_f32, two_matmul_switching, vanilla_matmul_32_32_64_bf16_f32.
...
```

So you can do `./run.py --tests=batch_matmul_1_128_128_256_bf16_f32 ... ` for example, to run just the one test. But you can also do `./run.py --tests=Matmul ...` to run all tests with matmuls. I think this is a better design for debugging.

### DIFF
--- a/build_tools/ci/cpu_comparison/matmul_template/matmul_generator.py
+++ b/build_tools/ci/cpu_comparison/matmul_template/matmul_generator.py
@@ -3,7 +3,7 @@ import re
 import os
 
 
-def generate_matmul_test(output_fn, input_fn, m, n, k, lhs_rhs_type, acc_type, b=0):
+def generate_matmul_test(output_fn, input_fn, m, n, k, lhs_rhs_type, output_type, b=0):
     """
     Generate mlir file (output_fn) from the template file (input_fn).
     """
@@ -13,12 +13,12 @@ def generate_matmul_test(output_fn, input_fn, m, n, k, lhs_rhs_type, acc_type, b
     replace["N"] = n
     replace["K"] = k
     replace["TYPE1"] = lhs_rhs_type
-    replace["TYPE2"] = acc_type
+    replace["TYPE2"] = output_type
 
     replace["B"] = b  # This is only used for batch matmul
-    accl_is_int = acc_type[0] == "i"
-    replace["ZERO"] = 0 if accl_is_int else 0.0
-    replace["ADD"] = "arith.addi" if accl_is_int else "arith.addf"
+    output_is_int = output_type[0] == "i"
+    replace["ZERO"] = 0 if output_is_int else 0.0
+    replace["ADD"] = "arith.addi" if output_is_int else "arith.addf"
 
     key_map = map(lambda s: "${" + s + "}", replace.keys())
     key_map_escaped = map(re.escape, key_map)
--- a/build_tools/ci/cpu_comparison/matmul_template/matmul_generator.py
+++ b/build_tools/ci/cpu_comparison/matmul_template/matmul_generator.py
@@ -3,7 +3,7 @@ import re
 import os
 
 
-def generate_matmul_test(output_fn, input_fn, m, n, k, lhs_rhs_type, output_type, b=0):
+def generate_matmul_test(output_fn, input_fn, m, n, k, lhs_rhs_type, acc_type, b=0):
     """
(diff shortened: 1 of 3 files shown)


build_tools/ci/cpu_comparison/matmul_template/matmul_generator.py
```diff
@@ -13,12 +13,12 @@ def generate_matmul_test(output_fn, input_fn, m, n, k, lhs_rhs_type, acc_type, b
     replace["N"] = n
     replace["K"] = k
     replace["TYPE1"] = lhs_rhs_type
-    replace["TYPE2"] = acc_type
+    replace["TYPE2"] = output_type
```
Abhishek-Varma, line 16: Here and elsewhere, I'd choose to keep the distinction as `acc_type` because `output_type` (as the name suggests) doesn't necessarily has to be different than `lhs_rhs_type` (eg: bf16 -> bf16 Matmul + Truncf case) - in which we get to have a higher bit-width accumulator type but the output type is same as LHS/RHS.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -11,38 +11,287 @@
...
+
+
+class MatmulThinBias(BaseMatmul):
+    """
+    A test of the form matmul(A,B) + C where A:MxK, B:KxN, C:N
+    """
+
+    def __init__(self, M, N, K, input_type, output_type, use_ukernel):
+        super().__init__(M, N, K, input_type, output_type)
+        self.name = f"matmul_thin_bias_{M}_{N}_{K}_{input_type}_{output_type}_ukernel{int(use_ukernel)}"
```
Abhishek-Varma, line 172: Nit:
```suggestion
        self.name = f"matmul_thin_bias_{M}_{N}_{K}_{input_type}_{output_type}{'_ukernel' if use_ukernel else ''}"
```

Abhishek-Varma (changes requested):
Thanks for this. Few change requests.

Also, for my understanding : Why do we still need to maintain `run_matmul_test.sh` ? I understand it has "repeat runs", but that shouldn't be the only motivating delta to keep it, right?

jtuyls:
> Also, for my understanding : Why do we still need to maintain `run_matmul_test.sh` ? I understand it has "repeat runs", but that shouldn't be the only motivating delta to keep it, right?

'repeat runs' is a very important 'delta'. When we're removing `run_matmul_test.sh`, we should keep at least the same coverage.

Abhishek-Varma:
> > Also, for my understanding : Why do we still need to maintain `run_matmul_test.sh` ? I understand it has "repeat runs", but that shouldn't be the only motivating delta to keep it, right?
> 
> 'repeat runs' is a very important 'delta'. When we're removing `run_matmul_test.sh`, we should keep at least the same coverage.

Absolutely! My main intention was to understand why maintain a separate file just for the repeat runs and not include that delta within `run.py` itself.

newling:
`run.py` has a `repeat_runs` option too. I intend to deprecate run_matmul_test.sh soon (move the tests to run.py)

Abhishek-Varma (approved):
LGTM.

PR #879, opened 2024-11-07
[Strix][Ukernel][ObjectFifo] Add Strix ukernel support

Abhishek-Varma:
-- This commit replaces `mm.cc` ukernel file with two separate files now : `mm_npu1.cc` for Phoenix and `mm_npu4.cc` for Strix.
    Therefore during xclbin's generation, based on the target device, the corresponding ukernel file is compiled.
-- It also makes changes to `lower-to-ukernels` pass to adapt to different target device.


Abhishek-Varma:
I have tried this on 64x64x64 bf16->f32 = it works with correct results - here is the [e2e IR](https://gist.github.com/Abhishek-Varma/54716d0974ca7707a5061fcd7311dc95)

I have tried this on 128x128x256 bf16 -> f32 = there are numerical mismatches in some places whilst most elements are correct - here is the [e2e IR](https://gist.github.com/Abhishek-Varma/209920d573393e08e5c753fe6eb37226) (Same case for 1024x1024x1024)
I speculated that this might be because of bf16 -> bfp16 but then this should've been the case for all values, but it's happening for only some of them - perhaps some tiling issue?!

NOTE: I have added the e2e test of 64x64x64 in `run_matmul_test.sh` for now for ease of adding the tests - it can be ported to `run.py` later.

build_tools/ci/run_matmul_test.sh
```diff
@@ -782,6 +782,21 @@ if [ -d "$VITIS" ]; then
       --num_repeat_runs "10" \
       --use_ukernel "1"
 
+# Strix ukernel chess test
+  run_matmul_test \
+      --name_prefix "chess_bf16_f32_npu4_ukernel" \
+     --target_device "npu4" \
```
jtuyls, line 788: This is not going to work, because the machine in CI is not strix. I can try adding it today, but it might be a bit cumbersome to support both strix and phoenix within the same bash script as some tests will only work on phoenix and others only on strix. The python tests should be easier for that.

newling:
For strix, wouldn't a 'simpler' matmul to test be i8->i32? Because that complexities around bf16 -> blf16 -> bf16 conversion.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -550,20 +562,23 @@ static LogicalResult generateCoreElfFiles(
...
             /*extraArgs*/ std::vector<std::string>{},
             /*workDir=*/tempDir,
             /*vitisDir=*/*maybeVitisDir,
             /*npuVersion*/ npuVersion, verbose);
         if (failed(mmObjectFilePath)) return failure();
       } else {
-        mmObjectFilePath = cwd / "mm.o";
+        mmObjectFilePath = cwd / ukernelObjectName;
       }
+      llvm::outs() << "Fetched compiled ukernel = " << ukernelObjectName
```
newling, line 580: Is this a WIP?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/mm_npu4.cc
```diff
@@ -0,0 +1,193 @@
...
+
+      for (unsigned j = 0; j < colB; j += 2)
+        chess_prepare_for_pipelining chess_loop_range(4, ) {
+          const bfloat16 *__restrict pA1 = pA + offsetA + (z)*size_A;
+          const bfloat16 *__restrict pA2 = pA + offsetA + ((z + 1)) * size_A;
+          const bfloat16 *__restrict pB1 = pB + offsetB + (j)*colA*size_B;
+          const bfloat16 *__restrict pB2 = pB + offsetB + ((j + 1))*colA * size_B;
+
+          aie::accum<accfloat, size_A>  accumA0 = load_v64<bfloat16>(pA1);
+          aie::bfp_vector<bfp16ebs8, size_A> A0 = accumA0.template to_vector<bfp16ebs8>();
```
newling, line 82: I don't think we'll ever want to do this in 'production' code, we'll be moving bfp16 in from L2 not casting to it on the core. For this reason, I wonder if it would be better use of time (and possibly simpler) to work on i8->i32 ukernels.

jtuyls: It's the best way to support bfloat16 on strix, even though it's exactly the same accuracy. I think this is quite useful for now until bfp16 will be supported, which can take a while.

jtuyls: But we should add an i8 kernel as well.

newling:
@Abhishek-Varma  I can't see anything obviously wrong with the ukernel which would cause numerical error for 128x128 (but pass for 64x64), so I think trying it through peano might be a good idea.

jtuyls:
@Abhishek-Varma For me we can merge this for the shape that works after rebase + fixing the print statements

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToUKernels.cpp
```diff
@@ -80,6 +80,28 @@ static FnNameAndDefAttrs getFnNameAndDefAttrs(RewriterBase &rewriter,
 /// ================== SAME UTILITIES AS IREE LLVMCPU ====================
 /// ======================================================================
 
+/// TODO(avarma): See if we already have a common utility and work accordingly.
```
jtuyls, line 83: Check inside `AMDAIEUtils` and either use/update there or move this there.

newling:
> @Abhishek-Varma For me we can merge this for the shape that works after rebase + fixing the print statements

Fine by me too. Please add a comment where the 64x64x64 test is created in run.py that failure is expected for the 128x128 case, and that we don't yet understand why.

jtuyls (approved):
LGTM

build_tools/ci/cpu_comparison/run.py
```diff
@@ -961,6 +961,7 @@ def __init__(self):
             )
         )
         self.register(VanillaMatmul(32, 32, 64, "bf16", "f32", use_ukernel=False))
+        # TODO: Failure is expected for the 128x128 case we don't yet understand why.
```
newling, line 964: lol thanks

PR #880, opened 2024-11-07
[CI][Test] Add `target_device` flag

jtuyls:
To start enabling strix, add an explicit flag to pass the target device to the test config.


Abhishek-Varma (approved):
LGTM.

PR #881, opened 2024-11-07
[CI][Test] Add Linux Strix job

jtuyls:


.github/workflows/ci-linux.yml
```diff
@@ -202,3 +202,66 @@ jobs:
...
+            test_aie_vs_cpu \
+            $PWD/iree-install \
+            /opt/xilinx/Vitis/2024.2 \
+            --vitis-dir /opt/xilinx/Vitis/2024.2 \
+            --target_device "npu4" \
+            --use_chess "1" \
+            --reset-npu-between-runs \
+            --xrt_lite_n_core_rows=$XRT_LITE_N_CORE_ROWS \
+            --xrt_lite_n_core_cols=$XRT_LITE_N_CORE_COLS \
+            --tests "vanilla_matmul_32_32_32_i32_i32" \
```
jtuyls, line 266: Just trying to get in a single initial test here, but this should be improved by for example annotating tests in `run.py` with the target devices they should be run on as tests might only work on some target devices.

jtuyls: Also, the strix tests are quite slow because of chess, so we should limit the number of tests we enable.

newling:
This is excellent!

PR #885, opened 2024-11-07
[Testing] Move some tests from run_matmul_test.sh to run.py

newling:
I've tested that the logic for approximate testing is correct (and like IREE's) but I should probably add a unit test. Tricky though, need to inject a failure. If people believe strongly we should/will just never test matmuls with `M*N*K >1e11` I can simplify this and scrap the sampling logic.

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -1280,8 +1280,6 @@ if __name__ == "__main__":
     parser.add_argument("--vitis-dir", type=abs_path)
     parser.add_argument("--xrt_lite_n_core_rows", type=int)
     parser.add_argument("--xrt_lite_n_core_cols", type=int)
-
-    # TODO: this must be documented whth a help string.
     parser.add_argument("--target_device", type=str, required=True)
 
     # TODO(newling) make bool options boolean, not integer (tried but had issues)
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -21,19 +21,9 @@ from input_generator import (
     load_input,
     get_output_type,
     np_from_binfile,
+    f32_to_bf16,
+    bf16_to_f32,
 )
-
-
-def run_conv_test(config, filename, n_repeats):
-    aie_vs_llvm_cpu(
-        config,
-        filename,
-        tile_pipeline="conv-decompose",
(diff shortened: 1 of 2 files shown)


newling:
Damn I just another incomplete effort to unify the testing so I'm not sure what the best way to clean things up is now 

 https://github.com/nod-ai/iree-amd-aie/pull/795/files#diff-86fa5267337e1cd5351558b17c572b525fa818ccc093cfe8dc2cd23d61ea7ce6

newling:
I think I/we should try the following approach (slow burn, expect many delays)

1) move and eliminate run_matmul_test.sh by porting the tests to run.py
2) move and eliminate run.py by porting the tests to https://github.com/nod-ai/iree-amd-aie/pull/795

Why not move all directly to https://github.com/nod-ai/iree-amd-aie/pull/795 ? Because steps 1 and 2 are each smaller jumps and I think the metric to use for difficulty here is more `l_inf` than `l_2`

jtuyls:
@newling Is this PR meant to be reviewed or is the port of all tests a WIP? I am not sure based on your latest comment.

newling:
> @newling Is this PR meant to be reviewed or is the port of all tests a WIP? I am not sure based on your latest comment.

I don't mind, if you have an opinion on the high-level plan feel free to post it here. So the question to answer is: Will moving the tests in `run_matmul_test.sh` to `run.py` improve the situation, and is it low/medium/high priority? I think "Yes" and "low". 

Also, I'm going to remove the complexity of approximate testing in this PR. Until we actually run very large matmuls, it's not helping at all:

```
In [6]: %time np.matmul(np.random.rand(1024, 4096), np.random.rand(4096, 1024))
CPU times: user 417 ms, sys: 839 ms, total: 1.26 s
Wall time: 116 ms
```

which compared to compile time is insignificant.

jtuyls:
> I don't mind, if you have an opinion on the high-level plan feel free to post it here. So the question to answer is: Will moving the tests in run_matmul_test.sh to run.py improve the situation, and is it low/medium/high priority? I think "Yes" and "low".

I agree, we should move the tests to run.py first, and I also think it's low priority. But all new tests should be added to run.py imo, like we have been doing lately already.

> Also, I'm going to remove the complexity of approximate testing in this PR. Until we actually run very large matmuls, it's not helping at all:

Yes, makes sense.

newling:
Closing as rebasing will be more effort than redoing

PR #886, opened 2024-11-07
Uncompiled file: remove or fix cmake? 

newling:
@makslevental please advise 

Background: I/we want to make a semi-standalone cpp file which doesn't use XRT for benchmarking kernels, ie something we can get CI without hacking print statements in https://github.com/nod-ai/iree-amd-aie/blob/9b0582f3cc7e04b9f1747e115e2ee43225ab77f0/runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc#L193 with which we can monitor matmul performance. This file looks like the right direction (or does it?)


newling:
set(ENABLE_XRT_LITE_CTS_TESTS ON CACHE BOOL "") 

Huh, ok.

makslevental:
https://github.com/nod-ai/iree-amd-aie/pull/858

PR #887, opened 2024-11-08
[Test] Update run.py to run only `target_device` specific tests

Abhishek-Varma:
-- This commit updates run.py to create test instances specifying
   the intended target device(s) to run them on. And accordingly run
   only those test instances whose intended target device list matches
   the supplied command-line arg `target_device`.

Keeping this PR logically separate from the Strix ukernel one - helps compartmentalizing.

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -2,6 +2,7 @@
 
 # Copyright 2024 The IREE Authors
 
+from abc import ABC, abstractmethod
 import argparse
 import os
 import platform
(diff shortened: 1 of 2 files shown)


build_tools/ci/cpu_comparison/run.py
```diff
@@ -58,23 +79,25 @@ def run(self, config):
...
-class ConvolutionNHWCQ:
-    def __init__(self):
+class ConvolutionNHWCQ(RunOnTargetDevice):
+    def __init__(self, run_on_target=["npu1_4col"]):
+        super().__init__(run_on_target)
         self.name = "convolution_nhwc_q"
         self.labels = ["Convolution", "ConvolutionNHWCQ"]
 
-    def run(self, config):
+    def _execute(self, config):
```
Abhishek-Varma, line 88: Please note single underscore preceding a member variable/method is an agreed upon convention (a hint) in Python community to indicate that the concerned member variable/method is not supposed to be used outside the class. Therefore we are not supposed to make a direct call to `_execute` for any of the test instances - we are going to be invoking `run` method itself which in turn is going to access this method from within the class.

I'd therefore prefer keeping it this way.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -101,12 +125,12 @@ class MatmulFullBias(BaseMatmul):
     A test of the form matmul(A,B) + C where A:MxK, B:KxN, C:MxN
     """
 
-    def __init__(self, M, N, K, input_type, acc_type):
-        super().__init__(M, N, K, input_type, acc_type)
+    def __init__(self, M, N, K, input_type, acc_type, run_on_target=["npu1_4col"]):
```
Abhishek-Varma, line 128: In order to have minimal changes in the file overall - I have chosen to keep `npu1_4col` as the default target device to run the instance on.

Therefore, while registering the instance test I had to change just one of the instance `Vanilla Matmul 32x32x32` to make it run on `npu4` as well.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -36,8 +37,28 @@ def run_conv_test(config, filename, n_repeats):
     return True
 
 
-class ConvolutionFromTemplate:
-    def __init__(self, params):
+# Base class to be inherited by any new dispatch/compute op template.
```
jtuyls, line 40: Class documentation should be inside the class within """ """.

newling (commented):
Looks like a good approach to me. 

I think the code will be simpler if `target_device` is a member of `config`.

I didn't know about ABC... interesting how python has added things to be more like C++!

Abhishek-Varma:
> Looks like a good approach to me.
> 
> I think the code will be simpler if `target_device` is a member of `config`.
> 
> I didn't know about ABC... interesting how python has added things to be more like C++!

Good catch. `target_device` was already a member of config. So have removed that from the function declaration/invocation now.

PR #889, opened 2024-11-08
Bump IREE to 8e5f218a1d73c39ccdebb4a1d19225d5329cc6f0

yzhang93:
Also bump MLIR-AIR to 8fe05179d83f04cf56965ae55da0f476ec7431ed

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -189,7 +189,7 @@ jobs:
       - name: Python tests
         run: |
           source .venv/bin/activate
-          pytest -v tests \
+          pytest -rv tests \
             --capture=tee-sys \
             --iree-install-dir=$PWD/iree-install \
             --peano-install-dir=$PWD/llvm-aie \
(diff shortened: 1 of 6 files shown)


makslevental:
I can figure out the reason for the segfault on Monday

yzhang93:
> I can figure out the reason for the segfault on Monday

Thanks! It might relate to this change in IREE? https://github.com/iree-org/iree/commit/0077358221e3fd2a52d114115ac9b9d17089fc16

makslevental:
The failing tests are on Linux where the XRT-LITE HAL is used but you're fixing/changing the XRT HAL...

yzhang93:
> The failing tests are on Linux where the XRT-LITE HAL is used but you're fixing/changing the XRT HAL...

There were windows failures as well https://github.com/nod-ai/iree-amd-aie/actions/runs/11749085186/job/32734566694, so the xrt hal changes were a fix for that. Linux failure has not been solved yet.

makslevental:
> , which I thought you'll help take a look.

Yup yup

PR #890, opened 2024-11-11
[xrt-lite] Update cts test to use objectFifo with bf16. 

newling:
Motivation: let's test objectFifo as that is now the pipeline we're supporting long-term.


makslevental:
> Does this not include all the extra configuration stuff which swamps the time to actually run the kernel?

I don't know exactly what's currently happening but I know [there are ways to make sure that overhead isn't counted](https://github.com/iree-org/iree/blob/60b65f30c932eaf967922785253a85a1aa14cebb/tools/iree-benchmark-module-main.cc#L42-L53).

newling:
@makslevental there does appear to be an issue, it is not just a caching mirage. The only change I've made since this PR passed is to change m=n=128 to m=n=512, and now (I expect) it fails.

makslevental:
Okay lemme take a look

runtime/src/iree-amd-aie/driver/xrt-lite/cts/executable_cache_test.mlir
```diff
@@ -8,26 +6,34 @@
...
+      %4 = flow.dispatch.tensor.load %1, offsets = [0, 0], sizes = [4096, 512], strides = [1, 1] : !fdt_rhs -> !t_rhs
+      %5 = tensor.empty() : !t_res
+      %6 = linalg.fill ins(%c0_f32 : f32) outs(%5 : !t_res) -> !t_res
+      %7 = linalg.matmul ins(%3, %4 : !t_lhs, !t_rhs) outs(%6 : !t_res) -> !t_res
+      flow.dispatch.tensor.store %7, %2, offsets = [0, 0], sizes = [512, 512], strides = [1, 1] : !t_res -> !fdt_res
       return
     }
   }
 }
+
```
jtuyls, line 38:
```suggestion
```
Just a single new line?

jtuyls (approved):
LGTM

PR #891, opened 2024-11-11
[Tests] Improve help and failure mode

newling:
Before if you passed an invalid string for target_device, the tests passed without running anything

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -1134,7 +1134,7 @@ if __name__ == "__main__":
         help="Number of AIE core columns of the xrt-lite device to use",
     )
 
-    # Takes from AMDAIEEnums.td
+    # Taken from AMDAIEEnums.td
     current_devices = [
         "xcvc1902",
         "xcve2302",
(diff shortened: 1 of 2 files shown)


build_tools/ci/cpu_comparison/run.py
```diff
@@ -1123,9 +1123,34 @@ def all_tests(
...
+        "xcvc1902",
+        "xcve2302",
+        "xcve2802",
+        "npu1",
+        "npu1_1col",
+        "npu1_2col",
+        "npu1_3col",
+        "npu1_4col",
+        "npu4",
+    ]
```
Abhishek-Varma, line 1148: I'm not sure if we've tested for all these devices.

Would it be okay to just go with `npu1_4col` and `npu4` as the two accepted devices ?

For instance, `run_on_target` has been defaulted to be `npu1_4col` for all tests and just one test adds `npu4` to the same list.

So, there should just be a correspondence between the two, I believe.

jtuyls (approved):
LGTM

build_tools/ci/cpu_comparison/run.py
Abhishek-Varma: Okay okay, that makes sense.

Abhishek-Varma (approved):
LGTM!

PR #892, opened 2024-11-11
Invasively print kernel execution time, percolate to CI output

newling:
{4C086AF3-4CB3-4F77-960C-ABEC14FCC4E6}

I'll remove some spacing, but above is where you can see the kernel execution time in the CI logs. 

Alternatives considered:

1) Suggestion here: https://github.com/nod-ai/iree-amd-aie/pull/892#discussion_r1838636704 very nearly worked, but it resulted in large amounts of trace information being written to terminal, which I was worried would swamp the CI logs. Played around with `IREE_TRACING_CONSOLE_FILE` but not really designed for this. 

2) Tried tracy, but there is no simple text file dumped with tracy (as far as I can tell) so this didn't work 

3) Tried making a standalone test based on https://github.com/nod-ai/iree-amd-aie/blob/main/runtime/src/iree-amd-aie/driver/xrt/cts/matmul_dispatch_test.cc but this got more complicated than I could handle. 

FWIW running the m=n=512 k=4096 matmul in a loop (ie having warm up runs) does not effect the time to run the kernel, at least that is what I've observed. 

If we're brave we can use the execution time as a threshold above which the test fails (thoughts on this?)

We could also create an artifact and upload these numbers to table and make nice plots of our progress (although I'd prefer someone more experienced with this stuff to take over from me)

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -163,11 +163,20 @@ class VanillaMatmul(BaseMatmul):
     """
 
     def __init__(
-        self, M, N, K, input_type, acc_type, use_ukernel, run_on_target=["npu1_4col"]
+        self,
+        M,
+        N,
+        K,
+        input_type,
+        acc_type,
+        use_ukernel,
+        run_on_target=["npu1_4col"],
+        additional_labels=[],
     ):
         super().__init__(M, N, K, input_type, acc_type, run_on_target)
         self.name = f"vanilla_matmul_{M}_{N}_{K}_{input_type}_{acc_type}"
         self.labels.append("VanillaMatmul")
+        self.labels += additional_labels
         self.use_ukernel = use_ukernel
 
     def _execute(self, config):
@@ -983,10 +992,23 @@ class Tests:
             )
         )
         self.register(VanillaMatmul(32, 32, 64, "bf16", "f32", use_ukernel=False))
+
         # TODO: Failure is expected for the 128x128 case we don't yet understand why.
         self.register(
             VanillaMatmul(
                 64, 64, 64, "bf16", "f32", use_ukernel=True, run_on_target=["npu4"]
+            )
+        )
+
+        self.register(
+            VanillaMatmul(
+                512,
+                512,
+                4096,
+                "bf16",
+                "f32",
+                use_ukernel=False,
+                additional_labels=["Performance"],
             )
         )
 
--- a/runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc
+++ b/runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc
@@ -6,15 +6,17 @@
 
 #include "iree-amd-aie/driver/xrt-lite/direct_command_buffer.h"
 
-#include <chrono>
-#include <iostream>
-
 #include "iree-amd-aie/driver/xrt-lite/buffer.h"
 #include "iree-amd-aie/driver/xrt-lite/executable.h"
 #include "iree-amd-aie/driver/xrt-lite/shim/linux/kmq/hwq.h"
 #include "iree-amd-aie/driver/xrt-lite/shim/linux/kmq/kernel.h"
 #include "iree-amd-aie/driver/xrt-lite/util.h"
 #include "iree/hal/utils/resource_set.h"
+
+#ifdef IREE_AMDAIE_TIME_KERNEL
+#include <chrono>
+#include <iostream>
+#endif
 
 struct iree_hal_xrt_lite_direct_command_buffer {
   iree_hal_command_buffer_t base;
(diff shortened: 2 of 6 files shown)


build_tools/ci/cpu_comparison/run.py
```diff
@@ -1123,9 +1123,34 @@ def all_tests(
...
+        "xcvc1902",
+        "xcve2302",
+        "xcve2802",
+        "npu1",
+        "npu1_1col",
+        "npu1_2col",
+        "npu1_3col",
+        "npu1_4col",
+        "npu4",
+    ]
```
makslevental, line 1148: https://stackoverflow.com/a/40324463/9045206

runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc
```diff
@@ -203,6 +207,12 @@ static iree_status_t iree_hal_xrt_lite_direct_command_buffer_dispatch(
     bo->sync(shim_xdna::direction::device2host);
   }
 
+  auto end = std::chrono::high_resolution_clock::now();
+  std::chrono::duration<double> elapsed_seconds = end - start;
+  double time_in_milliseconds = elapsed_seconds.count() * 1000;
+  std::cout << "\n\n\nTotal time was " << time_in_milliseconds  << " [ms]\n";
+  std::cerr << "\n\n\nTotal time was " << time_in_milliseconds  << " [ms]\n";
+
```
makslevental, line 215: i don't know what the intent is here but this seems like not a good idea - you don't want this kind of stuff in the HAL itself...

besides with `-DIREE_ENABLE_RUNTIME_TRACING=ON` you already get timing for any HAL function that uses a trace zone (`IREE_TRACE_ZONE_BEGIN(z0);`)

runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc
```diff
@@ -203,6 +207,12 @@ static iree_status_t iree_hal_xrt_lite_direct_command_buffer_dispatch(
     bo->sync(shim_xdna::direction::device2host);
   }
 
+  auto end = std::chrono::high_resolution_clock::now();
```
jtuyls, line 210: We should be measuring before the bo sync as we also do that on the input side. This way we get the pure AIE kernel time, without any runtime overheads.

runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc
jtuyls: The intent is to measure AIE kernel time, without any runtime syncs/copies/setup overheads. Therefore, this is not the same as the entire function. However, I agree that we shouldn't add this kind of logic here in the HAL. I think an alternative would be to create/extend a separate benchmark module that takes a vmfb and measures this.

makslevental: The way to measure smaller sets of calls, ie within trace zones, is just to add another trace zone

newling: Yeah, I know this is not not a good idea, and that we should be doing this with a standalone benchmark tool (or with additional trace zones). But I'm struggling to build the 'right' solution, still working through the complexity of hal/xrt-lite/zones (and sparsity of documentation + examples)

makslevental: try this

```c++
IREE_TRACE_ZONE_BEGIN_NAMED(z1, "kernel-dispatch);
hwq->issue_command(ebuf.get_exec_buf_bo());
hwq->wait_command(ebuf.get_exec_buf_bo(), 0);
IREE_TRACE_ZONE_END(z1);
```

and then build with

```
-DIREE_ENABLE_RUNTIME_TRACING=ON
-DIREE_TRACING_PROVIDER=console
```

which is (briefly) mentioned [here](https://github.com/nod-ai/iree-amd-aie?tab=readme-ov-file#debugging-hal).

> still working through the complexity of hal/xrt-lite/zones (and sparsity of documentation + examples)

Well I didn't like hoard/hide such docs - there simply aren't any. I had the same complaint and Ben graciously wrote the [`null` HAL](https://github.com/iree-org/iree/tree/main/runtime/src/iree/hal/drivers/null) to help me. So I studied that and the current HAL (and others) and bashed my head and this is what we got.

newling: lol, thanks @makslevental

runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc
```diff
@@ -191,10 +194,22 @@ static iree_status_t iree_hal_xrt_lite_direct_command_buffer_dispatch(
     ebuf.add_arg_bo(*bo);
   }
 
+  auto time0 = std::chrono::high_resolution_clock::now();
+
```
makslevental, line 198: it's ugly but i would guard these too ie
```suggestion
#ifdef IREE_AMDAIE_TIME_KERNEL
  auto time0 = std::chrono::high_resolution_clock::now();
#endif
```

newling: I tried like 100 variants on what to guard, and decided that keeping guards to the minimum was priority

yzhang93:
LGTM. Thanks for adding this feature.

> If we're brave we can use the execution time as a threshold above which the test fails (thoughts on this?)

This is a good idea. Main IREE also has some performance check in ci. We'd just need to be careful about the threshold.

> We could also create an artifact and upload these numbers to table and make nice plots of our progress (although I'd prefer > someone more experienced with this stuff to take over from me)

Yeah, I think this feature can be added later. There are some folks in the group have this experience.

makslevental (approved):

PR #896, opened 2024-11-12
[Vectorization] Remove coalescing from insert-loops-for-vectorization

Abhishek-Varma:
-- This commit removes coalescing from insert-loops-for-vectorization


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertLoopsForVectorization.cpp
```diff
@@ -97,8 +105,8 @@ class AMDAIEInsertLoopsForVectorizationPass
   /// this is optinal we need not return failure if the collapsing cannot take
   /// place. Eg: For <2x3x4> since there aren't any unit dimensions, it'd return
   /// failure, hence we can simply return.
-  static void collapseUnitDims(IRRewriter &rewriter,
-                               linalg::GenericOp &genericOp) {
+  void collapseUnitDims(IRRewriter &rewriter, linalg::GenericOp &genericOp) {
```
newling, line 108: nit: I think the code will be clearer if the check is done at the 2 places the function is called, ie outside the function. i.e. 


```
if (enableCollapsingUnitDims) collapseUnitDims(rewriter, genericOp);
```

build_tools/ci/cpu_comparison/run.py
```diff
@@ -327,6 +327,8 @@ def _execute(self, config):
             atol=0,
             n_repeats=1,
             output_type=get_output_type(self.filename),
+            coalesce_loops=True,
```
newling, line 330: Can you add a comment with these truncf tests indicating why these flags are enabled? Something like, "currently without enabling loop coalescing (and dimension collapsing) we run out of program memory, this is under investigation" (similarly in other PR for outlining)

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
```diff
@@ -155,6 +157,26 @@ struct AMDAIEOptions {
...
+            "Pass insert-loops-for-vectorization may disable/enable loop "
+            "coalescing depending on the pass flag. This flag therefore "
+            "ensures that we have a switch on that. It is intended for "
+            "development purposes only."));
+
+    binder.opt<bool>(
+        "iree-amdaie-enable-collapsing-unit-dims", enableCollapsingUnitDims,
+        llvm::cl::cat(category),
+        llvm::cl::desc(
+            "Pass insert-loops-for-vectorization may disable/enable collapsing "
```
newling, line 173: nit: string over more lines than needed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_loops_for_vectorization.mlir
```diff
@@ -312,27 +350,54 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
 // CHECK-SAME:  (%[[ARG0:.*]]: memref<1x1x4x8x4x8xbf16>,
 // CHECK-SAME:   %[[ARG1:.*]]: memref<1x1x8x4x8x4xbf16>
 // CHECK-SAME:   %[[ARG2:.*]]: memref<1x1x8x8x4x4xf32>)
+// COLLAPSE-LABEL: @matmul_bufferized
```
newling, line 353: Are these new matmul_bufferized lines testing anything new? If not, consider removing (to reduce work if this test ever needs updating).

newling (commented):
Looks good. As with the outlining PR, just a few small comments

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_loops_for_vectorization.mlir
newling: Sorry I wasn't clear: I was actually referring to these specific lines. I understand the need for COLLAPSE and COALESCE and how they work, I'm just concerned you're testing the same thing multiple times.

newling: But it's really a nitty comment, so please don't sink too much time into it!

newling (approved):
Could consider removing some lines from the lit test if there's duplication of testing, but otherwise LGTM. I also considered suggesting using more descriptive names, like `collapse_unit_dims_in_insert_loops_for_vectorization_pass` because there are lots of places where where we might collapse unit dims, but then thought we can change the names if ever there is clear ambiguity.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
```diff
@@ -155,6 +157,24 @@ struct AMDAIEOptions {
             "these vectorization passes. It is intended for development "
             "purposes only."));
 
+    binder.opt<bool>(
+        "iree-amdaie-enable-coalescing-loops", enableCoalescingLoops,
+        llvm::cl::cat(category),
+        llvm::cl::desc(
+            "Pass insert-loops-for-vectorization may disable/enable loop "
+            "coalescing depending on the pass flag. This flag therefore "
+            "ensures that we have a switch on that. It is intended for "
```
yzhang93, line 166: Nit: this can be more concise. "depending on this pass flag". Remove "This flag therefore ensures that we have a switch on that." Same for the next flag.

Abhishek-Varma: Sure, let me address this nit as part of the function-outlining PR to bank on the current CI progress for this PR.

PR #898, opened 2024-11-13
[AMDAIENoneAccessToTemporaryBuffer] Insert new alloc at block start to avoid SSA error

newling:
Consider the test added in this PR: it is not enough to insert the replacement alloc before any of the access ops, you will hit `operand #0 does not dominate this use` or end up with an alloc in the for loop. So this PR changes where the alloc is inserted : it is now inserted at the very start of the core's block. 

All other changes are here are tiny and non-functional (propagate op location for better diagnostics, don't do map look up twice when only once is needed, use assert instead of error if it shouldn't be reachable)


Abhishek-Varma (approved):
LGTM!

PR #899, opened 2024-11-13
[POC] Unpeel

newling:
This is actually pretty simple to do, and it currently compiles fine (if I don't try and unpeel the epilogue). But there is numerical error, which I haven't managed to understand yet. 

If you're looking at this I recommend taking a look at the .mlir file -- it should be clear what I'm trying to do in this pass, which matmul I'm trying to fuse into the loop. 

This is a WIP, something more rigorous is definitely needed!

Before unpeeling: 
[before_unpeel.txt](https://github.com/user-attachments/files/17734188/before_unpeel.txt)

After unpeeling:  
[after_unpeel.txt](https://github.com/user-attachments/files/17734147/after_unpeel.txt)

IR (module scope) before all with unpeeling:
[with_unpeel_before_all.txt](https://github.com/user-attachments/files/17757030/with_unpeel_before_all.txt)

All that changes is that the very first matmul, which is before the loop, goes into the loop (and the loop count goes up).

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/unpeel.mlir
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/unpeel.mlir
@@ -1,0 +1,182 @@
+// RUN: iree-opt --split-input-file --pass-pipeline="builtin.module(iree-amdaie-unpeel,canonicalize)" %s | FileCheck %s
+
+// -----// IR Dump Before AMDAIEUnpeel (iree-amdaie-unpeel) ('func.func' operation: @matmul_dispatch_0_matmul_8x8x128_bf16xbf16xf32) //----- //
+#config = #iree_codegen.lowering_config<tile_sizes = [[8, 8], [0, 0, 1], [1, 1, 0, 0, 0, 0]]>
+#executable_target_amdaie_pdi_fb = #hal.executable.target<"amd-aie", "amdaie-pdi-fb", {target_device = "npu1_4col", ukernels = "none"}>
+#map = affine_map<(d0, d1, d2, d3, d4, d5, d6, d7, d8) -> (d0, d2, d5, d3, d6, d8)>
+#map1 = affine_map<(d0, d1, d2, d3, d4, d5, d6, d7, d8) -> (d2, d1, d4, d5, d8, d7)>
+#map2 = affine_map<(d0, d1, d2, d3, d4, d5, d6, d7, d8) -> (d0, d1, d4, d3, d6, d7)>
+#map3 = affine_map<(d0) -> (d0 * 32)>
+#packingConfig = #amdaie.packing_config<packing_config = [{packedSizes = [4, 4, 32], transposePackIndices = [1], unpackEmpty = [false], innerPerm = [[1, 0]], outerPerm = [[0, 1]]}, {packedSizes = [0, 0, 0, 4, 4, 8], transposePackIndices = [0, 1, 2], unpackEmpty = [false, false, true], innerPerm = [[0, 1], [1, 0], [0, 1]], outerPerm = [[0, 1, 3, 2], [0, 1, 3, 2], [0, 1, 3, 2]]}]>
+#pipeline_layout = #hal.pipeline.layout<bindings = [#hal.pipeline.binding<storage_buffer, "ReadOnly|Indirect">, #hal.pipeline.binding<storage_buffer, "ReadOnly|Indirect">, #hal.pipeline.binding<storage_buffer, Indirect>], flags = Indirect>
+#translation = #iree_codegen.translation_info<Custom>
+#device_target_xrt_lite = #hal.device.target<"xrt-lite", [#executable_target_amdaie_pdi_fb]> : !hal.device
+module attributes {stream.affinity.default = #hal.device.affinity<@__device_0>} {
+  util.global private @__device_0 = #device_target_xrt_lite
+  hal.executable private @matmul_dispatch_0 {
+    hal.executable.variant public @amdaie_pdi_fb target(#executable_target_amdaie_pdi_fb) {
+      hal.executable.export public @matmul_dispatch_0_matmul_8x8x128_bf16xbf16xf32 ordinal(0) layout(#pipeline_layout) {
+      ^bb0(%arg0: !hal.device):
+        %x, %y, %z = flow.dispatch.workgroup_count_from_slice
+        hal.return %x, %y, %z : index, index, index
+      }
+      builtin.module {
+        func.func @matmul_dispatch_0_matmul_8x8x128_bf16xbf16xf32() attributes {translation_info = #translation} {
+          %c3 = arith.constant 3 : index
+          %c0 = arith.constant 0 : index
+          %cst = arith.constant 0.000000e+00 : f32
+          %c1 = arith.constant 1 : index
+          %alloc = memref.alloc() : memref<1x1x1x4x8x4xbf16, 2 : i32>
+          %alloc_0 = memref.alloc() : memref<1x1x4x1x4x8xbf16, 2 : i32>
+          %alloc_1 = memref.alloc() : memref<1x2x32x4xbf16, 1 : i32>
+          %alloc_2 = memref.alloc() : memref<2x1x4x32xbf16, 1 : i32>
+          %alloc_3 = memref.alloc() : memref<2x2x1x1x4x4xf32, 2 : i32>
+          %alloc_4 = memref.alloc() : memref<2x2x4x4xf32, 1 : i32>
+          %0 = hal.interface.binding.subspan layout(#pipeline_layout) binding(0) alignment(64) offset(%c0) flags("ReadOnly|Indirect") : !flow.dispatch.tensor<readonly:tensor<8x128xbf16>>
+          %1 = hal.interface.binding.subspan layout(#pipeline_layout) binding(1) alignment(64) offset(%c0) flags("ReadOnly|Indirect") : !flow.dispatch.tensor<readonly:tensor<128x8xbf16>>
+          %2 = hal.interface.binding.subspan layout(#pipeline_layout) binding(2) alignment(64) offset(%c0) flags(Indirect) : !flow.dispatch.tensor<writeonly:tensor<8x8xf32>>
+          %3 = flow.dispatch.tensor.load %0, offsets = [0, 0], sizes = [8, 128], strides = [1, 1] : !flow.dispatch.tensor<readonly:tensor<8x128xbf16>> -> tensor<8x128xbf16>
+          %4 = flow.dispatch.tensor.load %1, offsets = [0, 0], sizes = [128, 8], strides = [1, 1] : !flow.dispatch.tensor<readonly:tensor<128x8xbf16>> -> tensor<128x8xbf16>
+          %5 = tensor.empty() : tensor<8x8xf32>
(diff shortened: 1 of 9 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/unpeel.mlir
```diff
@@ -0,0 +1,182 @@
...
+  }
+}
+
+// Verify that the fill is written the peeled matmul is reabsoted, but the fill remains.
+// CHECK:         %[[FILL:.*]] = linalg.fill
+// CHECK:         scf.forall.in_parallel {
+// CHECK:           tensor.parallel_insert_slice %[[FILL]]
+// CHECK:         }
+// CHECK:         scf.for
+// CHECK-SAME:    %c0 to %c3 step %c1
```
yzhang93, line 179: This is not clear to me what the output is supposed to look like. Could you expand the checks or attach the IR after this pass?

newling: Added links to before and after in the comment

newling: >  Thanks for sharing the IR! So far the IR seems good to me. It's good to further check the IR before LowerToAIE to see if anything suspicious. 

I studied the IR for a few hours yesterday, but didn't see anything suspicious. I'd appreciate it if you or @jtuyls could take a look at some point, I'm still not fully up-to-speed with the semantics for all the dma/object-fifo/connection ops, so I could be missing something obvious. But yeah I agree -- the IR change done in this PR's unpeeling loop looks fine/harmless. 

> What kind of numerical issues you saw? All zeros / all results are non-zeros but wrong?

Yup, all zeros. I tried a few things to diagnose further: I tried filling with `1` instead of `0`,  because I thought maybe the linalg.fill was somehow happening after the matmul, but that wasn't the case -- filling with 1 instead of 0 still resulted in the output being all 0. Which suggests no data is being written back to L3, maybe? I also tried removing the prologue entirely, but still get all 0 too. 

> About the motivation, I didn't quite get why this would help reduce program memory usage?

It's basically the same reason as outlining reduces PM. The outlining that @Abhishek-Varma  implemented creates a function which does the full 32x32 matmul. Take a look at the IR generated with function outlining and you'll see something like:

```
...
%24 = amdaie.core(%tile, in : [%21, %22], out : []) {
        func.call @generic_matmul_0_outlined(%alloc_0, %alloc, %alloc_3) : 
           (memref<1x1x4x8x4x8xbf16, 2 : i32>, 
            memref<1x1x8x4x8x4xbf16, 2 : i32>, 
            memref<1x1x8x8x4x4xf32, 2 : i32>) -> ()
        amdaie.end
}
...
```

i.e. instead of the ` linalg.generic` there is a `func.call`. Currently, you will see 3 such function calls in the IR: 1 for the prologue, 1 for the main body (call inside the sc.for loop) and 1 for the epilogue. 

This results in PM reduction, because PM will only contains the instructions for doing 1 32x32 matmul. These are alot of instructions saved, because llvm/peano unrolls the 32x32 matmul into lots of 4x4 matmuls. Without function outlining, these unrolled instructions will be duplicated 3 times in PM: once for the prologue, once for main, and once for epilogue. 

With the approach in this PR, I'm completely eliminating the matmul from the prologue and epilogue. So that even before Abhishek's function outlining, there is only 1 32x32 matmul in PM. 

Note that the scf.for loop has a trip count which is 2 larger than before with this PR. In theory if peano unrolled this outer scf.for loop, the approach in this PR wouldn't work. But it doesn't unroll this outer scf.for loop -- I've checked this, the number of appearances of `llvm.aie2.bf.mac16.conf` in the optimized code is exactly the same for K=2048 and K=4096 (if you run tests in run.py, the optimized llvm IR after unrolling is in `module_matmul_dispatch_0_amdaie_pdi_fb/input.opt.ll`). 

It would seem crazy to me if peano did unroll this outer scf.for loop, because it contains lock accesses and releases, so I don't think there would be any advantage to peano unrolling it (@jtuyls might have some idea here). 

Finally I should mention what the advantage of this approach over function outlining: there will be no performance hit caused by function calls.

yzhang93: Okay, thanks for the explanation. In regarding of the all zeros, there could be different reasons behind that. Probably check the access patterns and locks in terms of racing condition? Could you please share the dump IR after all passes? I could take a look as well.

newling: >  Could you please share the dump IR after all passes? I could take a look as well.

Thanks! I've uploaded it https://github.com/user-attachments/files/17757030/with_unpeel_before_all.txt

jtuyls (changes requested):
> This is actually pretty simple to do, and it currently compiles fine (if I don't try and unpeel the epilogue). But there is numerical error, which I haven't managed to understand yet.

This is pretty simple because it's hardcoded to a specific case:
- Looking for prolog - main - epilog. What if peeled 2, 3, 4 times etc?
- Assumes the prolog, main and epilog are the same ops, that can be unpeeled. This is not generic as it doesn't proof that the generic op before the main is actually the prolog of the main (accesses could be different from main). The same issue exists for the epilog.

Also, could you argue why we need this? We discussed this before and we said we didn't need it as it would also not really solve the PM issue, just make it a slightly smaller problem and there are other simpler and more robust ways to tackle the PM issues.

newling:
> > This is actually pretty simple to do, and it currently compiles fine (if I don't try and unpeel the epilogue). But there is numerical error, which I haven't managed to understand yet.
> 
> This is pretty simple because it's hardcoded to a specific case:
> 
> * Looking for prolog - main - epilog. What if peeled 2, 3, 4 times etc?

Easily extends to that case. 

> * Assumes the prolog, main and epilog are the same ops, that can be unpeeled. This is not generic as it doesn't proof that the generic op before the main is actually the prolog of the main (accesses could be different from main). The same issue exists for the epilog.

I am aware of these issues, there is no need to shoot this down while it is still a proof-of-concept. There are tools in MLIR to check for block equivalence (which function outlining of Abhishek's already uses). FWIW the current function outlining is isn't general, hardcoded for matmul. 


> Also, could you argue why we need this? We discussed this before and we said we didn't need it as it would also not really solve the PM issue, just make it a slightly smaller problem and there are other simpler and more robust ways to tackle the PM issues.


See https://github.com/nod-ai/iree-amd-aie/pull/899#discussion_r1841016406

jtuyls:
> > * Assumes the prolog, main and epilog are the same ops, that can be unpeeled. This is not generic as it doesn't proof that the generic op before the main is actually the prolog of the main (accesses could be different from main). The same issue exists for the epilog.
> 
> I am aware of these issues, there is no need to shoot this down while it is still a proof-of-concept. There are tools in MLIR to check for block equivalence (which function outlining of Abhishek's already uses). FWIW the current function outlining is isn't general, hardcoded for matmul.

I think this is a conceptual issue so bringing up before we spend too much time on this. If the purpose of POC is to measure the PM reduction potential, then maybe, but that's not what's described in this PR. If it's meant to actually be merged as a pass, it needs to be really generic or it will lead to all kinds of issues. I think making this fully generic is not worth the effort, certainly not at this time, because I don't see much benefit, see below:

> > Also, could you argue why we need this? We discussed this before and we said we didn't need it as it would also not really solve the PM issue, just make it a slightly smaller problem and there are other simpler and more robust ways to tackle the PM issues.
> 
> See [#899 (comment)](https://github.com/nod-ai/iree-amd-aie/pull/899#discussion_r1841016406)

I am assuming you are referring to:

> Finally I should mention what the advantage of this approach over function outlining: there will be no performance hit caused by function calls.

I think this is a limited benefit at this point and I don't think the function call overhead will soon be the bottleneck. One data point on that is that the ukernel approach (which has function calls) performs significantly better than the vectorization approach w/o function calls.

newling:
>  If the purpose of POC is to measure the PM reduction potential

It's a POC in the sense that people said in the initial meeting where we discussed this that it couldn't be done, because of something to do with not being able to fuse the fill. But it can be done, that's what this POC is a "proof" of. 

> If it's meant to actually be merged as a pass, it needs to be really generic or it will lead to all kinds of issues.

I'm sorry but you are applying a double standard here -- many of the passes in the tiling pipeline are designed only to work together for a peeled matmul. But that's secondary  -- I've explicitly stated in the summary that this will be generalized once people get the basic idea of what the pass does. 

> I think making this fully generic is not worth the effort, certainly not at this time, because I don't see much benefit, see below:

The benefit is the same as function outlining (do you agree?) So if by "not much benefit" you mean over and above function outlining, then yes you are right, definitely not additive benefits. The motivation for trying this again was the apparent overhead of function outlining

> I think this is a limited benefit at this point and I don't think the function call overhead will soon be the bottleneck. One data point on that is that the ukernel approach (which has function calls) performs significantly better than the vectorization approach w/o function calls.

This is a good point. I wonder why the function call overhead of ukernel is lower than with our function outlining.  All I can think -- do we definitely know that we've used ukernel with multiple call sights, and seen good performance?

jtuyls:
> It's a POC in the sense that people said in the initial meeting where we discussed this that it couldn't be done, because of something to do with not being able to fuse the fill. But it can be done, that's what this POC is a "proof" of.

People in the initial meeting were thinking of doing this in general and that it is complicated/hard (no one said impossible) and not worth the effort. This PR doesn't proof that that is easy because it makes very big assumptions so that it becomes a simple problem.

> I'm sorry but you are applying a double standard here -- many of the passes in the tiling pipeline are designed only to work together for a peeled matmul. But that's secondary -- I've explicitly stated in the summary that this will be generalized once people get the basic idea of what the pass does.

Well, please list those passes that can only work on a peeled matmul and will break otherwise. 

This is not the same, but there are passes that will only work in limited circumstances, however, those should check for and only be activated in exactly those circumstances. The logic in this PR is not on the contrary and will have incorrect output for IR that is not expected, that's a very different issue, not a double standard.

> The benefit is the same as function outlining (do you agree?)

For peeled matmul, more or less (unpeeling probably has slightly less code, although minimal).
But not necessarily in general as function outlining can have more benefit for example `peeled matmul + relu + matmul`.

> So if by "not much benefit" you mean over and above function outlining, then yes you are right, definitely not additive benefits. The motivation for trying this again was the apparent overhead of function outlining

This is what's most important. Since function outlining is much simpler, we shouldn't need to look at unpeeling anytime soon. Again, fine if it's just a POC to measure/try something, but not useful imo to keep looking at.

> This is a good point. I wonder why the function call overhead of ukernel is lower than with our function outlining. All I can think -- do we definitely know that we've used ukernel with multiple call sights, and seen good performance?

There is a table in the doc that compares the ukernel approach with vectorization, showing significant difference. Unless you think the ukernel is inlined after all, which it shouldn't because 1) there is no inline on those functions 2) it's compiled separately through chess, so I don't see how that would work. Also, deriving from prior work, there shouldn't really be that much overhead from function calls and that's really the last of the last mile.

newling:
Ok, still a few points I don't fully agree with but I do totally agree that if function calls from outlining should have negligeable overhead then that's what we should work towards. And so I'll stop debating inconsequential things! 

> There is a table in the doc that compares the ukernel approach with vectorization
Cool, I'll get the ukernel into 'Performance benchmarks' in CI (https://github.com/nod-ai/iree-amd-aie/actions/runs/11864873206/job/33069601465?pr=906)

newling:
Closing for now, as if we can eliminate the overhead of the function call with function outlining then this pass is redundant.

PR #900, opened 2024-11-13
[LinalgFunctionOutlining] Enable/disable function outlining using pass flag

Abhishek-Varma:
-- This commit adds a pass flag `enable-function-outlining` to enable/
   disable function outlining.
-- It also adds a global flag for the same and makes changes to the tests.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/linalg_function_outlining.mlir
```diff
@@ -178,32 +208,3 @@ func.func @linalg_fill_copy(%A: memref<4xf32>, %B: memref<4xf32>) {
...
 }
-
-// -----
-
-func.func @unsupported_linalg_op(%A: memref<4x8xbf16>, %B: memref<8x4xbf16>, %C: memref<4x4xf32>) {
-  %c2 = arith.constant 2 : index
-  %c1 = arith.constant 1 : index
-  %tile = amdaie.tile(%c1, %c2)
-  %1 = amdaie.core(%tile, in : [], out : []) {
-    // expected-error@+1 {{unsupported linalg op for outlining}}
```
Abhishek-Varma, line 189: Couldn't find a way to selectively enable `expected-error` (besides just creating two separate files for the two `RUN`s). Need to ask around.

newling (commented):
Looks good, just a few small comments.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -458,6 +460,9 @@ def generate_aie_vmfb(
 
     if collapse_unit_dims:
         compilation_flags += ["--iree-amdaie-enable-collapsing-unit-dims"]
+    
+    if function_outline:
+        compilation_flags += ["--iree-amdaie-enable-function-outlining"]
```
yzhang93, line 465: It has to be two separate flags?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
```diff
@@ -177,6 +178,13 @@ struct AMDAIEOptions {
             "for "
             "development purposes only."));
 
+    binder.opt<bool>(
+        "iree-amdaie-enable-function-outlining", enableFunctionOutlining,
+        llvm::cl::cat(category),
+        llvm::cl::desc("Flag to use if we want to enable/disable "
```
yzhang93, line 184: Nit: can be more concise "Flag to enable/disbale..."

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
```diff
@@ -91,6 +96,7 @@ class AMDAIELinalgFunctionOutliningPass
 };
 
 void AMDAIELinalgFunctionOutliningPass::runOnOperation() {
+  if (!enableFunctionOutlining) return;
```
yzhang93, line 99: There's no point to use this flag within the pass. This should only be controlled outside in passes.cpp.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -270,6 +270,10 @@ def AMDAIELinalgFunctionOutlining :
     repeated codes.
   }];
   let constructor = "mlir::iree_compiler::AMDAIE::createAMDAIELinalgFunctionOutliningPass()";
+  let options = [
+    Option<"enableFunctionOutlining", "enable-function-outlining", "bool", /*default=*/"false",
+      "Enable/disable linalg function outlining">
+  ];
```
yzhang93, line 276: I don't think you need to add this option. See the reason above.

build_tools/ci/cpu_comparison/run.py
newling: My opinion is that it makes sense to have separate flags, and I don't see any harm in separate flags. @Abhishek-Varma maybe just add a comment with each of the flags that it "if enabled, it might reduce program memory usage"?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
newling: Make sense I think, like the `enableVectorizationPasses` approach.

build_tools/ci/cpu_comparison/run.py
newling: > but you have to be very familiar with the project in order to use all correct flags for compiling.

The defaults in this case are fine. My perspective is that these flags are just for the developer to play with and make progress without forcing them on all pipelines. Eventually, some of them will be expired. Coalescing for example: once the upstream changes are done and integrated (the pass to coalesce and collapse dimensions simultaneously which I think Mahesh mentioned) we can always enable coalescing, and the flag can be removed. 

If we had a single flag what would it be called 
(1)  `--iree-amdaie-optimize-for-program-memory` or 
(2) `--iree-amdaie-enable-outlining-and-dim-collapsing-and-coalescing`` ? 

I suppose we could go with (1) but it seems like we don't understand what these changes do well enough yet to give it such a bold name. 

Option (2) seems strange, why not just have them as separate flags.

LLVM has at 9 flags just to control loop unrolling https://llvm.org/doxygen/LoopUnrollPass_8h_source.html (!!)

yzhang93: I'm totally okay with separate flags for experimenting. Just saying it'd be good to keep less flags for production. @Abhishek-Varma could you add more comments on directions of when this flag `--iree-amdaie-enable-collapsing-unit-dims` is needed?

Abhishek-Varma: I've added it in https://github.com/nod-ai/iree-amd-aie/pull/896.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
Abhishek-Varma: Really nice point! Thanks @yzhang93 !

newling:
@Abhishek-Varma maybe add a test along the lines of https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/disable_vectorization.mlir

newling (changes requested):
Please add a test along the lines of https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/disable_vectorization.mlir

yzhang93:
> Please add a test along the lines of https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/disable_vectorization.mlir

Same request, otherwise LGTM.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/disable_linalg_function_outlining.mlir
```diff
@@ -0,0 +1,39 @@
...
+// pipeline (note that below the pipeline is not specified explicitly with
+// the flag --iree-amdaie-tile-pipeline). We check 3 paths:
+//
+// 1) Explicitly disabling linalg function outlining with
+//             --iree-amdaie-enable-function-outlining=0
+//
+// 2) Explicitly enabling linalg function outlining with
+//             --iree-amdaie-enable-function-outlining=1
+//
+// 3) Not specifying the flag at all, which should use the default value (1).
```
newling, line 11:
```suggestion
// 3) Not specifying the flag at all, which should use the default value (0).
```

newling (approved):
LGTM

PR #903, opened 2024-11-13
Split L2 input and output objectFifos for memTile/shimTile distribution

yzhang93:
This is the first PR needed to enable 4x4 AIE array. The L2 objectFifos are split to distribute on multiple memTiles and shimTiles for more channel usage. The shim/mem tile reassignment will be addressed in a separate PR.

Also note this PR doesn't change/combine the previous pass that splitting the third input (elementwise) for connection reuse with mamtul ops. What unclear to me is how matmul-elementwise path will be like for 4x4 AIE cores. If the existing logic is kept, we'll have to split the elementwise op to 16 objectFifos which will be a big challenge given the number of shimTile channels . I'd rather leave it as a separate thing for now before we have a clear path to move forward. 

@jtuyls I've simplified the functions a bit based on your initial version. Feel free to make modifications/push new commits.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifosForConnectionReuse.cpp
```diff
@@ -34,10 +33,20 @@ void AMDAIESplitLogicalObjFifosForConnectionReusePass::runOnOperation() {
   MLIRContext *context = &getContext();
   IRRewriter rewriter(context);
 
-  SmallVector<AMDAIE::DmaCpyNdOp> l2ToL1DmaOps =
-      fetchDmaCpyNdOpsToSplitOrCombine(moduleOp);
-
-  if (failed(splitLogicalObjectFifos(rewriter, l2ToL1DmaOps, context))) {
+  // Walk through CoreOps gathering 3rd input DmaOps (if applicable) which will
+  // be used to split L2 objectFifos of elementwise input for connection reuse.
+  SmallVector<AMDAIE::DmaCpyNdOp> l2ToL1DmaOps;
```
newling, line 38: Is this path through the code tested in a lit test?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifosForConnectionReuse.cpp
```diff
@@ -6,7 +6,6 @@
 #include "iree-amd-aie/IR/AMDAIEOps.h"
 #include "iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.h"
 #include "iree-amd-aie/Transforms/Passes.h"
-// #include "llvm/Support/Debug.h"
```
newling, line 9: remove

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -23,16 +23,15 @@
 
 namespace mlir::iree_compiler::AMDAIE {
 
-/// Utility to create a new logical objectfifo based on shape defined by
-/// `newSizesOpFoldResultArr`.
+/// Hardcoded the transposed dimensions for now.
+const SmallVector<size_t> transposeDim = {0, 2, 1, 3};
```
newling, line 27:
```suggestion
const static SmallVector<size_t> transposeDim = {0, 2, 1, 3};
```

newling: Comment on what `transposeDim` is?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -305,17 +298,34 @@ static LogicalResult checkWhetherSplitIsPossible(
...
-// 4. Delete old L2->L1, L3->L2 and corresponding AllocOps.
-LogicalResult splitLogicalObjectFifos(
+/// Utility to check if the L3->L2 dma is transposed on the L2 side.
+/// The intuition is if the sizes of L2 and L3 side are the same, then dma has
+/// transposition on L3 side, otherwise the transposition is on L2 side.
+static bool isL2DmaTransposed(AMDAIE::DmaCpyNdOp l3ToL2DmaOp, bool isL2Target) {
+  SmallVector<OpFoldResult> l2DmaSizes;
+  SmallVector<OpFoldResult> l3DmaSizes;
+  if (isL2Target) {
+    l2DmaSizes = l3ToL2DmaOp.getTargetMixedSizes();
```
newling, line 308: function body could maybe be
```
if (isL2Target) return l3ToL2DmaOp.getTargetMixedSizes().size() != l3ToL2DmaOp.getSourceMixedSizes().size();
return l2ToL3DmaOp.getTargetMixedSizes().size() != l2ToL3DmaOp.getSourceMixedSizes().size();
```

(or something else to avoid unnecessary possible heap allocations of SmallVectors)

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -305,17 +298,34 @@ static LogicalResult checkWhetherSplitIsPossible(
...
+  return l2DmaSizes.size() == l3DmaSizes.size() ? false : true;
+}
+
+/// Given a vector of L2->L1 Dma ops' perform the splitting :-
+/// 1. Check if the splitting can be performed or not. If not possible, bail
+/// out.
+/// 2. For the split dimension inferred set offset = 0 and size as 1 for L2 and
+///    L3.
+/// 3. Now traverse each L2->L1 Dma op and perform the following :-
+///    a) Create a new L2 AllocOp based on the updated size (step 3 above) and
```
newling, line 323:
```suggestion
///    a) Create a new L2 AllocOp based on the updated size (step 2 above) and
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -305,17 +298,34 @@ static LogicalResult checkWhetherSplitIsPossible(
...
+///       create a logicalobjectfifo using the same.
+///    b) Split L3->L2 Dma op.
+///    c) SPlit L2->L1 Dma op.
```
newling, line 326:
```suggestion
///    c) Split L2->L1 Dma op.
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -305,17 +298,34 @@ static LogicalResult checkWhetherSplitIsPossible(
...
+    l3DmaSizes = l3ToL2DmaOp.getSourceMixedSizes();
+  } else {
+    l2DmaSizes = l3ToL2DmaOp.getSourceMixedSizes();
+    l3DmaSizes = l3ToL2DmaOp.getTargetMixedSizes();
+  }
+  return l2DmaSizes.size() == l3DmaSizes.size() ? false : true;
+}
+
+/// Given a vector of L2->L1 Dma ops' perform the splitting :-
+/// 1. Check if the splitting can be performed or not. If not possible, bail
```
newling, line 318:
```suggestion
/// 1. Check if the splitting can be performed. If it can't, bail
```

newling (changes requested):
I think I understand the basic idea of the pass, thanks for the clear comments and code. I'd try to take another look later today to better understand the details.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -507,4 +495,254 @@ LogicalResult splitLogicalObjectFifos(
...
+      llvm::to_vector(op.getMemrefType().getShape());
+  assert(splitDim < memrefShape.size() &&
+         "the dimension to be split on should be smaller than the number of "
+         "dimensions in the objectfifo shape");
+  int64_t splitFactor = memrefShape[splitDim];
+  if (ShapedType::isDynamic(splitFactor)) {
+    return op.emitOpError()
+           << "a dynamic size on the split dimension is not supported";
+  }
+  memrefShape[splitDim] /= splitFactor;
```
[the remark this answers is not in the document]
yzhang93: If you look 5 lines above, splitFactor = memrefShape[splitDim]; So this just make the new memrefShape[splitDim] as 1.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifosForConnectionReuse.cpp
yzhang93: ? This is already removed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos.mlir
```diff
@@ -0,0 +1,181 @@
+// RUN: iree-opt --pass-pipeline="builtin.module(iree-amdaie-split-logical-objectfifos,cse,canonicalize)" --split-input-file --verify-diagnostics %s | FileCheck %s
```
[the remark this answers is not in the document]
yzhang93: Okay, I have split it into multiple tests as suggested.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -629,6 +629,12 @@ def AMDAIESinkIntoCore :
   let constructor = "mlir::iree_compiler::AMDAIE::createAMDAIESinkIntoCorePass()";
 }
 
+def AMDAIESplitLogicalObjFifos :
+  Pass<"iree-amdaie-split-logical-objectfifos", "ModuleOp"> {
+  let summary = "Pass to split L2 buffers to distribute on multiple shimTiles and memTiles.";
```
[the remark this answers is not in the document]
yzhang93: I've added more descriptions of the pass.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -507,4 +491,254 @@ LogicalResult splitLogicalObjectFifos(
...
+        sourceSizes, sourceStrides);
+    rewriter.replaceOp(consumer, newDmaOp);
+  }
+
+  return success();
+}
+
+/// Split DmaCpyNd ops between L2 and L3 memory spaces.
+LogicalResult splitDoublyStridedOp(IRRewriter &rewriter,
+                                   AMDAIE::DmaCpyNdOp op) {
```
newling, line 655: This function will be easier to reuse and understand if the signature is 

```
LogicalResult splitDoublyStridedOp(IRRewriter &rewriter, AMDAIE::DmaCpyNdOp op, uint32_t splitDimension, uint32_t splitSize, bool splitIsOnTargetSize); 
```

Please make this change.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -507,4 +491,254 @@ LogicalResult splitLogicalObjectFifos(
...
+FailureOr<size_t> getSplitDim(AMDAIE::LogicalObjectFifoFromMemrefOp op) {
+  if (op.getMemorySpaceAsUInt() != 1) {
+    return op.emitOpError() << "expected objectFifo from L2 memory space";
+  }
+  ArrayRef<int64_t> memrefShape = op.getMemrefType().getShape();
+  if (memrefShape.size() <= 2) {
+    return op.emitOpError() << "expected objectFifo shape larger than 2";
+  }
+  size_t splitDim;
+  if (memrefShape[0] != 1) {
```
newling, line 529: Is it intentional to not split on both dimensions if they are both not one? If this is run on array with 1 row or 1 column, it will fail, maybe don't fail?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos.mlir
```diff
@@ -0,0 +1,159 @@
+// RUN: iree-opt --pass-pipeline="builtin.module(iree-amdaie-split-logical-objectfifos,cse)" --split-input-file --verify-diagnostics %s | FileCheck %s
```
newling, line 1: Don't see cse either I've confirmed this: maybe cherry https://github.com/nod-ai/iree-amd-aie/commit/ad6b693c943bf23cd7bf35decb7b870d10589822 (I've got some either minor changes there).

newling (changes requested):

newling:
It would be nice to have include a motivation for this in the description. I imagine it is to split tensors equally across all memory tiles, rather than putting them entirely on one? If this is the case, are you assuming nrows = ncols in choosing splitFactor? If the array is 4x4 and you have 

```
 %alloc_0 = memref.alloc() : memref<4x1x32x32xi32, 1 : i32>
 %alloc_1 = memref.alloc() : memref<1x4x32x32xi32, 1 : i32>
```

I guess it works nicely. But is the array is 4x8 (8 cols) and the allocs are 4x1x... and 1x8x... it's not optimal?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos.mlir
yzhang93: Cool, thanks!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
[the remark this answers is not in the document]
yzhang93: This is a good question. Well, it should work for matmul transpose as it is going through the same tiling strategy as matmul. Currently this pass does base on the assumption that the outer two dimensions of the L2 allocation are the number of cores for distribution. I think it might not work for conv ops as the outer two dimensions don't necessary connect to the core being used? We'll need to investigate how this can be used for conv ops as a separate thread in the future.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos.mlir
yzhang93: I wasn't able to cherry-pick this commit. I tried to include most of your changes in the revision.

newling: Oh sorry, I wasn't sure if it would be possible to cherry-pick a commit. I think my mistake was rebasing your PR onto main, which meant that the collapsed commits included unrelated changes

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
yzhang93: I prefer not doing this. splitDim, splitFactor and splitDimTarget are determined by the input DmaCpyNdOp, each DmaCpyNdOp could have different sets of those values. I could split this function into two parts that first acquiring these values from the input dma and then pass the parameters with this dma to `splitDoublyStridedOp` again, but I think it would make the function call in `AMDAIESplitLogicalObjFifos` more complicated.

jtuyls: Yeah, I also feel like this function needs to be split up into one function that decides which DMA copy ops to split (and what the splitting factor should be) and another one that does the actual splitting.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -507,4 +512,266 @@ LogicalResult splitLogicalObjectFifos(
...
+    splitDim = memrefShape.size();
+  } else {
+    return op.emitOpError() << "failed to find a dimension for splitting";
+  }
+  return splitDim;
+}
+
+/// Split L2 space input and output logical objectFifos.
+LogicalResult splitLogicalObjectFifo(IRRewriter &rewriter,
+                                     AMDAIE::LogicalObjectFifoFromMemrefOp op) {
```
jtuyls, line 564: It would also be more reusable to split up this function into one which selects the logical objFifos to be split and the other one that actually does the splitting. The checks on which ones to split might actually be put together with the checks that select which DMA copy ops to split.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -507,4 +512,266 @@ LogicalResult splitLogicalObjectFifos(
   return success();
 }
 
+/// Utility to get the DoublyStridedCopyOp producers and consumers of a given
```
jtuyls, line 515:
```suggestion
/// Utility to get the `DmaCpyNdOp` producers and consumers of a given
```

yzhang93:
> It would be nice to have include a motivation for this in the description. I imagine it is to split tensors equally across all memory tiles, rather than putting them entirely on one? If this is the case, are you assuming nrows = ncols in choosing splitFactor? If the array is 4x4 and you have
> 
> ```
>  %alloc_0 = memref.alloc() : memref<4x1x32x32xi32, 1 : i32>
>  %alloc_1 = memref.alloc() : memref<1x4x32x32xi32, 1 : i32>
> ```
> 
> I guess it works nicely. But is the array is 4x8 (8 cols) and the allocs are 4x1x... and 1x8x... it's not optimal?

Yes, currently it aims for balance use of number of rows and columns (i.g., 2x2/4x4) and added basic support for that. Also it should work for unbalanced use of columns/rows such as A: 4x1x... B: 1x2x...., which will split A into 4 separate objectfifos and B into 2 objectfifos, but we can't control how to distribute these in this pass. I think it's the tiling strategy and the tile assignment strategy that should responsible for the distribution part.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -351,31 +375,13 @@ LogicalResult splitLogicalObjectFifos(
...
-  }
-
   OpFoldResult zeroVal = getAsIndexOpFoldResult(context, 0);
   OpFoldResult oneVal = getAsIndexOpFoldResult(context, 1);
 
-  if (dmaTransposeOnSource) {
+  FailureOr<bool> dmaTransposeOnTarget = isL2DmaTransposed(l3ToL2DmaOp, true);
+  if (failed(dmaTransposeOnTarget)) return failure();
+
+  if (!dmaTransposeOnTarget.value()) {
```
jtuyls, line 384: The changes in this function are refactorring changes for the matmul + elementwise and I think they should be put in a separate PR to make it easier to review the splitting additions needed for 4x4 in this PR.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -8,31 +8,28 @@
...
+/// The values are based on the results from ConvertToDma with option as
+/// transposed on target, e.g., dma size [1, 1, 32, 32] -> [1, 32, 1, 32].
+const static SmallVector<size_t> transposedL2Dims = {0, 2, 1, 3};
+
+/// Utility to create a new logical objectfifo.
 static AMDAIE::LogicalObjectFifoFromMemrefOp createNewLogicalObjectFifo(
     IRRewriter &rewriter,
     AMDAIE::LogicalObjectFifoFromMemrefOp &oldLogicalObjectFifo,
-    SmallVectorImpl<OpFoldResult> &newSizesOpFoldResultArr) {
+    const SmallVectorImpl<int64_t> &newSizes) {
```
jtuyls, line 31: `ArrayRef`?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -507,4 +512,266 @@ LogicalResult splitLogicalObjectFifos(
...
+  if (failed(maybeSplitDim)) return failure();
+  size_t splitDim = maybeSplitDim.value();
+
+  // Both outer dims are 1, no need to split, return success.
+  if (splitDim == memrefShape.size()) return success();
+  int64_t splitFactor = memrefShape[splitDim];
+  if (splitFactor == 1) return success();
+
+  FailureOr<bool> l2DmaTransposed = isL2DmaTransposed(op, isL2Target);
+  if (failed(l2DmaTransposed)) return failure();
```
jtuyls, line 710: Why is this needed?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -507,4 +512,266 @@ LogicalResult splitLogicalObjectFifos(
...
+  } else {
+    return op.emitOpError()
+           << "the input dma should have source or target in L2 memory space";
+  }
+  if (failed(maybeSplitDim)) return failure();
+  size_t splitDim = maybeSplitDim.value();
+
+  // Both outer dims are 1, no need to split, return success.
+  if (splitDim == memrefShape.size()) return success();
+  int64_t splitFactor = memrefShape[splitDim];
```
jtuyls, line 706: The splitting factor should really be derived from the number of columns as soon we might create larger L2 buffers that need to be split in a smaller factor (`[1, 16, 32, 32]`, split into 4 x `[1, 4, 32, 32]` for example). Might be ok to leave for future work though.

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
yzhang93: The original change for matmul-elementwise was introduced in this PR https://github.com/nod-ai/iree-amd-aie/pull/850. This utility is also used in 4x4 matmul splitting, so here is just small refactor.

jtuyls: It's not that small as there's quite a few changes in the top part of this file for it and it's completely independent, so I think it's really better for clarity/review purposes to put it in a separate PR to go in first.

yzhang93: Okay, moved the refactor to a separate PR https://github.com/nod-ai/iree-amd-aie/pull/910.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
yzhang93: It's because the shape of objectFifo and the target size of DmaCpNdOp can be different if we select transpose on target option within `ConvertToDma` pass. The splitDim we get earlier in this function is only the split dim in objectFifo, but not the split dim for the target size in DmaCpNdOp.

For example, in this example `%3 = amdaie.dma_cpy_nd(%0[0, 0, 0, 0] [1, 32, 2, 32] [2048, 32, 1024, 1], %1[0, %2] [32, 64] [128, 1]) : (!amdaie.logicalobjectfifo<memref<1x2x32x32xi32, 1 : i32>>, !amdaie.logicalobjectfifo<memref<128x128xi32>>)`,

the splitDim for objectfifo is index 1 while the splitDimTarget should be index 2.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
yzhang93: Okay, done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -507,4 +512,275 @@ LogicalResult splitLogicalObjectFifos(
...
+  splitFactor = memrefShape[splitDim];
+  if (ShapedType::isDynamic(splitFactor)) {
+    return op.emitOpError()
+           << "a dynamic size on the split dimension is not supported";
+  }
+  return success();
+}
+
+/// Utility to get the split dimension and factor from a L3->L2 dma op.
+LogicalResult getSplitDimAndFactorFromDma(AMDAIE::DmaCpyNdOp op,
```
jtuyls, line 574: This `getSplitDim...` function and the other one above are quite hardcoded, but as discussed offline I agree it's ok for now to get an initial 4x4 working. We will start generalizing this soon in follow-ups.

jtuyls (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -502,4 +502,277 @@ LogicalResult splitLogicalObjectFifoForElementwiseOp(
...
+           << splitFactor << sourceSize.value() << targetSize.value();
+  }
+  int64_t newSourceSize = sourceSize.value() / splitFactor;
+  int64_t newTargetSize = targetSize.value() / splitFactor;
+  sourceSizes[splitDim] = rewriter.getIndexAttr(newSourceSize);
+  targetSizes[splitDimInL2Dma] = rewriter.getIndexAttr(newTargetSize);
+
+  // Create `splitFactor` number of doubly stride ops.
+  rewriter.setInsertionPoint(op);
+  for (int i = 0; i < splitFactor; ++i) {
```
newling, line 758: Not blocking this PR on this question, but trying to understand how we can generalize this. 

It seems like @yzhang93 you are is assuming that the `size` of the dma is exactly the same as the shape of the memref of the objectfifo. This needn't be the case, right, they can be completely different? ( @jtuyls )

newling (changes requested):
> Yes, currently it aims for balance use of number of rows and columns (i.g., 2x2/4x4) and added basic support for that

It would be great if you can add a check on this in the code. My only remaining requests are

1) document the assumed relationship between first 2 dims of tensors and nrows, nols
2) add tests/safeguards for the cases nrows != ncols

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -662,6 +662,22 @@ def AMDAIESinkIntoCore :
...
+  let description = [{
+    Splitting L2 input and output logical objectFifos and their user dma operations
+    by a factor of the number of AIE columns being used, so that the logical objectFifos
```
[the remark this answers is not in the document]
jtuyls: No, but there's still a relationship with the number of columns. Suppose `rows: 4, cols: 2` and a different input being broadcasted to each row, then you likely still want to split it in `2` (number of columns), instead of 4.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
jtuyls: Yeah, that's not necessarily the case and soon won't be valid when we cache larger buffers on L2, but I think it's ok for now assume that to start off. We should make sure though to fail if the assumptions don't hold.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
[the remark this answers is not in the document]
newling: Exactly, that makes sense. Currently though it'd get split 4 ways afaict (I know the plan is to revisit this which is fine by me).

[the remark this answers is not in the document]
jtuyls: > Currently though it'd get split 4 ways afaict

Yes, indeed, and that's a 'valid' splitting, however, it will likely fail because of not enough channels.

yzhang93:
> > Yes, currently it aims for balance use of number of rows and columns (i.g., 2x2/4x4) and added basic support for that
> 
> It would be great if you can add a check on this in the code. My only remaining requests are
> 
> 1. document the assumed relationship between first 2 dims of tensors and nrows, nols
> 2. add tests/safeguards for the cases nrows != ncols

I've added more comments for the current assumptions. I'm not adding more tests specially for nrows != ncols, will leave it in the follow ups when the split factor is only depend on the ncols not the way it's hardcoded right now.

newling (approved):

PR #904, opened 2024-11-15
[AMDAIEAccessToAcquireRelease] Insert acquire/release pairs in same scope

newling:
Before this PR, the logic for read access was invalid. Code like 

```
for i in range(10):
    read_access
read_access
```

got transformed to 

```
for i in range(10):
    acquire
    read_access
release
acquire
read_access
release
```

As you can see above, the code ignored the scoping of the accesses. This PR ensures that acquires and releases are inserted in the same scope.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAccessToAcquireRelease.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAccessToAcquireRelease.cpp
@@ -4,10 +4,10 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+#include "iree-amd-aie/IR/AMDAIEAttrs.h"
 #include "iree-amd-aie/IR/AMDAIEOps.h"
+#include "iree-amd-aie/Transforms/AMDAIEUtils.h"
 #include "iree-amd-aie/Transforms/Passes.h"
-#include "mlir/IR/IRMapping.h"
-#include "mlir/IR/Iterators.h"
 
 #define DEBUG_TYPE "iree-amdaie-access-to-acquire-release"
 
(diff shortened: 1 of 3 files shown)


jtuyls:
> TBH I still don't understand what these ops actually mean. I would like to update the `write` access case too to be scope-safe, but am hesitant to do so before I understand the underlying semantics. I have added a lit test illustrating how our vanilla matmul's writes get transformed, it is not obvious to me why this is the desired place to put acquires and accesses. @jtuyls I would really appreciate if you can write some rough documentation of what these ops (acquire-access-release) actually signify, and why this pass does what it does.

These ops are documented here: https://github.com/nod-ai/iree-amd-aie/blob/076ea13e599d5c4283b989d0e6587b698220ef53/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td#L1102. They relate to the AIE semaphore locks, which you can find a lot of information about. Could you ask specific questions or tell me what's not clear? I am not going to speculate on what you're looking for and write further documentation at this point.

newling:
> Could you ask specific questions or tell me what's not clear?

Why does this pass insert acquire/release ops where it does? 

By the definition in https://github.com/nod-ai/iree-amd-aie/blob/076ea13e599d5c4283b989d0e6587b698220ef53/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td#L1102 it would seem fine to make the wrapping tight, i.e. insert the acquire just before the access, and the release just after. Or why not just specify that the access is atomic? But that's not what the pass seems to doing, and that also seem to work (numerical errors). 

What is happening with None accesses, specifically what is the reason for the transformation that takes place in epilogue_write_with_preceding_none_accesses (new lit test)? Why is the logic different for Write and Read, but the definition in https://github.com/nod-ai/iree-amd-aie/blob/076ea13e599d5c4283b989d0e6587b698220ef53/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td#L1102 there is no difference. 

Also, please can you review the change to the read case and let me know if the changes are valid.

jtuyls (changes requested):
> it would seem fine to make the wrapping tight, i.e. insert the acquire just before the access, and the release just after. Or why not just specify that the access is atomic? But that's not what the pass seems to doing, and that also seem to work (numerical errors).

These are sempahore acquire/release operations, which are atomic operations, not sure though what you mean with 'the access is atomic'? You have to acquire before the first use and release after the last use. Because once you release a buffer, DMAs can again acquire it and read/write access it. 

> What is happening with None accesses, specifically what is the reason for the transformation that takes place in epilogue_write_with_preceding_none_accesses (new lit test)?

This is the result of different core 'programs' being stitched together into a single core program (here it's the prolog, main and epilog, but it doesn't have to). In the prolog and main there is an access to an objFifo, which is can't be designated as read nor as write at that point. However, in the epilog it can, but the none accesses still need to be taken into account to insert the semaphore ops at the correct places.

> Why is the logic different for Write and Read

Maybe this can be consolidated, I would need to have another in-depth look.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAccessToAcquireRelease.cpp
```diff
@@ -4,88 +4,117 @@
...
+                                    newAccessOp.getResult());
+
+        // Insert the release op. The location of the release is as close to the
+        // following access op as possible, but always in the same block as the
+        // access op being released.
+        AMDAIE::LogicalObjectFifoAccessOp nextAccessOp;
+        if (i + 1 != accessOps.size()) nextAccessOp = accessOps[i + 1];
+        Operation *nextAccessOpsAncestor =
+            getAncestorInBlock(nextAccessOp, block);
+        if (nextAccessOpsAncestor && nextAccessOpsAncestor->getBlock() == block)
```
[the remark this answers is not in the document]
newling: First case: yeah this works (I added a new test case to check)
Second case: Is this the right pass to do this? I would have thought that the pass where accesses are inserted would be better

jtuyls:
> There is a problem with using the same algorithm for Write (it results in numerical errors, so definitely some problem). I want to get back to investigating this at a later point, and have left a TODO(newling) in the code for this. As such, I would like to leave the Write case untouched in this PR.

Yeah, this is what I remember, but I don't know exactly why that was the case. Maybe it's the 'none' handling.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAccessToAcquireRelease.cpp
```diff
@@ -4,88 +4,110 @@
...
+/// Some blocks have terminator ops, which must appear as the very last op in
+/// the block. If `block` has a terminator, set the insertion point of
+/// `rewriter` to just before the terminator, ready to create a new penultimate
+/// op in the block. Otherwise, set the insertion point to the very end of the
+/// block.
+void setInsertionToEnd(IRRewriter &rewriter, Block *block) {
+  if (block->back().hasTrait<OpTrait::IsTerminator>())
+    rewriter.setInsertionPoint(block->getTerminator());
+  else
+    rewriter.setInsertionPointToEnd(block);
```
jtuyls, line 27:
```suggestion
  if (block->back().hasTrait<OpTrait::IsTerminator>()) {
    rewriter.setInsertionPoint(block->getTerminator());
  } else {
    rewriter.setInsertionPointToEnd(block);
  }
```
Is this your formatter? You might need to set it on Google code style.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAccessToAcquireRelease.cpp
```diff
@@ -4,88 +4,110 @@
...
+        // Insert the release op. The location of the release is as close to the
+        // following access op as possible, but always in the same block as the
+        // access op being released.
+        AMDAIE::LogicalObjectFifoAccessOp nextAccessOp;
+        if (i + 1 != accessOps.size()) nextAccessOp = accessOps[i + 1];
+        Operation *nextAccessOpsAncestor =
+            getAncestorInBlock(nextAccessOp, block);
+        if (nextAccessOpsAncestor &&
+            nextAccessOpsAncestor->getBlock() == block) {
+          llvm::errs() << "Setting insertion point to nextAccessOpsAncestor\n";
```
jtuyls, line 103: This should be removed, no?

jtuyls (commented):
LGTM

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAccessToAcquireRelease.cpp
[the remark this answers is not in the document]
newling: Ok, thanks @jtuyls 

Maybe we need an canonicalize-accesses pass which does something like liveness analysis

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAccessToAcquireRelease.cpp
newling: No, I don't have any fancy formatter, just missed this one. 

Personally I wish C++ had never allowed dropping braces in the first place, but let's leave that can of worms closed...

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAccessToAcquireRelease.cpp
newling: Oof, I clearly should have gone over this a bit more carefully after my last change. Thanks for spotting

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/access_to_acquire_release.mlir
```diff
@@ -219,3 +220,180 @@ func.func @multiple_writes_deterministic_order(%arg0: !amdaie.logicalobjectfifo<
...
+      linalg.fill ins(%c0_i32 : i32) outs(%access2 : memref<32xi32, 2>)
+    }
+    %access3 = amdaie.logicalobjectfifo.access(%arg0, Write) : !amdaie.logicalobjectfifo<memref<32xi32, 2>> -> memref<32xi32, 2>
+    linalg.fill ins(%c0_i32 : i32) outs(%access3 : memref<32xi32, 2>)
+    amdaie.end
+  }
+  return
+}
+
+
```
jtuyls, line 399:
```suggestion
```

jtuyls (approved):
LGTM

PR #905, opened 2024-11-15
Bump IREE to 5318fce9dcc809b13420ab9ff22c4c4320e26746

yzhang93:
Bump mlir-air to 3bab1025d02ffd2b14c0e887bb3749b4836936b2


newling:
```
The following tests FAILED:
	130 - iree/target/amd-aie/Test/transform_dialect/conv_fill_spec_pad.mlir.test (Failed)
```

I'm ok with just removing this test (the actual conv pipeline is now different from the transform dialect version, anyway)

yzhang93:
> ```
> The following tests FAILED:
> 	130 - iree/target/amd-aie/Test/transform_dialect/conv_fill_spec_pad.mlir.test (Failed)
> ```
> 
> I'm ok with just removing this test (the actual conv pipeline is now different from the transform dialect version, anyway)

This test always has flaky failures on windows and macOs, and we already disabled it on macOs. Maybe we should only run this transform_dialect/ directory on linux.

newling (approved):

PR #908, opened 2024-11-18
Add pass that inserts infinite loops around core blocks

jtuyls:
Adds a new pass that inserts infinite looping around the `amdaie.core` blocks. This results in the cores running the same program over and over which is useful for measuring performance statistics like latency/throughput, averaged over a certain number of runs, while excluding core reconfiguration overhead.

Concretely, with this flag, it becomes possible to put a loop around the kernel command execution:

```
auto time0 = std::chrono::high_resolution_clock::now();
int nb_runs = 1;
for (int i = 0; i < nb_runs; i++) {
  ebuf.m_cmd_pkt->state = ERT_CMD_STATE_NEW;
  hwq->issue_command(ebuf.get_exec_buf_bo());
  hwq->wait_command(ebuf.get_exec_buf_bo(), 0);
}
auto time1 = std::chrono::high_resolution_clock::now();
std::chrono::duration<double> elapsed_s = (time1 - time0) / nb_runs;
```

at this location:  https://github.com/nod-ai/iree-amd-aie/blob/076ea13e599d5c4283b989d0e6587b698220ef53/runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc#L201

This way, every computation and data movement inside the NPU/AIE will be repeated N number of times.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.h
```diff
@@ -127,6 +125,9 @@ std::unique_ptr<Pass> createAMDAIEControlCodeToTransactionPass(
 /// Pass to convert `scf.forall` to `scf.for` within `aie.core`.
 std::unique_ptr<Pass> createAMDAIEConvertCoreForallToForPass();
 
+/// Pass to inserts an infinite loop around each `amdaie.core`'s block.
```
newling, line 128:
```suggestion
/// Pass to insert an infinite loop around each `amdaie.core`'s block.
```

newling (changes requested):
Does the core repeat the fill/matmul indefinitely, but the data is delivered from L3 only once? How is this intended to be used? It's not clear from the comments how an infinite loop can be used to "average over a certain number of runs". Please address this comment by improving the description in-code.

jtuyls:
> Does the core repeat the fill/matmul indefinitely, but the data is delivered from L3 only once? How is this intended to be used? It's not clear from the comments how an infinite loop can be used to "average over a certain number of runs". Please address this comment by improving the description in-code.

Done

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_infinite_loop_around_core_block.mlir
```diff
@@ -0,0 +1,238 @@
+// RUN: iree-opt --pass-pipeline="builtin.module(func.func(iree-amdaie-insert-infinite-loop-around-core-block),canonicalize)" --split-input-file %s | FileCheck %s
```
newling, line 1: Is it necessary to canonicalize in this lit test?

newling (changes requested):
Thanks, it's starting to make more sense to me now. Still a few questions 

Is this definitely useful, have you seen subsequent runs being significantly faster? If so, is it noticeably different from using the n_runs/n_repeats flags already exposed in run.py? If so, is the reason that the interval between runs is reduced with this new approach, and this reduced delay prevents the AIE going to sleep? Would be nice to confirm some of this I think before commiting code.  

So there seem to be 3 loops here: 

1) not in this PR, the developer will insert in runtime C++ 
2) The scf.for with the LCM for object fifo depths
3) The scf.while true, referred to in the pass name 
 
Is your reason for not including 1 that it will be too 'invasive'? Could it not be incorporated into the existing precompiler `#ifdef` logic using `<chrono>` and `<iostream>` ? 

I think if you could include 1 with some logic to ensure that 2 & 3 are automatically enabled, that would make this tool more widely useful to devs without background knowledge.  

Again, please reply to as much of the above in the docs.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_infinite_loop_around_core_block.mlir
jtuyls: Yes, leads to better and simpler code checks.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -37,18 +37,22 @@ def run_conv_test(config, filename, n_repeats):
...
     The derived instances would therefore be created specifying the intended target
     device(s) they're to be run; and would accordingly be `run` if the intended
     target device(s) contains the `target_device` supplied via command-line.
     """
 
-    def __init__(self, run_on_target):
+    def __init__(self, run_on_target, additional_aie_compilation_flags=None):
         self.run_on_target = run_on_target
+        self.additional_aie_compilation_flags = additional_aie_compilation_flags \
+            if additional_aie_compilation_flags is not None else []
```
yzhang93, line 51: Can you just do `__init__(self, run_on_target, additional_aie_compilation_flags=[])`
and `self.additional_aie_compilation_flags = additional_aie_compilation_flags`?

jtuyls:
> Is this definitely useful, have you seen subsequent runs being significantly faster? f so, is it noticeably different from using the n_runs/n_repeats flags already exposed in run.py?

Yes.

> If so, is the reason that the interval between runs is reduced with this new approach, and this reduced delay prevents the AIE going to sleep?

Yes, very likely.

> Is your reason for not including 1 that it will be too 'invasive'? Could it not be incorporated into the existing precompiler #ifdef logic using <chrono> and <iostream> ?

Yes, this can be done in a follow up. For now, manually adjusting works well for me and I would even have to do that after putting it inside the ifdef as I would have to adjust the number of runs anyway. So yeah, I prefer to do this well later.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertInfiniteLoopAroundCoreBlock.cpp
```diff
@@ -0,0 +1,129 @@
...
+  auto endOp = rewriter.create<AMDAIE::EndOp>(rewriter.getUnknownLoc());
+  rewriter.moveOpBefore(whileOp, endOp);
+  return success();
+}
+
+/// Insert a loop with extent equal to the least common multiple of the depths
+/// of the objFifos operated on. This ensures functional correctness after
+/// potential further loop unrolling based on the depths of the objFifos.
+/// Specifically, this anticipates the unrolling happening inside the
+/// `AcquireReleaseToUseLock` pass and is somewhat coupled with it.
```
yzhang93, line 55: Still not clear to me why this is needed for coreLoopUnroll. Since this pass will be run before `AcquireReleaseToUseLock`, is it possible/more clear to move `coreLoopUnroll` logic to this pass?

build_tools/ci/cpu_comparison/run.py
jtuyls: I prefer not doing that because it can lead to potential issues (although likely not in this case): https://stackoverflow.com/questions/43255481/python-default-list-in-function

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -320,6 +320,33 @@ def AMDAIEInsertCores :
   let constructor = "mlir::iree_compiler::AMDAIE::createAMDAIEInsertCoresPass()";
 }
 
+def AMDAIEInsertInfiniteLoopAroundCoreBlock :
+    Pass<"iree-amdaie-insert-infinite-loop-around-core-block", ""> {
+  let summary = "Inserts an infinite loop around each `amdaie.core`'s block.";
+  let description  = [{
+    This pass inserts infinite looping around the `amdaie.core` blocks. This
```
newling, line 327: Great description, thank you

newling (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertInfiniteLoopAroundCoreBlock.cpp
jtuyls: If I don't insert the loop with extent `lcm(depths)`, then there would be a while loop and unrolling wouldn't be possible, leading to numerical issues as different iterations of the loop potentially need to operate on different buffers (double buffering).

> Since this pass will be run before AcquireReleaseToUseLock, is it possible/more clear to move coreLoopUnroll logic to this pass?

No, because the pass in this PR is only run optionally and `coreLoopUnroll` always needs to be run.

yzhang93 (approved):

build_tools/ci/cpu_comparison/run.py
newling: Just noticed: you are overriding the additional flags for every test which runs subsequent to the one that appends. I'll try and think of a way to add flags to a test in isolation

PR #909, opened 2024-11-18
[CI] Add new shape and with/without ukernel for benchmark tracking

newling:
Now prints performance numbers for shapes

(M, N, K) = (512, 512, 4096) and (512, 4096, 512) and (4096, 512, 512)

with and without ukernels. Numbers with ukernels are interesting, take a look in CI, different gain with ukernel for the different shapes. 

But looks like there's some improvement we can make on the vectorization of matmul front.

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -241,6 +241,7 @@ class VanillaMatmul(BaseMatmul):
         run_on_target=["npu1_4col"],
         additional_labels=None,
         aie_compilation_flags=None,
+        n_repeats=1,
     ):
         super().__init__(
             run_on_target=run_on_target,
@@ -252,7 +253,7 @@ class VanillaMatmul(BaseMatmul):
             acc_type=acc_type,
             tile_pipeline="pack-peel",
             use_ukernel=use_ukernel,
-            n_repeats=1,
+            n_repeats=n_repeats,
         )
 
         self.name = f"vanilla_matmul_{M}_{N}_{K}_{input_type}_{acc_type}"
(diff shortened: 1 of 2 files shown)


build_tools/ci/cpu_comparison/run.py
```diff
@@ -1021,6 +1023,18 @@ def __init__(self):
             )
         )
 
+        self.register(
+            VanillaMatmul(
+                512,
+                512,
+                4096,
```
jtuyls, line 1030: It should be M == 512, N == 4096, K == 512 to correspond to the problem size we're tracking.

newling: Please tell me all the sizes we're tracking (sorry I lost the link to the doc again). I'd like to make this the official source of the tracker.

jtuyls: M == 512, N == 4096, K == 512
M == 4096, N == 512, K == 512

build_tools/ci/cpu_comparison/run.py
```diff
@@ -1056,6 +1058,19 @@ def __init__(self):
             )
         )
 
+        self.register(
+            VanillaMatmul(
+                "x",
```
yzhang93, line 1063: What does "x" represent?

yzhang93 (commented):
Any idea of the windows failure?

newling:
> Any idea of the windows failure?

Yes I'm 90% sure it'll be fixed after rebasing on https://github.com/nod-ai/iree-amd-aie/pull/911 (windows + ukernel = doesn't work)

build_tools/ci/cpu_comparison/run.py
```diff
@@ -1093,16 +1094,25 @@ def __init__(self):
...
-            VanillaMatmul(
-                512,
-                512,
-                4096,
-                "bf16",
-                "f32",
-                additional_labels=["Performance"],
+        # Some bf16 Performance tests:
+        for M, N, K, use_ukernel in [
+            (512, 512, 4096, False),
```
yzhang93, line 1099: I think @jtuyls suggested M=4096 N=512 K=512 for performance tracking as well.

yzhang93 (approved):

jtuyls (approved):
LGTM. Great to have these shapes benchmarked on every commit!

PR #910, opened 2024-11-18
Refactor LogicalObjFifoSplittingUtils

yzhang93:
This PR made some necessary changes in `LogicalObjFifoSplittingUtils` for matmul-elementwise ops as preparation for another splitting usage in https://github.com/nod-ai/iree-amd-aie/pull/903. Specially, the utility to check if the L3->L2 dma is transposed on the L2 side is rewritten.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.h
```diff
@@ -11,11 +11,8 @@
 
 namespace mlir::iree_compiler::AMDAIE {
 
-/// Utility to help fetch those input DmaCpyNd Ops which needs to be split.
-SmallVector<AMDAIE::DmaCpyNdOp> fetchDmaCpyNdOpsToSplitOrCombine(Operation *op);
-
 /// Utility to split logicalobjectfifos given a vector of L2->L1 dma ops.
-LogicalResult splitLogicalObjectFifos(
+LogicalResult splitThirdInputLogicalObjectFifos(
```
newling, line 15: Seems like the fact that the object fifos are all third inputs in the core isn't used in this function, so I propose removing it from the function name

yzhang93: Maybe the name is not accurate. But what it really means is it has the third input dma in `amdaie.core(in : [xx,  xx, xx], out : [xx]) `, e.g. https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos_for_connection_reuse.mlir#L135 vs https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos_for_connection_reuse.mlir#L36.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -23,16 +19,17 @@
...
+/// The values are based on the results from ConvertToDma with option as
+/// transposed on target, e.g., dma size [1, 1, 32, 32] -> [1, 32, 1, 32].
+const static SmallVector<size_t> transposedL2Dims = {0, 2, 1, 3};
+
+/// Utility to create a new logical objectfifo.
 static AMDAIE::LogicalObjectFifoFromMemrefOp createNewLogicalObjectFifo(
     IRRewriter &rewriter,
     AMDAIE::LogicalObjectFifoFromMemrefOp &oldLogicalObjectFifo,
-    SmallVectorImpl<OpFoldResult> &newSizesOpFoldResultArr) {
+    const SmallVectorImpl<int64_t> &newSizes) {
```
jtuyls, line 31:
```suggestion
    ArrayRef<int64_t> newSizes) {
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -305,17 +296,50 @@ static LogicalResult checkWhetherSplitIsPossible(
...
+/// Utility to check if the L3->L2 dma is transposed on the L2 side.
+/// The intuition is if the sizes of L2 and L3 side are the same and L2 dma
+/// strides are in non-decreasing order from right to left, then the dma has
+/// transposition on L3 side, otherwise the transposition is on L2 side.
+static FailureOr<bool> isL2DmaTransposed(AMDAIE::DmaCpyNdOp l3ToL2DmaOp,
+                                         bool isL2Target) {
+  SmallVector<OpFoldResult> l2DmaSizes;
+  SmallVector<OpFoldResult> l2DmaStrides;
+  SmallVector<OpFoldResult> l3DmaSizes;
+  bool l2DmaTransposed = true;
```
jtuyls, line 308: I suggest moving this closer to the use. I.e. before https://github.com/nod-ai/iree-amd-aie/pull/910/files#diff-591f54662e604611ca835020aa99f74410d5e332fda12400ccf2eca421d2283bR325

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -305,17 +296,50 @@ static LogicalResult checkWhetherSplitIsPossible(
...
+  } else {
+    l2DmaSizes = l3ToL2DmaOp.getSourceMixedSizes();
+    l3DmaSizes = l3ToL2DmaOp.getTargetMixedSizes();
+    l2DmaStrides = l3ToL2DmaOp.getSourceMixedStrides();
+  }
+  std::optional<SmallVector<int64_t>> staticStrides =
+      getConstantIntValues(l2DmaStrides);
+  if (!staticStrides) {
+    LLVM_DEBUG(llvm::dbgs() << "expected static L2 strides\n");
+    return failure();
```
jtuyls, line 322:
```suggestion
    return l3ToL2DmaOp.emitOpError() << "expected static L2 strides";
```
I think it's better to return this as an error because you're returning a failure anyway.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifosForConnectionReuse.cpp
```diff
@@ -34,10 +32,20 @@ void AMDAIESplitLogicalObjFifosForConnectionReusePass::runOnOperation() {
...
-
-  if (failed(splitLogicalObjectFifos(rewriter, l2ToL1DmaOps, context))) {
+  // Walk through CoreOps gathering 3rd input DmaOps (if applicable) which will
+  // be used to split L2 objectFifos of elementwise input for connection reuse.
+  SmallVector<AMDAIE::DmaCpyNdOp> l2ToL1DmaOps;
+  moduleOp->walk([&](AMDAIE::CoreOp coreOp) {
+    SmallVector<Value> inputDmas = coreOp.getInputDmas();
+    if (inputDmas.size() != 3) return WalkResult::skip();
+    auto dmaCpyNdOp = inputDmas[2].getDefiningOp<AMDAIE::DmaCpyNdOp>();
+    assert(dmaCpyNdOp && "expected an amdaie.dma_cpy_nd op");
```
jtuyls, line 42: Here we should throw an error as core ops operating on `DmaCpyNdOp` is not guaranteed.

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -305,17 +296,50 @@ static LogicalResult checkWhetherSplitIsPossible(
...
-//       create a logicalobjectfifo using the same.
-//    b) Split L3->L2 Dma op.
-//    c) SPlit L2->L1 Dma op.
-// 4. Delete old L2->L1, L3->L2 and corresponding AllocOps.
-LogicalResult splitLogicalObjectFifos(
+/// Utility to check if the L3->L2 dma is transposed on the L2 side.
+/// The intuition is if the sizes of L2 and L3 side are the same and L2 dma
+/// strides are in non-decreasing order from right to left, then the dma has
+/// transposition on L3 side, otherwise the transposition is on L2 side.
+static FailureOr<bool> isL2DmaTransposed(AMDAIE::DmaCpyNdOp l3ToL2DmaOp,
```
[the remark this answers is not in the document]
newling: Just realised, we should actually also ignore dimensions of size 1 and 0 when checking for decreasing dimensions, because there stride values mean nothing. Maybe that is why the shape comparison is needed

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.h
newling: Very simplistic example explaining my problem with the name:

```
def doubled_values_less_than_ten(X):
  return [2*v for v in X]
  
def main():
  vals = [1, 20]
  vals_less_than_ten = [x for x in vals if x < 10]
  return doubled_values_less_than_ten(vals_less_than_ten)
```

`double_values_less_than_ten` should be called `doubled_values`

i.e. where is the fact they're third arguments used, afaict it's not relevant.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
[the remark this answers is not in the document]
newling: Ok, I've updated with the 0/1 check on dimsize and it seems to work without the rank-shape check, which is hopefully an improvement and more reusable whenever we want to check for transposition in dma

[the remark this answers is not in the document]
newling: Vivian, hopefully this time you can just merge this PR into your branch (if you think it's a useful contribution)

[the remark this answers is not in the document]
yzhang93: @newling Thanks for your contribution! What I don't understand is why the stride is non-static, then this might be a transpose. And if we use word "maybe" which means we are not sure if it's transposed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.h
yzhang93: How about name "splitLogicalObjectFifosForElementwiseOp" or "splitLogicalObjectFifosForThirdInputInCore"?

newling: I think those names have the same issue -- where inside the function is it used that this comes from an elementwise op, or from the third arg in a core?  A function name should say just what it does, not who for :) Maybe this func shouldn't be exposed as a utility until we've generalized it more, will it be used outside of the original pass in a follow up PR? But anyway I know this isn't original code in PR, so I'm happy to accept it with any name and we can get back to a refactor / rename later, my vote of the 3 goes for splitLogicalObjectFifosForElementwiseOp

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELogicalObjFifoSplittingUtils.cpp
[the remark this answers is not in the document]
newling: My thinking is, if the stride is `?` it could be a very large or a very small stride, which means this might be transpose. i.e. it is "maybe" a transpose. 

Example 

sizes = [4,4,4] 
strides = [16, ?, 1]

If the `?` is greater than 16, then this represents a transpose. If it <= 16, then it does not represent a transpose. 

I guess we could have had signature `isDefinitelyTranspose` and returned false for `?` (because with `?` it isn't _definitely_ a transpose)

It's all very edge-casey though as I don't think we'll ever have dynamic strides.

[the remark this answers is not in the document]
newling: Yeah, I thought that's what you had. I just think returning failure increases the complexity _slightly_, but I don't have a strong preference either way if it's documents what the true/failure/false cases denote.

newling (approved):
Looks good to me.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifosForConnectionReuse.cpp
```diff
@@ -34,10 +32,22 @@ void AMDAIESplitLogicalObjFifosForConnectionReusePass::runOnOperation() {
...
-  if (failed(splitLogicalObjectFifos(rewriter, l2ToL1DmaOps, context))) {
+  // Walk through CoreOps gathering 3rd input DmaOps (if applicable) which will
+  // be used to split L2 objectFifos of elementwise input for connection reuse.
+  SmallVector<AMDAIE::DmaCpyNdOp> l2ToL1DmaOps;
+  moduleOp->walk([&](AMDAIE::CoreOp coreOp) {
+    SmallVector<Value> inputDmas = coreOp.getInputDmas();
+    if (inputDmas.size() != 3) return WalkResult::skip();
+    auto dmaCpyNdOp = inputDmas[2].getDefiningOp<AMDAIE::DmaCpyNdOp>();
+    if (!dmaCpyNdOp) {
+      coreOp->emitOpError() << "failed to get a DmaCpyNdOp from the input";
```
jtuyls, line 43: You should return a pass failure here I think

jtuyls: Or return an interrupt and catch it afterwards to return a pass failure.

yzhang93: Oh, my bad. Thanks for catching it.

jtuyls (approved):
LGTM

PR #911, opened 2024-11-19
[CI] run.py improvements 

newling:
- make `aie_compilation_flags` test specific (undoes mixing of local/global state of https://github.com/nod-ai/iree-amd-aie/pull/908)

- use above to simplify code controlling function outlining, coalescing, etc 

- reuse logic of when to run ukernel (move to base class) 

- remove return-on-fail flag (always return on fail)

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -49,14 +49,17 @@ class BaseTest(ABC):
     An instance of this class has a member `aie_compilation_flags`
     which are additional flags to be passed to the AIE backend compiler.
 
-    Compilation flags can therefore be injected in 2 ways:
+    Compilation flags can therefore be injected into tests in 2 ways:
     1) via the constructor of this base class
     2) via the `add_aie_compilation_flags` method
     """
 
-    def __init__(self, run_on_target=["npu1_4col"], aie_compilation_flags=[]):
-        self.run_on_target = run_on_target
-        self.aie_compilation_flags = aie_compilation_flags
+    def __init__(self, run_on_target=["npu1_4col"], aie_compilation_flags=None):
+        self.run_on_target = [] if run_on_target is None else run_on_target
+        self.aie_compilation_flags = (
+            [] if aie_compilation_flags is None else aie_compilation_flags
+        )
+
         # NB: derived classes should add labels to this list in their
         # constructor, never overwrite it.
         self.labels = ["All"]
@@ -146,9 +149,9 @@ class BaseMatmul(BaseTest):
         K,
         input_type,
         acc_type,
-        tile_pipeline="pack-peel",
-        lower_to_aie_pipeline="objectFifo",
         use_ukernel=False,
+        lower_to_aie_pipeline="objectFifo",
+        tile_pipeline="pack-peel",
         n_repeats=1,
     ):
         super().__init__(run_on_target, aie_compilation_flags)
@@ -201,7 +204,7 @@ class MatmulFullBias(BaseMatmul):
     def __init__(self, M, N, K, input_type, acc_type, run_on_target=["npu1_4col"]):
         super().__init__(
             run_on_target=run_on_target,
-            aie_compilation_flags=[],
+            aie_compilation_flags=None,
             M=M,
             N=N,
             K=K,
@@ -236,8 +239,8 @@ class VanillaMatmul(BaseMatmul):
         name_suffix="",
         use_ukernel=False,
         run_on_target=["npu1_4col"],
-        additional_labels=[],
-        aie_compilation_flags=[],
+        additional_labels=None,
+        aie_compilation_flags=None,
     ):
         super().__init__(
             run_on_target=run_on_target,
@@ -248,7 +251,6 @@ class VanillaMatmul(BaseMatmul):
             input_type=input_type,
             acc_type=acc_type,
             tile_pipeline="pack-peel",
-            lower_to_aie_pipeline="objectFifo",
             use_ukernel=use_ukernel,
             n_repeats=1,
         )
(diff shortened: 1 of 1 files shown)


build_tools/ci/cpu_comparison/run.py
```diff
@@ -37,35 +38,48 @@ def run_conv_test(config, filename, n_repeats):
...
 
-    def __init__(self, run_on_target, additional_aie_compilation_flags=None):
+    def __init__(self, run_on_target=["npu1_4col"], aie_compilation_flags=[]):
         self.run_on_target = run_on_target
-        self.additional_aie_compilation_flags = additional_aie_compilation_flags \
-            if additional_aie_compilation_flags is not None else []
+        self.aie_compilation_flags = aie_compilation_flags
+        # NB: derived classes should add labels to this list in their
+        # constructor, never overwrite it.
+        self.labels = ["All"]
```
yzhang93, line 62: What is "All" used for in the base class? Why it's not empty?

newling: If tests are run with for example

```
./run.py ...   --target_device=npu1_4col --xrt_lite_n_core_rows=4 --xrt_lite_n_core_cols=4 --tests=VanillaMatmul
```

then all the tests which have label `VanillaMatmul` will be run. 

All the tests inherit from this base class, and so they all have the label `All`. 

So if you run 

```
./run.py ...   --target_device=npu1_4col --xrt_lite_n_core_rows=4 --xrt_lite_n_core_cols=4 --tests=VanillaMatmul
```
then all the tests will run. Well, there is additional check for tests which use microkernel that chess is available, but that's the basic idea

build_tools/ci/cpu_comparison/run.py
```diff
@@ -1180,8 +1221,8 @@ def all_tests(
     for test in tests.tests:
 
         # Determine if the test is a match for the test_set provided by caller
-        match = "All" in test_set
-        match = match or test.name in test_set
+        # match = "All" in test_set
```
newling, line 1224: @yzhang93 before there was a special case check for "All" here (which I meant to completely remove) but this PR removes the special case and makes it just another label, but one which all tests have. 

Maybe it was simpler before this change

build_tools/ci/cpu_comparison/run.py
newling: (tests can have multiple labels)

yzhang93: What is not clear to me is all other classes will append to the labels from the baseClass. So if I create an instance of VanillaMatmul class, it will run "All" and "VanillaMatmul"?

build_tools/ci/cpu_comparison/run.py
```diff
@@ -37,35 +38,48 @@ def run_conv_test(config, filename, n_repeats):
...
+    An instance of this class has a member `aie_compilation_flags`
+    which are additional flags to be passed to the AIE backend compiler.
+
+    Compilation flags can therefore be injected in 2 ways:
+    1) via the constructor of this base class
+    2) via the `add_aie_compilation_flags` method
     """
 
-    def __init__(self, run_on_target, additional_aie_compilation_flags=None):
+    def __init__(self, run_on_target=["npu1_4col"], aie_compilation_flags=[]):
```
jtuyls, line 57: Did you specifically change this on purpose from `None` to `[]`, and if so, why? In general, I try to avoid this because it can lead to quirky bahviour/issues: https://stackoverflow.com/questions/43255481/python-default-list-in-function

I saw your comment on the other PR (see below), but I am not sure what that has to do with this? Every object has it's own `self.aie_compilation_flags`?

> Just noticed: you are overriding the additional flags for every test which runs subsequent to the one that appends. I'll try and think of a way to add flags to a test in isolation

newling: >  config.add_additional_aie_compilation_flag(flag)

After this, the config has more flags for all subsequent tests. It's the appending to config from a specific test which looks bad. The config should only contain things which are common across all tests.

newling: > Did you specifically change this on purpose from None to [], and if so, why?

It was intentional to be consistent with run_on_target, but I didn't realize there were potential issues. I'll change it back to None

build_tools/ci/cpu_comparison/run.py
newling: If you create an instance of VanillaMatmul, it will have labels "All" and "VanillaMatmul". It will be run if the user specifies `tests=All` or `tests=VanillaMatmul`. At least that's what I've designed it to do.

yzhang93: So if I understand correctly, if I specify `tests=VanillaMatmul`, it will only run those tests within VanillaMatmul class, but it has the `labels = ["All", "VanillaMatmul"]`?

newling: https://github.com/nod-ai/iree-amd-aie/blob/4c3bbb37e3dc5173dfd10332c10ef94c663fd97e/build_tools/ci/cpu_comparison/run.py#L1186

> So if I understand correctly, if I specify tests=VanillaMatmul, it will only run those tests within VanillaMatmul class,

Yes

> but it has the labels = ["All", "VanillaMatmul"]

Yes, so all tests will run if you specify `tests=All`

build_tools/ci/cpu_comparison/run.py
```diff
@@ -309,56 +370,63 @@ def __init__(
...
         currently without enabling loop coalescing and unit dimension collapsing
-        we run out of program memory, this is under investigation.
+        we run out of program memory, this is under investigation. We also 
+        enable function outlining.
         """
+        self.aie_compilation_flags += [
+            "--iree-amdaie-enable-coalescing-loops",  # enable loop coalescing
+            "--iree-amdaie-enable-collapsing-unit-dims",  # enable unit dimension collapsing
+            "--iree-amdaie-enable-function-outlining",  # enable function outlining
+        ]
```
jtuyls, line 414: This is a new addition? Do we want to enable these flags on all tests?

jtuyls (approved):
LGTM

yzhang93 (approved):

PR #912, opened 2024-11-20
Fix bug in PushToQueue address generation

jtuyls:
The address generation for the `PushToQueue` op is incorrect for `column != 0` and `row == 0` as the condition `col && row` is not true. There is not need for this condition and the base tile address can always be computed as `((col & 0xff) << colShift) | ((row & 0xff) << rowShift)`, see:  https://github.com/Xilinx/aie-rt/blob/3fdf8d8d9a7a2abcd8df76155f950680607eac8a/driver/src/common/xaie_helper.h#L159

Note that we don't see this issue in main right now and this is prefetching an issue that's occurring on 4x4.


Abhishek-Varma (approved):
LGTM.

PR #913, opened 2024-11-20
[Utils] Finding ancestor in block 

newling:
Chipped off from review comment in https://github.com/nod-ai/iree-amd-aie/pull/904

#include changes: include the files that are used directly not indirectly


jtuyls (approved):
LGTM

PR #914, opened 2024-11-21
Split distribute-cores-and-objectfifos pass in 2

newling:
I know @Abhishek-Varma proposed a more dramatic version of this PR previously https://github.com/nod-ai/iree-amd-aie/pull/452 but I think that the place I'm choosing to split this pass in 2 in reasonable -- basically I'm just separating out the logic which assigns tiles to logicalobjectfifos from everything else i.e. the only thing that changes in the second pass is 

```
 %0 = amdaie.logicalobjectfifo.from_memref %alloc, {} : ... 
 ```
 
 becomes 
 
 ```
  %0 = amdaie.logicalobjectfifo.from_memref %alloc, {%tile} : ... 
 ```
 
 The code copied between .cpp files is completely unmodified, a 100% cut-paste. 
 
 I'd like to improve the tests in a second PR. 

I will polish this (add comments for the TODOs) if the proposal seems ok.


yzhang93:
I think @jtuyls was also working on tile assignments as a separate pass and already had a draft. We need something more sophisticated when assigning multiple memTiles and shimTiles.

jtuyls:
> I think @jtuyls was also working on tile assignments as a separate pass and already had a draft. We need something more sophisticated when assigning multiple memTiles and shimTiles.

Yes, I will be putting up a PR later today which has this, but with a bunch of changes in the tile assignment for 4x4.

newling:
Ok, will close this and take a look at https://github.com/nod-ai/iree-amd-aie/pull/915

PR #915, opened 2024-11-21
Add pass to assign tiles to logical objectFifos

jtuyls:
Adds a pass that assigns tiles to logical objectFifos based on tile usage by trying to to distribute logical objectFifos as evenly as possible across available tiles. The new pass is based on the tile assignment logic that was originally part of `DistributeCoresAndObjectFifos` and refactors the latter pass to use the extracted functions to keep the logic the same for now. 

Note that I am not completely removing the tile assignment functions from `DistributeCoresAndObjectFifos` for two reasons:
- To keep the tests largely the same for now.
- Because the output of `DistributeCoresAndObjectFifos` before tile assignment is not stable, i.e. calling cse/canonicalize in between `DistributeCoresAndObjectFifos` and `AssignTiles` leads to different functional (incorrect) results.

Until the output of `DistributeCoresAndObjectFifos` before tile assignment is stable, the tile assignment utility functions should be called within the pass itself and not be separated out.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -605,6 +605,10 @@ void addAMDAIEObjectFifoLoweringPasses(
   passManager.addPass(createCSEPass());
   passManager.addPass(createCanonicalizerPass());
 
+  passManager.addPass(createAMDAIEAssignTilesPass());
```
newling, line 608: Is this pass supposed to be added to the pipeline here? Thought you wanted to still do everything in `SplitLogicalObjFifosForConnectionReusePass`

newling:
If you want to semi-ensure that 2 passes are always run back-to-back, you can use the aievec approach of adding to the pipeline with a helper function: https://github.com/nod-ai/iree-amd-aie/blob/59d991ee1c63d89ca5a31fd18d10c7b74f938e62/compiler/plugins/target/AMD-AIE/aievec/Passes.h#L97 

(minus) doesn't technically ensure a user doesn't misuse the pass 
(plus) integrates all code into the main pipeline just once
(plus) easier for us to see what transformation happened in the different non-trivial functions with --mlir-print-ir-before-all

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
jtuyls: This pass is now run after `SplitLogicalObjFifosForConnectionReusePass`, so everything done in that pass (and the one from @yzhang93 we will enable soon) because those passes might cause us to want to assign new tile locations.

jtuyls:
> If you want to semi-ensure that 2 passes are always run back-to-back, you can use the aievec approach of adding to the pipeline with a helper function:
> 
> https://github.com/nod-ai/iree-amd-aie/blob/59d991ee1c63d89ca5a31fd18d10c7b74f938e62/compiler/plugins/target/AMD-AIE/aievec/Passes.h#L97
> 
> (minus) doesn't technically ensure a user doesn't misuse the pass (plus) integrates all code into the main pipeline just once (plus) easier for us to see what transformation happened in the different non-trivial functions with --mlir-print-ir-before-all

Yeah, but the disadvantage is a real issue for me. A pass not being stable on itself is just an invite for bugs and misuse. The workaround hides the real underlying problem that needs to be tackled, which is ensuring that the output without tile assignment is stable on itself. I have some ideas for this, but I don't think it's a priority right now because the pass has held up pretty well as it stands and if needed, you can easily debug it with `--debug-only`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
```diff
@@ -0,0 +1,437 @@
...
+      auto newTarget = rewriter.create<AMDAIE::LogicalObjectFifoFromMemrefOp>(
+          rewriter.getUnknownLoc(),
+          cast<LogicalObjectFifoType>(target.getOutput().getType()),
+          target.getMemref());
+      rewriter.replaceUsesWithIf(
+          target.getOutput(), newTarget.getOutput(), [&](OpOperand &use) {
+            return use.getOwner() == copyOp.getOperation();
+          });
+    }
+  });
```
yzhang93, line 119: These are almost duplicate codes as the above. Maybe extract these as a separate function.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
```diff
@@ -0,0 +1,437 @@
...
+            llvm::map_to_vector(logicalObjectFifo.getTiles(), [](Value tile) {
+              return dyn_cast_if_present<TileOp>(tile.getDefiningOp());
+            });
+        AMDAIE::TileOp assignedTileOp =
+            *std::min_element(tiles.begin(), tiles.end(), tileLocAndUsageCmp);
+
+        // Increase usage of the chosen tile.
+        int64_t col = getConstantIndexOrAssert(assignedTileOp.getCol());
+        int64_t row = getConstantIndexOrAssert(assignedTileOp.getRow());
+        tileLocToUsage[TileLoc(col, row)] += 1;
```
yzhang93, line 353: Can you add more comments to explain these?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
newling: Makes sense

compiler/plugins/target/AMD-AIE/iree-amd-aie/Test/samples/matmul_pack_peel_objectfifo.mlir
```diff
@@ -20,7 +20,7 @@
 // CHECK:         aie.use_lock
 // Check a bit of the aiex.runtime_sequence:
 // CHECK:       aiex.runtime_sequence @matmul_i32()
-// CHECK:       } {npu_instructions = dense_resource<npu_instructions> : tensor<174xui32>, runtime_sequence_name = "matmul_i32"}
+// CHECK:       } {npu_instructions = dense_resource<npu_instructions> : tensor<208xui32>, runtime_sequence_name = "matmul_i32"}
```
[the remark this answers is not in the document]
jtuyls: This is the controlcode instructions tensor. It was a bit weird that this changes, but I had a look at it and it seems like earlier channels were shared when they shouldn't have been and now they are not anymore, leading to more shim channels being used and therefore, more control code.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -67,6 +67,11 @@ def AMDAIEAssignPacketIds :
   let constructor = "mlir::iree_compiler::AMDAIE::createAMDAIEAssignPacketIdsPass()";
 }
 
+def AMDAIEAssignTiles : Pass<"iree-amdaie-assign-tiles", ""> {
+  let summary = "Assign physical tile locations to logical objFifos.";
```
newling, line 71:
```suggestion
  let summary = "Assign physical tile locations to logical objFifos. Existing assignments will be ignored/replaced.";
```
maybe?

newling (commented):
Looks good, no blocking comments from me

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
```diff
@@ -0,0 +1,437 @@
...
+  LLVM_DEBUG(llvm::dbgs() << "After assignLocalTiles: \n" << *parentOp << "\n");
+
+  // Duplicate global objFifos for each strided copy-like operation user to
+  // allow global logical objectFifos to be assigned to different tile
+  // locations.
+  if (failed(duplicateGlobalObjFifos(rewriter, parentOp))) {
+    parentOp->emitOpError() << "failed duplicating global object fifos";
+    return signalPassFailure();
+  }
+  if (failed(verify(parentOp, true))) {
```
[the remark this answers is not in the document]
jtuyls: > Question 1: The additional prints and verifications: I know they're copied from another pass, but do you think maybe they're not needed here? Because they're not in other passes. They suggest to me that this pass is still not to be considered stable, is that accurate?

Yeah, good point, I think this part is actually stable and they can be removed.

> Question 2: I assume duplicateGlobalObjFifos is something you explicitly do not want to use in DistributeCoresAndObjectFifos` ? If that's not the case, you could have moved this copied code into another util function.

The reason for this was mainly to keep the behaviour of `DistributeCoresAndObjectFifos` as close as possible to what it was to avoid more test changes. Maybe this could be tried and changed in a follow-up.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
jtuyls: Done

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeCoresAndObjectFifos.cpp
```diff
@@ -694,6 +441,18 @@ class AMDAIEDistributeCoresAndObjectFifosPass
...
   MLIRContext *context = &getContext();
   ModuleOp moduleOp = getOperation();
+  IRRewriter rewriter(moduleOp.getContext());
+  auto targetAttr = IREE::HAL::ExecutableTargetAttr::lookup(moduleOp);
+  std::optional<AMDAIEDevice> maybeDevice = getConfigAMDAIEDevice(targetAttr);
+  if (!maybeDevice) {
+    moduleOp->emitOpError()
+        << "has no AMDAIEDevice in the target attribute configuration. This "
+           "device-specific information is required to determine when loops "
+           "can be subsumed into DMA operations, and must be attached to a "
```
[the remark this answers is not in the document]
jtuyls: Thanks for the catch, updated it,

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
```diff
@@ -0,0 +1,437 @@
...
+    auto createTileLocations =
+        [&](SmallVector<AMDAIE::TileOp, 16> &tiles) -> LogicalResult {
+      // For deterministic and canonical output, sort on column index and erase
+      // duplicates.
+      std::sort(tiles.begin(), tiles.end(),
+                AMDAIE::TileOp::tileColumnComparator);
+      tiles.erase(std::unique(tiles.begin(), tiles.end()), tiles.end());
+      for (AMDAIE::TileOp tile : tiles) {
+        std::optional<int64_t> column = getConstantIntValue(tile.getCol());
+        if (!column) return tile.emitOpError() << "found non-constant column";
```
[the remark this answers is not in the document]
jtuyls: Yeah, I changed this to `notifyMatchFailure`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
jtuyls: Thanks, updated it.

yzhang93 (approved):

PR #918, opened 2024-11-21
[DistributeL1Allocations] Fix logic for conv and correctly propagate errors

newling:
There was what appeared to be a false positive error in CI. i.e. an op emits an error somewhere, but the end-to-end numerical test still passes:

{7764D0FA-0A0A-44F6-8247-7F984C7C5EC5}

I tracked this down and fixed it. 2 things: 

1) error wasn't being propagated all the way to the signal pass failure
2) the convolution tiling doesn't use the approach where L1 result tensors are initially of shape <4x4x*> and then distributed to tensors of shape <1x1x*> but the logic wasn't quite right and it was trying to do this for conv. 

1 was hiding 2, causing default behaviour which just happened to work.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeL1Allocations.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeL1Allocations.cpp
@@ -57,7 +57,7 @@ MemRefType getDistributedType(memref::AllocOp alloc,
       // that if a subview has an offset which is not a constant and not a
       // thread id, it's not 'distributing'.
       Operation::operand_range offsets = subview.getOffsets();
-      int nIndVars {0};
+      int nIndVars{0};
       for (Value offset : offsets) {
         bool isConst = matchPattern(offset, m_Constant());
         bool isIndVar = llvm::is_contained(indVars, offset);
@@ -99,8 +99,6 @@ SmallVector<Value> substitute(Container toUpdate,
 /// smaller memory. This is ultimately needed because cores can't operate on
 /// one shared L1 memory.
 LogicalResult distributeLocalMemory(ModuleOp moduleOp) {
-
-
   FailureOr<DenseSet<Value>> maybeIndVars = getThreadIndVars(moduleOp);
   if (failed(maybeIndVars)) return failure();
   const DenseSet<Value> &indVars = maybeIndVars.value();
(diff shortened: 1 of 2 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeL1Allocations.cpp
```diff
@@ -91,11 +99,14 @@ SmallVector<Value> substitute(Container toUpdate,
 /// smaller memory. This is ultimately needed because cores can't operate on
 /// one shared L1 memory.
 LogicalResult distributeLocalMemory(ModuleOp moduleOp) {
+
+
```
yzhang93, line 103: Remove these empty lines.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/distribute_l1_allocations.mlir
```diff
@@ -90,4 +90,27 @@ func.func @non_distributing_subview(%index : index) {
...
 
+// -----
+
+// Example where the subview type is a complete view of the alloc: unchanged IR
+
+// CHECK-LABEL: @complete_view_subview
+// CHECK: %[[ALLOC:.+]] = memref.alloc() : memref<4xbf16, 2>
+// CHECK: scf.forall (%[[ARG0:.+]]) in (4) {
+// CHECK: %[[C0_BF16:.+]] = arith.constant
+// CHECK: %[[SUBVIEW:.+]] = memref.subview
```
yzhang93, line 101: I think you can remove these CHECKS?

yzhang93 (commented):
Nice catch! I have seen this error when I run tests and thought it's just some kind of warning :)

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/distribute_l1_allocations.mlir
newling: I wanted to check that the IR was unchanged (i.e. no new created). Is there a problem with including these CHECKS? I changed them to to CHECK-NEXTs to be absolutely sure no new allocs are created.

yzhang93: I was thinking checking the alloc and subview shapes are enough to proof the IR is unchanged. There seems no need to check scf.for ops and arith.constant/

yzhang93: Okay, I don't want to block on this, but I think it's better to check the alloc shape as what you did in the first version. You can use CHECK-NEXT as is in between.

```
// CHECK: %[[ALLOC:.+]] = memref.alloc() : memref<4xbf16, 2>
...
// CHECK-SAME: %[[ALLOC]][0] [4] [1] : memref<4xbf16, 2> to memref<4xbf16, 2>
```

yzhang93 (approved):

PR #923, opened 2024-11-22
[CI] Refactor tests with chess 

newling:
- Make flag usage consistent (use_this_format instead of this-format uniformly)
- Simplify run_matmul_tests.sh -- remove all peano/vitis/ukernel path testing, move all these tests to run.py 
- Make the use_chess flag test specific, rather that a global config flag 
- Change `VanillaMatmul` label to just `Matmul`

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -40,21 +40,31 @@ def run_conv_test(config, aie_compilation_flags, filename, n_repeats):
 
 class BaseTest(ABC):
     """
-    Base class to be inherited by all tests. The derived instances will be
-    created specifying the intended target device(s) they're to be run on; and
-    will accordingly be `run` if the intended target device(s) contains
-    the `target_device` found in `config`. The default set of targets to run on
-    is the singleton set `["npu1_4col"]`.
+    Base class to be inherited by all tests.
 
-    An instance of this class has a member `aie_compilation_flags`
-    which are additional flags to be passed to the AIE backend compiler.
+    Args:
+       run_on_target: The derived instances will be created specifying the
+          intended target device(s) they're to be run on; and will accordingly
+          be `run` if the intended target device(s) contains the `target_device`
+          found in `config`. The default set of targets to run on is the
+          singleton set `["npu1_4col"]`.
 
-    Compilation flags can therefore be injected into tests in 2 ways:
-    1) via the constructor of this base class
-    2) via the `add_aie_compilation_flags` method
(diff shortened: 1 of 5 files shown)


build_tools/ci/cpu_comparison/run.py
```diff
@@ -78,9 +95,29 @@ def add_aie_compilation_flags(self, flags):
...
-        return False
+
+        # If the target device is not in the set of devices to run on, then
+        # return False. ie. don't raise an error because is legitimate,
+        # we just won't run the test.
+        if config.target_device not in self.run_on_target:
+            return False
+
+        # If use_chess=1, and config has not provided a valid
+        # path to vitis, then don't run the test. The asymettry between
```
yzhang93, line 106: Typo: asymettry -> asymmetry

build_tools/ci/cpu_comparison/run.py
```diff
@@ -1122,6 +1182,68 @@ def __init__(self):
                 )
             )
 
+        self.register(
```
yzhang93, line 1185: Add a comment above to show what this test is for.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -1122,6 +1182,68 @@ def __init__(self):
...
+                32,
+                "i32",
+                "i32",
+                name_suffix="chess",
+                use_chess=True,
+                n_repeats=10,
+            )
+        )
+
+        self.register(
```
yzhang93, line 1233: Add comments for these tests as well. It will be easier for users to quickly get the purpose of the test.

yzhang93 (commented):
LGTM. Just some minor comments.

build_tools/ci/cpu_comparison/run.py
newling: In future can you please make comments like this using the "Add suggestion" button? That way I can just land time directly in the web browser

{5DFBAE28-FC09-4F07-ABE7-405C75D751D9}

build_tools/ci/cpu_comparison/run.py
newling: I don't know what this test is for, there was no comment in run_matmul_test.sh

build_tools/ci/cpu_comparison/run.py
newling: Can you suggest a comment, please?

yzhang93: // ukernel test for chess

PR #924, opened 2024-11-22
Bump IREE to 7850ea99eebadf84e91963da12a49236fdd613f5

yzhang93:


newling:
These tests were added in https://github.com/nod-ai/iree-amd-aie/pull/795

newling:
Probably this commit:

https://github.com/llvm/llvm-project/commit/3ad0148020ca91cc288bffd8ad36e25f7555a3bb#diff-72ebcdb4d3db4feaffadbb116379626ea14f4383014ef24792e2b6c37765143d

Annoyingly there isn't a single example of using linalg.matmul in any python tests in MLIR upstream. 

Makes me wonder if these tests will easy to maintain going forward, or if we should disable this directory for now and reconsider it at a later point.

yzhang93:
> Probably this commit:
> 
> [llvm/llvm-project@3ad0148#diff-72ebcdb4d3db4feaffadbb116379626ea14f4383014ef24792e2b6c37765143d](https://github.com/llvm/llvm-project/commit/3ad0148020ca91cc288bffd8ad36e25f7555a3bb#diff-72ebcdb4d3db4feaffadbb116379626ea14f4383014ef24792e2b6c37765143d)
> 
> Annoyingly there isn't a single example of using linalg.matmul in any python tests in MLIR upstream.
> 
> Makes me wonder if these tests will easy to maintain going forward, or if we should disable this directory for now and reconsider it at a later point.

Just saw your comment after sending mine :) yeah, I think we should probably want to disable the tests unless @makslevental has a better idea to refactor these tests.

newling (approved):

PR #930, opened 2024-11-25
[AssignPacketIds] Base packet ID assignment on source channels

jtuyls:
Updates the packet ID assignment pass to perform assignment of packet IDs based on the source channels of the flow ops. I.e. `amdaie.flow` ops with the same source channel will get a different packet IDs assigned to accommodate multiple data packets being routed through the same ports.


yzhang93 (approved):
LGTM. Thanks for the fix!

newling:
@jtuyls looks good. I made a PR with what I think are small improvements on this

PR #943, opened 2024-11-28
[AssignTiles] Generalize pass to use LogicalObjFifoInterface and CopyOpInterface

jtuyls:
Generalize `AMDAIEAssignTiles` to use `LogicalObjFifoInterface` and `CopyOpInterface` where possible. This allows this pass to be moved around more within the pass pipeline.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
@@ -1143,7 +1143,8 @@ def AMDAIE_LogicalObjectFifoAcquire:
 
 def AMDAIE_LogicalObjectFifoFromBuffersOp
     : AMDAIE_Op<"logicalobjectfifo.from_buffers", 
-    [LogicalObjFifoOpInterface, Pure, AttrSizedOperandSegments]> {
+    [DeclareOpInterfaceMethods<LogicalObjFifoOpInterface, ["replaceWithNewTiles"]>,
+     Pure, AttrSizedOperandSegments]> {
   let summary = "Create a logical objectFifo from a set of buffers";
   let description = [{
     Creates a logical objectFifo which encapsulates a set of memref `buffers`.
@@ -1220,7 +1221,8 @@ def AMDAIE_LogicalObjectFifoFromBuffersOp
 
 def AMDAIE_LogicalObjectFifoFromMemrefOp
     : AMDAIE_Op<"logicalobjectfifo.from_memref", 
-    [LogicalObjFifoOpInterface, Pure]> {
+    [DeclareOpInterfaceMethods<LogicalObjFifoOpInterface, ["replaceWithNewTiles"]>, 
+     Pure]> {
   let summary = "Create a logical objectFifo from a memref";
   let description = [{
     Creates a logical objectFifo which encapsulates a memref. The logical objectFifo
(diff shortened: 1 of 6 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
```diff
@@ -1212,6 +1212,11 @@ def AMDAIE_LogicalObjectFifoFromBuffersOp
 
     // Return the locks on the producer side (write) on the requested tile.
     llvm::SmallVector<LockOp> getProducerLocksOnTile(TileOp tileOp);
+
+    // A utility to replace this logical objectFifo operation with a new one 
+    // with new tiles.
+    FailureOr<LogicalObjFifoOpInterface> replaceWithNewTiles(
+      ::mlir::RewriterBase &rewriter, ::mlir::ValueRange tiles);
```
Abhishek-Varma, line 1219: If it's already declared in the Interface `LogicalObjFifoOpInterface`, wouldn't these already be part of the ops inheriting them?

What I mean to ask is if the change in `AMDAIEOps.td` is necessary.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
```diff
@@ -73,15 +77,18 @@ LogicalResult findUsersInCoreAndAddTiles(
...
     if (objFifo.getMemorySpaceAsUInt() != 2) {
       rewriter.setInsertionPoint(objFifo);
       SmallVector<Value> tiles;
-      rewriter.replaceOpWithNewOp<AMDAIE::LogicalObjectFifoFromMemrefOp>(
-          objFifo, cast<LogicalObjectFifoType>(objFifo.getOutput().getType()),
-          objFifo.getMemref(), tiles);
+      if (failed(objFifo.replaceWithNewTiles(rewriter, tiles))) {
+        objFifo.emitOpError() << "could not replace its tiles.";
+        return WalkResult::interrupt();
+      }
```
Abhishek-Varma, line 87: Is there any scenario where we might hit a `LogicalObjectFifoFromBuffersOp` in this flow or the one below ? If yes, a lit test might be warranted.
Else we could just do a `(void)(objFifo.replaceWithNewTiles(rewriter, tiles))`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
```diff
@@ -373,7 +396,7 @@ class AMDAIEAssignTilesPass
     : public impl::AMDAIEAssignTilesBase<AMDAIEAssignTilesPass> {
  public:
   void getDependentDialects(DialectRegistry &registry) const override {
-    registry.insert<AMDAIEDialect>();
+    registry.insert<AMDAIEDialect, arith::ArithDialect>();
```
Abhishek-Varma, line 399: Some test must've triggered this addition as it wasn't caught earlier? If yes, then it'd be nice to add a short lit test for this as well.

Abhishek-Varma (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
jtuyls: No, this isn't triggered and I thought this was needed, but I can remove this as well and it works, which is unclear to me.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
jtuyls: The methods need to be declared in the ops that implement the interface, but I was able to simplify it with: `DeclareOpInterfaceMethods<LogicalObjFifoOpInterface, ["replaceWithNewTiles"]>`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
jtuyls: Good point. I added a lit test for `logicalobjectfifo.from_buffers` and `logicalobjectfifo.placeholder`.

Abhishek-Varma (approved):
LGTM % two change requests.

PR #944, opened 2024-11-28
[FuseConsumer] Update consumer fusion pass to be more generic

Abhishek-Varma:
-- This commit updates consumer fusion pass to be more generic whilst trying to pick the compute op/loop to fuse with.
-- The idea is to hook on to the innermost SCF loop -> find a compute op within it -> try to aggressively fuse the consumer.
-- As a result it removes `useSCFFor` pass flag.
-- Helps in [increasing L2 tile size](https://github.com/nod-ai/iree-amd-aie/tree/increase_L2_tile_sizes) task.

Before : [IR dump](https://gist.github.com/Abhishek-Varma/b1f7b31d8d74c45da49a7108dafb5a27) - (refer [this e2e IR chain](https://gist.github.com/yzhang93/aa66bddeee0f229d436ba942b30f6182))
After this fix : [IR dump](https://gist.github.com/Abhishek-Varma/a45d8d595e39a1b3d5aa9d1ffe91e8e2)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseConsumerIntoLoop.cpp
```diff
@@ -49,72 +46,74 @@ void AMDAIEFuseConsumerIntoLoopPass::runOnOperation() {
...
+  // Step 4. Based on the `fuseDepth`, we would greedily fuse the consumer ops.
+  for (unsigned depth = 1; depth <= fuseDepth; depth++) {
+    do {
+      Value::user_range users = computeOp->getResult(0).getUsers();
+      if (!llvm::hasSingleElement(users)) {
+        computeOp->emitOpError("Expected only one user of the compute op");
+        return signalPassFailure();
+      }
+
+      Operation *terminatorStoreOp = *(users.begin());
```
jtuyls, line 90: Why is this called `teminatorStoreOp`? What about just `userOp`?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseConsumerIntoLoop.cpp
```diff
@@ -49,72 +46,74 @@ void AMDAIEFuseConsumerIntoLoopPass::runOnOperation() {
...
-    // Search the compute op and its consumer slices.
-    linalg::LinalgOp linalgOp;
-    scfLoopOp->walk<WalkOrder::PostOrder, ReverseIterator>(
-        [&](linalg::LinalgOp op) {
-          linalgOp = op;
+  // Step 2. Find the first scf loop in postorder walk.
+  Operation *scfLoopOp = nullptr;
+  funcOp->walk<WalkOrder::PostOrder, ReverseIterator>(
+      [&](LoopLikeOpInterface op) {
+        if (isa<scf::ForOp>(op)) {
```
jtuyls, line 53:
```suggestion
        if (isa<scf::ForOp, scf::ForallOp>(op)) {
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseConsumerIntoLoop.cpp
```diff
@@ -49,72 +46,74 @@ void AMDAIEFuseConsumerIntoLoopPass::runOnOperation() {
...
-
-    Value::user_range users = linalgOp->getResult(0).getUsers();
-    if (!llvm::hasSingleElement(users)) {
-      linalgOp->emitOpError("Expected only one user of the compute op");
-      return signalPassFailure();
-    }
+        } else if (isa<scf::ForallOp>(op)) {
+          scfLoopOp = op;
+          return WalkResult::interrupt();
+        }
```
jtuyls, line 59:
```suggestion
        }
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/fuse_consumer_into_loop.mlir
```diff
@@ -0,0 +1,468 @@
...
+      %3 = tensor.empty() : tensor<128x128xi32>
+      %unpack = tensor.unpack %1 outer_dims_perm = [0, 1, 3, 2] inner_dims_pos = [2, 3] inner_tiles = [4, 4] into %2 : tensor<4x4x8x8x4x4xi32> -> tensor<4x4x32x32xi32>
+      %unpack_0 = tensor.unpack %unpack inner_dims_pos = [0, 1] inner_tiles = [32, 32] into %3 : tensor<4x4x32x32xi32> -> tensor<128x128xi32>
+      scf.forall.in_parallel {
+        tensor.parallel_insert_slice %unpack_0 into %arg6[0, 0] [128, 128] [1, 1] : tensor<128x128xi32> into tensor<128x128xi32>
+      }
+    }
+    return %0 : tensor<128x128xi32>
+  }
+}
```
jtuyls, line 468:
```suggestion
}

```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseConsumerIntoLoop.cpp
```diff
@@ -49,72 +46,74 @@ void AMDAIEFuseConsumerIntoLoopPass::runOnOperation() {
...
-      return signalPassFailure();
-    }
-    fusedConsumer->origConsumerOperand->getOwner()->erase();
+  Operation *computeOp = linalgOp;
+  // Step 4. Based on the `fuseDepth`, we would greedily fuse the consumer ops.
+  for (unsigned depth = 1; depth <= fuseDepth; depth++) {
+    do {
+      Value::user_range users = computeOp->getResult(0).getUsers();
+      if (!llvm::hasSingleElement(users)) {
+        computeOp->emitOpError("Expected only one user of the compute op");
```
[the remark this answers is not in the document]
Abhishek-Varma: I guess for both pure Matmul as well as Matmul + Elemwise, we always expect this pass to succeed.

A silent bailing on an inherent (and strong) assumption that this pass makes might not be a good idea - I believe all kind of input dispatches generated via IREE would maintain the "one user of the compute op" and as such it might be worth it to raise an error if any of the preceding transformations we do in our pipeline changes that in any way.

I don't mind deleting this (and the corresponding lit tests) - let me know your thoughts.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseConsumerIntoLoop.cpp
Abhishek-Varma: The upstream consumer fusion API `scf::tileAndFuseConsumerOfSlice` takes in either `tensor.insert_slice` or `tensor.parallel_insert_slice` - both of these are actually expected to be the final operation in the `scf.for*` op (it in turn verifies these claims).

So, the naming of this variable as `teminatorStoreOp` is a preferential choice - I believe it helps to know what is the nature of this user op.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseConsumerIntoLoop.cpp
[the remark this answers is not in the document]
jtuyls: > A silent bailing on an inherent (and strong) assumption that this pass makes might not be a good idea

What's the inherent strong assumption? If we bail here, we're just not doing any fusion, right? With this error you're basically throwing an error on a potentially correct input we haven't seen before, assuming it will be incorrect, but which will not break this pass afaict, so I don't think this is the right place to throw an error.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseConsumerIntoLoop.cpp
jtuyls: Are `tensor::InsertSliceOp, tensor::ParallelInsertSliceOp` terminator store operations? The `scf:;tileAndFuseConsumerOfSlice` function calls them `candidateSliceOp` instead. I am nitting here but I don't find `terminatorStoreOp` a clear name while reading the code and I couldn't find it anywhere else in the code base. What about `candidateSliceOp` like the scf API?

jtuyls (approved):
LGTM

Abhishek-Varma:
Merging this in as this is anyway an improvement into the existing functionality of the pass.

PR #946, opened 2024-12-02
[AMDAIEAssignNpuDmaBdIds] Support both constant and semi-affine expressions for BD IDs

Yu-Zhewen:
- Motivated by DMA chaining (#931), this update aims to encourage the use of different BD IDs wherever feasible.
- When the DMA operation resides in the innermost loop of the control code, the BD ID will naturally form a semi-affine expression.
- After loop unrolling, these BD IDs will simplify to constant.
- The `value` operand for `BdIdOp` is changed from `UI32Attr` to `Index` to accommodate either a constant or expression. **_(question: Why use a separate `BdIdOp` instead of directly integrating as an operand into the DMA operation?)_**
- Add an `incremental` mode for `ChannelBdIdGenerator` that prioritizes increasing BD IDs during assignment.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/assign_npu_dma_bd_ids.mlir
```diff
@@ -269,7 +280,7 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
 #map = affine_map<(d0) -> (d0 * 16)>
 #executable_target_amdaie_xclbin_fb = #hal.executable.target<"amd-aie", "amdaie-xclbin-fb", {target_device = "npu1_4col", ukernels = "none"}>
 module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb} {
-  func.func @nested_loops(%arg0: memref<8x16xi32>, %arg1: memref<8x16xi32>, %arg2: memref<8x16xi32>, %arg3: memref<1x1x8x16xi32, 1>) {
+  func.func @nested_loops_1(%arg0: memref<8x16xi32>, %arg1: memref<8x16xi32>, %arg2: memref<8x16xi32>, %arg3: memref<1x1x8x16xi32, 1>) {
```
jtuyls, line 283: Instead of `_1`/`_2`, try to describe the difference between them instead.

runtime/src/iree-amd-aie/aie_runtime/Utils/test/ChannelBdIdGeneratorTest.cpp
```diff
@@ -102,10 +93,26 @@ TEST(ChannelBdIdGeneratorTest, Release) {
   EXPECT_EQ(generator.isBdIdAssigned(1), true);
 }
 
+TEST(ChannelBdIdGeneratorTest, IncrementalAssign) {
```
jtuyls, line 96: Add a couple of test as well for:
- Checking correctness when the max value is overflown (wrap).
- Checking correctness when all BDs are assigned/in use.

runtime/src/iree-amd-aie/aie_runtime/Utils/ChannelBdIdGenerator.h
```diff
@@ -41,11 +50,25 @@ class ChannelBdIdGenerator {
   /// reused.
   void releaseBdId(uint32_t bdId) { assignedBdIds.erase(bdId); }
 
+  uint32_t getAvailableBdIdNum(uint32_t channel) {
```
jtuyls, line 53: This function could use documentation.

jtuyls, line 53:
```suggestion
  uint32_t getNumAvailableBdIds(uint32_t channel) {
```
This is more consistent with naming in the codebase (`type.getNumElements`, `getNumArguments`, etc)

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -158,8 +262,18 @@ LogicalResult assignNpuDmaBdIds(AMDAIE::WorkgroupOp workgroupOp) {
...
             shimTileToGeneratorMap[tileOp.getResult()];
-        uint32_t value = bdIdOp.getValue();
-        generator.releaseBdId(value);
+        Value value = bdIdOp.getValue();
+        if (auto addOp = value.getDefiningOp<arith::AddIOp>()) {
+          // If the BD ID is a semi-affine expression, retrieve the BD ID for
+          // the first iteration.
+          FailureOr<uint32_t> bdId = retriveBdId(addOp);
+          if (failed(bdId)) return WalkResult::interrupt();
+          generator.releaseBdId(*bdId);
```
jtuyls, line 271: Shouldn't all of them be released?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -9,14 +9,165 @@
...
+      getGeneratorTileOp(npuDmaOp, shimTileToGeneratorMap);
+  if (failed(tileOp)) return failure();
+
+  ChannelBdIdGenerator &generator = shimTileToGeneratorMap[tileOp->getResult()];
+  std::optional<uint32_t> bdId =
+      generator.getAndAssignBdId(channel, BdIdAssignmentMode::Incremental);
+  if (!bdId) return failure();
+  AMDAIE::BdIdOp bdIdOp;
+  rewriter.setInsertionPoint(npuDmaOp);
+  if (isInMostInnerLoop(npuDmaOp)) {
```
jtuyls, line 105: Why do you need to check whether this is the innermost loop? Shouldn't this always work?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -9,14 +9,165 @@
...
+    // If the DMA is in the innermost loop, assign a BD ID using the
+    // semi-affine expression:
+    // `(iv * step + bdId - offset) % size + offset`,
+    // `step` represents the number of BD IDs needed per loop iteration,
+    // `bdId` is the BD ID assigned by the generator,
+    // `offset` is the BD ID assigned to the first DMA in the loop,
+    // `size` is the number of BD IDs available.
```
jtuyls, line 112: I don't think I am fully understanding this expression. 

Assuming the loop is normalized (we can normalize and ensure this before), I would have expected:
```
size = max(<available_incremental_bd_ids>, <loop_extent>)
bd_id = offset + (iv % size)
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -9,14 +9,165 @@
...
+        rewriter.create<arith::AddIOp>(rewriter.getUnknownLoc(), mul,
+                                       diff);  // iv * step + bdId - offset
+    auto mod = rewriter.create<arith::RemUIOp>(
+        rewriter.getUnknownLoc(), add1,
+        size);  // (iv * step + bdId - offset) % size
+    auto add2 = rewriter.create<arith::AddIOp>(
+        rewriter.getUnknownLoc(), mod,
+        offset);  // (iv * step + bdId - offset) % size + offset
+    bdIdOp = rewriter.create<AMDAIE::BdIdOp>(rewriter.getUnknownLoc(), *tileOp,
+                                             add2.getResult());
```
jtuyls, line 146: You could use an `AffineMap` instead.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -9,14 +9,165 @@
...
+    // If the DMA is not in the innermost loop, assign a constant BD ID
+    auto constant = rewriter.create<arith::ConstantOp>(
+        rewriter.getUnknownLoc(), rewriter.getIndexAttr(bdId.value()));
+    bdIdOp = rewriter.create<AMDAIE::BdIdOp>(rewriter.getUnknownLoc(), *tileOp,
+                                             constant.getResult());
+  }
+  return bdIdOp;
+};
+
+FailureOr<uint32_t> retriveBdId(arith::AddIOp add2) {
```
jtuyls, line 157:
```suggestion
FailureOr<uint32_t> retrieveBdId(arith::AddIOp add2) {
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -9,14 +9,165 @@
...
+    tiles = logicalObjFifo.getTiles();
+  }
+  if (tiles.size() != 1) {
+    if (tiles.empty()) {
+      return npuDmaOp.emitOpError() << "no tiles found";
+    } else {
+      return npuDmaOp.emitOpError()
+             << "operating on multiple tiles is not supported";
+    }
+  }
```
jtuyls, line 51:
```suggestion
    return npuDmaOp.emitOpError()
             << "expected to operate on a singe tile, but found: " << tiles.size();
  }
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -9,14 +9,165 @@
...
+  SmallVector<Value> tiles;
+  if (npuDmaOp.getSource()) {
+    auto logicalObjFifo =
+        dyn_cast_if_present<AMDAIE::LogicalObjFifoOpInterface>(
+            npuDmaOp.getSource().getDefiningOp());
+    if (!logicalObjFifo)
+      return npuDmaOp.emitOpError() << "expected a source logical objectFifo";
+    tiles = logicalObjFifo.getTiles();
+  }
+  if (npuDmaOp.getTarget()) {
```
jtuyls, line 35: I think this doesn't result in exactly the same logic as before:

Before:

````
if (npuDmaOp.getSource()) {
  tiles = npuDmaOp.getSource()...getTiles();
  getGeneratorTileOp(tiles, ...);
}
if (npuDmaOp.getTarget()) {
  tiles = npuDmaOp.getSource()...getTiles();
  getGeneratorTileOp(tiles, ...);
}
```

Here, both source and target could be set and handled, while after this change:

````
getGeneratorTileOp() {
  tiles;
  if (npuDmaOp.getSource()) {
     tiles = npuDmaOp.getSource()...getTiles();
   }
   if (npuDmaOp.getTarget()) {
      tiles = npuDmaOp.getTarget()...getTiles();
   }
}
```

So, if target is set, tiles will always be overridden with the target tiles, so a potential `NpuDmaCpyNdOp` with two sides isn't supported anymore (we will need that in the future).

You could solve this for example by using `template <CopyOpOperateOn OperateOn>` on this function.

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
Yu-Zhewen: that's true if there is only one DMA copy inside the loop, where in my expression `step = 1` and `bdId-offset=0`. Otherwise, in a different example `step=2`, `offset=1`, `size=16-1=15`

```ll
dma_cpy_nd(bd_id = 0) // generator assign bd_id 0, so `offset=1`
scf.for iv {
   dma_cpy_nd(bd_id = (2 * iv + 1 - 1) % 15 + 1) // generator assign bd_id 1 for the first iteration, then it is converted to expression
   dma_cpy_nd(bd_id = (2 * iv + 2 - 1) % 15 + 1)   // generator assign bd_id 2 for the first iteration, then it is converted to expression
}
```

and I can't move `+-offset` inside or outside the bracket due to the modulo operation

jtuyls: Ok, I see, that's more general towards multiple `dma_cpy_nd`, cool!

A few follow-up question: where do you ensure that all these BDs being used in the loop are actually assigned?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
Yu-Zhewen: Assume the number of DMAs per iteration of the loop is even, while the total number of available BD IDs is odd. In this case, the ownership of BD IDs might get interleaved. Since the logic here was assigning on DMA copy and releasing on DMA wait, it can't handle this "complex" case. As a hacky approach, I only assign / release BD ID for the first iteration. It's probably fine if it is the innermost loop, but I will think about a better solution.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
Yu-Zhewen: As explained before

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
jtuyls: > Assume the number of DMAs per iteration of the loop is even, while the total number of available BD IDs is odd. In this case, the ownership of BD IDs might get interleaved.

Could you show this case with an example? I would expect every `dma_cpy_nd` to have a fixed range of BD ids assigned for the loop iterations, so if `nb_dma_cpy_nd_ops * nb_bd_ids_per_dma_cpy_nd < nb_available_bd_ids`, there shouldn't be an issue?

I think annotating the full range of BD ids as assigned should be a safe thing to do, even with nested inner loops:

```
for iv = 0 to 4 step 1 {
  %0 = dma_cpy_nd(bd_id = iv)  // simplified
  %1 = dma_cpy_nd(bd_id = 4 +  iv)  // simplified
  for iv2 = 0 to 2 step 1 {
    %2 = dma_cpy_nd(bd_id = 8 +  iv2)  // simplified
    dma_wait(%2)
  }
  dma_wait(%0)
  dma_wait(%1)
}
```

Here, the most nested loop can't use BD ranges [0, 4[ and [4, 8[ because they are already used.

As the wait ops release the full range of BD ids, this works as well with:

```
for iv = 0 to 4 step 1 {
  %0 = dma_cpy_nd(bd_id = iv)  // simplified
  %1 = dma_cpy_nd(bd_id = 4 +  iv)  // simplified
  dma_wait(%0)
  dma_wait(%1)
  for iv2 = 0 to 2 step 1 {
    %2 = dma_cpy_nd(bd_id = iv2)  // simplified
    dma_wait(%2)
  }
}
```
Here, the BD ranges [0, 4[ and [4, 8[ are released before the inner loop, so they can be reused.

Yu-Zhewen: I think there are currently two ways:

The main difference might be when number of available BD is not divisible by number of DMA. For Method I, I guess the intention is also to leave some BD never utilized. For example, if the allowed BD range is [0,8], 8 is never used. I am using Method II for now, but the issue is this statement is no longer true.... `I would expect every dma_cpy_nd to have a fixed range of BD ids assigned for the loop iterations,`

Method I:
```
for iv = 0 to 64 step 1 {
  %0 = dma_cpy_nd(bd_id = iv mod 8)            // get 0, 1, 2, 3,   0, 1, 2, 3
  %1 = dma_cpy_nd(bd_id = (4 +  iv) mod 8)     // get 4, 5, 6, 7,    4, 5, 6, 7
}
```

Method II:
```
for iv = 0 to 64 step 1 {
  %0 = dma_cpy_nd(bd_id = (2*iv) mod 9)      // get 0, 2, 4, 6, 8,    1, 3, 5, 7 ....
  %1 = dma_cpy_nd(bd_id = (2*iv+1) mod 9)    // get 1, 3, 5, 7, 0,    2, 4, 6, 8 ....
}
```

Yu-Zhewen: Now I am thinking Method I is probably better, given the cost is small (losing 1 BD in this case)

jtuyls: > I guess the intention is also to leave some BD never utilized. For example, if the allowed BD range is [0,8], 8 is never used. 

That's true in this example, but it doesn't have a performance impact and it might be useful to have this additional BD in other cases:

```
%0 = dma_cpy_nd(bd_id = 0)  
for iv = 0 to 64 step 1 {
  %1 = dma_cpy_nd(bd_id = 1 + iv mod 8)            // get 1, 2, 3, 4,   1, 2, 3, 4
  %2 = dma_cpy_nd(bd_id = 1+  (4 +  iv) mod 8)     // get 5, 6, 7, 8,   5, 6, 7, 8
}
```

```
for iv = 0 to 64 step 1 {
  %0 = dma_cpy_nd(bd_id = iv mod 8)            // get 0, 1, 2, 3,   0, 1, 2, 3
  %1 = dma_cpy_nd(bd_id = (4 +  iv) mod 8)     // get 4, 5, 6, 7,    4, 5, 6, 7
  %2 = dma_cpy_nd(bd_id = 8)                        // always 8, even though we can't use different BD ids and we won't be able to chain, it's still a valid solution
}
```

Yu-Zhewen: And for the nested loop, I haven't seen any actual example has this pattern
```ll
for iv = 0 to 4 step 1 {
  %0 = dma_cpy_nd(bd_id = iv)  // simplified
  %1 = dma_cpy_nd(bd_id = 4 +  iv)  // simplified
  dma_wait(%0)
  dma_wait(%1)
  for iv2 = 0 to 2 step 1 {
    %2 = dma_cpy_nd(bd_id = iv2)  // simplified
    dma_wait(%2)
  }
}
```

Instead, this happens a lot
```ll
for iv = 0 to 4 step 1 {
  %0 = dma_cpy_nd(bd_id = iv)  // simplified
  %1 = dma_cpy_nd(bd_id = 4 +  iv)  // simplified
  for iv2 = 0 to 2 step 1 {
    %2 = dma_cpy_nd(bd_id = iv2)  // simplified
    dma_wait(%2)
  }
  dma_wait(%0)
  dma_wait(%1)
}
```
Theoretically, for the second case, we can still have the semi-affine expression for the outer loop, as long as (inner loop iteration * inner loop per iteration DMA) < available BD. 

Given that, maybe I can stay with the constraint about InnerMostLoop, and add a TODO there?

jtuyls: Maybe we haven't seen the first example

````

> And for the nested loop, I haven't seen any actual example has this pattern
> 
> ```llvm
> for iv = 0 to 4 step 1 {
>   %0 = dma_cpy_nd(bd_id = iv)  // simplified
>   %1 = dma_cpy_nd(bd_id = 4 +  iv)  // simplified
>   dma_wait(%0)
>   dma_wait(%1)
>   for iv2 = 0 to 2 step 1 {
>     %2 = dma_cpy_nd(bd_id = iv2)  // simplified
>     dma_wait(%2)
>   }
> }
> ```

Maybe we haven't seen this, but it's valid IR that could come up, and I don't see the issue in supporting it? I think that if you assign the range of values used for a loop when you encounter the `dma_cpy_nd` and release them on wait, it should handle both cases above, or am I missing something?

Yu-Zhewen: Yeah, so I guess when encountering the `dma_cpy_nd`, just adding some extra logic to check whether `dma_wait` is before/after any innerloop. And how about the case of wait after innerloop, shall we also try to have expression if possible or leave it for now?

jtuyls: > Yeah, so I guess when encountering the dma_cpy_nd, just adding some extra logic to check whether dma_wait is before/after any innerloop.

Why would you need that additional logic? I think you can loop through the range of values when you encounter `dma_cpy_nd` and assign them/annotate them as assigned and then release all of them when you encounter `dma_wait`. I think this should avoid needing special logic for innermost, wait before/after inner loop.

Yu-Zhewen: Because when encountering `dma_cpy_nd`, I need to decide beforehand whether to use BD as a constant or an expression. If it's an expression, I also need to determine its range, as I want to avoid leaving the inner loop without any BD to assign. However, all of this depends on the location of `dma_wait`.

jtuyls: > Because when encountering dma_cpy_nd, I need to decide beforehand whether to use BD as a constant or an expression. If it's an expression, I also need to determine its range, as I want to avoid leaving the inner loop without any BD to assign. However, all of this depends on the location of dma_wait.

You could traverse the controlcode from inner to outer?

Yu-Zhewen: Potentially yes, but the BD release might be getting tricky. When the inner loop releases BDs, the outer loop actually needs avoid using them.

jtuyls (approved):
LGTM

PR #949, opened 2024-12-02
[AMDAIEConvertToDma] Small code simplification

newling:
Simplification to not pass more flags into functions than are actually used (logic for pack vs unpack handled before function call).


jtuyls (approved):
LGTM, but could you remove [NFC] from the commit message before merging?

PR #950, opened 2024-12-02
Default flags: enable function outlining and use `-O3` for peano's opt

newling:
This PR changes 2 default compiler settings to 

1: _enable_ function outlining by default
2: use `-O3` in peano's opt by default 

For one one our benchmarks (bf16 -> f32 matmul, M=4096 N=K=512), I observe the following times (in milliseconds). 

| Optimization Level | Outlining      | No Outlining      |
|---------------------|----------------|--------------------|
| -O2                | 10.7           | 5.4               |
| -O3                | 4.0            | 5.4               |

This PR effectively goes from the top-right above to the bottom-left above. The 1.4 ms speed-up is not in the margin of noise. I can't explain it yet, still investigating... to verify locally, you can try

```
./run.py outdir ${IREE} --peano_dir=$PEANO} --target_device=npu1_4col \
 --xrt_lite_n_core_rows=4 --xrt_lite_n_core_cols=4 \
 --tests=vanilla_matmul_benchmark_4096_512_512_bf16_f32 --verbose \
 --aie_compilation_flags="--iree-amdaie-enable-function-outlining=0  --iree-amd-aie-additional-peano-opt-flags="-O3\""
```

And vary the value of `--iree-amdaie-enable-function-outlining=.` and `--iree-amd-aie-additional-peano-opt-flags=.`

Perf numbers also in CI.


MaheshRavishankar:
Nice!!!!

jtuyls (approved):
LGTM. Maybe at some point we don't need to keep track of the non-outlined performance, but I think it's good to keep it for a while for reference.

newling:
> LGTM. Maybe at some point we don't need to keep track of the non-outlined performance, but I think it's good to keep it for a while for reference.

Yeah, I agree. We should keep adapting the set of tests being tracked for performance based on what our goal is, and consider having another set of less visible performance tests with thresholds above which PRs fail

PR #954, opened 2024-12-03
Do not outline elementwise operation

newling:
What observations led to this PR? 

1) I compared the input.opt.ll files with and without coalescing. One difference I notices was the number of calls to the elementwise outlined function. Specifically, grepping in the output directory of the run.py script after 2 runs creating directories `no_coalescing` and `with_coalescing`:

```
grep "call void @generic_elementwise" no_coalescing/matmul_truncf_128_256_bf16_f32/input.opt.ll | wc -l
16
grep "call void @generic_elementwise" with_coalescing/matmul_truncf_128_256_bf16_f32/input.opt.ll | wc -l
0
```
Above notice that there are no function calls to the elementwise func when coalescing is enabled. i.e. the calls must have been inlined. Note that for matmul the number of calls is the same:
```
grep "call void @generic_matmul" no_coalescing/matmul_truncf_128_256_bf16_f32/input.opt.ll | wc -l
128
grep "call void @generic_matmul" with_coalescing/matmul_truncf_128_256_bf16_f32/input.opt.ll | wc -l
128
```

Background fact: we go out of program memory quite badly without coalescing (we hit 30 kB, much more than the allowed 16 kB).  With coalescing, we fit. 

So it would seem that _inlining the function calls dramatically reduces the memory required_. I can't explain this. 

With the above observation, this PR removes function outlining for the elementwise ops, so that they are effectively always inline. With this change, program fits in memory even without coalescing.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/linalg_function_outlining.mlir
```diff
@@ -72,88 +72,38 @@ func.func @matmul_example(%A: memref<4x8xbf16>, %B: memref<8x4xbf16>, %C: memref
...
+// CHECK-SAME:  %[[A1:.*]]: memref<4x6xbf16>, %[[B1:.*]]: memref<6x4xbf16>,
+// CHECK-SAME:   %[[C:.*]]: memref<4x4xf32>) {
+// CHECK:      amdaie.core
+// CHECK-NEXT: func.call @[[MATMUL_K4]](%[[A0]], %[[B0]], %[[C]])
+// CHECK-NEXT: func.call @[[MATMUL_K4]](%[[A0]], %[[B0]], %[[C]])
+// CHECK-NEXT: func.call @[[MATMUL_K4]](%[[B0]], %[[A0]], %[[C]])
+// CHECK-NEXT: func.call @[[MATMUL_K6]](%[[A1]], %[[B1]], %[[C]])
+// CHECK-NEXT: amdaie.end
+// CHECK:      return
+func.func @matmul_example_2(%A0: memref<4x4xbf16>, %B0: memref<4x4xbf16>,
```
Abhishek-Varma, line 93: No need for this example. If required, you can add one `linalg.matmul` to an already existing test case above.

Abhishek-Varma (approved):
This seems like a very good find @newling ! Thank you!

Another supporting argument that adds to what you mentioned above: since we don't outline `linalg.fill` (a producer fused to matmul - part of a single loop - PROLOGUE), there shouldn't be any need to outline an elementwise (a consumer fused to matmul - part of a single loop - EPILOGUE).

LGTM % one review comment.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/linalg_function_outlining.mlir
newling: I like this test, I think it's good to have a sequence of tests which slowly change what's being tested from the preceding test, makes diagnosing failures easier. I should use a more descriptive name though...

PR #955, opened 2024-12-03
[LoweringStrategy] Refactor to take num of rows/cols as inputs

yzhang93:
This PR added support to take the number of cores (using flag `--iree-amdaie-num-rows --iree-amdaie-num-cols`) as input and generate L2 tile sizes accordingly. This PR also addressed some remaining issues when switching to 4x4 array usage.

- In AIR pipeline, for matmul ops switch to use 4x4 array by default. However, the matmul-elementwise tests don't work with 4x4 array, so added flags to use 2x2 array in ci. CC @erwei-xilinx 
- Also change option name `usePassPipeline` to `useTilePipeline`.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lowering_strategy.mlir
```diff
@@ -217,7 +217,7 @@ builtin.module {
 // -----
 
 // CHECK-PACK-PEEL{LITERAL}: #config = #iree_codegen.lowering_config<tile_sizes = [[44, 128], [0, 0, 1], [1, 1, 0, 0, 0, 0]]>
-// CHECK-PACK-PEEL{LITERAL}: #packingConfig = #amdaie.packing_config<packing_config = [{packedSizes = [44, 64, 64], transposePackIndices = [1], unpackEmpty = [false], innerPerm = [[1, 0]], outerPerm = [[0, 1]]}, {packedSizes = [0, 0, 0, 4, 4, 8], transposePackIndices = [0, 1, 2], unpackEmpty = [false, false, true], innerPerm = [[0, 1], [1, 0], [0, 1]], outerPerm = [[0, 1, 3, 2], [0, 1, 3, 2], [0, 1, 3, 2]]}]>
+// CHECK-PACK-PEEL{LITERAL}: #packingConfig = #amdaie.packing_config<packing_config = [{packedSizes = [44, 32, 64], transposePackIndices = [1], unpackEmpty = [false], innerPerm = [[1, 0]], outerPerm = [[0, 1]]}, {packedSizes = [0, 0, 0, 4, 4, 8], transposePackIndices = [0, 1, 2], unpackEmpty = [false, false, true], innerPerm = [[0, 1], [1, 0], [0, 1]], outerPerm = [[0, 1, 3, 2], [0, 1, 3, 2], [0, 1, 3, 2]]}]>
```
Abhishek-Varma, line 220: Why did it change from `64` -> `32` ?

Wouldn't this warrant inclusion of the row/col flags ?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -150,6 +150,39 @@ class AIETargetBackend final : public IREE::HAL::TargetBackend {
...
+      case AMDAIEDevice::npu1_3col:
+        maxCoreRows = 4;
+        maxCoreCols = 3;
+        break;
+      case AMDAIEDevice::npu4:
+        maxCoreRows = 4;
+        maxCoreCols = 8;
+        break;
+      default:
+        llvm::errs() << "unhandled NPU partitioning.\n";
```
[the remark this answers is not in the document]
newling: Yes I'd put an `assert(false && "unhandled target device, we must specify the array size here")`

But isn't this info available already from somewhere like https://github.com/nod-ai/iree-amd-aie/blob/7c4b98571d86d7ec25daa20f1d0ac9e2be8a8c05/runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h#L284 ?

build_tools/ci/cpu_comparison/run.py
```diff
@@ -467,9 +440,50 @@ def _execute(self, config):
         matmul_template_dir = config.file_dir / "matmul_template"
         template_name = matmul_template_dir / "matmul_bias_MxK_KxN_N.mlir"
         self.generate(config, template_name)
+        self.add_aie_compilation_flags(
+            [
+                "--iree-amdaie-matmul-elementwise-fusion",
+                "--iree-amdaie-num-rows=2",
```
newling, line 446: Doesn't this change the default for all matmuls from 4x4 to 2x2? Maybe not what we want? 

I'd be interested to know how many of the tests 
```
for n_rows in [1,4]:
  for n_cols in [1,2,3,4]:
    # run matmul with these values  (on npu1_4col for say M=N=K=3*256).
```
work

newling:
Can the flags `--xrt_lite_n_core_rows`  and `--xrt_lite_n_core_cols` be reused, instead of adding new flags `--iree-amdaie-num-rows` and `--iree-amdaie-num-cols` ? When would I use different values for these?

jtuyls:
> Can the flags `--xrt_lite_n_core_rows` and `--xrt_lite_n_core_cols` be reused, instead of adding new flags `--iree-amdaie-num-rows` and `--iree-amdaie-num-cols` ? When would I use different values for these?

 `--xrt_lite_n_core_rows` and `--xrt_lite_n_core_cols`  are `iree-run-module` flags. I don't think they can be reused in `iree-compile`.

build_tools/ci/cpu_comparison/run.py
jtuyls: It being shown in a confusion way here, but these changes are actually in `MatmulThinBias`, not `Matmul`

yzhang93: Yes, it's added to `MatmulThinBias` class and used for matmul-elementwise tests in AIR.

newling: Ha, I totally missed that... 

{8C56B1BC-EB69-4D62-9344-F07F07905261}

newling:
> > Can the flags `--xrt_lite_n_core_rows` and `--xrt_lite_n_core_cols` be reused, instead of adding new flags `--iree-amdaie-num-rows` and `--iree-amdaie-num-cols` ? When would I use different values for these?
> 
> `--xrt_lite_n_core_rows` and `--xrt_lite_n_core_cols` are `iree-run-module` flags. I don't think they can be reused in `iree-compile`.

That makes sense.  Maybe (unrelated to this PR) the `xrt_lite_n_core_rows` and `xrt_lite_n_core_cols` can be inferred from the device? i.e. always just take the maximum, so for npu1_4col, infer that `xrt_lite_n_core_row=`xrt_lite_n_core_col=4` ? There was probably a reason Maks didn't do it this way though

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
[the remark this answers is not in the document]
yzhang93: It's not just printing the message, it will bail out like `assert`. And there are `llvm::errs()` used through out this file.

@newling @jtuyls Yes, we can use `deviceModel.configPtr.NumCols` to get num_rows/cols, but it still need some hardcode adjustments to show number of rows/cols for cores. For example, `npu1_4col` returns 5 cols and 6 rows, so we need to -1 cols and -2 rows for cores. While `npu4` returns 8 cols and 6 rows, so we only need to -2 rows for cores.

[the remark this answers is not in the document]
jtuyls: You can create a new function in the device model `getNumCoreRows` that returns the number of rows with cores and for the 5 columns, we probably just need to change that in device model.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lowering_strategy.mlir
yzhang93: No, it's because now we change 2x2 core usage to 4x4 by default for AIR path. If you use flag `--iree-amdaie-num-rows=2 --iree-amdaie-num-cols=2` you will see the previous sizes.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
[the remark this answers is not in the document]
yzhang93: For number of columns, we might be able to use `partitionNumCols`? https://github.com/nod-ai/iree-amd-aie/blob/40628f5528fae75d8a51978247355e42ed954588/runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.cc#L589

But the value for `npu1` is 5, and we want 4 instead.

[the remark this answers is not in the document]
yzhang93: Okay, I figured out we can use `deviceModel.cols()` to get correct number of cols from the device. It returns 4 cols for `npu1_4col`, and 5 cols for `npu1`. But I think it's okay, we'd never use 5 cols for npu1 device.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -368,6 +368,9 @@ struct AMDAIEDeviceModel {
     return deviceConfig.maxVectorSizeBits;
   }
 
+  uint32_t getNumCoreRows() const { return rows() - 2; }
```
newling, line 371: https://github.com/nod-ai/iree-amd-aie/blob/cbb17c9397f59de1ebd9bfea67281f87fd8899d7/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.cpp#L113

I'm wondering if there will might ever be more than 1 row of L2 memory tiles, or some other reason this might not work in the future

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -150,6 +151,19 @@ class AIETargetBackend final : public IREE::HAL::TargetBackend {
...
     SmallVector<NamedAttribute> configItems;
 
+    // Make sure the input number of rows/cols is smaller or equal to the max
+    // number of rows/cols from the device.
+    AMDAIEDeviceModel deviceModel =
+        AMDAIE::getDeviceModel(options.AMDAIETargetDevice);
+    uint32_t maxCoreRows = deviceModel.getNumCoreRows();
+    uint32_t maxCoreCols = deviceModel.getNumCoreCols();
+    if (options.AMDAIENumRows <= 0 || options.AMDAIENumRows > maxCoreRows) {
+      llvm::report_fatal_error("option numRows is out of range\n");
```
newling, line 161:
```suggestion
            llvm::report_fatal_error(llvm::Twine("Invalid number of core rows (") +
                               std::to_string(options.AMDAIENumRows) +
                               "), must be in the range [1, " +
                               std::to_string(maxCoreRows) + "] for device " +
                               stringifyEnum(deviceModel.device));
```

Might save you in the future! A lit test of this would be nice if possible.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
```diff
@@ -59,6 +59,8 @@ struct AMDAIEOptions {
   bool insertLoopAroundCoreBlock{false};
   bool matmulElementwiseFusion{false};
   AMDAIEDevice AMDAIETargetDevice{AMDAIEDevice::npu1_4col};
+  int AMDAIENumRows{4};
```
newling, line 62: These default values can be inferred from the default AMDAIETargetDevice: create device model and then call `getNumCoreRows()` and `getNumCoreCols()`

My thinking is that the fewer defaults to twiddle when we one day change to npu4 as the default, the better.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
```diff
@@ -231,6 +233,14 @@ struct AMDAIEOptions {
             clEnumValN(AMDAIEDevice::npu4, "npu4",
                        "Strix B0 NPU with 8 columns and 6 rows")));
 
+    binder.opt<int>("iree-amdaie-num-rows", AMDAIENumRows,
+                    llvm::cl::cat(category),
+                    llvm::cl::desc("Number of rows used in an AIE core array"));
```
newling, line 238:
```suggestion
                    llvm::cl::desc("Number of rows used in an AIE core array. The compiler will choose a tiling strategy that uses no more than this number of rows. However, some workloads (like convolution) currently ignore this flag, and use a hardcoded number of rows."));
```

or something like that, for users without the context

newling (commented):
Looks good, thanks for this. 

I tried this locally with run.py with t M=N=512 K=4096  with different numbers of colums and rows, and I see

```
--iree-amdaie-num-rows=4", "--iree-amdaie-num-cols=1"
error: 'amdaie.logicalobjectfifo.from_memref' op should have at least one tile candidate

--iree-amdaie-num-rows=4", "--iree-amdaie-num-cols=2"
<unknown>:0: error: 'aie.memtile_dma' op could not find and assign a valid BD id

--iree-amdaie-num-rows=4", "--iree-amdaie-num-cols=3"
error: 'amdaie.logicalobjectfifo.from_memref' op should have at least one tile candidate

--iree-amdaie-num-rows=1", "--iree-amdaie-num-cols=2"
error: 'amdaie.logicalobjectfifo.from_memref' op should have at least one tile candidate
```

@yzhang93 have you got any example with n_rows != n_cols to run end to end? I'm just curious, I'm happy for progress to made on this as a follow-up.

yzhang93:
> Looks good, thanks for this.
> 
> I tried this locally with run.py with t M=N=512 K=4096 with different numbers of colums and rows, and I see
> 
> ```
> --iree-amdaie-num-rows=4", "--iree-amdaie-num-cols=1"
> error: 'amdaie.logicalobjectfifo.from_memref' op should have at least one tile candidate
> 
> --iree-amdaie-num-rows=4", "--iree-amdaie-num-cols=2"
> <unknown>:0: error: 'aie.memtile_dma' op could not find and assign a valid BD id
> 
> --iree-amdaie-num-rows=4", "--iree-amdaie-num-cols=3"
> error: 'amdaie.logicalobjectfifo.from_memref' op should have at least one tile candidate
> 
> --iree-amdaie-num-rows=1", "--iree-amdaie-num-cols=2"
> error: 'amdaie.logicalobjectfifo.from_memref' op should have at least one tile candidate
> ```
> 
> @yzhang93 have you got any example with n_rows != n_cols to run end to end? I'm just curious, I'm happy for progress to made on this as a follow-up.

Thanks for trying different combinations! I think we have e2e test running on 4x2 previously in a dev branch, but now as I tried and these failed with the same errors above. @jtuyls It looks like we'll need to generalize `AssignTiles` pass.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
jtuyls: It should just be `configPtr.AieTileNumRows`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.h
```diff
@@ -36,8 +36,8 @@ void addMLIRAIELoweringPasses(OpPassManager &passManager);
 /// the structured ops path. The pass manager `pm` here operate on the module
 /// within the IREE::HAL::ExecutableOp.
 void buildAMDAIETransformPassPipeline(
-    OpPassManager &variantPassManager, AMDAIEDevice device,
-    TilePassPipeline useTilePipeline,
+    OpPassManager &variantPassManager, AMDAIEDevice device, int numRows,
+    int numCols, TilePassPipeline useTilePipeline,
```
Abhishek-Varma, line 40:
```suggestion
    OpPassManager &variantPassManager, AMDAIEDevice device, uint32_t numRows,
    uint32_t numCols, TilePassPipeline useTilePipeline,
```

Abhishek-Varma (approved):
Thanks @yzhang93 !

LGTM % one comment.

newling (commented):
LGTM

PR #958, opened 2024-12-04
[LowerToAIE] Make size folding hardware-aware

jtuyls:
Make the dimension folding inside `LowerToAIE` partially hardware-aware by retrieving max sizes from the device model. This avoids overflowing the number of available bits.

A few notes:
- This folding inside `LowerToAIE` is needed at this place and can't be done separately earlier right now because offsets can be assumed all zero here because they are already handled earlier in this pass. We should try to streamline this further in the future so that this one-off folding is not needed. This can be done for example by canonicalizing on the lowered ops instead.
- Similar checks should be done for strides as well. I can do this in a follow-up.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/aie/AIEDialect.cpp
+++ b/compiler/plugins/target/AMD-AIE/aie/AIEDialect.cpp
@@ -494,10 +494,10 @@ LogicalResult DMABDOp::verify() {
                << "exceeds memref size "
                << std::to_string(bufferType.getNumElements());
       }
-      if (dim.getSize() >= (1UL << 9) + 1) {
-        return emitOpError() << "Size may not exceed 1023.";
+      if (dim.getSize() >= ((1UL << 10) - 1)) {
+        return emitOpError() << "Size may not exceed " << ((1UL << 10) - 1);
       }
-      if (dim.getStride() >= (1UL << 19)) {
+      if (dim.getStride() >= (1UL << 20)) {
         return emitOpError() << "Stride may not exceed " << (1 << 20);
       }
     }
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.h
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.h
@@ -30,7 +30,8 @@ namespace mlir::iree_compiler::AMDAIE {
 /// `amdaie.workgroup`.
 class AIEDeviceBuilder {
  public:
-  AIEDeviceBuilder(MLIRContext *ctx) : rewriter(ctx) {}
+  AIEDeviceBuilder(MLIRContext *ctx, AMDAIEDevice device)
+      : rewriter(ctx), device(device), deviceModel(getDeviceModel(device)) {}
 
   LogicalResult lowerToAIE(ModuleOp moduleOp);
 
@@ -67,7 +68,7 @@ class AIEDeviceBuilder {
   /// `AIE::BDDimLayoutArrayAttr`.
   AIE::BDDimLayoutArrayAttr convertSizeStrideToBDDimLayoutArrayAttr(
       const SmallVector<OpFoldResult> &sizes,
-      const SmallVector<OpFoldResult> &strides);
+      const SmallVector<OpFoldResult> &strides, uint8_t memSpace);
 
   /// Utility to create DMA blocks and add them to `memOp`.
   void createDMA(Operation *memOp, AIE::DMAChannelDir channelDir,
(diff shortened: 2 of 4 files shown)


compiler/plugins/target/AMD-AIE/aie/AIEDialect.cpp
```diff
@@ -494,10 +494,10 @@ LogicalResult DMABDOp::verify() {
                << "exceeds memref size "
                << std::to_string(bufferType.getNumElements());
       }
-      if (dim.getSize() >= (1UL << 9) + 1) {
-        return emitOpError() << "Size may not exceed 1023.";
+      if (dim.getSize() >= ((1UL << 10) - 1)) {
```
jtuyls, line 497: Without this fix, a convolution lit test actually fails.

compiler/plugins/target/AMD-AIE/aie/AIEDialect.cpp
```diff
@@ -494,10 +494,10 @@ LogicalResult DMABDOp::verify() {
...
+        return emitOpError() << "Size may not exceed " << ((1UL << 10) - 1);
       }
-      if (dim.getStride() >= (1UL << 19)) {
+      if (dim.getStride() >= (1UL << 20)) {
```
jtuyls, line 500: This one was wrong as well. This whole verifier needs to be refactored, but I am leaving that for future work.

compiler/plugins/target/AMD-AIE/aie/AIEDialect.cpp
newling: Great find!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
```diff
@@ -620,6 +620,58 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
...
+      %channel = amdaie.channel(%tile_0_1, 0, port_type = DMA, direction = MM2S)
+      %channel_1 = amdaie.channel(%tile_0_2, 0, port_type = DMA, direction = S2MM)
+      %4 = amdaie.flow({%channel} -> {%channel_1}) {is_packet_flow = false}
+      %5 = amdaie.connection(%1 {%channel_1}, %0 {%channel}, flow = %4) : (!amdaie.logicalobjectfifo<memref<4096xi32, 2 : i32>, 1>, !amdaie.logicalobjectfifo<memref<4096xi32, 1 : i32>, 1>)
+      %channel_2 = amdaie.channel(%tile_0_1, 1, port_type = DMA, direction = MM2S)
+      %channel_3 = amdaie.channel(%tile_0_2, 1, port_type = DMA, direction = S2MM)
+      %6 = amdaie.flow({%channel_2} -> {%channel_3}) {is_packet_flow = false}
+      %7 = amdaie.connection(%3 {%channel_3}, %2 {%channel_2}, flow = %6) : (!amdaie.logicalobjectfifo<memref<4096xi32, 2 : i32>, 1>, !amdaie.logicalobjectfifo<memref<4096xi32, 1 : i32>, 1>)
+      amdaie.controlcode {
+        // CHECK: aie.dma_bd(%{{.+}} : memref<4096xi32, 1 : i32>) {dimensions = #aie<bd_dim_layout_array[<size = 32, stride = 32>, <size = 32, stride = 1>]>, len = 1024 : i32}
```
newling, line 662: Is this testing the size=512 is valid? So before this change, this would have failed with the error about the size being too large?

jtuyls: This one is testing that `[32, 32] [32, 1]` isn't transformed into `[1024, 1]`, because `1024` exceeds `1023`. The one below is testing that `[16, 32] [32, 1]` is transformed into `[512, 1]`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.h
```diff
@@ -111,6 +112,10 @@ class AIEDeviceBuilder {
 
   IRRewriter rewriter;
   IRMapping mapper;
+  /// The device and device model. The device is needed separately to convert to
+  /// the `AIEDevice`.
+  AMDAIEDevice device;
```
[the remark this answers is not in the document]
jtuyls: Oh, wow, I overlooked that, thanks!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lower_to_aie.mlir
newling: That makes sense, thanks

newling (approved):

PR #959, opened 2024-12-04
[AMDAIELinalgFunctionOutlining] Don't outline if operand has non-identity layout 

newling:
This is a fix for a bug I introduced in https://github.com/nod-ai/iree-amd-aie/pull/951 . I thought that the approach of removing the offset from the memref function signature worked, but it doesn't seem to. I must have been running a test other than the convolution end-to-end test when I was confirming it worked - now with function outlining enabled in gives a numerical error. 

This PR simplifies logic: any non-identity layout results in no outlining.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
@@ -20,15 +20,6 @@ namespace mlir::iree_compiler::AMDAIE {
 
 namespace {
 
-/// Return true if `type` is a memref with an identity layout.
-bool isMemRefWithIdentityLayout(Type type) {
-  auto memRefType = dyn_cast<MemRefType>(type);
-  if (!memRefType) return false;
-  MemRefLayoutAttrInterface layout = memRefType.getLayout();
-  if (!layout) return true;
-  return layout.isIdentity();
-}
-
 /// Utility to outline the linalg compute op.
 static FailureOr<func::FuncOp> outline(IRRewriter &rewriter, ModuleOp moduleOp,
                                        linalg::LinalgOp computeOp,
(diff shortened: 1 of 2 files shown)


yzhang93:
> This is a fix for a bug I introduced in #951 . I thought that the approach of removing the offset from the memref function signature worked, but it doesn't seem to. I must have been running a test other than the convolution end-to-end test when I was confirming it worked - now with function outlining enabled in gives a numerical error.
> 
> This PR simplifies logic: any non-identity layout results in no outlining.

The CI doesn't capture the numerical error? Is there a test you can add to check it?

newling:
> The CI doesn't capture the numerical error? Is there a test you can add to check it?

When I turn outlining on by default it will show. I was planning on doing that in the next PR: https://github.com/nod-ai/iree-amd-aie/pull/950

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
```diff
@@ -20,48 +20,29 @@ namespace mlir::iree_compiler::AMDAIE {
...
-Type getTypeWithoutLayout(Type type) {
+/// Return true if `type` is a memref with an identity layout.
+bool isMemRefWithIdentityLayout(Type type) {
   auto memRefType = dyn_cast<MemRefType>(type);
-  if (!memRefType) return {};
-  if (!isContiguousMemRef(memRefType)) return {};
-  return MemRefType::get(memRefType.getShape(), memRefType.getElementType(),
-                         MemRefLayoutAttrInterface{},
-                         memRefType.getMemorySpace());
+  if (!memRefType) return false;
```
yzhang93, line 26: I think it should throw an error instead of return false.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
```diff
@@ -183,21 +161,9 @@ void AMDAIELinalgFunctionOutliningPass::runOnOperation() {
...
-        Type type = func.getArgumentTypes()[iter.index()];
-        Value cast =
-            rewriter.createOrFold<memref::CastOp>(loc, type, iter.value());
-        castOperands.push_back(cast);
-      }
-      rewriter.create<func::CallOp>(loc, func, castOperands);
-    }
+    rewriter.setInsertionPoint(computeOp);
+    rewriter.create<func::CallOp>(computeOp.getLoc(), func,
+                                  computeOp->getOperands());
```
yzhang93, line 166: This is a nice simplification. But I'm trying to understand why it doesn't need `cast` any more?

yzhang93 (approved):
LGTM. Just minor comments/questions.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
newling: The cast was there to cast from a type with an offset, to a type without an offset. This PR changes the logic so that if a memref has an offset, then it cannot be outlined. So we never need to cast.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
newling: I guess I could refactor this function to take a memref argument.

PR #962, opened 2024-12-05
Add a pass to fold DMA waits

Yu-Zhewen:
Each DMA channel has a task queue with the depth of 4. DMA wait is only required for every 4 pushes, reducing unnecessary synchronization.

Example: https://gist.github.com/Yu-Zhewen/5f569b56c7b1f1a8715a7c4c3bf9e609

Results compared to 7c4b98571d86d7ec25daa20f1d0ac9e2be8a8c05:
| Test (MxKxN)  | Instruction Size Before (Words) | Instruction Size After (Words) |
|---------------|---------------------------------|--------------------------------|
| 512x4096x512  | 1228                            | 1132                           |
| 512x512x4096  | 820                             | 772                            |
| 4096x512x512  | 4628                            | 4244                           |

This optimization is orthogonal to DMA chaining #931.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESimplifyDmaWaits.cpp
```diff
@@ -0,0 +1,158 @@
...
+          rewriter.eraseOp(op);
+        }
+      }
+    }
+  }
+
+  return success();
+}
+
+class AMDAIESimplifyDmaWaitsPass
```
jtuyls, line 116: `Simplify` is a quite broad term. I think for example `AMDAIEFoldDmaWaitsPass` is more descriptive.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESimplifyDmaWaits.cpp
```diff
@@ -0,0 +1,158 @@
...
+              return WalkResult::interrupt();
+            }
+            // Get the maximum queue size.
+            uint32_t col = getConstantIndexOrAssert(tileOp.getCol());
+            uint32_t row = getConstantIndexOrAssert(tileOp.getRow());
+            uint32_t maxQueueSize = deviceModel.getDmaMaxQueueSize(col, row);
+            // Keep wait op if reaches the maximum queue size or there is a
+            // duplicate BD ID.
+            uint32_t bdId = getConstantIndexOrAssert(bdIdOp.getValue());
+            auto &bdIdQueue = connectionToBdIdQueues[connectionOp];
```
jtuyls, line 76:
```suggestion
            SmallVector<uint32_t> &bdIdQueue = connectionToBdIdQueues[connectionOp];
```
Following LLVM conventions, we avoid `auto` unless the type is clear from the context (`cast<Type>` etc).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESimplifyDmaWaits.cpp
```diff
@@ -0,0 +1,158 @@
...
+            if (bdIdQueue.size() >= maxQueueSize) bdIdQueue.clear();
+            if (bdIdQueue.empty() || llvm::is_contained(bdIdQueue, bdId)) {
+              toErase = false;
+              bdIdQueue = {bdId};
+            } else {
+              bdIdQueue.push_back(bdId);
+            }
```
jtuyls, line 83: Isn't `bdIdQueue` empty on the first encounter, but the wait can be erased right?

Also, you can simplify this logic:
```suggestion
            if (bdIdQueue.empty() || bdIdQueue.contains(bdId)) toErase = false;
            bdIdQueue.push_back(bdId);
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESimplifyDmaWaits.cpp
```diff
@@ -0,0 +1,158 @@
...
+  if (res.wasInterrupted()) return signalPassFailure();
+}
+
+}  // namespace
+
+std::unique_ptr<Pass> createAMDAIESimplifyDmaWaitsPass() {
+  return std::make_unique<AMDAIESimplifyDmaWaitsPass>();
+}
+
+}  // namespace mlir::iree_compiler::AMDAIE
```
jtuyls, line 158:
```suggestion
}  // namespace mlir::iree_compiler::AMDAIE

```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESimplifyDmaWaits.cpp
```diff
@@ -0,0 +1,158 @@
...
+#define DEBUG_TYPE "iree-amdaie-simplify-dma-waits"
+
+namespace mlir::iree_compiler::AMDAIE {
+
+namespace {
+
+/// Traverses the control code in reverse, ensuring that for each connection,
+/// only one DMA wait op is retained for every maximum queue size.
+LogicalResult simplifyDmaWaits(AMDAIE::AMDAIEDeviceModel deviceModel,
+                               AMDAIE::WorkgroupOp workgroupOp) {
```
jtuyls, line 22: This is a nit, but you could take a `AMDAIE::ControlCodeOp` as parameter instead as the workgroup is only needed to retrieve the control code op, so this would simplify the function a bit and make in more easily reusable.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/fold_dma_waits.mlir
```diff
@@ -0,0 +1,224 @@
...
+    amdaie.workgroup {
+      amdaie.controlcode {
+        amdaie.end
+      }
+    }
+    return
+  }
+}
+
+// -----
```
newling, line 30: Please add comments about what is being tested in each case. I can see there are fewer dma_wait operations after the pass, but it's not clear to me which ones are being removed. Also a bit surprised that there are no CHECK-NOT or CHECK-NEXT statements.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -378,6 +378,8 @@ struct AMDAIEDeviceModel {
   DenseMap<uint32_t, SmallVector<uint32_t>> getChannelToValidBdIds(
       AMDAIETileType tileType) const;
 
+  uint8_t getDmaMaxQueueSize(uint8_t col, uint8_t row);
```
newling, line 381: Really unfortunate that this can't be `const`.  I see that getTileType uses a const_cast to workaround the lack of const-correctness in aie_rt, seems like first done here 

https://github.com/nod-ai/iree-amd-aie/commit/6c4f9050d8b9166845d95dc0a404e47d24938fd3#diff-17008229092a63d5df9831105108aa99c799a226441b0dd9d8327708e57fc2aeR218

I tracked this down because I saw you passing `AMDAIE::AMDAIEDeviceModel deviceModel` by value, which isn't great as this isn't a pointer type so there is overhead. But if getDmaMaxQueueSize isn't const, deviceModel can't be const ref where you use it...

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
```diff
@@ -0,0 +1,164 @@
...
+            AMDAIE::ConnectionOp connectionOp = maybeConnectionOp.value();
+            // Retrieve the flow op.
+            std::optional<AMDAIE::FlowOp> maybeFlowOp =
+                maybeConnectionOp->getFlowOp();
+            if (!maybeFlowOp) {
+              maybeConnectionOp->emitOpError()
+                  << "expected to operate on an `amdaie.flow`";
+              return WalkResult::interrupt();
+            }
+            if (maybeFlowOp->getIsPacketFlow()) return WalkResult::advance();
```
newling, line 52: This advance -- so you skip to the next waitOp, effectively making toErase = false for this waitOp? I think the logic would be easier to follow if you made standalone functions. Maybe a function like

LogicalResult<bool> canFoldWaitOp(WaitOp waitOp, AMDAIE::AMDAIEDeviceModel deviceModel, ...) {

}

newling (dismissed):

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
jtuyls, line 381: I know this class isn't completely organized, but I think it helps to put new functions at the most sensible place in the class. In this case, I think `getDmaMaxQueueSize` fits together with other DMA information retrieval methods, like `getDmaBdProp and `getDmaProp`, so maybe move it after those? Same for the function in .cpp.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/fold_dma_waits.mlir
```diff
@@ -0,0 +1,230 @@
...
+// while folding the others.
+// CHECK-LABEL: @fold_dma_waits_max_queue_size
+// CHECK:       %[[OBJECT_FIFO_0:.+]] = amdaie.logicalobjectfifo.from_buffers
+// CHECK:       %[[CHANNEL_0:.+]] = amdaie.channel
+// CHECK:       %[[CHANNEL_1:.+]] = amdaie.channel
+// CHECK:       %[[CONNECTION:.+]] = amdaie.connection
+// CHECK:         %[[OBJECT_FIFO_1:.+]] = amdaie.logicalobjectfifo.from_memref
+// CHECK:         %[[BD_ID_0:.+]] = amdaie.bd_id
+// CHECK:         %[[TOKEN_0:.+]] = amdaie.npu.half_dma_cpy_nd async %[[CONNECTION]](%[[OBJECT_FIFO_1]] [] [] [] bd_id = %[[BD_ID_0]] channel = %[[CHANNEL_0]]) : !amdaie.logicalobjectfifo<memref<2048xi32>>
+// CHECK:         amdaie.npu.dma_wait(%[[TOKEN_0]] : !amdaie.async_token)
```
jtuyls, line 92: Why do we have a wait on the first `dma_cpy_nd`?

I think we should expect:
```
dma_cpy_nd
dma_cpy_nd
dma_cpy_nd
%0 = dma_cpy_nd
dma_wait(%0)
```

Instead of:

```
%0 = dma_cpy_nd
dma_wait(%0)
dma_cpy_nd
dma_cpy_nd
%1 = dma_cpy_nd
dma_wait(%0)
```

Yu-Zhewen: I am traversing controlcode in reverse order, and the example actually is:
```ll
%0 = dma_cpy_nd
dma_wait(%0)
dma_cpy_nd
dma_cpy_nd
dma_cpy_nd
%1 = dma_cpy_nd
dma_wait(%0)
```
there are four `dma_cpy_nd` ops between.

jtuyls: Sorry, not only for `nb_dma_cpy_nd_ops <= 4`, but also in the example above, you're using two waits in what could be implemented with one.

Yu-Zhewen: but if we don't have a wait on the first op, then we will have to do this:
```
dma_cpy_nd
dma_cpy_nd
dma_cpy_nd
dma_cpy_nd
%0 = dma_cpy_nd
dma_wait(%0)
%1 = dma_cpy_nd
dma_wait(%1)
```
we still have two waits, since anyway we need one wait at the end of controlcode?

jtuyls: > we still have two waits, since anyway we need one wait at the end of controlcode?

That's not necessarily true, because typically the outputs depend on the inputs, so just waiting for the outputs is good. But it's good to keep the wait on the inputs for now because it's hard to **proof** that we're already waiting on a dependent output DMA, so we don't need to wait on this DMA.

Even if we keep the waits, there are cases in which you will generate too many waits:

```
%0 = dma_cpy_nd
dma_wait(%0)
dma_cpy_nd
dma_cpy_nd
dma_cpy_nd
%1 = dma_cpy_nd
dma_wait(%1)
%2 = dma_cpy_nd
dma_wait(%2)
```
Here, 3 waits are used, but it can be implemented with 2:
```
dma_cpy_nd
dma_cpy_nd
dma_cpy_nd
%0 = dma_cpy_nd
dma_wait(%0)
dma_cpy_nd
%1 = dma_cpy_nd
dma_wait(%1)
```

Yu-Zhewen: actually running the above example with my current code, I get the following, which has two waits:
```ll
dma_cpy_nd
%0 = dma_cpy_nd
dma_wait(%0)
dma_cpy_nd
dma_cpy_nd
dma_cpy_nd
%1 = dma_cpy_nd
dma_wait(%1)
```
(one thing is all these dma_cpy_nd need diffent BD ID)

jtuyls: Ok, that's good, but I still think this is non-intuitive output, leading to confusion, and it seems like an arbitrary choice unless I am missing something? So why do you feel strongly about returning it in this order?

Yu-Zhewen: it is just making the logic simpler by traversing backward. Otherwise, I need to check if the current wait is the last one for each connection

jtuyls: > Otherwise, I need to check if the current wait is the last one for each connection

You can traverse in forward order and keep track of every 4th and last DMA on a connection, then only keep those waits. I don't see how that's more complex?

jtuyls: The reason I think this is important is that the output IR should be as one would intuitively expect. A lot of the times I have to debug issues by just reading and understanding the IR and non-intuitive output is no fun when you're doing that.

Yu-Zhewen: I tried something, but then I realized that traversing in forward order complicates the management of BD IDs. When encountering a duplicate BD ID, we need to keep the wait for the last DMA whichused that BD ID. Also, all the erasion decisions between those two DMAs need to be updated accordingly.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
```diff
@@ -0,0 +1,167 @@
...
+                npuHalfDmaCpyNdOp.getConnectionOp();
+            if (!maybeConnectionOp) {
+              npuHalfDmaCpyNdOp.emitOpError()
+                  << "expected to operate on an `amdaie.connection`";
+              return WalkResult::interrupt();
+            }
+            AMDAIE::ConnectionOp connectionOp = maybeConnectionOp.value();
+            // Retrieve the flow op.
+            std::optional<AMDAIE::FlowOp> maybeFlowOp =
+                maybeConnectionOp->getFlowOp();
```
newling, line 47:
```suggestion
                connectionOp.getFlowOp();
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
```diff
@@ -0,0 +1,167 @@
...
+  std::vector<AMDAIE::NpuDmaWaitOp> waitOpsToErase;
+  DenseMap<std::pair<AMDAIE::TileOp, AMDAIE::ConnectionOp>,
+           SmallVector<uint32_t>>
+      tileConnectionToBdIdQueueMap;
+  // Traverse the control code in reverse.
+  WalkResult res = controlCodeOp->walk<WalkOrder::PostOrder, ReverseIterator>(
+      [&](AMDAIE::NpuDmaWaitOp waitOp) {
+        bool toErase = true;
+        for (Value token : waitOp.getAsyncTokens()) {
+          if (auto npuHalfDmaCpyNdOp =
```
newling, line 33: I still kinda think a function at the level 

FailureOr<bool> canFoldBasedOnNpuHalfDmaCpyNdOp(...)


would make for slightly easier to read (less indented) code.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
```diff
@@ -0,0 +1,167 @@
...
+            if (!tileOp) {
+              bdIdOp.emitOpError() << "must operate on an `amdaie.tile`";
+              return WalkResult::interrupt();
+            }
+            // Get the maximum queue size.
+            uint32_t col = getConstantIndexOrAssert(tileOp.getCol());
+            uint32_t row = getConstantIndexOrAssert(tileOp.getRow());
+            uint32_t maxQueueSize = deviceModel.getDmaMaxQueueSize(col, row);
+            // Keep wait op if, either reaches the maximum queue size, or there
+            // is a duplicate BD ID in the same tile.
```
newling, line 76:
```suggestion
            // is a duplicate BD ID in the same tile, or packet flow, or the queue is empty
```
?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/fold_dma_waits.mlir
```diff
@@ -0,0 +1,230 @@
...
+      }
+    }
+    return
+  }
+}
+
+// -----
+
+// Expect no DMA waits to be folded, since the same BD ID is used.
+// CHECK-LABEL: @fold_dma_waits_same_bd_id
```
newling, line 33: Could just be 
CHECK-COUNT-2: dma_wait
CHECK-NOT: dma_wait

newling (commented):
Thanks for the comments in the lit tests, they helped me

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/fold_dma_waits.mlir
jtuyls: Ok, could you add documentation to the function on why you iterate in reverse order and an example to show expected output for one of these more quirky cases above?

Yu-Zhewen: Thanks, added now

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
Yu-Zhewen: Yes, made it a function now

jtuyls (approved):
LGTM

PR #965, opened 2024-12-05
[Tests] Simplify test flags

yzhang93:
This PR did some cleanup of tests including

- Changed the test flags in `Test/samples` to use `--compile-to=executable-targets` instead of `--compile-to=executable-sources` and get rid of a mixture use of `iree-compile` and `iree-opt`.
- Removed redundant (with default value) flags.
- Modified some `//CHECK` tags to keep consistent format.


newling (approved):
LGTM

PR #966, opened 2024-12-05
[CI] improve performance printout

newling:
- Do not duplicate running of the tests with the Performance label (shouldn't be run in numerical testing)
- Ensure that if there is a failure with Performance tests, the trace is provided (see https://github.com/nod-ai/iree-amd-aie/pull/960 where this is tested)
- Print program memory at sensible place (before this PR it looked like the line printed came from the subsequent test)


.github/workflows/ci-linux.yml
```diff
@@ -200,12 +203,21 @@ jobs:
...
+            --tests=Performance > performance.log 2>&1
+          exit_code=$?  # Capture the exit code of run.py
+          set -e  # Re-enable exit on error for subsequent commands
+
+          # If there was an error, print the contents of the log to terminal and
+          # exit.
+          if [ $exit_code -ne 0 ]; then
+            cat performance.log
+            exit 1
+          fi
```
jtuyls, line 215: I think we should try to keep the yml files as simple as possible and put this in a script instead.

newling: This does add complexity (not good) but I don't think putting this in a script is the best option, it'll add another layer of passing parameters around and it hides the complexity somewhere less visible. 

So I've just simplified it. Only downside is that now if there is an error in the test you'll still the python exception stack, just not the full log.

jtuyls: > This does add complexity (not good) but I don't think putting this in a script is the best option, it'll add another layer of passing parameters around and it hides the complexity somewhere less visible.

Yeah, I was more thinking about piping the output to another script that then takes care of the error logging functionality.

jtuyls (approved):

PR #967, opened 2024-12-05
[Utils] Move all util files to a separate directory

yzhang93:
As the number of passes is growing in `Transform`, a flat structure of files is hard for looking up particular utility function. So this PR moved all util files to a separate directory under `Transform`.


yzhang93:
> I'm not really sure what you mean by this, doesn't your IDE help jumping to function definitions/declarations?

It's not about using IDE for developing. I'm just try to make the file structure more clear and easy for browse. It's okay not to merge this if all of you like the original structure.

newling:
> > Is it possible to get Utils next to Transforms, rather than nested in it? Like for example
> > /iree/compiler/src/iree/compiler/Codegen
> 
> It depends on which level the utils should be in. I was referring to the LLVMGPU file structures. https://github.com/iree-org/iree/tree/main/compiler/src/iree/compiler/Codegen/LLVMGPU
> 

Actually that's a better reference for us, so in that sense this PR makes sense. I'm going to accept, but maybe wait for others to say if they're onboard. 

> > I'm not really sure what you mean by this, doesn't your IDE help jumping to function definitions/declarations?
> 
> It's not about using IDE for development. My goal is simply to improve the file structure to make it clearer and easier for browsing. If everyone prefers the original structure, that’s perfectly fine, and I’m okay with not merging this change.


👍

newling (approved):

Abhishek-Varma (approved):
LGTM.

jtuyls (approved):

PR #963, opened 2024-12-05
[CombineStridedOps] Add support for same values

jtuyls:
Add support in the `CombineStridedOpsPass` for the case when all dimensions, except possibly the first one, are the same.

Example:
```
amdaie.npu.dma_cpy_nd %0([] [] [], [0, 0] [8, 16] [8, 1])
amdaie.npu.dma_cpy_nd %0([] [] [], [0, 0] [8, 16] [8, 1])
```
is transformed into:
```
amdaie.npu.dma_cpy_nd %0([] [] [], [0, 0, 0] [2, 8, 16] [0, 8, 1])
```

Another example:
```
amdaie.npu.dma_cpy_nd %0([] [] [], [0, 0, 0] [16, 8, 16] [32, 8, 1])
amdaie.npu.dma_cpy_nd %0([] [] [], [1, 0, 0, 0] [2, 16, 8, 16] [64, 32, 8, 1])
```
is transformed into:
```
amdaie.npu.dma_cpy_nd %0([] [] [], [0, 0, 0, 0] [3, 16, 8, 16] [64, 32, 8, 1])
```

PR #968, opened 2024-12-06
[CombineStridedOps] Add case for size(A) > size(B)

jtuyls:
I missed a case for `size(A) > size(B)` in https://github.com/nod-ai/iree-amd-aie/pull/963.

If `size(A) > size(B)` and `stridesA[0] == 0` and all other dimensions are equal, then access pattern A contains N repetitions of access pattern B and they can be combined into a single access pattern with N+1 repetitions.

Example:
```
A: offsets: [0, 0], sizes: [7, 16], strides: [0, 1]
B: offsets: [0], sizes: [16], strides: [1]
```
is transformed into:
```
offsets: [0, 0], sizes: [8, 16], strides: [0, 1]
```

Note that outer offset of A can have any value as it's anyway multiplied with `strideA[0] == 0`.


Abhishek-Varma (approved):
LGTM!

PR #971, opened 2024-12-09
[TargetCDO] Add support for implicit locks in BD chaining

jtuyls:
Adds support for a chain of DMA BD configurations with implicitly disabled locks:

```
^bb3:  // pred: ^bb5
  %1 = aie.dma_start(S2MM, 1, ^bb4, ^bb2, repeat_count = 1)
^bb4:  // 2 preds: ^bb3, ^bb5
  aie.use_lock(%lock_0_2, AcquireGreaterEqual, 1)
  aie.dma_bd(%buf1 : memref<8x2x8x4xi32>) {bd_id = 1 : i32, len = 256 : i32, next_bd_id = 2 : i32}
  aie.next_bd ^bb5
^bb5:  // 2 preds: ^bb4
  aie.dma_bd(%buf1 : memref<8x2x8x4xi32>) {bd_id = 2 : i32, len = 256 : i32, next_bd_id = 1 : i32, offset = 256 : i32}
  aie.use_lock(%lock_0_2_3, Release, 1)
  aie.next_bd ^bb4
```
Here, there are two circular chained BDs (bd_id 1 calls bd_id 2 as next bd, which calls bd_id 1 again etc). In this case, the first BD only needs to acquire a lock and the second one needs to release one. The first BD will have its release lock fields disabled and the second one will have its acquire lock fields disabled.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIETargetCDODirect.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIETargetCDODirect.cpp
@@ -71,10 +71,8 @@ LogicalResult configureLocksAndBd(Block &block, const TileLoc &tileLoc,
                                   const AMDAIEDeviceModel &deviceModel) {
   FailureOr<XAie_DmaDesc> dmaTileBd = initDMADesc(deviceModel, tileLoc);
   if (failed(dmaTileBd)) return failure();
-  assert(!block.getOps<UseLockOp>().empty() && "BD block has no lock-usage");
   std::optional<int> acqValue, relValue, acqLockId, relLockId;
   bool acqEn;
-  // switch (lock->getAc)
   for (auto op : block.getOps<UseLockOp>()) {
     // Only dyn_cast if you are going to check if it was of the type
     // expected; if you aren't checking use cast instead as it will at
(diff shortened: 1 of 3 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIETargetCDODirect.cpp
```diff
@@ -71,10 +71,8 @@ LogicalResult configureLocksAndBd(Block &block, const TileLoc &tileLoc,
                                   const AMDAIEDeviceModel &deviceModel) {
   FailureOr<XAie_DmaDesc> dmaTileBd = initDMADesc(deviceModel, tileLoc);
   if (failed(dmaTileBd)) return failure();
-  assert(!block.getOps<UseLockOp>().empty() && "BD block has no lock-usage");
```
Yu-Zhewen, line 74: why this line is deleted, and are we still expecting at least one acquire or one release?

Yu-Zhewen (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIETargetCDODirect.cpp
jtuyls: Added it: https://github.com/nod-ai/iree-amd-aie/pull/971/files#diff-14d46268eadd9d7fca77f66d72a2ab285546a8f3369b4f3b1ab8c3f5c15185c7R55

PR #972, opened 2024-12-09
[BufferizeToAllocation] Handle pack operands dependent on other op types

jtuyls:
Refactor `getOperandsFromDefOp` to search for and return `pack` ops up to a certain depth away from the linalg op's inputs and rename to `getPackOperands`. This enables more complex cases with intermediate `extract_slice` operations, see the new test cases being added.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
@@ -67,25 +67,30 @@ static SmallVector<Value> getInputOperands(linalg::LinalgOp &linalgOp) {
   return operands;
 }
 
-/// Utility to fetch operands from the defining ops of LinalgOp's input
-/// operands. For example, we want to fetch the input operand of %pack0 and
-/// %pack1 as shown in below, and promote them to memory.
-/// %pack0 = tensor.pack % arg0
-/// %pack1 = tensor.pack % arg1
-/// %pack2 = tensor.pack % pack0
-/// %pack3 = tensor.pack % pack1
-/// %generic = linalg.generic ins(%pack2, %pack3)
-static FailureOr<SmallVector<Value>> getOperandsFromDefOp(
-    linalg::LinalgOp &linalgOp) {
+/// Utility to fetch pack operands at a specified depth from the LinalgOp's
+/// input operands.
+static FailureOr<SmallVector<Value>> getPackOperands(linalg::LinalgOp linalgOp,
+                                                     uint32_t depthLevel) {
   SmallVector<Value> operands;
-  for (Value input : linalgOp.getDpsInputs()) {
-    auto defOp = input.getDefiningOp<tensor::PackOp>();
+  for (auto input : llvm::enumerate(linalgOp.getDpsInputs())) {
+    uint32_t currentLevel{0};
+    Operation *currentOp = input.value().getDefiningOp();
+    while (currentLevel < depthLevel && currentOp != nullptr) {
+      if (dyn_cast<tensor::PackOp>(currentOp)) {
+        currentLevel++;
+        if (currentLevel == depthLevel) break;
+      }
+      currentOp = currentOp->getOperand(0).getDefiningOp();
+    }
     // The defining op has to be a pack op, fail otherwise.
-    if (!defOp) {
-      return failure();
+    if (!currentOp) {
+      return linalgOp.emitOpError()
+             << "operand #" << input.index() << " only has pack ops to depth "
+             << currentLevel << ", but request is for a depth " << depthLevel
+             << " pack op.";
     }
     // We only want to fetch the input operand of the pack op.
-    operands.push_back(defOp->getOperand(0));
+    operands.push_back(currentOp->getResult(0));
   }
   return operands;
 }
@@ -94,7 +99,8 @@ static FailureOr<SmallVector<Value>> getOperandsFromDefOp(
 // ops, based on which operands the caller wants to bufferize via
 // `bufferizeOperand` parameter.
 static FailureOr<SmallVector<Value>> getOperandsToBufferize(
-    BufferizeOperand bufferizeOperand, linalg::LinalgOp &linalgOp) {
+    BufferizeOperand bufferizeOperand, linalg::LinalgOp &linalgOp,
+    uint32_t packDepth) {
   switch (bufferizeOperand) {
     /// Create new allocations for Lhs, Rhs and Out.
     case BufferizeOperand::InputOutput:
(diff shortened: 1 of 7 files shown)


newling:
Likes good, minor suggestions in https://github.com/jtuyls/iree-amd-aie/pull/15

One question -- why is the default depth 2, shouldn't it be 1 because the the bufferization in the current pipeline is always on the closest pack op to the matmul? Or maybe there should just be no default.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -99,9 +99,12 @@ def AMDAIEBufferizeToAllocation :
                    "Create new allocations for lhs, rhs of a linalg op."),
         clEnumValN(mlir::iree_compiler::AMDAIE::BufferizeOperand::Output, "output",
                    "Create new allocations for output of a linalg op."),
-        clEnumValN(mlir::iree_compiler::AMDAIE::BufferizeOperand::DefOp, "def-op",
-                   "Create new allocations for operands from the def ops of a linalg op.")
-    )}]>
+        clEnumValN(mlir::iree_compiler::AMDAIE::BufferizeOperand::Pack, "pack",
```
yzhang93, line 102: I feel `pack` may not be a good name for this situation. Maybe we should rename these options to something like "LinalgInput"/"LinalgOutput"/"PackInput"?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/bufferize_to_allocation.mlir
```diff
@@ -1,7 +1,7 @@
 // RUN: iree-opt --pass-pipeline='builtin.module(func.func(iree-amdaie-bufferize-to-allocation{memory-space=2 bufferize-operand=input-output}))' --split-input-file %s | FileCheck %s --check-prefix=INPUT-OUTPUT
 // RUN: iree-opt --pass-pipeline='builtin.module(func.func(iree-amdaie-bufferize-to-allocation{memory-space=2 bufferize-operand=input}))' --split-input-file %s | FileCheck %s --check-prefix=INPUT
 // RUN: iree-opt --pass-pipeline='builtin.module(func.func(iree-amdaie-bufferize-to-allocation{memory-space=2 bufferize-operand=output}))' --split-input-file %s | FileCheck %s --check-prefix=OUTPUT
-// RUN: iree-opt --pass-pipeline='builtin.module(func.func(iree-amdaie-bufferize-to-allocation{memory-space=1 bufferize-operand=def-op}))' --split-input-file %s | FileCheck %s --check-prefix=DEF-OP
+// RUN: iree-opt --pass-pipeline='builtin.module(func.func(iree-amdaie-bufferize-to-allocation{memory-space=1 bufferize-operand=pack pack-depth=2}))' --split-input-file --verify-diagnostics %s | FileCheck %s --check-prefix=PACK
```
yzhang93, line 4: I don't think you need to add `--verify-diagnostics`?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
```diff
@@ -67,25 +67,30 @@ static SmallVector<Value> getInputOperands(linalg::LinalgOp &linalgOp) {
...
+      currentOp = currentOp->getOperand(0).getDefiningOp();
+    }
     // The defining op has to be a pack op, fail otherwise.
-    if (!defOp) {
-      return failure();
+    if (!currentOp) {
+      return linalgOp.emitOpError()
+             << "operand #" << input.index() << " only has pack ops to depth "
+             << currentLevel << ", but request is for a depth " << depthLevel
+             << " pack op.";
```
yzhang93, line 90: I don't know if it is easy to add a test for this.

yzhang93 (changes requested):

yzhang93:
> > Looks good, minor suggestions in [jtuyls#15](https://github.com/jtuyls/iree-amd-aie/pull/15)
> > One question -- why is the default depth 2, shouldn't it be 1 because the the bufferization in the current pipeline is always on the closest pack op to the matmul? Or maybe there should just be no default.
> 
> Yeah, maybe it's better to set default to `1` instead.

I think currently `PackDepth` is only used within `getPackOperands` function and the tests are only with `pack-depth=2`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp
jtuyls: It's here: https://github.com/nod-ai/iree-amd-aie/pull/972/files#diff-08dc5392bd52ce08c6e138c89646fbccbec2f0c782d79323431ea1322919539dR102

jtuyls:
> I think we may keep default as 2, as currently `PackDepth` is only used within `getPackOperands` function and the existing tests are only with `pack-depth=2`. And option `bufferize-operand=pack pack-depth=1` seems to be a specific case for option `bufferize-operand=input`. Note the linalg input operand could be other operations than pack ops.

That's why I initially had it as `2`, but I think the default is not necessarily the one we're using, but the one that makes sense from a tranformation's perspective, for which I think `1` is a better default as in principal there are more cases with one level of packs than two.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
jtuyls: Done, renamed it.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -99,9 +99,12 @@ def AMDAIEBufferizeToAllocation :
                    "Create new allocations for lhs, rhs of a linalg op."),
         clEnumValN(mlir::iree_compiler::AMDAIE::BufferizeOperand::Output, "output",
                    "Create new allocations for output of a linalg op."),
-        clEnumValN(mlir::iree_compiler::AMDAIE::BufferizeOperand::DefOp, "def-op",
-                   "Create new allocations for operands from the def ops of a linalg op.")
-    )}]>
+        clEnumValN(mlir::iree_compiler::AMDAIE::BufferizeOperand::PackInput, "pack",
```
yzhang93, line 102: I would change this to `pack-input`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
yzhang93: Thanks! I see you didn't change the other options. Do you think it's more clear if we change option "Input" to "LinalgInput" and "Output" to "LinalgOutput", etc? This is not critical, so can be addressed in the future if needed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
jtuyls: Done

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
jtuyls: Yeah, I think that makes sense, but I wanted to avoid convoluting this PR with those changes.

yzhang93 (approved):

PR #975, opened 2024-12-09
[DistributeL1Allocations] Include cases when the offset is a user of affineApplyOp

yzhang93:
Without this fix, the L1 allocation in the new added test would fail to distribute.


newling (changes requested):
This change will require an update to the pass explanation I think, it doesn't mention adding of thread ids and block ids. This is a new idea to me, what is the context? 


https://github.com/nod-ai/iree-amd-aie/blob/f5ab91ebc3eb9fd7ffd8e53bd53aed434e893c5f/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td#L222

yzhang93:
> This change will require an update to the pass explanation I think, it doesn't mention adding of thread ids and block ids. This is a new idea to me, what is the context?
> 
> https://github.com/nod-ai/iree-amd-aie/blob/f5ab91ebc3eb9fd7ffd8e53bd53aed434e893c5f/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td#L222

I think the change doesn't affect the explanations. The context can be self-explained from the added test which has an additional nested loop around the inner/thread-level loop.

newling:
> > This change will require an update to the pass explanation I think, it doesn't mention adding of thread ids and block ids. This is a new idea to me, what is the context?
> > https://github.com/nod-ai/iree-amd-aie/blob/f5ab91ebc3eb9fd7ffd8e53bd53aed434e893c5f/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td#L222
> 
> I think the change doesn't affect the explanations. The context can be self-explained from the added test which has an additional nested loop around the inner/thread-level loop.

This line in the pass explanation:
> Each AIE core/tile is uniquely identified by gpu thread ids, usually 'y' (for AIE row) and 'x' (for AIE column).

As far as I can tell, there are 16 AIE cores in your example, but thread grid only enumerates 4 things. How can 16 things be uniquely identified by a set of 4 things?

newling (approved):
I'm looking forward to seeing how this is used!

PR #978, opened 2024-12-10
Remove testing of manual pipeline with iree-opt

newling:
Updating this test is causing issues, and I'd rather remove it than maintain it. The problem I hit is about an unregistered aievec pipeline. @Yu-Zhewen is hitting an error about `'memref.cast' op is an unsupported operation.` We can probably figure this out with some time, but I'm not sure it's worth it?


yzhang93:
> Updating this test is causing issues, and I'd rather remove it than maintain it. The problem I hit is about an unregistered aievec pipeline. @Yu-Zhewen is hitting an error about `'memref.cast' op is an unsupported operation.` We can probably figure this out with some time, but I'm not sure it's worth it?

About unregistered aievec pipeline, you probably should pass flag `--allow-unregistered-dialect `.

yzhang93 (approved):

jtuyls (approved):
All right, I though this was useful to be able to run and debug partial pipelines, but I haven't needed it a while, so if this causes issues in maintaining, I am ok removing it entirely.

PR #982, opened 2024-12-11
Fix max sizes in DmaDimConfig and refactor

jtuyls:
Some incorrect max size calculation in `DmaDimConfig` was blocking linear dimension folding and causing the following error when trying to increase the L1 matmul problem from 32x32x32 to 32x32x64 (MxNxK) for bf16:

```
<unknown>:0: error: 'aie.dma_bd' op Cannot give more than 3 dimensions for step sizes and wraps in this  tile (got 4 dimensions).
```

With this PR, that error should be resolved, but I am leaving the addition of that change to a separate PR. cc @yzhang93


newling (approved):
Looks good

yzhang93 (approved):
Thanks for fixing the issue.

PR #983, opened 2024-12-11
[AMDAIEFusePackIntoLoop] Handle case where there is no extract_slice 

newling:
This is very similar to https://github.com/nod-ai/iree-amd-aie/pull/976 

When the AIE grid we're using is `m x n` where one or both of `m` and `n` is `1`, for a matmul we get pack ops that produce operands for matmuls directly. i.e. as opposed to `pack->extract_slice->matmul` we have `pack->matmul`. 

TODO(newling) add a test and see if the pass can be simplified further.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
@@ -19,40 +19,30 @@ namespace mlir::iree_compiler::AMDAIE {
 
 namespace {
 
-FailureOr<Operation *> getParentBeforeLoop(BlockArgument blkArg) {
-  Operation *parent = blkArg.getOwner()->getParentOp();
-  LoopLikeOpInterface loop = dyn_cast<LoopLikeOpInterface>(parent);
-  if (!loop) return failure();
-  Operation *operandParent = loop.getTiedLoopInit(blkArg)->getOwner();
-  return operandParent;
-}
-
 /// A utility function specific to this pass which, given a value `operand`,
 /// traverses the def-chain till it finds a tensor.extract_slice. The 2 cases
 /// where it successfully finds and returns an extract_slice (SLICE) are:
 ///
 /// Case 1)
-/// pack -> block arg -> SLICE  -> pack -> pack -> pack -> operand
-///                                ^^^^^^^^^^^^^^^^^^^^
-///                                any number (>= 0) of trailing packs
-///
-/// Case 2)
 /// pack -> SLICE -> pack -> pack -> pack -> operand
 ///                  ^^^^^^^^^^^^^^^^^^^^
 ///                  any number (>= 0) of trailing packs
 ///
+/// Case 2)
+/// pack -> block arg -> SLICE  -> pack -> pack -> pack -> operand
+///                                ^^^^^^^^^^^^^^^^^^^^
+///                                any number (>= 0) of trailing packs
+///
+/// Case 2 only matches where `block arg` is for a loop operation.
 static FailureOr<tensor::ExtractSliceOp> getTensorExtractSliceDefiningOp(
     Value operand) {
   // roll back through all the packs immediately preceding `operand`.
-  while (operand.getDefiningOp() &&
-         isa<tensor::PackOp>(operand.getDefiningOp())) {
+  while (isa_and_present<tensor::PackOp>(operand.getDefiningOp())) {
     operand = operand.getDefiningOp()->getOperand(0);
   }
 
-  // If the parent of `operand` is not an extract_slice, return failure.
-  Operation *defOp = operand.getDefiningOp();
-  if (!defOp) return failure();
-  tensor::ExtractSliceOp sliceOp = dyn_cast<tensor::ExtractSliceOp>(defOp);
+  tensor::ExtractSliceOp sliceOp =
+      dyn_cast_if_present<tensor::ExtractSliceOp>(operand.getDefiningOp());
   if (!sliceOp) return failure();
 
   // Case 1 outlined above.
@@ -62,9 +52,11 @@ static FailureOr<tensor::ExtractSliceOp> getTensorExtractSliceDefiningOp(
 
   // Case 2 outlined above.
   else if (auto blkArg = dyn_cast<BlockArgument>(sliceOp.getSource())) {
-    FailureOr<Operation *> operandParent = getParentBeforeLoop(blkArg);
-    if (failed(operandParent)) return failure();
-    if (isa_and_present<tensor::PackOp>(operandParent.value())) return sliceOp;
+    Operation *parent = blkArg.getOwner()->getParentOp();
+    LoopLikeOpInterface loop = dyn_cast<LoopLikeOpInterface>(parent);
+    if (!loop) return failure();
+    Operation *operandParent = loop.getTiedLoopInit(blkArg)->getOwner();
+    if (isa_and_present<tensor::PackOp>(operandParent)) return sliceOp;
   }
 
   return failure();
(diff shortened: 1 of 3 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
```diff
@@ -18,37 +19,54 @@ namespace mlir::iree_compiler::AMDAIE {
...
-    }
-    // We perform further traversal only if we have tensor.pack op in the
-    // def-chain.
-    if (!isa<tensor::PackOp>(defOp)) {
-      break;
-    }
-    operand = defOp->getOperand(0);
+  // roll back through all the packs immediately preceding `operand`.
+  while (operand.getDefiningOp() &&
+         isa<tensor::PackOp>(operand.getDefiningOp())) {
```
yzhang93, line 48: I think you can use `while (isa_and_present<tensor::PackOp>(operand.getDefiningOp()))`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
```diff
@@ -18,37 +19,54 @@ namespace mlir::iree_compiler::AMDAIE {
...
+    operand = operand.getDefiningOp()->getOperand(0);
+  }
+
+  // If the parent of `operand` is not an extract_slice, return failure.
+  Operation *defOp = operand.getDefiningOp();
+  if (!defOp) return failure();
+  tensor::ExtractSliceOp sliceOp = dyn_cast<tensor::ExtractSliceOp>(defOp);
+  if (!sliceOp) return failure();
```
yzhang93, line 56: Like commented above, you can simply check `dyn_cast_if_present<tensor::ExtractSliceOp>(operand.getDefiningOp())`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
```diff
@@ -122,29 +135,32 @@ void AMDAIEFusePackIntoLoopPass::runOnOperation() {
...
-                                          MutableArrayRef(&loops, 1));
-      if (!fusedProducer) {
-        funcOp->emitOpError("Failed to fuse pack ops into for loop.");
-        return signalPassFailure();
+      // Case where operand of generic is a pack op which is in a different
+      // block than the generic's block.
+      else {
+        if (auto parent =
+                dyn_cast_if_present<tensor::PackOp>(operand.getDefiningOp())) {
+          Block *genericBlock = genericOp->getBlock();
```
yzhang93, line 158: I think here you need to use `else if`, and then add another `else` in which adding some debug message and return.

yzhang93 (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
newling: done

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
newling: done

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
```diff
@@ -18,37 +19,54 @@ namespace mlir::iree_compiler::AMDAIE {
 
 namespace {
 
-/// A utility function specific to this pass which, given a value, would
-/// traverse the def-chain till it either finds a tensor.extract_slice op or a
-/// BlockArgument.
+FailureOr<Operation *> getParentBeforeLoop(BlockArgument blkArg) {
+  Operation *parent = blkArg.getOwner()->getParentOp();
+  LoopLikeOpInterface loop = dyn_cast<LoopLikeOpInterface>(parent);
+  if (!loop) return failure();
```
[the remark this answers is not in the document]
newling: No failure here isn't necessary

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
```diff
@@ -18,37 +19,54 @@ namespace mlir::iree_compiler::AMDAIE {
 
 namespace {
 
-/// A utility function specific to this pass which, given a value, would
-/// traverse the def-chain till it either finds a tensor.extract_slice op or a
-/// BlockArgument.
+FailureOr<Operation *> getParentBeforeLoop(BlockArgument blkArg) {
```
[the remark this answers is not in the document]
newling: I've inlined the function

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
yzhang93: Logically the else condition is not needed, but I was thinking it may good to add some debug message so people know why the pack is not fused into the loop in some specific situation. You can decide whether to add it.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
```diff
@@ -18,37 +19,46 @@ namespace mlir::iree_compiler::AMDAIE {
...
+      dyn_cast_if_present<tensor::ExtractSliceOp>(operand.getDefiningOp());
+  if (!sliceOp) return failure();
+
+  // Case 1 outlined above.
+  if (isa_and_present<tensor::PackOp>(sliceOp.getSource().getDefiningOp())) {
+    return sliceOp;
+  }
+
+  // Case 2 outlined above.
+  else if (auto blkArg = dyn_cast<BlockArgument>(sliceOp.getSource())) {
```
yzhang93, line 54: I would use `if` instead.

yzhang93 (approved):
Thanks. Just one more comment.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
newling: We don't want it to enter both statements which might happen if this is `if`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
newling: The operand might not even come from a pack though, it could be anything...

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFusePackIntoLoop.cpp
yzhang93: How would it enter both statements? After it enters the first it will return.

newling: no it doesn't?

PR #984, opened 2024-12-11
[AMDAIESplitLogicalObjFifos] Handle case where the AIE array used in `m x n` where `m=1` or `n=1` 

newling:
Without this change, the split dimension has size equal to a matmul tiling size, in the case I saw 32. This is not definitely bad! i,e, from shape `1x1x32x64` the algorithm finds the first non-unit dimension from the left which is 32, and splits the memory and objectfifo into 32 smaller ones. 

TODO(newling) add a test, and see if you can generalize this further to make the pass more robust 

with this + https://github.com/nod-ai/iree-amd-aie/pull/976 + https://github.com/nod-ai/iree-amd-aie/pull/983 the lowering for "skinny grids" is better, hitting error: 'aie.memtile_dma' op could not find and assign a valid BD id (see also https://github.com/nod-ai/iree-amd-aie/pull/955#pullrequestreview-2479790539 )

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
@@ -6,7 +6,6 @@
 #include "iree-amd-aie/IR/AMDAIEOps.h"
 #include "iree-amd-aie/Transforms/Passes.h"
 #include "iree-amd-aie/Transforms/Utils/AMDAIELogicalObjFifoSplittingUtils.h"
-#include "mlir/IR/Iterators.h"
 #include "mlir/Pass/Pass.h"
 
 #define DEBUG_TYPE "iree-amdaie-split-logical-objectfifos"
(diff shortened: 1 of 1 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -59,9 +59,9 @@ LogicalResult collectSplittingDims(
     auto iter = std::find_if(memrefShape.begin(), memrefShape.end(),
                              [](int64_t size) { return size > 1; });
     size_t objFifoSplitDim = std::distance(memrefShape.begin(), iter);
-    // If all dimensions are unit (1), no splitting can be done, so continue to
-    // the next pair.
-    if (objFifoSplitDim >= memrefShape.size()) continue;
+    // Only consider splitting on dimensions 0 and 1.
+    if (objFifoSplitDim >= 2) continue;
```
jtuyls, line 63: It would be better to just assume that we can only split on the outer dimension (dim 0) and skip/continue if that one is a unit dimension (no splitting). The packing permutation can/have been adjusted so that this should work and it shouldn't have any performance impact.

jtuyls (changes requested):

newling:
Currently fails with 

```
error: 'amdaie.npu.dma_cpy_nd' op no BD ID available
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
newling: It seems like there's been a change since I last tried this that makes it obsolete -- nrows=ncols=1 now works on main :-)

PR #986, opened 2024-12-12
[AMDAIEFoldDmaWaits] Fold DMA wait operations across multi columns

Yu-Zhewen:
This is an enhancement for https://github.com/nod-ai/iree-amd-aie/pull/962.

In the previous PR, DMA waits on the same `connection` (and the same tile) could be folded, exploiting the fact that each DMA channel has a queue size of 4.

In this PR, DMA waits across multiple `columns` can also be folded, provided their corresponding `row`, `channel`, and `direction` are the same. This optimization leverages the ability to specify `colNum` in `TCTSync`, where the range `[col, col + colNum)` can be addressed.

The numbers in the following table show the instruction size in words.
| Test (MxKxN)  | No Folding | Only Fold by Connection | Only Fold by Column | Fold Both |
|---------------|------------|--------------------|----------------|-----------|
| 512x4096x512  | 1228       | 1132               | 1120           | 1096      |
| 512x512x4096  | 820        | 772                | 748            | 736       |
| 4096x512x512  | 4628       | 4244               | 4220           | 4124      |


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEControlCodeToTransaction.cpp
```diff
@@ -200,16 +200,34 @@ LogicalResult convertOp(AMDAIE::NpuAddressPatchOp op,
...
+        asyncToken.getDefiningOp());
     if (!pushToQueueOp) {
       return op.emitOpError()
-             << "should operate on an `amdaie.push_to_queue` op";
+             << "should operate on an `amdaie.push_to_queue` op async token";
     }
+    if (!columnBatches.empty()) {
+      auto &[lastPushOp, lastColNum] = columnBatches.back();
+      if (lastPushOp.getRow() == pushToQueueOp.getRow() &&
+          lastPushOp.getCol() + lastColNum == pushToQueueOp.getCol() &&
```
jtuyls, line 216: Shouldn't you sort the tokens on column number for this to always result the largest batches? Suppose the tokens are provided as (all have same row, direction, channel):

```
token_0: col == 0
token_1: col == 2
token_2: col == 1
```
With the current logic, won't this result in two TCT syncs:

```
TCT sync 1: token_0, token_2
TCT sync 2: token_1
```

While if you sort on column first:
```
token_0: col == 0
token_2: col == 1
token_1: col == 2
```
This would result in one TCT sync?
```
TCT sync 1: token_0, token_2, token_1
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEControlCodeToTransaction.cpp
```diff
@@ -200,16 +200,34 @@ LogicalResult convertOp(AMDAIE::NpuAddressPatchOp op,
 }
 
 LogicalResult convertOp(AMDAIE::NpuDmaWaitOp op, TransactionBuilder &builder) {
-  for (Value token : op.getAsyncTokens()) {
-    auto pushToQueueOp =
-        dyn_cast_if_present<AMDAIE::NpuPushToQueueOp>(token.getDefiningOp());
+  // Batch DMA operations with the same row, channel, and direction into a
+  // single TCT sync operation, as long as they have consecutive columns.
+  SmallVector<std::pair<AMDAIE::NpuPushToQueueOp, uint32_t>> columnBatches;
```
jtuyls, line 205: Shouldn't this be a map?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/controlcode_to_transaction.mlir
```diff
@@ -153,6 +153,59 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
...
+#executable_target_amdaie_xclbin_fb = #hal.executable.target<"amd-aie", "amdaie-xclbin-fb", {target_device = "npu1_4col", ukernels = "none"}>
+module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb} {
+  func.func @async_push_to_queue_and_wait_col_num() {
+    amdaie.workgroup {
+      amdaie.controlcode {
+        %0 = amdaie.npu.push_to_queue async {bd_id = 0 : ui32, channel = 0 : ui32, col = 0 : ui32, direction = 1 : i32, repeat_count = 1 : ui32, row = 0 : ui32}
+        %1 = amdaie.npu.push_to_queue async {bd_id = 0 : ui32, channel = 0 : ui32, col = 1 : ui32, direction = 1 : i32, repeat_count = 1 : ui32, row = 0 : ui32}
+        %2 = amdaie.npu.push_to_queue async {bd_id = 0 : ui32, channel = 0 : ui32, col = 2 : ui32, direction = 1 : i32, repeat_count = 1 : ui32, row = 0 : ui32}
+        %3 = amdaie.npu.push_to_queue async {bd_id = 0 : ui32, channel = 0 : ui32, col = 3 : ui32, direction = 1 : i32, repeat_count = 1 : ui32, row = 0 : ui32}
+        amdaie.npu.dma_wait(%0, %1, %2, %3 : !amdaie.async_token, !amdaie.async_token, !amdaie.async_token, !amdaie.async_token)
```
jtuyls, line 199: Could you add test cases as well for:
- Tokens being provided in a different order to the `dma_wait`
- Different `push_to_queue` ops with different columns, channels, directions

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
```diff
@@ -152,6 +153,147 @@ LogicalResult foldDmaWaits(const AMDAIE::AMDAIEDeviceModel &deviceModel,
...
+  // Sorted by column.
+  std::map<uint32_t, AMDAIE::NpuDmaWaitOp> colWaitOpMap;
+};
+
+/// Updates a batch of asynchronous DMA wait operations by combining their
+/// async tokens into a single NpuDmaWaitOp.
+void updateColumnBatchTokens(
+    IRRewriter &rewriter,
+    std::map<uint32_t, AMDAIE::NpuDmaWaitOp> &colWaitOpMap) {
+  if (colWaitOpMap.size() < 2) return;
```
jtuyls, line 170: What does this do?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
```diff
@@ -152,6 +153,147 @@ LogicalResult foldDmaWaits(const AMDAIE::AMDAIEDeviceModel &deviceModel,
...
+    rewriter.create<AMDAIE::NpuDmaWaitOp>(waitOps.back().getLoc(), asyncTokens);
+    for (AMDAIE::NpuDmaWaitOp waitOp : waitOps) {
+      rewriter.eraseOp(waitOp);
+    }
+  }
+}
+
+/// Utility function to determine if a DMA wait operation can be folded.
+/// This is achieved by verifying whether it shares the same row, channel,
+/// and direction with preceding wait operations.
```
jtuyls, line 203: You don't necessarily need to check whether they share the same row, channel, direction and you could just batch all tokens until some other operation touches the same connection and then leave the column batch discovery to `ControlCodeToTransactionLowering`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/fold_dma_waits.mlir
```diff
@@ -220,3 +220,93 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
...
+        memref.assume_alignment %7, 64 : memref<64x32xi32>
+        %bd_id = amdaie.bd_id(%tile_0_0, %c0)
+        %18 = amdaie.npu.half_dma_cpy_nd async %10(%15 [] [] [] bd_id = %bd_id channel = %channel) : !amdaie.logicalobjectfifo<memref<2048xi32>>
+        %bd_id_15 = amdaie.bd_id(%tile_1_0, %c0)
+        %19 = amdaie.npu.half_dma_cpy_nd async %12(%16 [] [] [] bd_id = %bd_id_15 channel = %channel_11) : !amdaie.logicalobjectfifo<memref<2048xi32>>
+        %bd_id_16 = amdaie.bd_id(%tile_3_0, %c0)
+        %20 = amdaie.npu.half_dma_cpy_nd async %14(%17 [] [] [] bd_id = %bd_id_16 channel = %channel_13) : !amdaie.logicalobjectfifo<memref<2048xi32>>
+        amdaie.npu.dma_wait(%18 : !amdaie.async_token)
+        amdaie.npu.dma_wait(%19 : !amdaie.async_token)
+        amdaie.npu.dma_wait(%20 : !amdaie.async_token)
```
jtuyls, line 306: Could you add tests for more complex cases, cases that should fail and potential corner cases? For example:

Assuming `%connection_0` and `%connection_1` use the same channel, direction, row:
```
%0 = amdaie.npu.half_dma_cpy_nd %connection_0 
amdaie.npu.dma_wait(%0 : !amdaie.async_token)
%1 = amdaie.npu.half_dma_cpy_nd %connection_0 
%2 = amdaie.npu.half_dma_cpy_nd %connection_1
amdaie.npu.dma_wait(%2 : !amdaie.async_token)
```
In this case, the `%0` and `%2` tokens should not be batched as `%1 = amdaie.npu.half_dma_cpy_nd %connection_0` touches the same connection in between.

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEControlCodeToTransaction.cpp
Yu-Zhewen: The sorting is actually done inside through this map.
https://github.com/nod-ai/iree-amd-aie/blob/022bdf813d38d9888893519af1c991ce9c2d3dd3/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp#L161-L162

In the current implementation, both batching and sorting are performed in `AMDAIEFoldDmaWaits`.  Later, in `AMDAIEControlCodeToTransaction`, the batching condition is rechecked (matching `row`, `channel`, and `direction`)  to ensure that no other passes have altered the operations in between. Sorting, however, is not repeated.

For debugging purposes, it might be better to enforce this as an assertion: a single `dma_wait` operation 
should not result in multiple `TCT_sync` operations. This would prevent potential confusion when debugging 
by maintaining a clear one-to-one correspondence between `dma_wait` and its resulting control operations.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEControlCodeToTransaction.cpp
Yu-Zhewen: It's to maintain the insertion order for `NpuPushToQueueOp`, which will be used to generate the transaction code.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
Yu-Zhewen: Similar reason as before, I feel maintaining a clear one-to-one correspondence between `dma_wait` and `TCT_sync` probably is easier for debugging.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEControlCodeToTransaction.cpp
jtuyls: > In the current implementation, both batching and sorting are performed in AMDAIEFoldDmaWaits.

That's a different pass, so we shouldn't make that assumption and just sort again here.

> For debugging purposes, it might be better to turn this to an assertion: a single dma_wait operation
should not result in multiple TCT_sync operations. 

Why not? You could put wait for multiple tokens, which is what the current implementation supports.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEControlCodeToTransaction.cpp
jtuyls: You can maintain insertion order with a map as well, like you're doing at other places.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
jtuyls: It would be good to add a comment here that this skips in case `batch == 1`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
jtuyls: There is not such one-to-one correspondence assumed in the `NpuDmaWaitOp` as you can wait on multiple tokens at the same time, possibly resulting in multiple TCT syncs. This is supported in the current implementation and enables more compact IR as well as the optimization you're adding here. I don't see a need to limit  `NpuDmaWaitOp` to a single TCT sync.

Yu-Zhewen: I understand it allows a more compact IR, but inside `ControlCodeToTransaction`, `NpuDmaWaitOp` is the only operation that does not have one-to-one correspondence. Unlike a DMA copy, which follows a clear transformation path (`dmaCpy` → `halfDmaCpy` → `pushToQueue` -> `TXN_OPC_WRITE`), I feel it is unclear what `NpuDmaWaitOp` represents, as its definition is a bit far away from `TCTSync`. 

This lack of one-to-one correspondence also puts the `FoldDmaWaits` pass in an awkward position. Although the pass reduces the number of `NpuDmaWaitOp`, there is no guarantee that this will result in fewer `TCTSync` operations.

May I suggest introducing an intermediate operation between `NpuDmaWaitOp` and `TCTSync`? This intermediate operation would only allow multiple tokens if they can be grouped into a single `TCTSync`. The `FoldDmaWaits` pass could then focus on optimizing at this intermediate level, while `ControlCodeToTransaction` only handles one-to-one conversion.

jtuyls: > May I suggest introducing an intermediate operation between NpuDmaWaitOp and TCTSync? This intermediate operation would only allow multiple tokens if they can be grouped into a single TCTSync. The FoldDmaWaits pass could then focus on optimizing at this intermediate level, while ControlCodeToTransaction only handles one-to-one conversion.

I think a lower-level TCT-like operation makes sense, with `NpuDmaWaitOp` to `TCTSyncOp` lowering part of the ControlCodeLowering pass. If you go this route, create a separate PR though that introduces `TCTSyncOp` with rationale you have above. 

I am not convinced though that `FoldDmaWaits` should operate on the `TCTSyncOps` though. I kind of like that it's done at the higher level and just tries to batch as many syncs as possible because:
1. As a by-product, it results in more compact IR with clear boundaries
2. If new/future sync optimization opportunities become available (different channels), this pass can just stay the same and more low-level checks/transformations can take care of finding and leveraging the new capabilities within a batch. 
3. I don't see a case now or in the future in which we wouldn't want to batch waits.

> Although the pass reduces the number of NpuDmaWaitOp, there is no guarantee that this will result in fewer TCTSync operations.

I think that's fine, the DMA wait batching is general enough to potentially help in a lot of different (future) cases, without any obvious disadvantages, see reasoning above.

Yu-Zhewen: Sure, I will make a separate PR for adding such `TCTSyncOp`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
```diff
@@ -152,6 +149,116 @@ LogicalResult foldDmaWaits(const AMDAIE::AMDAIEDeviceModel &deviceModel,
...
+LogicalResult updateBatchTokens(IRRewriter &rewriter,
+                                SmallVector<AMDAIE::NpuDmaWaitOp> &waitOps) {
+  // Skip if there are less than two DMA wait operations.
+  if (waitOps.size() < 2) return success();
+
+  SmallVector<Value> asyncTokens;
+  Operation *parentOp = waitOps[0]->getParentOp();
+  for (AMDAIE::NpuDmaWaitOp waitOp : waitOps) {
+    if (waitOp->getParentOp() != parentOp) {
+      return waitOp.emitError(
```
jtuyls, line 162: Could you add a test case for this? And I think this doesn't have to fail and could just result in no rewrite if you put it in `canFoldByBatch` instead.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
```diff
@@ -152,6 +149,116 @@ LogicalResult foldDmaWaits(const AMDAIE::AMDAIEDeviceModel &deviceModel,
...
+    }
+    asyncTokens.append(waitOp.getAsyncTokens().begin(),
+                       waitOp.getAsyncTokens().end());
+  }
+
+  rewriter.setInsertionPointAfter(waitOps.back());
+  rewriter.create<AMDAIE::NpuDmaWaitOp>(waitOps.back().getLoc(), asyncTokens);
+  for (AMDAIE::NpuDmaWaitOp waitOp : waitOps) {
+    rewriter.eraseOp(waitOp);
+  }
```
jtuyls, line 173:
```suggestion
  for (AMDAIE::NpuDmaWaitOp waitOp : waitOps) rewriter.eraseOp(waitOp);
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
Yu-Zhewen: New test added
https://github.com/nod-ai/iree-amd-aie/blob/fb6d4d28130f48120d73fe535daf54def6542acc/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/fold_dma_waits.mlir#L73

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/fold_dma_waits.mlir
```diff
@@ -141,8 +141,171 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
...
+        memref.assume_alignment %4, 64 : memref<64x32xi32>
+        %bd_id = amdaie.bd_id(%tile_0_0, %c0)
+        %12 = amdaie.npu.half_dma_cpy_nd async %7(%10 [] [] [] bd_id = %bd_id channel = %channel) : !amdaie.logicalobjectfifo<memref<2048xi32>>
+        amdaie.npu.dma_wait(%12 : !amdaie.async_token)
+        %bd_id_9 = amdaie.bd_id(%tile_0_0, %c0)
+        %13 = amdaie.npu.half_dma_cpy_nd async %7(%10 [] [] [] bd_id = %bd_id_9 channel = %channel) : !amdaie.logicalobjectfifo<memref<2048xi32>>
+        %bd_id_10 = amdaie.bd_id(%tile_1_0, %c0)
+        %14 = amdaie.npu.half_dma_cpy_nd async %9(%11 [] [] [] bd_id = %bd_id_10 channel = %channel_7) : !amdaie.logicalobjectfifo<memref<2048xi32>>
+        amdaie.npu.dma_wait(%13 : !amdaie.async_token)
+        amdaie.npu.dma_wait(%14 : !amdaie.async_token)
```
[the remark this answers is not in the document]
Yu-Zhewen: Added here
https://github.com/nod-ai/iree-amd-aie/blob/fb6d4d28130f48120d73fe535daf54def6542acc/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/fold_dma_waits.mlir#L292

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
```diff
@@ -62,93 +107,242 @@ FailureOr<bool> canFoldBasedOnHalfDmaCpy(
...
+      connectionOps.empty() || !isSameScope) {
+    // Clear the BD IDs for all the connections in the batch.
+    for (auto &entry : dmaBdIdsMap) {
+      ConnectionOp connectionOp = entry.first.second;
+      DenseSet<uint32_t> &bdIds = entry.second;
+      if (connectionOps.contains(connectionOp)) bdIds.clear();
+    }
+    connectionOps.clear();
+    canFold = false;
+  }
```
jtuyls, line 282: I think it would be more clear to just return a bool from this function and do the clearing outside this function. This makes the function more standalone (single responsibility) and limited to what the name suggests (`canFoldByBatch`).

jtuyls: And then you can also return early, for example:
```
bool isSameScope = npuHalfDmaCpyNdOp->getParentOp() == batchParentOp;
if (!isSameScope) return false;
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
```diff
@@ -62,93 +107,242 @@ FailureOr<bool> canFoldBasedOnHalfDmaCpy(
...
-    bdIdQueue.clear();
+  // Can't fold wait op if:
+  // (1) the current BD ID on the same tile already occurs in the queue, or
+  // (2) the current operation is a packet flow, or
+  // (3) reaches the maximum queue size, or
+  // (4) the queue is empty, or
+  // (5) the current operation is not in the same scope as the queue.
+  if (isDuplicateBdId || isPacketFlow || bdIds.size() >= maxQueueSize ||
+      bdIds.empty() || !isSameScope) {
+    bdIds.clear();
```
jtuyls, line 124: I know this is pre-existing logic, but try to avoid changing data in `can/is...` functions to make more descriptive and clear of what is going on (not for this PR).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
```diff
@@ -16,13 +16,58 @@ namespace mlir::iree_compiler::AMDAIE {
...
 
-/// Utility function to determine whether a DMA wait op can be folded based on
-/// its half DMA copy operation.
-FailureOr<bool> canFoldBasedOnHalfDmaCpy(
-    const AMDAIE::AMDAIEDeviceModel &deviceModel,
+using DmaBdIdKey = std::pair<AMDAIE::TileOp, AMDAIE::ConnectionOp>;
+
+/// Utility function to erase the DMA wait operations in the queue, except for
+/// the last one.
+LogicalResult eraseQueueOperations(IRRewriter &rewriter,
```
jtuyls, line 23: I think this is more of a refactoring change and isn't specifically needed for DMA folding across multiple columns? If so, try to avoid combining multiple changes in one PR as it makes reviewing harder and it makes it unclear where these changes come from if someone needs to look at this in the future.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
Yu-Zhewen: It is now separated into two functions, one just returns a bool

https://github.com/nod-ai/iree-amd-aie/blob/cc1ae9e027fc2a580421325de63725b90896cdfd/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp#L236

and the other does the update/clear

https://github.com/nod-ai/iree-amd-aie/blob/cc1ae9e027fc2a580421325de63725b90896cdfd/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp#L267

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
Yu-Zhewen: will have them in a separate PR

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
```diff
@@ -152,6 +151,179 @@ LogicalResult foldDmaWaits(const AMDAIE::AMDAIEDeviceModel &deviceModel,
...
+  });
+
+  // Can't fold wait op if:
+  // (1) the current connection op already occurs in the batch, or
+  // (2) the current BD ID on the same tile already occurs in the batch, or
+  // (3) the current operation is a packet flow, or
+  // (4) the batch is empty, or
+  // (5) the current operation is not in the same scope as the batch.
+  return !(isDuplicateConnection || isDuplicateBdId || isPacketFlow ||
+           connectionOps.empty() || !isSameScope);
```
jtuyls, line 237: Could you put these checks earlier? For example:
```
bool isSameScope = currHalfDmaCpyNdOp->getParentOp() == batchParentOp;
if (!isSameScope) return false;
```
This makes the code easier to read (variables are defined close to where they are used, boolean checks are trivial) and also avoids retrieving information that is not needed anymore. For example, you don't need to retrieve the connection op and do all the other checks if the ops are not in the same scope.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
```diff
@@ -152,6 +151,179 @@ LogicalResult foldDmaWaits(const AMDAIE::AMDAIEDeviceModel &deviceModel,
...
+  AMDAIE::BdIdOp bdIdOp = maybeBdIdOp.value();
+  currBdIdVal = getConstantIndexOrAssert(bdIdOp.getValue());
+
+  // Retrieve the tile op.
+  AMDAIE::TileOp tileOp =
+      dyn_cast_if_present<AMDAIE::TileOp>(bdIdOp.getTile().getDefiningOp());
+  if (!tileOp) {
+    return bdIdOp.emitOpError() << "must operate on an `amdaie.tile`";
+  }
+  currBdIdKey = {tileOp, connectionOp};
```
jtuyls, line 224: Could you also avoid updating `currBdIdKey` and `currBdIdVal` inside this function and leave that outside as these are unclear side effects of the function?

Yu-Zhewen: Also making this as a separate functino now.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
```diff
@@ -152,6 +184,162 @@ LogicalResult foldDmaWaits(const AMDAIE::AMDAIEDeviceModel &deviceModel,
...
+/// (1) the current operation is not in the same scope as the batch, or
+/// (2) the current connection op already occurs in the batch, or
+/// (3) the batch is empty, or
+/// (4) the current operation is a packet flow, or
+/// (5) the current BD ID on the same tile already occurs in the batch.
+FailureOr<bool> canFoldByBatch(
+    const Operation *batchParentOp,
+    const DenseSet<AMDAIE::ConnectionOp> &connectionOps,
+    const DenseMap<DmaBdIdKey, DenseSet<uint32_t>> &dmaBdIdsMap,
+    AMDAIE::NpuHalfDmaCpyNdOp currHalfDmaCpyNdOp, DmaBdIdPair &currBdIdPair) {
```
jtuyls, line 222:
```suggestion
    AMDAIE::NpuHalfDmaCpyNdOp currHalfDmaCpyNdOp, DmaBdIdPair currBdIdPair) {
```
Non-expensive pairs are preferably passed by value as it's more simple, avoids reference issues (e.g. dangling, currBdIdPair shouldn't be changed ) and is as fast/faster if the contents are small.

jtuyls (approved):
LGTM

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFoldDmaWaits.cpp
Yu-Zhewen: implemented in #990.

PR #987, opened 2024-12-12
[Tests] Add missing tests for matmul-transpose-b and 2x2 cores

yzhang93:


build_tools/ci/run_matmul_test.sh
```diff
@@ -677,6 +675,31 @@ run_matmul_test_on_shapes ${bf16_i8_shapes_medium[@]} \
...
+    --do_transpose_rhs "1"
+
+run_matmul_test_on_shapes ${transpose_shapes[@]} \
+    --name_prefix "transpose_i8" \
+    --lower_to_aie_pipeline "objectFifo" \
+    --tile_pipeline "pack-peel" \
+    --lhs_rhs_type "i8" \
+    --acc_type "i32" \
+    --num_repeat_runs "2" \
+    --do_transpose_rhs "1"
```
jtuyls, line 702: As we're trying to move everything to `run.py`, we should avoid adding more tests to the bash script and add them in `run.py` instead.

jtuyls (approved):
LGTM

PR #990, opened 2024-12-16
Add NpuTctSync operation

Yu-Zhewen:
This PR introduces the `tct_sync` operation, designed to replace `dma_wait` during the `ControlCodeLowering` pass. The primary motivation for this change is to ensure a one-to-one conversion in the subsequent `ControlCodeToTransaction` pass. 

Checks for the same row, channel, and direction are now handled during the `ControlCodeLowering` pass instead of the `ControlCodeToTransaction` pass. Some unit tests have also been relocated.


compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
```diff
@@ -1022,6 +1022,31 @@ def AMDAIE_NpuWriteBdOp: AMDAIE_Op<"npu.write_bd"> {
   let assemblyFormat = [{ attr-dict }];
 }
 
+def AMDAIE_NpuTctSyncOp: AMDAIE_Op<"npu.tct_sync"> {
```
jtuyls, line 1025: Could you add a roundtrip test for this op as well?

jtuyls (approved):
LGTM

PR #995, opened 2024-12-18
[Tests] Add matmul-transpose-b performance run

Yu-Zhewen:


build_tools/ci/cpu_comparison/run.py
```diff
@@ -1660,8 +1843,20 @@ def __init__(self):
             if outline:
                 name_suffix += "_outline"
 
+            if (transpose_a, transpose_b) == (False, False):
+                correctness_test_class = Matmul
```
jtuyls, line 1847: Maybe use camelcase here for the class name for clarity and consider `NumericTestClass` instead of `CorrectnessTestClass`.

jtuyls (approved):

PR #996, opened 2024-12-19
Add vectorization support for matmul_transpose_a ops

yzhang93:
With this PR, `mamtul_transpose_a` ops work with bf16->f32 vectorization. But i8->i32 tests don't work because of the following op
```
%10 = vector.transpose %9, [1, 0] : vector<8x4xi8> to vector<4x8xi8>
```
failed to lower to an `aievec.shuffle` op. 

 https://github.com/nod-ai/iree-amd-aie/blob/befab5e1adb9a884d08f96bffb0b189f80f19aaa/compiler/plugins/target/AMD-AIE/aievec/VectorToAIEVecConversions.cpp#L244

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_loops_for_vectorization.mlir
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_loops_for_vectorization.mlir
@@ -140,30 +140,6 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb} 
     return %0 : !t3_f32
   }
 
-  // Check that the final 3 dimensions do have the pattern of a matmul (or matmul transpose)
-  // CHECK-LABEL: batched1
-  // CHECK-NOT:     scf.for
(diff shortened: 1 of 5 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_loops_for_vectorization.mlir
```diff
@@ -140,30 +140,6 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
     return %0 : !t3_f32
   }
 
-  // Check that the final 3 dimensions do have the pattern of a matmul (or matmul transpose)
```
jtuyls, line 143: Why does this test need to be removed?

jtuyls (approved):
LGTM

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_loops_for_vectorization.mlir
yzhang93: You are right, this shouldn't be removed. I'll just modify the checks.

PR #1002, opened 2024-12-30
[ControlCodeToTransaction] Use air-rt's serializer to generate transaction

Yu-Zhewen:


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/controlcode_to_transaction.mlir
```diff
@@ -15,7 +15,7 @@ module {
 // -----
 
 // CHECK:       0x06030100
-// CHECK:       0x00000105
+// CHECK:       0x00000104
```
Yu-Zhewen, line 18: It was previously hardcoded as `0x00000105`, which was wrong as the last byte stands for the number of columns, and we are using `npu1_4col`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/controlcode_to_transaction.mlir
```diff
@@ -68,10 +68,10 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
...
 // CHECK:       0x00000001
 // CHECK:       0x00000028
-// CHECK:       0x00000000
+// CHECK:       0x00140000
```
Yu-Zhewen, line 74: This is a workaround to accommodate `aie-rt`.

I believe the following implementation of `_XAie_GetRowfromRegOff` is actually wrong, as it behaves incorrectly by always preserving the lowest 8 bits.
https://github.com/Xilinx/aie-rt/blob/1470f6c1d0e765a73c741ad51bc761515d74457e/driver/src/common/xaie_helper.c#L1024-L1027

In contrast, `_XAie_GetColfromRegOff` appears to be correctly implemented and performs as expected.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_configure.cc
```diff
@@ -162,14 +164,19 @@ LogicalResult configureDMABD(
     TRY_XAIE_API_LOGICAL_RESULT(XAie_DmaSetPadding, &dmaDesc, &dmaPadTensor);
   }
 
+  if (maybeIter.has_value()) {
+    auto iter = maybeIter.value();
```
jtuyls, line 168:
```suggestion
    BDIterLayout iter = maybeIter.value();
```
Use explicit type.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEControlCodeToTransaction.cpp
```diff
@@ -223,33 +181,22 @@ LogicalResult convertOp(AMDAIE::NpuWriteBdOp op, TransactionBuilder &builder) {
...
-      std::max((int64_t)strides[strides.size() - 2] - 1, (int64_t)0);
-  uint32_t d2Stride =
-      std::max((int64_t)strides[strides.size() - 3] - 1, (int64_t)0);
-  uint32_t iterationSize =
-      std::max((int64_t)op.getIterationSize() - 1, (int64_t)0);
+  // Strides and iteration_size will be encoded as `actual - 1`, so we need to
+  // ensure they are at least 1.
+  SmallVector<int32_t> stridesNew(strides.begin(), strides.end());
+  for (auto &stride : stridesNew)
+    stride = std::max(int64_t(stride), int64_t(1));
```
jtuyls, line 192:
```suggestion
  std::for_each(stridesNew.begin(), stridesNew.end(), [](int32_t &stride) { stride = std::max(stride, (int32_t)1)});
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEControlCodeToTransaction.cpp
```diff
@@ -223,33 +181,22 @@ LogicalResult convertOp(AMDAIE::NpuWriteBdOp op, TransactionBuilder &builder) {
...
-      std::max((int64_t)strides[strides.size() - 1] - 1, (int64_t)0);
-  uint32_t d1Stride =
-      std::max((int64_t)strides[strides.size() - 2] - 1, (int64_t)0);
-  uint32_t d2Stride =
-      std::max((int64_t)strides[strides.size() - 3] - 1, (int64_t)0);
-  uint32_t iterationSize =
-      std::max((int64_t)op.getIterationSize() - 1, (int64_t)0);
+  // Strides and iteration_size will be encoded as `actual - 1`, so we need to
+  // ensure they are at least 1.
+  SmallVector<int32_t> stridesNew(strides.begin(), strides.end());
```
jtuyls, line 190: You could perform the copy on `strides` directly to avoid having both `strides` and `stridesNew` around, for example:

```
SmallVector<int32_t> strides(op.getStrides());
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/controlcode_to_transaction.mlir
jtuyls: Where exactly do you perform this workaround?

jtuyls (commented):
This is great! Just a few nits.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEControlCodeToTransaction.cpp
jtuyls: This is just a suggestion, if you prefer the style you have, that works as well.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/controlcode_to_transaction.mlir
Yu-Zhewen: I mean the correct value should be `0x00000000` by definition (as it operates on row 0, col 0). However, because the current aie-rt implementation behaves differently, the CHECK value is updated accordingly. This discrepancy currently does not cause any hardware errors, as these values are discarded later in the process. Nonetheless, I will report this issue to the aie-rt repository.

jtuyls: Ok, I see, thanks for the info, would indeed be good to report this.

jtuyls (approved):
LGTM

PR #1007, opened 2025-01-04
[SplitLogicalObjectFifos] Modify the split factor to depend on num_cols

yzhang93:
The old logic always split the outermost dimension of logical objectFifo to 1 without considering the number of Shim/MemTile used and could result in more channel usage than the limit. This PR modifies the splitting logic to depend on number of columns, so for example on a 4x2 AIE array, `amdaie.logicalobjectfifo<memref<4x1x32x32xi32, 1 : i32>>` will be split to two `amdaie.logicalobjectfifo<memref<2x1x32x32xi32, 1 : i32>>` .

This makes the e2e tests working for `numCols != numRows` cases.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
@@ -151,11 +151,15 @@ class AMDAIESplitLogicalObjFifosPass
     : public impl::AMDAIESplitLogicalObjFifosBase<
           AMDAIESplitLogicalObjFifosPass> {
  public:
-  using AMDAIESplitLogicalObjFifosBase::AMDAIESplitLogicalObjFifosBase;
-
   void getDependentDialects(DialectRegistry &registry) const override {
     registry.insert<AMDAIEDialect>();
   }
+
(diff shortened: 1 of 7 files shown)


Abhishek-Varma (approved):
Nice! LGTM.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -215,8 +219,9 @@ void AMDAIESplitLogicalObjFifosPass::runOnOperation() {
 
 }  // namespace
 
-std::unique_ptr<Pass> createAMDAIESplitLogicalObjFifosPass() {
-  return std::make_unique<AMDAIESplitLogicalObjFifosPass>();
+std::unique_ptr<Pass> createAMDAIESplitLogicalObjFifosPass(
+    AMDAIESplitLogicalObjFifosOptions options) {
```
jtuyls, line 223: Instead of providing `numCols` as an option, I think we can retrieve this information from the target attribute instead. I have this change in a follow-up.

jtuyls (approved):
LGTM. I have related changes that further generalize the splitting logic, that partially conflict with the logic added in this PR, but I will submit that as a follow-up.

PR #1008, opened 2025-01-06
[Router] Fix shim tile connections for non-DMA/NOC ports

Yu-Zhewen:
This fix addresses an issue in the shim tile circuit flow for non-DMA/NOC ports. Previously, the router was producing incorrect connections in the generated switchbox configuration.

For a shim tile circuit flow defined as:
```ll
aie.flow(%t00, CTRL : 0, %t00, SOUTH : 0)
```

The router incorrectly generated:
```ll
aie.switchbox(%t00) {
  aie.connect<SOUTH : 0, SOUTH : 0>
}
```

Instead, the correct output should be:
```ll
aie.switchbox(%t00) {
  aie.connect<CTRL : 0, SOUTH : 0>
}
```

A new router unit test has been added to validate this behavior.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/aie/test/unit_ctrl_routing.mlir
+++ b/compiler/plugins/target/AMD-AIE/aie/test/unit_ctrl_routing.mlir
@@ -1,0 +1,115 @@
+// RUN: iree-opt --amdaie-create-pathfinder-flows %s | FileCheck %s
+
+// CHECK-LABEL:   aie.device(npu1_4col) {
+// CHECK:           %[[TILE_0_0:.*]] = aie.tile(0, 0)
+// CHECK:           %[[SHIM_MUX_0_0:.*]] = aie.shim_mux(%[[TILE_0_0]]) {
+// CHECK:             aie.connect<DMA : 0, NORTH : 3>
+// CHECK:           }
+// CHECK:           %[[TILE_0_1:.*]] = aie.tile(0, 1)
+// CHECK:           %[[SWITCHBOX_0_1:.*]] = aie.switchbox(%[[TILE_0_1]]) {
+// CHECK:             %[[AMSEL_0:.*]] = aie.amsel<0> (0)
+// CHECK:             %[[AMSEL_1:.*]] = aie.amsel<1> (0)
+// CHECK:             %[[MASTERSET_CTRL:.*]] = aie.masterset(CTRL : 0, %[[AMSEL_1]])
+// CHECK:             %[[MASTERSET_NORTH:.*]] = aie.masterset(NORTH : 1, %[[AMSEL_0]])
+// CHECK:             aie.packet_rules(SOUTH : 1) {
+// CHECK:               aie.rule(24, 0, %[[AMSEL_0]])
+// CHECK:             }
+// CHECK:             aie.packet_rules(SOUTH : 4) {
+// CHECK:               aie.rule(31, 1, %[[AMSEL_1]])
+// CHECK:             }
+// CHECK:           }
+// CHECK:           %[[TILE_0_2:.*]] = aie.tile(0, 2)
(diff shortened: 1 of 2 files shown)


compiler/plugins/target/AMD-AIE/aie/test/unit_ctrl_routing.mlir
```diff
@@ -0,0 +1,115 @@
...
+  aie.device(npu1_4col) {
+    %t00 = aie.tile(0, 0)
+    %t01 = aie.tile(0, 1)
+    %t02 = aie.tile(0, 2)
+    %t03 = aie.tile(0, 3)
+    %t04 = aie.tile(0, 4)
+    %t05 = aie.tile(0, 5)
+
+    // For Task Completion Tokens (TCTs).
+    aie.flow(%t00, CTRL : 0, %t00, SOUTH : 0)
```
jtuyls, line 87: This would be for the control response right?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
```diff
@@ -639,7 +639,9 @@ std::map<TileLoc, std::vector<Connect>> emitConnections(
       int channel = dst.channel;
       // handle special shim connectivity
       if (curr == srcTileLoc &&
-          deviceModel.isShimNOCorPLTile(srcTileLoc.col, srcTileLoc.row)) {
+          deviceModel.isShimNOCorPLTile(srcTileLoc.col, srcTileLoc.row) &&
+          (srcBundle == StrmSwPortType::DMA ||
+           srcBundle == StrmSwPortType::NOC)) {
```
jtuyls, line 644: Could you add a comment here to explain in which cases this special connectivity is is needed in which not?

compiler/plugins/target/AMD-AIE/aie/test/unit_ctrl_routing.mlir
Yu-Zhewen: https://github.com/Xilinx/mlir-aie/pull/1705 introduces a new pass to automatically insert the following two groups of flows:

(1) `shim-DMA-to-tile-CTRL`, handles sending control packets into each tile.

(2) `tile-CTRL-to-shim-SOUTH`, which is not specific for control packets, but instead to receive any TCT for DMA BD. This was previously handled by a hack in CDO generation, and `mlir-aie`  relocates this feature into the new pass.

(we currently still have this hack) https://github.com/nod-ai/iree-amd-aie/blob/332e8d19722bbd1501be9313e297b2bf51bdbd86/runtime/src/iree-amd-aie/aie_runtime/iree_aie_configure.cc#L270-L279

---
There is also a possibility if we want to retrieve data from CTRL port, then we need to set up  `tile-CTRL-to-shim-DMA` instead of  `tile-CTRL-to-shim-SOUTH`. This is currently not covered by https://github.com/Xilinx/mlir-aie/pull/1705, but an example has been provided to do so.

https://github.com/Xilinx/mlir-aie/blob/30300fadd08e0a6189f1f493ada7948d99241034/test/npu-xrt/add_one_ctrl_packet/aie.mlir#L32

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
Yu-Zhewen: done

compiler/plugins/target/AMD-AIE/aie/test/unit_ctrl_routing.mlir
jtuyls: I see, this makes sense, thanks for the explanation! Are you intending to get rid of that hack as part of this control packet effort?

Yu-Zhewen: Yes, I am currently working on a PR based on Xilinx/mlir-aie#1705, where I will also remove that hack

PR #1011, opened 2025-01-08
[KernelDispatch] Enable unpack outer permutation for (batch)matmul

jtuyls:
This PR enables outer permutation of the level 0 unpack, which results in a logical objectFifo with outer dimension aligning with the columns instead of rows. This helps with splitting and finding a routing solution after splitting (as we always split logical objectFifos on the outermost non-unit dimension). 

For example, when targeting strix with an 8 column array, this results in an output L2 objectFifo with shape `8x4x32x32` instead of `4x8x32x32`. Splitting therefore happens on the dimension with size 8 instead of size 4, which helps with assigning one output buffer to every column and finding enough channels for routing the data from L1 to L2.


Abhishek-Varma (approved):
LGTM! Thanks Jorn!

PR #1012, opened 2025-01-08
Add a new pass to generate column-wise control overlay

Yu-Zhewen:
Adapted from https://github.com/Xilinx/mlir-aie/pull/1705. 

Introduces a new pass to automatically insert the following two groups of flows:

1. `route-shim-to-tct`, circuit flows from shim `CTRL` to shim `SOUTH` ports, for sending TCTs.
2. `route-shim-to-tile-ctrl`, packet flows between shim `DMA` to shim/mem/compute tile `CTRL` ports, for sending control packets.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEGenerateColumnControlOverlay.cpp
```diff
@@ -0,0 +1,187 @@
...
+  // Exclude those channels that are already used by a circuit flow.
+  workgroupOp->walk([&](AMDAIE::FlowOp flowOp) {
+    if (flowOp.getIsPacketFlow()) return WalkResult::advance();
+    SmallVector<AMDAIE::ChannelOp> sourceChannels;
+    for (auto value : flowOp.getSources()) {
+      if (auto channelOp = dyn_cast<AMDAIE::ChannelOp>(value.getDefiningOp())) {
+        sourceChannels.push_back(channelOp);
+      }
+    }
+    for (auto channelOp : sourceChannels) {
```
jtuyls, line 47:
```suggestion
    for (AMDAIE::ChannelOp channelOp : sourceChannels) {
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEGenerateColumnControlOverlay.cpp
```diff
@@ -0,0 +1,187 @@
...
+      AMDAIETileType::SHIMNOC, AMDAIEDmaProp::NumChannels);
+  std::for_each(shimTileOps.begin(), shimTileOps.end(), [&](TileOp shimTileOp) {
+    shimTileToGeneratorMap[shimTileOp.getResult()] =
+        ChannelGenerator(numShimDmaChannels, numShimDmaChannels);
+  });
+  // Exclude those channels that are already used by a circuit flow.
+  workgroupOp->walk([&](AMDAIE::FlowOp flowOp) {
+    if (flowOp.getIsPacketFlow()) return WalkResult::advance();
+    SmallVector<AMDAIE::ChannelOp> sourceChannels;
+    for (auto value : flowOp.getSources()) {
```
jtuyls, line 42:
```suggestion
    for (Value value : flowOp.getSources()) {
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEGenerateColumnControlOverlay.cpp
```diff
@@ -0,0 +1,187 @@
...
+    SmallVector<TileOp> shimTileOps = llvm::to_vector<4>(llvm::map_range(
+        columnToShimTile, [](auto pair) { return pair.second; }));
+    if (failed(getAvailableShimChannels(workgroupOp, shimTileOps,
+                                        shimTileToGeneratorMap))) {
+      return failure();
+    }
+    WalkResult res = workgroupOp->walk([&](AMDAIE::TileOp tileOp) {
+      uint32_t col = getConstantIndexOrAssert(tileOp.getCol());
+      TileOp shimTileOp = columnToShimTile[col];
+      // Get the available channel, but do not assigning it. Allow it to be
```
jtuyls, line 108:
```suggestion
      // Get the available channel, but do not assign it. Allow it to be
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEGenerateColumnControlOverlay.cpp
```diff
@@ -0,0 +1,187 @@
...
+          rewriter.getUnknownLoc(), ValueRange{shimCtrlChannelOp},
+          ValueRange{shimSouthChannelOp},
+          /*isPacketFlow*/ false, /*packetId*/ nullptr);
+    }
+  }
+
+  return success();
+}
+
+class AMDAIEGenerateColumnControlOverlayPass
```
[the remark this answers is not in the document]
Yu-Zhewen: I was using the same pass name as `mlir-aie`,  probably because the overlay is constructed in a column-wise fashion, where the `CTRL` ports are connected to the corresponding shimtile within the same column.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_configure.cc
```diff
@@ -267,7 +267,7 @@ LogicalResult configureStreamSwitch(const AMDAIEDeviceModel &deviceModel,
                                     const TileLoc &tileLoc,
                                     const std::vector<Connect> &connects) {
   auto devInst = const_cast<XAie_DevInst *>(&deviceModel.devInst);
-  // FIXME hack for TCT routing
+  // mlir-air legacy, hack for TCT routing
```
Yu-Zhewen, line 270: Unfortunately, since the new pass operates on `AMDAIE::TileOp` and `FlowOp`,  this hack cannot be removed for tests using AIR.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEGenerateColumnControlOverlay.cpp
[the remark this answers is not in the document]
jtuyls: Right, but I am not even seeing a use case for it and even if we want to support different control overlays in the future, I think we will just be using the same pass with a different option? As there is also no need to have the same name as mlir-aie, I think dropping the 'column' part would be simpler, more clear and versatile towards future changes.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEGenerateColumnControlOverlay.cpp
```diff
@@ -0,0 +1,191 @@
...
+      AMDAIE::TileOp tileOp = channelOp.getTileOp();
+      uint8_t channel = channelOp.getValue();
+      StrmSwPortType portType = channelOp.getPortType();
+      AMDAIE::DMAChannelDir direction = channelOp.getDirection();
+      if (shimTileOps.contains(tileOp) && portType == StrmSwPortType::DMA) {
+        // Assign to exclude.
+        if (direction == AMDAIE::DMAChannelDir::MM2S) {
+          shimTileToGeneratorMap[tileOp.getResult()].assignProducerDMAChannel(
+              channel);
+        } else {
```
jtuyls, line 52: Use an `else if` here and add an assert in the else case to make sure we're only ever handling `MM2S` and `S2MM`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEGenerateColumnControlOverlay.cpp
```diff
@@ -0,0 +1,191 @@
...
+  AMDAIEDeviceModel deviceModel = AMDAIE::getDeviceModel(device.value());
+
+  IRRewriter rewriter(workgroupOp->getContext());
+  DenseSet<uint32_t> occupiedCols;
+  DenseMap<uint32_t, AMDAIE::TileOp> columnToShimTile;
+  workgroupOp->walk([&](AMDAIE::TileOp tileOp) {
+    uint32_t col = getConstantIndexOrAssert(tileOp.getCol());
+    uint32_t row = getConstantIndexOrAssert(tileOp.getRow());
+    occupiedCols.insert(col);
+    if (deviceModel.getTileType(col, row) == AMDAIETileType::SHIMNOC)
```
jtuyls, line 81:
```suggestion
    if (deviceModel.isShimNOCTile(col, row))
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEGenerateColumnControlOverlay.cpp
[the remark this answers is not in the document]
Yu-Zhewen: Yeah, that makes sense. Done

jtuyls (approved):
LGTM

Yu-Zhewen:
On hold, to check a random error https://github.com/nod-ai/iree-amd-aie/actions/runs/12694938797/job/35386212688

PR #1013, opened 2025-01-08
[ChannelGenerator] Assign the channel only for circuit flow

Yu-Zhewen:


runtime/src/iree-amd-aie/aie_runtime/Utils/ChannelGenerator.h
```diff
@@ -22,45 +22,61 @@ class ChannelGenerator {
   ChannelGenerator() {}
   ChannelGenerator(uint8_t numProducerChannels, uint8_t numConsumerChannels)
       : numProducerChannels(numProducerChannels),
-        numConsumerChannels(numConsumerChannels) {}
+        numConsumerChannels(numConsumerChannels) {
+    assert(numProducerChannels > 0 && numConsumerChannels > 0 &&
+           "Invalid number of producer/consumer channels.");
+    // Initialize to the last channel for round-robin usage.
```
jtuyls, line 28: I think it makes sense to have an `AssignmentMode` similar to the `ChannelBdIdGenerator`:

```
ChannelAssignmentMode {
  FirstAvailable,
  RoundRobin
}
```

Or if the logic differs quite a bit, use specialized child classes instead: `RoundRobinChannelGenerator`, `FirstAvailableChannelGenerator`.

jtuyls (changes requested):
Could you document in the description and as a comment in the code what the assignment strategy is for packet routing (round robin) and why?

runtime/src/iree-amd-aie/aie_runtime/Utils/ChannelGenerator.h
```diff
@@ -16,51 +16,83 @@ using namespace llvm;
...
     assignedConsumerChannels.insert(channel);
   }
 
  private:
   uint8_t numProducerChannels = 0;
   uint8_t numConsumerChannels = 0;
+  // Tracks the channels that are used by circuit flows.
   DenseSet<uint8_t> assignedProducerChannels;
   DenseSet<uint8_t> assignedConsumerChannels;
+  // Tracks the last used channel, for both circuit and packet flows.
```
jtuyls, line 93: There might be a bit of unclarity about what 'using' a DMA channel means, so I think 'lastRetrievedDMAChannel' might be a more clear name. And maybe document this as well.

Yu-Zhewen: done

jtuyls (approved):
LGTM

PR #1014, opened 2025-01-08
Bump peano forward by 3 months

newling:
Using this more recent version fixes a numerical error I've observed. Performance numbers appear unchanged


Yu-Zhewen (approved):
Thanks!

PR #1016, opened 2025-01-09
[KernelDispatch] Add matmul RHS outer permutation

jtuyls:
Permutes the level 0 RHS pack operations to always get RHS/B on L2 into the packed layout [N K ? ?] instead of [K N ? ?]. This is needed for splitting on the outer dimension when caching more of K in L2 (K > 1).


yzhang93 (approved):
LGTM.

PR #1020, opened 2025-01-10
[CI] Use pre-commit (copy IREE) for linting

newling:
This PR aligns iree-amd-aie's linting with IREE's. It copies files https://github.com/iree-org/iree/blob/main/.pre-commit-config.yaml and https://github.com/iree-org/iree/blob/main/.github/workflows/lint.yml , with the only change being a removal of basel and yml linting, as we don't have basel or yml in iree-amd-aie (and the basel linter was failing). 

This PR also includes the changes needed to pass the linter. Running  `pre-commit run --all-files` basically did everything needed, with 2 places where a `clang-format off/on` pair was needed to avoid header rearrangement. 

Note to developers: if your PR fails linting after this lands, it's very easy to fix locally: `pip install pre-commit` and then run pre-commit on all or a subset of the project files. 

{0378F65F-6FDD-4765-9E8E-378380638A74}


jtuyls (approved):
LGTM, thanks!

PR #1021, opened 2025-01-10
[IR] Update how tile's with partial info are printed 

newling:
Will now be `%tile_c_3` and `%tile_2_r` if one or other of the column and row are integers. As opposed to just `%tile` and `%tile_1`. This aligns tile printing with the (not landed) printing of logicalobjectfifos: https://github.com/nod-ai/iree-amd-aie/pull/1017


jtuyls (approved):
LGTM

PR #1027, opened 2025-01-14
[FuseConsumerIntoLoop] Extend to multiple results and generalize depth fusion

jtuyls:
This PR extends the consumer fusion to work for matmul + elemwise in case of an additional level of tiling (see added lit test).


yzhang93 (approved):
LGTM

PR #1028, opened 2025-01-14
[IR] Add info to SSA names: what operand of the linalg op does this correspond to? 

newling:
This is quite a heavy hammer to apply for SSA name generation but I've found it useful when trying to get my head around the IR. With this change, the operand of the high-level operation is included in the low level op name. For example, now a `.logicalobjectfifo.from_buffers` on column 0 and row 2 which is for moving data of operand number 2 of the top-level operation might look like:

```
%lof_0_2_C =  amdaie.logicalobjectfifo.from_buffers(
              {%buffer_0_2_C}, {%lock_0_2_10}, {%lock_0_2_11}) 
               : memref<1024xf32, 2 : i32> -> !amdaie.logicalobjectfifo<memref<1024xf32, 2 : i32>>
```

The trailing `_C` is added because of a traversal through producers/consumers until compute ops (linalg.generics and func.calls) are found, and indexes determined. The `_C` specifically is because the index is 2, and 'C' = 'A' + 2. If it's not unambiguously clear what the index is, this new suffix is just not added. 

Take a look at the round-trip test added here for more info!

ps. I'm expecting push-back on this one, so I'll say upfront that I'm also happy just to cherry-pick this onto my work branch when I can't follow IR.

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -19,7 +19,7 @@ concurrency:
 jobs:
   build_and_ctest:
     name: Build and Test (linux, ASSERTIONS)
-    runs-on: nod-ai-shared-cpubuilder-manylinux-x86_64
+    runs-on: azure-cpubuilder-linux-scale
     strategy:
       fail-fast: true
     env:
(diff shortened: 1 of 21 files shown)


makslevental:
FYI you could just use this 

https://github.com/llvm/llvm-project/pull/119996

 functionality plus `NamedLoc` to get this same effect "for free".

newling:
> FYI you could just use this
> 
> [llvm/llvm-project#119996](https://github.com/llvm/llvm-project/pull/119996)
> 
> functionality plus `NamedLoc` to get this same effect "for free".

Nice I'll try it, but I'll be very surprised if we propagate locations when new ops are created in passes.

```
iree-amd-aie/Transforms $ grep -r "UnknownLoc()" | wc -l
102
```

PR #1029, opened 2025-01-14
Compilation error fix for matmul transpose B with pack-peel-4-level-tiling pipeline

newling:
This fixes the error 

```
error: 'amdaie.core' op failed to coalesce for loops
```

in the test 

```
MatmulTransposeB(
    128,
    256,
    128,
    "i8",
    "i32",
    tile_pipeline="pack-peel-4-level-tiling"
)
```

The source of error is some very defensive code in MLIR: https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/SCF/Utils/Utils.cpp#L903 (Why shouldn't I do a no-op coalesce of 1 for op?). 

Note: compilation gets further with this fix, but doesn't compile to completion

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertCoreForallToFor.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertCoreForallToFor.cpp
@@ -38,9 +38,12 @@ LogicalResult coreForallToFor(RewriterBase &rewriter, AMDAIE::CoreOp coreOp) {
     SmallVector<scf::ForOp> forOps = llvm::map_to_vector(
         forOpResults,
         [](Operation *loop) { return dyn_cast<scf::ForOp>(loop); });
-    if (failed(coalesceLoops(rewriter, forOps))) {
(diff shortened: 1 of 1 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertCoreForallToFor.cpp
```diff
@@ -38,9 +38,12 @@ LogicalResult coreForallToFor(RewriterBase &rewriter, AMDAIE::CoreOp coreOp) {
     SmallVector<scf::ForOp> forOps = llvm::map_to_vector(
         forOpResults,
         [](Operation *loop) { return dyn_cast<scf::ForOp>(loop); });
-    if (failed(coalesceLoops(rewriter, forOps))) {
-      coreOp.emitOpError() << "failed to coalesce for loops";
-      return WalkResult::interrupt();
+
+    if (forOps.size() > 1) {
```
jtuyls, line 42: Maybe add a comment on why this is needed FFR.

jtuyls (approved):
LGTM

PR #1031, opened 2025-01-15
[AssignTiles][CI] Fix assign-tiles for L3 buffers with multiple tile candidate + add e2e test targeting 4x8 AIE array on Strix

Abhishek-Varma:
This PR logically contains two commits :-

**Commit 1**
-- Required to enable 4x8 AIE array on Strix.
-- Before this commit `iree-amdaie-assign-tiles` pass ends up allocating/distributing LHS L3 buffers on (0,0) -> (7,0).
-- Since this leads to consumer DMA channel exhaustion later on, this commit aims to address the same and tries to assign same tile set to L3 buffers based on the L3 LOF count for L3<->L2 Copy.

**Commit 2**
-- Updates `mm_npu4.cc` for targeting 4x8 intrinsic for ukernel.
-- Adds two e2e tests for CI : a normal execution and a ukernel based execution.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
```diff
@@ -365,9 +367,34 @@ LogicalResult assignNonLocalTiles(RewriterBase &rewriter, Operation *op,
...
+          memrefToTileMap[memref].erase(memrefToTileMap[memref].begin());
+          memrefToTileMap[memref].push_back(assignedTileOp);
+        } else {
+          assignedTileOp =
+              *std::min_element(tiles.begin(), tiles.end(), tileLocAndUsageCmp);
+          if (hasMultipleTileCandidates && memSpace == 0) {
+            if (!memrefToTileMap.contains(memref)) memrefToTileMap[memref] = {};
+            memrefToTileMap[memref].push_back(assignedTileOp);
+          }
+        }
```
jtuyls, line 397: What is going wrong without this?  I didn't get this:

> // Test to demonstrate that we go ahead with the same tile assignment for different blocks in case
// the L3 logicalObjectFifo has multiple tile candidates.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -804,9 +804,7 @@ LogicalResult splitDoublyStridedOp(IRRewriter &rewriter,
                             ? maybeSplitFactor.value()
                             : std::gcd(sourceSize, targetSize);
   if (sourceSize % splitFactor != 0 || targetSize % splitFactor != 0) {
-    return op.emitOpError() << "the target or source size is not divisible by "
-                               "the provided splitting factor: "
-                            << splitFactor;
+    splitFactor = std::gcd(sourceSize, targetSize);
```
jtuyls, line 807: This would be a good place to add a `LLVM_DEBUG` statement that mentions that the splitting factor is changed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
Abhishek-Varma: Column: 0, 1, 2, 3, 4, 5, 6, 7

Whenever we assign tiles we increment its "usage" count by 1.

Now, whenever we've a LOF with multiple tile candidates we decide a particular tile by picking the one with the least "usage" count.

L3 buffers we have :-
LHS: 4 (Prologue)+ 4 (Main) + 4 (Epilogue)
RHS: 8 (Prologue)+ 8 (Main) + 8 (Epilogue)
OUT: 8 (Epilogue)

For RHS and OUT L3 buffers, it works well because they're indeed to be assigned to (0,0) -> (7,0), so picking the least count tile on every occurrence gets us from 0 -> 7 incrementally.

But for LHS, since we want it to be on (0,0) -> (4,0), the above logic doesn't make that distinction and ends up assigning (5,0) -> (7,0) for the LHS L3 buffers on Main, which later on in the pipeline causes consumer DMA channel exhaustion.

So, the intention of the above logic is to have a "similar" tile assignment strategy for the buffers of each of Prologue + Main + Epilogue.

jtuyls: > But for LHS, since we want it to be on (0,0) -> (4,0), the above logic doesn't make that distinction and ends up assigning (5,0) -> (7,0) for the LHS L3 buffers on Main, which later on in the pipeline causes consumer DMA channel exhaustion.

They don't necessarily need to be mapped to (0,0) -> (4,0), (5,0) -> (7,0) should be fine as well. I don't see how that would result in exhaustion of consumer DMA channels. Could you share the IR dump for that case?

Abhishek-Varma: > > But for LHS, since we want it to be on (0,0) -> (4,0), the above logic doesn't make that distinction and ends up assigning (5,0) -> (7,0) for the LHS L3 buffers on Main, which later on in the pipeline causes consumer DMA channel exhaustion.
> 
> They don't necessarily need to be mapped to (0,0) -> (4,0), (5,0) -> (7,0) should be fine as well. I don't see how that would result in exhaustion of consumer DMA channels. Could you share the IR dump for that case?

Oops. I forgot to write "as well" above along with incorrect tile mentions. My bad!
Let me rectify and explain :-

Basically, LHS should be four tiles on row 0 but due without this fix the pass will end up assigning other four tiles as well.

So, this is what it ends up looking for LHS without the above fix :-
```
Prologue
     tile(0,0)
     tile(1,0)
     tile(2,0)
     tile(3,0)
Main
     tile(4,0)
     tile(5,0)
     tile(6,0)
     tile(7,0)
Epilogue
     tile(0,0)
     tile(1,0)
     tile(2,0)
     tile(3,0)
```

With the current fix we get :-
```
Prologue
     tile(0,0)
     tile(1,0)
     tile(2,0)
     tile(3,0)
Main
     tile(0,0)
     tile(1,0)
     tile(2,0)
     tile(3,0)
Epilogue
     tile(0,0)
     tile(1,0)
     tile(2,0)
     tile(3,0)
```

Here is the e2e IR dump without the fix from this PR: https://gist.github.com/Abhishek-Varma/6335a2fad91e9298a08b6aee29951550

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
```diff
@@ -365,6 +367,19 @@ LogicalResult assignNonLocalTiles(RewriterBase &rewriter, Operation *op,
             llvm::map_to_vector(logicalObjectFifo.getTiles(), [](Value tile) {
               return dyn_cast_if_present<TileOp>(tile.getDefiningOp());
             });
+
+        auto fromMemrefOp = dyn_cast<AMDAIE::LogicalObjectFifoFromMemrefOp>(
```
jtuyls, line 371: Could you add a comment here explaining that you're limiting the number of tile options to buffer count to avoid repeated accesses of the same buffer being assigned to different tiles?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
```diff
@@ -365,6 +367,19 @@ LogicalResult assignNonLocalTiles(RewriterBase &rewriter, Operation *op,
...
+            logicalObjectFifo.getOperation());
+        if (fromMemrefOp) {
+          Operation *defOp = fromMemrefOp.getMemref().getDefiningOp();
+          if (defOp && l3BufferCount.contains(defOp)) {
+            SmallVector<AMDAIE::TileOp> effectiveTiles;
+            for (unsigned i = 0, n = l3BufferCount[defOp], m = tiles.size();
+                 i < n && i < m; i++)
+              effectiveTiles.push_back(tiles[i]);
+            tiles = effectiveTiles;
```
jtuyls, line 380:
```suggestion
            tiles.truncate(std::max((size_t)l3BufferCount[defOp], tiles.size()));
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
```diff
@@ -433,8 +448,67 @@ void AMDAIEAssignTilesPass::runOnOperation() {
...
                           << *parentOp << "\n");
 
+  // Analyse the count of each L3 logicalObjectFifos appearing in L3<->L2
+  // CopyOp. We would be maintaining this count in a map `l3BufferCount` and
+  // using this later to pick one tile to assign to the L3 logicalObjectFifo.
+  // TODO(avarma): Have an analysis pass instead and annotate the L3 buffers
+  // with their count. OR better to have an analysis pass that annotates
+  // LHS/RHS/OUT buffers. That might come handy for other passes.
+  bool visitedCopyOp = false;
+  DenseMap<Operation *, int64_t> l3BufferCount;
```
jtuyls, line 458:
```suggestion
  DenseMap<Operation *, size_t> l3BufferCount;
```
I think `size_t` is a better data type here for a count and it avoids type casting when comparing with `tiles.size()`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/assign_tiles.mlir
```diff
@@ -509,3 +509,215 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
     return
   }
 }
+
+// -----
```
jtuyls, line 513: Could you add the same two tests for L2->L3 as well?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
```diff
@@ -433,8 +448,67 @@ void AMDAIEAssignTilesPass::runOnOperation() {
...
+      }
+      return WalkResult::advance();
+    } else if (isa<AMDAIE::CoreOp>(op)) {
+      return WalkResult::interrupt();
+    }
+    return WalkResult::advance();
+  });
+
+  // We don't need the following. Have added this just for sake of completeness
+  // in the entire L3 buffer count analysis.
```
jtuyls, line 484: Remove this note as it's bound to get stale at some point when we silently need it. It's good to always add logic for completeness and a test case for it as well, even though we haven't seen the case in an e2e example.

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
```diff
@@ -433,8 +448,67 @@ void AMDAIEAssignTilesPass::runOnOperation() {
...
+  res = parentOp->walk<WalkOrder::PostOrder, ReverseIterator>(
```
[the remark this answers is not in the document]
Abhishek-Varma: For L2 -> L3 (Output), we indeed need to make a reverse order traversal.

The walk prior to this is fetching L3 -> L2.

The following is the structure currently :-
```
scf.forall {
    // PROLOGUE                                 |
    // L3 -> L2                                      |
    <amdaie.core ops>                       V  End of first walk
    
    // MAIN
    scf.for {
        // L3 -> L2
        <amdaie.core ops>
    }
    
    // EPILOGUE
    // L3 -> L2
    <amdaie.core ops>                      /\  End of second walk
    // L2 -> L3                                     |   
}
```

To get L3 buffer usage count for INPUTS, we only need to look at PROLOGUE - the former walk performs that and breaks out at the first instance of `amdaie.core` op
To get L3 buffer usage count for OUTPUT, we only need to look at EPILOGUE- the latter walk performs that and breaks out at the first instance of `amdaie.core` op.

To consolidate the two walks into one, we can focus only on EPILOGUE for both INPUTS and OUTPUT but that adds unnecessary checks and makes the walk more complicated than it needs to be, because after processing OUTPUT L3, we need to go through ALL those amdaie.core ops in between and then start processing INPUTS.

Hence I feel it's better to keep the two walks logically separate.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/assign_tiles.mlir
Abhishek-Varma: I've added L2 -> L3 within the same and have marked the Copy section with comments in input case.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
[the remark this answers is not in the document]
jtuyls: > To get L3 buffer usage count for INPUTS, we only need to look at PROLOGUE - the former walk performs that and breaks out at the first instance of amdaie.core op

I didn't notice this earlier, but we shouldn't be making the assumption that all L3->L2 copy ops are before the `amdaie.core` ops and similarly with the L2->L3 ops.

You could easily have something like:
```
L3->L2 copy op
amdaie.core
L3->L2 copy op
amdaie.core
```

I understand that we're currently not seeing this case as the CI is passing, but this is perfectly valid IR and we shouldn't prohibit this.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp
```diff
@@ -365,6 +367,22 @@ LogicalResult assignNonLocalTiles(RewriterBase &rewriter, Operation *op,
...
+        // more DMA channels on those new tiles than needed, and as a result we
+        // will end up exhausting the DMA channels. Currently the following fix
+        // works for L3 buffers.
+        auto fromMemrefOp = dyn_cast<AMDAIE::LogicalObjectFifoFromMemrefOp>(
+            logicalObjectFifo.getOperation());
+        if (fromMemrefOp) {
+          Operation *defOp = fromMemrefOp.getMemref().getDefiningOp();
+          if (defOp && uniqueL3L2Pair.contains(defOp))
+            tiles.truncate(
+                std::min((size_t)uniqueL3L2Pair[defOp].size(), tiles.size()));
```
jtuyls, line 384:
```suggestion
                std::min(uniqueL3L2Pair[defOp].size(), tiles.size()));
```

jtuyls (approved):
LGTM, just one nit.

Thanks for updating to analyzing based on L2 buffers now.

PR #1032, opened 2025-01-15
[CombineStridedOps] Generalize dimension checking

jtuyls:
This PR generalizes dimension checking in `CombineStridedOps`. This allows difference ops to provide their own dimension checking implementation. This is needed because circular and non-circular `dma_cpy_nd` ops have different logic for checking whether the number of dimensions of the combined DMA op would exceed the available number of dimensions.

With this fix, BatchMatmul works with the `pack-peel-4-level-tiling` pipeline.

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -708,7 +708,17 @@ class BatchMatmul(BaseMatmul):
     A test of the form batch_matmul(A,B) where A:BxMxK, B:BxKxN
     """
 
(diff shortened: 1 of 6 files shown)


yzhang93 (approved):
LGTM.

newling:
I guess this will solve https://github.com/nod-ai/iree-amd-aie/issues/1036 

What is the logic behind circular dma's being allowed unlimited dma dimensions?

newling:
> What is the logic behind circular dma's being allowed unlimited dma dimensions?

Answered privately: having unlimited is actually too lenient, but the thinking is that you can unlimited dimensions as long as they have stride = 0 (I guess because they can all be collapsed).

PR #1033, opened 2025-01-15
[CtrlPkt] Convert `aie.device` to `amdaie.npu.control_packet`

Yu-Zhewen:
Using the `aie-rt` APIs, convert the `aie.device` into a sequence of transaction operations ( `write`, `blockwrite`, and `maskwrite`). Then, translate these transaction operations into control packet headers and payloads, which are represented by `amdaie.npu.control_packet`.

P.S. `amdaie.npu.control_packet` will be further lowered to binary, along with the shim DMA `memcpy` and `sync` operations in follow-up PRs.

references:
https://github.com/Xilinx/mlir-aie/pull/1709
https://github.com/Xilinx/mlir-aie/pull/1768


compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.cpp
```diff
@@ -1363,6 +1363,19 @@ SmallVector<AMDAIE::NpuDmaCpyNdOp> NpuDmaWaitOp::getDmaOps() {
...
 
+//===----------------------------------------------------------------------===//
+// AMDAIE_NpuControlPacketOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult NpuControlPacketOp::verify() {
+  std::optional<ArrayRef<int32_t>> data = getData();
+  if (data.has_value() && data.value().size() != getLength()) {
+    return emitOpError()
+           << "data length does not match the specified attribute";
```
jtuyls, line 1374:
```suggestion
           << "data length does not match the specified `length` attribute";
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -516,6 +517,17 @@ LogicalResult AIETargetBackend::serializeExecutable(
       return failure();
     }
 
+    if (options.convertAieToCtrlPkt) {
```
jtuyls, line 520: Can't we put this in a pass instead so we can inspect the IR for easier debugging?

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
```diff
@@ -1047,6 +1047,31 @@ def AMDAIE_NpuTctSyncOp: AMDAIE_Op<"npu.tct_sync"> {
...
+
+    ```mlir
+    amdaie.npu.control_packet {address = 0 : ui32, data = array<i32: 2>
+      length = 1 : ui32, opcode = 0 : ui32, stream_id = 0 : ui32}
+    ```
+  }];
+  let arguments = (
+    ins UI32Attr:$address,
+        UI32Attr:$length,
+        UI32Attr:$opcode,
```
jtuyls, line 1067: What are the opcode and stream_id?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIEControlPacket.cpp
```diff
@@ -0,0 +1,171 @@
...
+  uint32_t NumOps = txn_header->NumOps;
+  txn_ptr += sizeof(XAie_TxnHeader);
+
+  // Create a new MLIR module.
+  MLIRContext context;
+  context.loadDialect<AMDAIEDialect>();
+  context.loadDialect<func::FuncDialect>();
+  context.loadDialect<IREE::HAL::HALDialect>();
+  OpBuilder builder(&context);
+  ModuleOp newModuleOp = ModuleOp::create(builder.getUnknownLoc());
```
jtuyls, line 50: Why do we need to create a new module?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/test/AMDAIEControlPacketTest.cpp
```diff
@@ -0,0 +1,133 @@
...
+      moduleOp.getOps<xilinx::AIE::DeviceOp>());
+  if (deviceOps.size() != 1) {
+    std::cerr << "Error: Expected exactly one xilinx.aie.device op\n";
+    return 1;
+  }
+  xilinx::AIE::DeviceOp deviceOp = deviceOps[0];
+  AMDAIEDevice device = deviceOp.getDevice();
+
+  // TODO(zhewen): The logic here is currently copied from AIETarget.cpp.
+  // Refactor this to use enums for `npuVersion` and `targetArch`.
```
jtuyls, line 76: Can't we create a `DeviceModel` from the `DeviceOp` and retrieve `targetArch`, `npuVersion` from that?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIEControlPacket.cpp
```diff
@@ -0,0 +1,171 @@
...
+      case XAie_TxnOpcode::XAIE_IO_BLOCKWRITE: {
+        XAie_BlockWrite32Hdr *bw_header =
+            reinterpret_cast<XAie_BlockWrite32Hdr *>(txn_ptr);
+        uint64_t addr = bw_header->RegOff;
+        auto payload = reinterpret_cast<uint32_t *>(
+            txn_ptr + sizeof(XAie_BlockWrite32Hdr));
+        // Payload length in 32-bit words.
+        size_t length = (bw_header->Size - sizeof(XAie_BlockWrite32Hdr)) / 4;
+        // Split payload into beats of 4 or less, in int32_t.
+        for (size_t i = 0; i < length; i += 4) {
```
jtuyls, line 109: This might create a lot of control packets, can we initially create large control packets (potentially using heap resource blobs) and split the packets later? That would also make this more extensible towards changes in splitting size in future generations. This comment is related to the comment above on creating a separate pass.

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
Yu-Zhewen: Ideally yes...but I am not sure how to make this a pass, given that the elf files of the cores are required for the conversion. Or if we make it a pass, where would be the best place to put it in the pipeline?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIEControlPacket.cpp
Yu-Zhewen: https://github.com/nod-ai/iree-amd-aie/blob/1076d63e0de11b8e0e17286cf422c5134240b951/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp#L524-L526

And based on where it is currently called, it suggests the possibility of having multiple `device`s within a single `module`, so I can't reuse the original `module` by erasing its contents.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIEControlPacket.cpp
Yu-Zhewen: Yeah, that makes sense, I will separate this part.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
jtuyls: The elf binaries could be created in a pass as well and that would basically have to happen at the end of the current pipelines somewhere to avoid too much collision. I see that this might be some work, so this would probably be best split up into pieces.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIEControlPacket.cpp
jtuyls: Yeah, I see, this is not beautiful, because we would be going amdaie -> aie -> amdaie, but I think this works in the short term until we morph AIE and AMDAIE.

> And based on where it is currently called, it suggests the possibility of having multiple devices within a single module, so I can't reuse the original module by erasing its contents.

Some thoughts:
- You could also loop through the devices within `convertAieToControlPacket` itself.
- `DeviceOp` is created from `Workgroup` and we only allow one workgroup for now, so you could also check for that and error out if multiple devices are found.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIEControlPacket.cpp
Yu-Zhewen: The logic moved to the deviceModel. Will submit a separate PR.
https://github.com/nod-ai/iree-amd-aie/blob/81ae843899af2e52395071c012cf0dbb1702edc4/runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h#L390-L411

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/test/AMDAIEControlPacketTest.cpp
Yu-Zhewen: The logic moved to the deviceModel. Will submit a separate PR.
https://github.com/nod-ai/iree-amd-aie/blob/81ae843899af2e52395071c012cf0dbb1702edc4/runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h#L390-L411

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIEControlPacket.cpp
Yu-Zhewen: After removing the split, the example IR looks like this
https://github.com/nod-ai/iree-amd-aie/blob/81ae843899af2e52395071c012cf0dbb1702edc4/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/test/convert_aie_to_ctrl_pkts.mlir#L15-L40

Regarding the data, the length is typically 1, except for the `XAie_LoadElf` case. As such, I'm uncertain which representation is more appropriate: `HeapAsmResourceBlob` or `DenseArray`.  Any preference?

jtuyls: > Regarding the data, the length is typically 1, except for the XAie_LoadElf case. As such, I'm uncertain which representation is more appropriate: HeapAsmResourceBlob or DenseArray. Any preference?

> Alternatively, having control_packet for the scalar data, and introducing a separate control_packet_block operation for length > 1?

Yeah, or support both `DenseArray` and `HeapAsmResourceBlob`?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/test/AMDAIEControlPacketTest.cpp
Yu-Zhewen: implemented in #1040

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
Yu-Zhewen: Implemented as a pass here
https://github.com/nod-ai/iree-amd-aie/blob/06899b76e42dc59de60977512692e8a6d0cb97f0/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertDeviceToControlPackets.cpp#L1

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIEControlPacket.cpp
Yu-Zhewen: Added a single `DeviceOp` check here,
https://github.com/nod-ai/iree-amd-aie/blob/06899b76e42dc59de60977512692e8a6d0cb97f0/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertDeviceToControlPackets.cpp#L186
and the original `moduleOp` is now reused.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertDeviceToControlPackets.cpp
```diff
@@ -0,0 +1,203 @@
...
+                                           xilinx::AIE::DeviceOp deviceOp,
+                                           const std::string &pathToElfs) {
+  AMDAIEDeviceModel deviceModel = getDeviceModel(deviceOp.getDevice());
+
+  // Start collecting transations.
+  TRY_XAIE_API_LOGICAL_RESULT(XAie_StartTransaction, &deviceModel.devInst,
+                              XAIE_TRANSACTION_DISABLE_AUTO_FLUSH);
+  if (failed(addAllAieElfs(deviceModel, deviceOp, Path{pathToElfs},
+                           /*aieSim=*/false)))
+    return failure();
```
jtuyls, line 34:
```suggestion
                           /*aieSim=*/false))) {
    return failure();
  }
```
Because the condition has multiple lines.

jtuyls (approved):
LGTM, just a few nits.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
jtuyls: Cool, looks good, thanks!

Yu-Zhewen: As a follow-up, ideally, we should move the ELF binary and object file generation to the end of the main pipeline? This would allow the control packet-related passes to be integrated into the main pipeline as well.

https://github.com/nod-ai/iree-amd-aie/blob/06899b76e42dc59de60977512692e8a6d0cb97f0/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp#L1230-L1239

However, there seem to be known issues with shifting these passes from `XCLBinGen` to the main pipeline (#838).

I will have a look at this.

PR #1034, opened 2025-01-16
[BufferizeToAllocation] Add support to bufferize linalg.copy ops given input depth

yzhang93:
The defining ops of the inputs of computation op can be copy or pack ops. This PR extends the support to cover the copy ops given depth to the computational op.


jtuyls (approved):
LGTM

PR #1042, opened 2025-01-20
[DmaComposition] Rework logic to enable new composition opportunity

newling:
- slight modification to logic used for compositing DMAs, makes use of global offset difference (please see comments in code)
- more aggressive folding of unit dimensions 
- new unit test in combine_strided_ops.mlir shows the case that was being missed in an end-to-end matmul transpose b with the new pipeline pack-peel-4-level-tiling

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
@@ -7,201 +7,251 @@
 #include "AMDAIEDmaUtils.h"
 
 #include "AMDAIEUtils.h"
+#include "iree-amd-aie/IR/AMDAIEOps.h"
 #include "iree-amd-aie/Transforms/Utils/AMDAIEUtils.h"
 #include "llvm/ADT/SmallPtrSet.h"
+#include "llvm/ADT/SmallVector.h"
 #include "mlir/Dialect/Utils/StaticValueUtils.h"
 
 #define DEBUG_TYPE "iree-amdaie-dma-utils"
 
 namespace mlir::iree_compiler::AMDAIE {
 
-static bool isEqualConstantIntOrValueArrayFromIndices(
-    ArrayRef<OpFoldResult> ofrsA, ArrayRef<OpFoldResult> ofrsB,
-    size_t indexA = 0, size_t indexB = 0) {
-  if ((ofrsA.size() - indexA) != (ofrsB.size() - indexB)) return false;
-  return isEqualConstantIntOrValueArray(ofrsA.drop_front(indexA),
-                                        ofrsB.drop_front(indexB));
+// Functions in namespace `detail` are not intended to be used outside of this
+// file, but are exposed in the .h file for testing purposes.
+namespace detail {
+
+void matchStridesOfUnitDims(MLIRContext *ctx, ArrayRef<OpFoldResult> sizesX,
+                            SmallVector<OpFoldResult> &stridesX,
+                            SmallVector<OpFoldResult> &offsetsX,
+                            ArrayRef<OpFoldResult> stridesY) {
+  for (int i = 0; i < sizesX.size(); ++i) {
+    if (stridesX[i] != stridesY[i]) {
+      std::optional<int64_t> maybeIntSize = getConstantIntValue(sizesX[i]);
+      if (maybeIntSize.has_value() && maybeIntSize.value() == 1) {
+        std::optional<int64_t> maybeOffset = getConstantIntValue(offsetsX[i]);
+        std::optional<int64_t> maybeStrideX = getConstantIntValue(stridesX[i]);
+        std::optional<int64_t> maybeStrideY = getConstantIntValue(stridesY[i]);
+        if (maybeOffset.has_value() && maybeStrideX.has_value() &&
+            maybeStrideY.has_value()) {
+          int64_t offset = maybeOffset.value();
+          int64_t strideX = maybeStrideX.value();
+          int64_t strideY = maybeStrideY.value();
+          int64_t numerator = offset * strideX;
+          int64_t denominator = strideY;
+          if (numerator % denominator == 0) {
+            offsetsX[i] = getAsIndexOpFoldResult(ctx, numerator / denominator);
+            stridesX[i] = stridesY[i];
+          }
+        }
+      }
+    }
+  }
 }
 
-bool areAccessPatternsEqualFromIndices(ArrayRef<OpFoldResult> offsetsA,
-                                       ArrayRef<OpFoldResult> sizesA,
-                                       ArrayRef<OpFoldResult> stridesA,
-                                       ArrayRef<OpFoldResult> offsetsB,
-                                       ArrayRef<OpFoldResult> sizesB,
-                                       ArrayRef<OpFoldResult> stridesB,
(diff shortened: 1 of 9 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
```diff
@@ -511,6 +480,91 @@ LogicalResult moveNpuDmaSyncUsersAfterAncestorInSameBlock(
   return success();
 }
 
+/// Try to add `offsetToMerge` to one of the offsets in `offsets`.  This
+/// function assumes an effective stride for `offsetToMerge` of 1one so to add
```
yzhang93, line 484: I'm not sure what it should be. 1one -> one?

yzhang93 (commented):
LGTM. Some minor comments.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
newling: eleven ;) 
I'll fix, thanks.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/AMDAIEDmaUtilsTest.cpp
```diff
@@ -326,6 +332,9 @@ TEST_F(AccessPatternCombinationTest, CombineAccessPatterns) {
   checkCombineAccessPatterns({0, 0}, {16, 32}, {16, 1}, {1, 0, 0}, {3, 16, 32},
                              {64, 16, 1}, {0, 0, 0}, {4, 16, 32}, {64, 16, 1},
                              3);
+  // stride matching
```
jtuyls, line 335: Could you add a few more test cases for this case with more dimensions and with non-zero offsets for A? E.g. something like this:
```
checkCombineAccessPatterns({4, 0}, {8, 16}, {64, 1}, {4 * 64 + 8 * 64, 0}, {1, 16},
                             {1, 1}, {4, 0}, {9, 16}, {64, 1}, 3, true);
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
```diff
@@ -7,201 +7,275 @@
...
+///
+/// This function computes the difference between the global offsets of two
+/// access patterns. If it is not constant, i.e. if the difference contains
+/// an MLIR value which is not a constant, then nullopt is returned.
+///
+/// This function is useful when determining if the access pattern A, followed
+/// by the access pattern B, can be merged into a single access pattern.
+///
+/// \return global_offset(X) - global_offset(Y).
+std::optional<int64_t> getGlobalOffsetDifference(
```
jtuyls, line 100: I think it would be useful to add unit tests for this utility (and other similar utilities).

jtuyls (changes requested):
This is a great refactor with additional combinations being enabled, thanks for the thorough logic and great explanations!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
```diff
@@ -7,201 +7,275 @@
...
+      globalOffset += sign * o.value() * s.value();
+    } else if (o.has_value()) {
+      increment(cast<Value>(stride), sign * o.value());
+    } else if (s.has_value()) {
+      increment(cast<Value>(offset), sign * s.value());
     }
+    return true;
+  };
+
+  for (uint32_t i = 0; i < offsetsX.size(); ++i) {
```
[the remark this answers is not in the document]
newling: That was what I had originally, but then there are missed opportunities where the difference is constant, but the individual global offsets are not. I hit these cases with some of the end-to-end matmuls. 

Something like 
X = offsets:  [%v] sizes: [10] strides: [1]
Y = offsets:  [%v]  sizes: [10] strides: [1]

for example: the global offset of X (and Y) is %v  (i.e. non-constant), but the difference is %v - %v = 0 is constant. So by computing the global difference directly we see we can merge to 

merged = offsets: [0, %v] sizes: [2, 10] strides: [0, 1]. 

Another option could be to actually build a global offset expression with arith ops for X and Y, and then hope to somehow canonicalize the difference between them and hope to get a constant. This is what I had originally, but it didn't work.  

I will add more tests with non-constant offsets illustrating this.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/AMDAIEDmaUtilsTest.cpp
newling: I've added this and a few others

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
newling: I've added some in the .cpp test file for the constant cases. For the non-constant cases, I'm in the process of adding in the combine_strided_ops.mlir, need to add some negative cases though (non-constant global offset difference). Tomorrow..

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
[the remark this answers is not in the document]
jtuyls: Thanks for the explanation! I think it would be useful to add this to the docstring of the function as well. Can be brief, as a reference for someone touching that function in the future.

newling:
> This is a great refactor with additional combinations being enabled, thanks for the thorough logic and great explanations!

Thanks! I think I've addressed all the comments now, let me know if you think it should have more testing etc.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/AMDAIEDmaUtilsTest.cpp
```diff
@@ -6,15 +6,27 @@
...
 #include "llvm/ADT/SmallVectorExtras.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Utils/StaticValueUtils.h"
 
 namespace {
 
 using namespace mlir;
 using namespace mlir::iree_compiler::AMDAIE;
 
+SmallVector<int64_t> fromOpFoldResults(SmallVector<OpFoldResult> ofrs) {
```
jtuyls, line 18: Can't you use `getConstantIntValues`?

jtuyls (approved):
LGTM, just one nit above.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/AMDAIEDmaUtilsTest.cpp
newling: Oops second time. Changed

PR #1047, opened 2025-01-21
Disable failing batch_matmul tests

jtuyls:
The batch_matmul tests with the pack-peel-4-level-tiling pipeline result in intermittent numerics issues, making the CI unstable. I was able to reproduce this offline and will need to investigate this further.


Yu-Zhewen (approved):
LGTM

PR #1049, opened 2025-01-21
[CI] Disable turbo mode tests due to instability

jtuyls:
Locally, on a Strix machine, I have been able to crash the NPU in turbo mode, requiring a cold reboot. I am not sure that this is the same issue as we're occasionally seeing with the CI runners, but the result is similar, and I think it's worth a try disabling these tests for a while to verify.


Yu-Zhewen (approved):
LGTM

PR #1050, opened 2025-01-22
Update FusePackIntoLoop -> FuseProducerIntoLoop to include linalg.copy op

yzhang93:
In cases that `linalg.copy` ops are used instead of `tensor.pack` ops as the producers of the computation op, they should be able to be fused into the loops in the same way. This PR updates the pass to handle such cases.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseProducerIntoLoop.cpp
```diff
@@ -143,24 +151,31 @@ void AMDAIEFusePackIntoLoopPass::runOnOperation() {
...
           Operation *firstOpInBlock = &genericBlock->front();
           rewriter.moveOpBefore(parent, firstOpInBlock);
         }
+      } else if (auto parent = dyn_cast_if_present<linalg::CopyOp>(
+                     operand.getDefiningOp())) {
+        Block *genericBlock = genericOp->getBlock();
+        if (parent->getBlock() != genericBlock && parent->hasOneUse()) {
+          Operation *firstOpInBlock = &genericBlock->front();
+          rewriter.moveOpBefore(parent, firstOpInBlock);
+        }
```
Abhishek-Varma, line 169: This can be merged into the previous `if` itself as `parent` is being used as a generic operation.
```
if (isa_and_present<tensor::PackOp, linalg::CopyOp>(operand.getDefiningOp())) {
  Operation* parent = operand.getDefiningOp();
  . . .
}
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/fuse_producer_into_loop.mlir
```diff
@@ -1,7 +1,7 @@
...
-// RUN: iree-opt --split-input-file --pass-pipeline='builtin.module(func.func(iree-amdaie-fuse-pack-into-loop{fuse-pack-depth=2}))' %s | FileCheck %s --check-prefix=DEPTH-2
-// RUN: iree-opt --split-input-file --pass-pipeline='builtin.module(func.func(iree-amdaie-fuse-pack-into-loop{fuse-pack-depth=2 use-scf-for=false}))' %s | FileCheck %s --check-prefix=FORALL-DEPTH-2
+// RUN: iree-opt --split-input-file --pass-pipeline='builtin.module(func.func(iree-amdaie-fuse-producer-into-loop))' %s | FileCheck %s --check-prefix=DEPTH-1
+// RUN: iree-opt --split-input-file --pass-pipeline='builtin.module(func.func(iree-amdaie-fuse-producer-into-loop{use-scf-for=false}))' %s | FileCheck %s --check-prefix=FORALL-DEPTH-1
+// RUN: iree-opt --split-input-file --pass-pipeline='builtin.module(func.func(iree-amdaie-fuse-producer-into-loop{fuse-depth=2}))' %s | FileCheck %s --check-prefix=DEPTH-2
+// RUN: iree-opt --split-input-file --pass-pipeline='builtin.module(func.func(iree-amdaie-fuse-producer-into-loop{fuse-depth=2 use-scf-for=false}))' %s | FileCheck %s --check-prefix=FORALL-DEPTH-2
```
Abhishek-Varma, line 4: Could you add a lit test targeting `linalg.copy` ?

Abhishek-Varma (changes requested):

newling:
Alternative option: factorize [this](https://github.com/nod-ai/iree-amd-aie/blob/a43a23fe7a926f71f61a26c082d87fcc5a616a28/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertToDma.cpp#L347) into a new pass `iree-amdaie-copy-to-pack`, run that after after any pass that might introduce linalg.copy (ire-amdaie-pad etc), and then we can always assume there are no linalg.copy

yzhang93:
> This looks good to me, but an alternative option would be to factorize [this](https://github.com/nod-ai/iree-amd-aie/blob/a43a23fe7a926f71f61a26c082d87fcc5a616a28/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertToDma.cpp#L347) into a new pass `iree-amdaie-copy-to-pack`, run that after after any pass that might introduce linalg.copy (ire-amdaie-pad etc), and then we can always assume there are no linalg.copy

I know this trick would work as we used it as an intermediate step in `ConvertToDmaPass`, but I'm not sure if it's a good idea to write it as a formal pass at tensor level. In the definition of `PackOp`, the original tensor of rank `n` is packed into a result tensor of rank `n + k` where `k>0` as shown here https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/Tensor/IR/TensorOps.td#L1887

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/fuse_producer_into_loop.mlir
yzhang93: Done

newling:
:-0 so if those docs are correct you can have an identity IREE::LinalgExt::PackOp but not an identity Linalg::PackOp...

newling (approved):

Abhishek-Varma (approved):
LGTM.

PR #1052, opened 2025-01-22
[CtrlPkt] Add pass to split control packet payload into 4-byte chunks

Yu-Zhewen:


runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -206,6 +206,24 @@ inline ::XAie_TxnOpcode txnToTxn(XAie_TxnOpcode t) {
...
 
+/// Enum representing the format of the control packet header, which includes
+/// the following fields:
+/// - [19:0] Address,
+/// - [21:20] Beat, the number of 32-bit words data in the packet,
+/// - [23:22] Operation,
+/// - [28:24] Stream ID, for return packet,
+/// - [30:29] Reserved,
+/// - [31] Odd parity bit.
+enum class AMDAIECtrlPktHeader : uint8_t {
```
jtuyls, line 217: I guess we can't retrieve this information from aie-rt? 
Could you put this in the device model so it becomes changeable for different hardware generations?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_control_packet_data.mlir
```diff
@@ -0,0 +1,83 @@
...
+  dialect_resources: {
+    builtin: {
+      ctrl_pkt_data_0: "0x040000001501002000005500E00C07000100010001000100378803000000000000000000B7880300C00000000000000001000100BB100018C0010008000019800208010015010010000019200338997EFC0F99C2FC0F59F63C1FBB48031808100070C6FF1908001001000100010001000100D9C2FC0701000100010001000100D97EFC0719180010010001000100010019E0FF3F"
+    }
+  }
+#-}
+
+// -----
+
+// For control packet reads (where `opcode=1` and `data` are not present), the split still occurs, with only the `address` and `length` attributes being updated.
```
jtuyls, line 68: What's the rationale here? Shouldn't we throw an error?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitControlPacketData.cpp
```diff
@@ -0,0 +1,94 @@
...
+struct SplitControlPacketData
+    : public OpRewritePattern<AMDAIE::NpuControlPacketOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(AMDAIE::NpuControlPacketOp op,
+                                PatternRewriter &rewriter) const override {
+    // The maximum length (beats) of a control packet is determined by the
+    // number of bits allocated to the `beat` field in the control packet
+    // header.
+    unsigned maxLength =
```
jtuyls, line 32: If you put this information in the device model, you could provide a utility `getCtrlPktMaxLength` instead.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
Yu-Zhewen: Done

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitControlPacketData.cpp
Yu-Zhewen: Done

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_control_packet_data.mlir
Yu-Zhewen: We can send a control packet to request the core to send data back. In this case, the `opcode` is set to 1 (indicating a read request), the `length` field specifies the number of words to be returned, and the `data` attribute is left empty, as only the header is sent to the core.

Yu-Zhewen: An example is here:
https://github.com/Xilinx/mlir-aie/blob/0d03400dca76782df37ef9f945ce9820904d887f/test/npu-xrt/add_one_ctrl_packet/test.cpp#L150

jtuyls: Ok, thanks for the explanation, makes sense!

Also, just thinking about this now, but I think using an enum attribute instead of an int for the opcode would be helpful: `CtrlPktType::Rea` instead of `1` etc.

jtuyls (approved):
LGTM

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_control_packet_data.mlir
Yu-Zhewen: Thanks, I will make a separate PR for that.

Yu-Zhewen: implemented in #1056

PR #1053, opened 2025-01-22
UB fix 

newling:
Since https://github.com/nod-ai/iree-amd-aie/pull/1033 I have not been able to compile all the way to xclbin, because the value of  options.emitCtrlPkt over [here](https://github.com/nod-ai/iree-amd-aie/blob/a43a23fe7a926f71f61a26c082d87fcc5a616a28/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp#L1250) in the code when I print it is 17, even though the default value is false (0). 

I don't know where the bug is, iree-amd-aie, iree, or the clang C++ compiler I'm using. I've spent some time trying to find some invalid code, but without any success. The change in this PR fixes my problem, although it's definitely just a workaround for another deeper issue (initializing with {false} instead of = false should be valid afaik). 

The only nice thing about this PR is that IREE initializes options with `= false` instead of `{false}`, so at least we're aligning with IREE.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
@@ -27,10 +27,10 @@ struct AMDAIEOptions {
   std::string vitisInstallDir;
 
   // Dump system commands used during compilation
-  bool showInvokedCommands{false};
+  bool showInvokedCommands = false;
 
   // Use the chess compiler. The default is to use peano.
-  bool useChess{false};
+  bool useChess = false;
 
   // Additional flags to run peano's opt with (if peano is the backend compiler
   // selected). These are mostly appended on the end of the default flags, but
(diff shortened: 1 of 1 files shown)


jtuyls (commented):
Euh, this shouldn't make a difference afaik.. 

Not sure about IREE, and in most cases it doesn't matter too much, but list initialization is preferred by the developer of C++ and it should be the version that always works: https://stackoverflow.com/questions/18222926/what-are-the-advantages-of-list-initialization-using-curly-braces?rq=1

newling:
> Euh, this shouldn't make a difference afaik..
> 
> Not sure about IREE, and in most cases it doesn't matter too much, but list initialization is preferred by the developer of C++ and it should be the version that always works: https://stackoverflow.com/questions/18222926/what-are-the-advantages-of-list-initialization-using-curly-braces?rq=1

Yeah... agree. My UB woes seem to have mysteriously vanished now, so I'm going to close this...

PR #1060, opened 2025-01-27
[SplitLogicalObjFifo] Fix split-logicalobjfifo pass to analyse unique producers/consumers ObjFifos

Abhishek-Varma:
In order to decide the split factor we were basing the inference solely on the number of columns available.
Because of this, for 4x8 array and for the new pipeline, we were splitting L2 buffers :-
1. LHS - 8 times.
2. RHS - 8 times.
3. OUT - 8 times.

As a result, the tiles being assigned were :-
1. LHS : (0,0) -> (7,0)
2. RHS : (0,0) -> (7,0)
3. OUT : (0,0) -> (7,0)

This causes exhaustion of DMA channels. Refer to [this](https://github.com/nod-ai/iree-amd-aie/pull/1031#discussion_r1920237380) thread for the discussion thread.

The L2 buffer split should be :-
1. LHS : (0,0) -> (3,0)
2. RHS : (0,0) -> (7,0)
3. OUT : (0,0) -> (7,0)

So that later on when the tiles are being assigned, the expected no. of tile assignments for LHS/RHS/OUT matches the corresponding L2 buffers.

This PR aims to analyse the number of unique producers/consumers ObjFifos for the ObjFifo being split..

e2e CI test for Matmul both with/without ukernel via `pack-peel-4-level-tiling` pipeline targeting 4x8 array on Strix have been added.

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -1805,6 +1805,26 @@ class Tests:
                 additional_labels=["I8UKernel"],
             )
         )
+        self.register(
+            Matmul(
+                64,
+                64,
+                64,
+                "bf16",
+                "f32",
+                name_suffix="4rows_8cols_npu4_pack_peel_4_level_tiling_ukernel",
+                use_ukernel=True,
+                tile_pipeline="pack-peel-4-level-tiling",
+                run_on_target=["npu4"],
+                aie_compilation_flags=[
+                    "--iree-amdaie-num-rows=4",
+                    "--iree-amdaie-num-cols=8",
+                    "--iree-amd-aie-additional-peano-opt-flags=-O3",
+                    "--iree-amdaie-enable-function-outlining=True",
+                ],
+                use_chess=True,
+            )
+        )
 
         # Matmul test on 2(rows)x2(cols) cores
         self.register(
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -1721,6 +1721,25 @@ class Tests:
                 use_chess=False,
             )
         )
+        self.register(
+            Matmul(
+                512,
+                512,
+                256,
+                "i32",
+                "i32",
+                name_suffix="4rows_8cols_npu4_pack_peel_4_level_tiling",
+                tile_pipeline="pack-peel-4-level-tiling",
+                run_on_target=["npu4"],
+                aie_compilation_flags=[
+                    "--iree-amdaie-num-rows=4",
+                    "--iree-amdaie-num-cols=8",
+                    "--iree-amd-aie-additional-peano-opt-flags=-O3",
+                    "--iree-amdaie-enable-function-outlining=True",
+                ],
+                use_chess=True,
+            )
+        )
 
         for target in ["npu1_4col", "npu4"]:
             self.register(
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -1734,10 +1734,7 @@ class Tests:
                 aie_compilation_flags=[
                     "--iree-amdaie-num-rows=4",
                     "--iree-amdaie-num-cols=8",
-                    "--iree-amd-aie-additional-peano-opt-flags=-O3",
-                    "--iree-amdaie-enable-function-outlining=True",
                 ],
-                use_chess=True,
             )
         )
 
@@ -1819,10 +1816,7 @@ class Tests:
                 aie_compilation_flags=[
                     "--iree-amdaie-num-rows=4",
                     "--iree-amdaie-num-cols=8",
-                    "--iree-amd-aie-additional-peano-opt-flags=-O3",
-                    "--iree-amdaie-enable-function-outlining=True",
                 ],
-                use_chess=True,
             )
         )
 
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -1817,6 +1817,7 @@ class Tests:
                     "--iree-amdaie-num-rows=4",
                     "--iree-amdaie-num-cols=8",
                 ],
+                use_chess=True,
             )
         )
 
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIELogicalObjFifoOpInterface.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIELogicalObjFifoOpInterface.cpp
@@ -17,10 +17,10 @@ SmallVector<mlir::CopyOpInterface> getCopyLikeConsumers(
     LogicalObjFifoOpInterface op) {
   SmallVector<mlir::CopyOpInterface> copyLikOps;
   for (Operation *userOp : op->getUsers()) {
-    if (auto copyOp = dyn_cast<CopyOpInterface>(userOp)) {
-      if (dyn_cast_if_present<LogicalObjFifoOpInterface>(
-              copyOp.getSource().getDefiningOp()) == op)
-        copyLikOps.push_back(copyOp);
+    if (auto copyOp = dyn_cast<CopyOpInterface>(userOp);
+        copyOp && dyn_cast_if_present<LogicalObjFifoOpInterface>(
+                      copyOp.getSource().getDefiningOp()) == op) {
+      copyLikOps.push_back(copyOp);
     }
   }
   return copyLikOps;
@@ -30,10 +30,10 @@ SmallVector<mlir::CopyOpInterface> getCopyLikeProducers(
     LogicalObjFifoOpInterface op) {
   SmallVector<mlir::CopyOpInterface> copyLikOps;
   for (Operation *userOp : op->getUsers()) {
-    if (auto copyOp = dyn_cast<CopyOpInterface>(userOp)) {
-      if (dyn_cast_if_present<LogicalObjFifoOpInterface>(
-              copyOp.getTarget().getDefiningOp()) == op)
-        copyLikOps.push_back(copyOp);
+    if (auto copyOp = dyn_cast<CopyOpInterface>(userOp);
+        copyOp && dyn_cast_if_present<LogicalObjFifoOpInterface>(
+                      copyOp.getTarget().getDefiningOp()) == op) {
+      copyLikOps.push_back(copyOp);
     }
   }
   return copyLikOps;
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIELogicalObjFifoOpInterface.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIELogicalObjFifoOpInterface.cpp
@@ -17,10 +17,10 @@ SmallVector<mlir::CopyOpInterface> getCopyLikeConsumers(
     LogicalObjFifoOpInterface op) {
   SmallVector<mlir::CopyOpInterface> copyLikOps;
   for (Operation *userOp : op->getUsers()) {
-    if (auto copyOp = dyn_cast<CopyOpInterface>(userOp);
-        dyn_cast_if_present<LogicalObjFifoOpInterface>(
-            copyOp.getSource().getDefiningOp()) == op) {
-      copyLikOps.push_back(copyOp);
+    if (auto copyOp = dyn_cast<CopyOpInterface>(userOp)) {
+      if (dyn_cast_if_present<LogicalObjFifoOpInterface>(
+              copyOp.getSource().getDefiningOp()) == op)
+        copyLikOps.push_back(copyOp);
     }
   }
   return copyLikOps;
@@ -30,10 +30,10 @@ SmallVector<mlir::CopyOpInterface> getCopyLikeProducers(
     LogicalObjFifoOpInterface op) {
   SmallVector<mlir::CopyOpInterface> copyLikOps;
   for (Operation *userOp : op->getUsers()) {
-    if (auto copyOp = dyn_cast<CopyOpInterface>(userOp);
-        dyn_cast_if_present<LogicalObjFifoOpInterface>(
-            copyOp.getTarget().getDefiningOp()) == op) {
-      copyLikOps.push_back(copyOp);
+    if (auto copyOp = dyn_cast<CopyOpInterface>(userOp)) {
+      if (dyn_cast_if_present<LogicalObjFifoOpInterface>(
+              copyOp.getTarget().getDefiningOp()) == op)
+        copyLikOps.push_back(copyOp);
     }
   }
   return copyLikOps;
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
@@ -349,7 +349,23 @@ void AMDAIESplitLogicalObjFifosPass::runOnOperation() {
   }
 
   // Fetch and store all unique pairs of L2<->L1 Copy ops. This would helps us
-  // figure out the split factor for all LogicalObjectFifos.
+  // figure out the split factor for all LogicalObjectFifos. Basically we get to
+  // decide how many splits to perform for a particular L2 ObjectFifo based on
+  // the total unique L2<->L1 Copy ops.
+  // Eg:
+  //      %lhs = LOF_on_L2
+  //      %a = LOF_on_L1_0
+  //      %b = LOF_on_L1_1
+  //      %c = LOF_on_L1_2
+  //      DMA(%a, %lhs)
+  //      DMA(%b, %lhs)
+  //      DMA(%c, %lhs)
+  //      DMA(%b, %lhs)
+  //      DMA(%c, %lhs)
+  //
+  //    In the above snippet although we have 5 DMA ops for L2<->L1, only 3 of
+  //    them are unique. Hence we'd split %lhs into 3 unique splits, instead
+  //    of 5.
   DenseMap<Operation *, DenseSet<Operation *>> uniqueL2L1Pair;
   moduleOp->walk([&](Operation *op) -> WalkResult {
     if (auto copyOp = dyn_cast<CopyOpInterface>(op)) {
@@ -371,16 +387,19 @@ void AMDAIESplitLogicalObjFifosPass::runOnOperation() {
       }
       Operation *l2DefOp = nullptr;
       Operation *l1DefOp = nullptr;
-      if (source.getMemorySpaceAsUInt() == 1 &&
-          target.getMemorySpaceAsUInt() == 2) {
+      // L2 -> L1.
+      if (target.getMemorySpaceAsUInt() == 2) {
         l2DefOp = sourceFromMemrefOp.getMemref().getDefiningOp();
         l1DefOp = targetFromMemrefOp;
-      } else if (target.getMemorySpaceAsUInt() == 1 &&
-                 source.getMemorySpaceAsUInt() == 2) {
+      } else if (source.getMemorySpaceAsUInt() == 2) {
+        // L1 -> L2.
         l2DefOp = targetFromMemrefOp.getMemref().getDefiningOp();
         l1DefOp = sourceFromMemrefOp;
       } else {
         return WalkResult::advance();
+      }
+      if (!l2DefOp || !l1DefOp) {
+        return WalkResult::interrupt();
       }
       uniqueL2L1Pair[l2DefOp].insert(l1DefOp);
       return WalkResult::advance();
@@ -392,30 +411,31 @@ void AMDAIESplitLogicalObjFifosPass::runOnOperation() {
   /// dimensions.
   DenseMap<Operation *, int64_t> splitFactorOfLOF;
   for (auto &&[dmaOp, dmaSplitInfo] : dmaSplitInfoMap) {
-    auto stridedOp =
-        cast<AMDAIE::DoublyStridedOpInterface>(dmaOp.getOperation());
     auto dmaCpyNd = cast<AMDAIE::DmaCpyNdOp>(dmaOp.getOperation());
     int64_t splitFactor = dmaSplitInfo.splitSize;
-    if (stridedOp.getSourceMemorySpaceAsUInt() == 0) {
-      splitFactor =
-          uniqueL2L1Pair
-              [dmaCpyNd.getTarget()
-                   .getDefiningOp<AMDAIE::LogicalObjectFifoFromMemrefOp>()
-                   .getMemref()
-                   .getDefiningOp()]
-                  .size();
-    } else if (stridedOp.getTargetMemorySpaceAsUInt() == 0) {
-      splitFactor =
-          uniqueL2L1Pair
-              [dmaCpyNd.getSource()
-                   .getDefiningOp<AMDAIE::LogicalObjectFifoFromMemrefOp>()
-                   .getMemref()
-                   .getDefiningOp()]
-                  .size();
+    auto sourceDefOp =
+        dmaCpyNd.getSource()
+            .getDefiningOp<AMDAIE::LogicalObjectFifoFromMemrefOp>();
+    auto targetDefOp =
+        dmaCpyNd.getTarget()
+            .getDefiningOp<AMDAIE::LogicalObjectFifoFromMemrefOp>();
+    if (!sourceDefOp || !targetDefOp) {
+      LLVM_DEBUG(llvm::dbgs()
+                 << "Expected defining op of source/target for : " << dmaOp);
+      return signalPassFailure();
     }
+    if (dmaCpyNd.getSourceMemorySpaceAsUInt() == 0) {
+      if (Operation *l2DefOp = targetDefOp.getMemref().getDefiningOp())
+        splitFactor = uniqueL2L1Pair[l2DefOp].size();
+    } else if (dmaCpyNd.getTargetMemorySpaceAsUInt() == 0) {
+      if (Operation *l2DefOp = sourceDefOp.getMemref().getDefiningOp())
+        splitFactor = uniqueL2L1Pair[l2DefOp].size();
+    }
+    // In cases where the number of available columns < the inferred split
+    // factor, we'll cap the final split factor by the lower bound.
     splitFactor = std::gcd(dmaSplitInfo.splitSize, splitFactor);
     FailureOr<int64_t> maybeSplitFactor = splitDoublyStridedOp(
-        rewriter, stridedOp, dmaSplitInfo.sourceSplitDim,
+        rewriter, dmaCpyNd, dmaSplitInfo.sourceSplitDim,
         dmaSplitInfo.targetSplitDim, splitFactor, dmaSplitInfo.newSourceStride,
         dmaSplitInfo.newTargetStride);
     if (failed(maybeSplitFactor)) {
@@ -423,8 +443,11 @@ void AMDAIESplitLogicalObjFifosPass::runOnOperation() {
                  << "Failed to perform splitting of the DMA op: " << dmaOp);
       return signalPassFailure();
     }
-    splitFactorOfLOF[dmaCpyNd.getTarget().getDefiningOp()] = *maybeSplitFactor;
-    splitFactorOfLOF[dmaCpyNd.getSource().getDefiningOp()] = *maybeSplitFactor;
+    // The above function might change the split factor based on divisibility
+    // with source/target. Therefore here we maintain the final split factor
+    // which we'll use later to split the LogicalObjectFifo.
+    splitFactorOfLOF[targetDefOp] = *maybeSplitFactor;
+    splitFactorOfLOF[sourceDefOp] = *maybeSplitFactor;
   }
   for (auto &&[objFifo, splitInfo] : objFifoSplitInfoMap) {
     if (failed(splitLogicalObjectFifo(rewriter, objFifo, splitInfo.splitDim,
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
@@ -137,20 +137,23 @@ FailureOr<int64_t> getSplitStride(ArrayRef<AMDAIE::DmaCpyNdOp> dmaOps,
 ///      DMA(%c, %lhs)
 ///
 ///    In the above snippet although we have 5 DMA ops for L2<->L1, only 3 of
-///    them are unique. Hence we'd split %lhs into 3 unique splits, instead of 5.
-static FailureOr<int64_t> fetchTotalUniqueL2L1(SmallVector<CopyOpInterface> copyLikeOps, bool fetchTarget) {
-  DenseSet<Operation*> uniqueLof;
+///    them are unique. Hence we'd split %lhs into 3 unique splits, instead
+///    of 5.
+static FailureOr<int64_t> fetchTotalUniqueL2L1(
+    SmallVector<CopyOpInterface> copyLikeOps, bool fetchTarget) {
+  DenseSet<Operation *> uniqueLof;
   for (CopyOpInterface copyOp : copyLikeOps) {
     AMDAIE::LogicalObjectFifoFromMemrefOp lof = nullptr;
     if (fetchTarget) {
       lof = dyn_cast_if_present<AMDAIE::LogicalObjectFifoFromMemrefOp>(
-            copyOp.getTarget().getDefiningOp());
+          copyOp.getTarget().getDefiningOp());
     } else {
       lof = dyn_cast_if_present<AMDAIE::LogicalObjectFifoFromMemrefOp>(
-            copyOp.getSource().getDefiningOp());
+          copyOp.getSource().getDefiningOp());
     }
     if (!lof) {
-      return copyOp.emitOpError()<< "could not retrieve source/target objectFifo";
+      return copyOp.emitOpError()
+             << "could not retrieve source/target objectFifo";
     }
     uniqueLof.insert(lof);
   }
@@ -181,7 +184,8 @@ LogicalResult collectSplittingDims(
     ModuleOp &moduleOp, const SmallVector<DmaObjFifoPairT> &dmaObjFifoPairs,
     DenseMap<AMDAIE::DmaCpyNdOp, DmaSplitInfo> &dmaSplitInfoMap,
     DenseMap<AMDAIE::LogicalObjectFifoFromMemrefOp, ObjFifoSplitInfo>
-        &objFifoSplitInfoMap, int64_t numCols) {
+        &objFifoSplitInfoMap,
+    int64_t numCols) {
   for (auto [dmaOp, objFifo] : dmaObjFifoPairs) {
     LLVM_DEBUG(llvm::dbgs() << "dmaOp: " << dmaOp << "\n");
     LLVM_DEBUG(llvm::dbgs() << "objFifo: " << objFifo << "\n");
@@ -256,7 +260,8 @@ LogicalResult collectSplittingDims(
       // Calculate the new source stride to be used for splitting the DMA.
       int64_t newSourceStride =
           splitStride != 1 ? splitDimSize / splitStride : 1;
-      FailureOr<int64_t> maybeUniqueL2L1 = fetchTotalUniqueL2L1(objFifo.getCopyLikeConsumers(), /*fetchTarget=*/true);
+      FailureOr<int64_t> maybeUniqueL2L1 = fetchTotalUniqueL2L1(
+          objFifo.getCopyLikeConsumers(), /*fetchTarget=*/true);
       if (failed(maybeUniqueL2L1)) {
         objFifo.emitOpError()
             << "could not retrieve total unique L2<->L1 pairs";
@@ -277,7 +282,8 @@ LogicalResult collectSplittingDims(
       LLVM_DEBUG(llvm::dbgs() << "splitFactor: " << splitFactor << "\n");
       dmaSplitInfoMap[dmaOp] = {sourceSplitDim, newSourceStride, targetSplitDim,
                                 1, splitFactor};
-      objFifoSplitInfoMap[objFifo] = {objFifoSplitDim, splitFactor, splitStride};
+      objFifoSplitInfoMap[objFifo] = {objFifoSplitDim, splitFactor,
+                                      splitStride};
     } else if (dmaOp.getSourceObjectFifo() == objFifo) {
       // Find outermost dimension in the access pattern that has stride ==
       // sizeAfterSplit and size != 1.
@@ -323,7 +329,8 @@ LogicalResult collectSplittingDims(
       // Calculate the new target stride to be used for splitting the DMA.
       int64_t newTargetStride =
           splitStride != 1 ? splitDimSize / splitStride : 1;
-      FailureOr<int64_t> maybeUniqueL2L1 = fetchTotalUniqueL2L1(objFifo.getCopyLikeProducers(), /*fetchTarget=*/false);
+      FailureOr<int64_t> maybeUniqueL2L1 = fetchTotalUniqueL2L1(
+          objFifo.getCopyLikeProducers(), /*fetchTarget=*/false);
       if (failed(maybeUniqueL2L1)) {
         objFifo.emitOpError()
             << "could not retrieve total unique L2<->L1 pairs";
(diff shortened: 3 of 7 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos.mlir
```diff
@@ -436,38 +461,133 @@ module attributes {hal.executable.target = #executable_target_amdaie_pdi_fb} {
...
     %2 = amdaie.dma_cpy_nd(%0[0, 0] [256, 32] [128, 1], %1[0, 0, 0, 0] [4, 32, 1, 32] [1024, 32, 1024, 1]) : (!amdaie.logicalobjectfifo<memref<256x128xi32>>, !amdaie.logicalobjectfifo<memref<4x1x32x32xi32, 1 : i32>>)
     memref.dealloc %alloc_0 : memref<4x1x32x32xi32, 1 : i32>
     memref.dealloc %alloc : memref<1x1x4x8x4x8xi32, 2 : i32>
     return
   }
 }
+
+// -----
+
+// CHECK-LABEL: @pack_peel_4_level_4x8_Strix
```
yzhang93, line 501: I think for the CHECKs, the focus should be the L2 buffers and logical objectfifos. Could you remove those L3 and tile checks? Also could you add some comments on what we result IR we should expect? I don't understand how the L2 buffers are splitted. I thought the original L2 buffer should be split into 8 L2 buffers?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos.mlir
```diff
@@ -436,38 +461,133 @@ module attributes {hal.executable.target = #executable_target_amdaie_pdi_fb} {
...
+    %alloc_1 = memref.alloc() : memref<1x1x8x8x4x8xi32, 2 : i32>
+    %alloc_2 = memref.alloc() : memref<16x8x32x32xi32, 1 : i32>
+    %tile_0_1 = amdaie.tile(%c0, %c1)
+    %lof_0_1 = amdaie.logicalobjectfifo.from_memref %alloc_2, {%tile_0_1} : memref<16x8x32x32xi32, 1 : i32> -> !amdaie.logicalobjectfifo<memref<16x8x32x32xi32, 1 : i32>>
+    %alloc_3 = memref.alloc() : memref<16x8x64x32xi32, 1 : i32>
+    %tile_1_1 = amdaie.tile(%c1, %c1)
+    %lof_1_1 = amdaie.logicalobjectfifo.from_memref %alloc_3, {%tile_1_1} : memref<16x8x64x32xi32, 1 : i32> -> !amdaie.logicalobjectfifo<memref<16x8x64x32xi32, 1 : i32>>
+    %alloc_4 = memref.alloc() : memref<8x8x32x64xi32, 1 : i32>
+    %tile_2_1 = amdaie.tile(%c2, %c1)
+    %lof_2_1 = amdaie.logicalobjectfifo.from_memref %alloc_4, {%tile_2_1} : memref<8x8x32x64xi32, 1 : i32> -> !amdaie.logicalobjectfifo<memref<8x8x32x64xi32, 1 : i32>>
```
yzhang93, line 552: If these IR are from a real IR dump, it seems to be some redundant allocations. Why there are both `memref<16x8x64x32xi32, 1 : i32>` and `memref<8x8x32x64xi32, 1 : i32>` in L2 memory space?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -348,23 +348,87 @@ void AMDAIESplitLogicalObjFifosPass::runOnOperation() {
...
   /// dimensions.
+  DenseMap<Operation *, int64_t> splitFactorOfLOF;
   for (auto &&[dmaOp, dmaSplitInfo] : dmaSplitInfoMap) {
     auto stridedOp =
         cast<AMDAIE::DoublyStridedOpInterface>(dmaOp.getOperation());
-    if (failed(splitDoublyStridedOp(
-            rewriter, stridedOp, dmaSplitInfo.sourceSplitDim,
-            dmaSplitInfo.targetSplitDim, dmaSplitInfo.splitSize,
-            dmaSplitInfo.newSourceStride, dmaSplitInfo.newTargetStride))) {
+    auto dmaCpyNd = cast<AMDAIE::DmaCpyNdOp>(dmaOp.getOperation());
```
yzhang93, line 397: It seems redundant and weird to me that you cast the dma operation to both `DoublyStridedOpInterface` and `DmaCpyNdOp`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -348,23 +348,87 @@ void AMDAIESplitLogicalObjFifosPass::runOnOperation() {
...
+    int64_t splitFactor = dmaSplitInfo.splitSize;
+    if (stridedOp.getSourceMemorySpaceAsUInt() == 0) {
+      splitFactor =
+          uniqueL2L1Pair
+              [dmaCpyNd.getTarget()
+                   .getDefiningOp<AMDAIE::LogicalObjectFifoFromMemrefOp>()
+                   .getMemref()
+                   .getDefiningOp()]
```
yzhang93, line 405: We don't need any checks when calling `getDefiningOp()`? Also please add some comments on the logic to update split factor.

yzhang93 (changes requested):

build_tools/ci/cpu_comparison/run.py
```diff
@@ -1712,6 +1712,25 @@ def __init__(self):
...
+                name_suffix="4rows_8cols_npu4_pack_peel_4_level_tiling",
+                tile_pipeline="pack-peel-4-level-tiling",
+                run_on_target=["npu4"],
+                aie_compilation_flags=[
+                    "--iree-amdaie-num-rows=4",
+                    "--iree-amdaie-num-cols=8",
+                    "--iree-amd-aie-additional-peano-opt-flags=-O3",
+                    "--iree-amdaie-enable-function-outlining=True",
+                ],
+                use_chess=True,
```
jtuyls, line 1731: We should have most tests on peano as the compilation is faster.
```suggestion
```

build_tools/ci/cpu_comparison/run.py
```diff
@@ -1777,6 +1796,26 @@ def __init__(self):
...
+                use_ukernel=True,
+                tile_pipeline="pack-peel-4-level-tiling",
+                run_on_target=["npu4"],
+                aie_compilation_flags=[
+                    "--iree-amdaie-num-rows=4",
+                    "--iree-amdaie-num-cols=8",
+                    "--iree-amd-aie-additional-peano-opt-flags=-O3",
+                    "--iree-amdaie-enable-function-outlining=True",
+                ],
+                use_chess=True,
```
jtuyls, line 1816:
```suggestion
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIELogicalObjFifoSplittingUtils.h
```diff
@@ -38,9 +38,11 @@ LogicalResult splitLogicalObjectFifo(
     int64_t splitStride = 1);
 
 /// Split doubly strided operations on a source and target split dimension with
-/// the provided split factor. If no split factor is provided, the doubly
-/// strided operation will be split on the size of the dimension being split.
-LogicalResult splitDoublyStridedOp(
+/// the provided split factor which might get updated. On success, return the
+/// split factor to the caller, else return failure.
```
jtuyls, line 42: Now that the `splittingFactor` is returned and expected to be used in the user, this is getting too coupled with the user function. We should extract the `gcd` function and update logic from this function and put it in the user instead.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -348,23 +348,87 @@ void AMDAIESplitLogicalObjFifosPass::runOnOperation() {
...
+          dyn_cast_if_present<AMDAIE::LogicalObjectFifoFromMemrefOp>(
+              copyOp.getTarget().getDefiningOp());
+      if (!sourceFromMemrefOp || !targetFromMemrefOp) {
+        return WalkResult::interrupt();
+      }
+      Operation *l2DefOp = nullptr;
+      Operation *l1DefOp = nullptr;
+      if (source.getMemorySpaceAsUInt() == 1 &&
+          target.getMemorySpaceAsUInt() == 2) {
+        l2DefOp = sourceFromMemrefOp.getMemref().getDefiningOp();
```
jtuyls, line 376: Why do you need the defining op of the memref? Can't you just work with objectFifos?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -348,23 +348,87 @@ void AMDAIESplitLogicalObjFifosPass::runOnOperation() {
...
+      auto targetFromMemrefOp =
+          dyn_cast_if_present<AMDAIE::LogicalObjectFifoFromMemrefOp>(
+              copyOp.getTarget().getDefiningOp());
+      if (!sourceFromMemrefOp || !targetFromMemrefOp) {
+        return WalkResult::interrupt();
+      }
+      Operation *l2DefOp = nullptr;
+      Operation *l1DefOp = nullptr;
+      if (source.getMemorySpaceAsUInt() == 1 &&
+          target.getMemorySpaceAsUInt() == 2) {
```
jtuyls, line 375: I think you can avoid relying on the memory spaces by collecting all copy ops that operate on the objectFifos being split.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -348,23 +348,87 @@ void AMDAIESplitLogicalObjFifosPass::runOnOperation() {
     return signalPassFailure();
   }
 
+  // Fetch and store all unique pairs of L2<->L1 Copy ops. This would helps us
+  // figure out the split factor for all LogicalObjectFifos.
+  DenseMap<Operation *, DenseSet<Operation *>> uniqueL2L1Pair;
+  moduleOp->walk([&](Operation *op) -> WalkResult {
```
jtuyls, line 354: I think this could be put in a utility function. Also, instead of walking all the copy ops, I think you can walk the objectFifos being split and their producers/consumers, there should be existing functions: `getCopyLikeProducers`, `getCopyLikeConsumers`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -348,23 +348,87 @@ void AMDAIESplitLogicalObjFifosPass::runOnOperation() {
     return signalPassFailure();
   }
 
+  // Fetch and store all unique pairs of L2<->L1 Copy ops. This would helps us
+  // figure out the split factor for all LogicalObjectFifos.
+  DenseMap<Operation *, DenseSet<Operation *>> uniqueL2L1Pair;
```
jtuyls, line 353: As you're just using the `.size()` of the `DenseSet`, you don't need to keep track of all ops, just how many there are.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -348,23 +348,87 @@ void AMDAIESplitLogicalObjFifosPass::runOnOperation() {
...
                  << "Failed to perform splitting of the DMA op: " << dmaOp);
       return signalPassFailure();
     }
+    splitFactorOfLOF[dmaCpyNd.getTarget().getDefiningOp()] = *maybeSplitFactor;
+    splitFactorOfLOF[dmaCpyNd.getSource().getDefiningOp()] = *maybeSplitFactor;
   }
   for (auto &&[objFifo, splitInfo] : objFifoSplitInfoMap) {
     if (failed(splitLogicalObjectFifo(rewriter, objFifo, splitInfo.splitDim,
-                                      splitInfo.splitSize,
+                                      splitFactorOfLOF[objFifo],
```
jtuyls, line 431: You're basically computing and discarding the `splitInfo.splitSize`. Instead, you can provide the map with producers/consumers to the `collectSplittingDims` function (or compute the necessary info inside) and use it to calculate a better `splitInfo.splitSize`.

jtuyls (changes requested):

build_tools/ci/cpu_comparison/run.py
```diff
@@ -1712,6 +1712,25 @@ def __init__(self):
...
+                "i32",
+                "i32",
+                name_suffix="4rows_8cols_npu4_pack_peel_4_level_tiling",
+                tile_pipeline="pack-peel-4-level-tiling",
+                run_on_target=["npu4"],
+                aie_compilation_flags=[
+                    "--iree-amdaie-num-rows=4",
+                    "--iree-amdaie-num-cols=8",
+                    "--iree-amd-aie-additional-peano-opt-flags=-O3",
+                    "--iree-amdaie-enable-function-outlining=True",
```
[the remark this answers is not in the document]
Abhishek-Varma: Function outlining is but Peano-opt-flag isn't default I see.

But I've removed them both anyway - it works.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos.mlir
Abhishek-Varma: >Could you remove those L3 and tile checks?

Done.

> Also could you add some comments on what we result IR we should expect? 

Done.

>  I don't understand how the L2 buffers are splitted. I thought the original L2 buffer should be split into 8 L2 buffers?

We should not split LHS into 8 L2 buffers. I've updated the PR's description to highlight that better along with a discussion thread.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
Abhishek-Varma: Updated comments on the logic for split factor.

build_tools/ci/cpu_comparison/run.py
Abhishek-Varma: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
Abhishek-Varma: Good point. Updated.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
Abhishek-Varma: Good point.

I did try that but it seems to be doing a similar thing which the current `walk` achieves.

For LHS/RHS I need L2 -> L1 (so I need `getCopyLikeConsumers`)
For OUT I need L1 -> L2 (so I need `getCopyLikeProducers`)

Since, currently, there is no way to identify which is what by walking just the objectFifos, I'll have to invoke both the APIs and then filter out L2<->L1.

> I think you can avoid relying on the memory spaces by collecting all copy ops that operate on the objectFifos being split.

In the `walk` which I've implemented, a similar feat is being achieved where I have all the copy ops and I'm filtering L2<->L1. I don't think I can get rid of the memory checks. In case I missed something, let me know.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
Abhishek-Varma: Responded in the [following](https://github.com/nod-ai/iree-amd-aie/pull/1060#discussion_r1933355809) thread.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
Abhishek-Varma: Great point! Updated.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
jtuyls: You don't need to filter out the `L2<->L1` connections as you just need to take the other side of the DMA being split: if the producer of an objectFifo and the objectFifo is being split, then you need to retrieve copy-like consumers and the other way around. It doesn't matter whether the producer/consumer is L1<->L2, L3<->L2, L1<->L3 or something else, so we should avoid making that assumption. In general, we should avoid checking memory space hardcoding etc whenever possible as it makes the logic fragile towards future changes (a memory space being added, skipped, removed, renamed).

build_tools/ci/cpu_comparison/run.py
jtuyls: The `use_chess=True,` is still there?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIELogicalObjFifoSplittingUtils.h
```diff
@@ -38,8 +38,10 @@ LogicalResult splitLogicalObjectFifo(
     int64_t splitStride = 1);
 
 /// Split doubly strided operations on a source and target split dimension with
-/// the provided split factor. If no split factor is provided, the doubly
-/// strided operation will be split on the size of the dimension being split.
+/// the provided split factor which might get updated. On success, return the
+/// split factor to the caller, else return failure.
```
jtuyls, line 42: Not true anymore?

build_tools/ci/cpu_comparison/run.py
jtuyls: Did you try removing it? I don't think it needs to be there as even when using peano for general compilation, the ukernel will still be compiled using chess.

jtuyls: And not all ukernel tests have `use_chess=True`: https://github.com/nod-ai/iree-amd-aie/blob/7a059cdd2a40f99921e1283c6a340c2a7cf0973e/build_tools/ci/cpu_comparison/run.py#L1771

Abhishek-Varma: Yep, I did.
It lead to the failure only in that e2e test where it didn't find some `mm_primitive` in the `.o` (that's what I can recollect from the CI) - but I can reconfirm the same and revert.

Abhishek-Varma: Here's the error after removing `use_chess=True` :-
```
ld.lld: error: undefined symbol: me_primitive::control_rnd
>>> referenced by mm_npu4.o
>>>               /home/shark-strix-1/workspace/iree-amd-aie/test_matmuls/matmul_64_64_64_bf16_f32_4rows_8cols_npu4_pack_peel_4_level_tiling_ukernel_ukernel/mm_npu4.o:(void matmul_vectorized_bf16_f32<2u, 8u, 1u, 8u, 8u, 8u>(bfloat16 const*, unsigned int, bfloat16 const*, unsigned int, float*, unsigned int))
```

jtuyls: Interesting, might not be implemented in peano yet I guess, let's keep `use_chess=True` then. It might be good to track this error down a bit further and create an issue on the llvm-aie repo if it's indeed unsupported.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIELogicalObjFifoSplittingUtils.cpp
```diff
@@ -800,15 +801,9 @@ LogicalResult splitDoublyStridedOp(IRRewriter &rewriter,
...
-                            : std::gcd(sourceSize, targetSize);
-  if (sourceSize % splitFactor != 0 || targetSize % splitFactor != 0) {
-    int64_t newSplitFactor = std::gcd(sourceSize, targetSize);
-    LLVM_DEBUG(llvm::dbgs() << "split factor has been changed from "
-                            << splitFactor << " to " << newSplitFactor);
-    splitFactor = newSplitFactor;
-  }
+  assert(maybeSplitFactor.has_value() &&
+         "expected split factor to be sent by the caller");
+  int64_t splitFactor = maybeSplitFactor.value();
```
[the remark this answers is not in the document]
Abhishek-Varma: I have chosen the latter. Updated.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
Abhishek-Varma: Aah yes! I missed "`instead of walking all the copy ops, I think you can walk the objectFifos being split`".

Makes sense. Updated. Thanks!

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIELogicalObjFifoOpInterface.cpp
```diff
@@ -17,10 +17,10 @@ SmallVector<mlir::CopyOpInterface> getCopyLikeConsumers(
...
   SmallVector<mlir::CopyOpInterface> copyLikOps;
   for (Operation *userOp : op->getUsers()) {
-    if (auto copyOp = dyn_cast<CopyOpInterface>(userOp);
-        dyn_cast_if_present<LogicalObjFifoOpInterface>(
-            copyOp.getSource().getDefiningOp()) == op) {
-      copyLikOps.push_back(copyOp);
+    if (auto copyOp = dyn_cast<CopyOpInterface>(userOp)) {
+      if (dyn_cast_if_present<LogicalObjFifoOpInterface>(
+              copyOp.getSource().getDefiningOp()) == op)
+        copyLikOps.push_back(copyOp);
```
jtuyls, line 23: Why the change? This is exactly the same logically?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -121,6 +121,45 @@ FailureOr<int64_t> getSplitStride(ArrayRef<AMDAIE::DmaCpyNdOp> dmaOps,
...
+///      DMA(%a, %lhs)
+///      DMA(%b, %lhs)
+///      DMA(%c, %lhs)
+///      DMA(%b, %lhs)
+///      DMA(%c, %lhs)
+///
+///    In the above snippet although we have 5 DMA ops for L2<->L1, only 3 of
+///    them are unique. Hence we'd split %lhs into 3 unique splits, instead
+///    of 5.
+static FailureOr<int64_t> fetchTotalUniqueL2L1(
```
jtuyls, line 142: Update the name, we shouldn't mention `L2L1`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -121,6 +121,45 @@ FailureOr<int64_t> getSplitStride(ArrayRef<AMDAIE::DmaCpyNdOp> dmaOps,
...
+///      %c = LOF_on_L1_2
+///      DMA(%a, %lhs)
+///      DMA(%b, %lhs)
+///      DMA(%c, %lhs)
+///      DMA(%b, %lhs)
+///      DMA(%c, %lhs)
+///
+///    In the above snippet although we have 5 DMA ops for L2<->L1, only 3 of
+///    them are unique. Hence we'd split %lhs into 3 unique splits, instead
+///    of 5.
```
jtuyls, line 141: Update the description, don't mention L2, L1 etc

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -121,6 +121,45 @@ FailureOr<int64_t> getSplitStride(ArrayRef<AMDAIE::DmaCpyNdOp> dmaOps,
...
+static FailureOr<int64_t> fetchTotalUniqueL2L1(
+    SmallVector<CopyOpInterface> copyLikeOps, bool fetchTarget) {
```
jtuyls, line 143: You could use `template <CopyOpOperateOn OperateOn>` instead of `bool fetchTarget` to make this a compile-time check.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIELogicalObjFifoOpInterface.cpp
jtuyls: Right, the same would be:

```
if (auto copyOp = dyn_cast<CopyOpInterface>(userOp);
        copyOp  && dyn_cast_if_present<LogicalObjFifoOpInterface>(
            copyOp.getSource().getDefiningOp()) == op) {
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -274,17 +330,31 @@ LogicalResult collectSplittingDims(
       // Calculate the new target stride to be used for splitting the DMA.
       int64_t newTargetStride =
           splitStride != 1 ? splitDimSize / splitStride : 1;
+      FailureOr<int64_t> maybeUniqueL2L1 =
```
jtuyls, line 333:
```suggestion
      FailureOr<int64_t> maybeNumUniqueConsumers =
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -218,17 +260,31 @@ LogicalResult collectSplittingDims(
       // Calculate the new source stride to be used for splitting the DMA.
       int64_t newSourceStride =
           splitStride != 1 ? splitDimSize / splitStride : 1;
+      FailureOr<int64_t> maybeUniqueL2L1 =
```
jtuyls, line 263:
```suggestion
      FailureOr<int64_t> maybeNumUniqueProducers =
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -121,6 +121,45 @@ FailureOr<int64_t> getSplitStride(ArrayRef<AMDAIE::DmaCpyNdOp> dmaOps,
...
+///      DMA(%b, %lhs)
+///      DMA(%c, %lhs)
+///      DMA(%b, %lhs)
+///      DMA(%c, %lhs)
+///
+///    In the above snippet although we have 5 DMA ops for L2<->L1, only 3 of
+///    them are unique. Hence we'd split %lhs into 3 unique splits, instead
+///    of 5.
+template <CopyOpOperateOn OperateOn>
+static FailureOr<int64_t> fetchTotalUniqueLogicalObjFifos(
```
jtuyls, line 142:
```suggestion
static FailureOr<int64_t> fetchTotalUniqueLogicalObjFifoUsers(
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos.mlir
yzhang93: Thanks! This makes more sense to me.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos.mlir
```diff
@@ -436,38 +461,127 @@ module attributes {hal.executable.target = #executable_target_amdaie_pdi_fb} {
...
+    %alloc_4 = memref.alloc() : memref<8x8x32x64xi32, 1 : i32>
+    %alloc_5 = memref.alloc() : memref<512x512xi32>
+    %alloc_6 = memref.alloc() : memref<512x4096xi32>
+    %alloc_7 = memref.alloc() : memref<512x4096xi32>
+    %tile_0_1 = amdaie.tile(%c0, %c1)
+    %tile_1_1 = amdaie.tile(%c1, %c1)
+    %tile_2_1 = amdaie.tile(%c2, %c1)
+    %tile_0_0 = amdaie.tile(%c0, %c0)
+    %tile_1_0 = amdaie.tile(%c1, %c0)
+    %tile_2_0 = amdaie.tile(%c2, %c0)
```
yzhang93, line 547: I don't think you need to add these tile ops. Refer to the tests above.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos.mlir
```diff
@@ -436,38 +461,127 @@ module attributes {hal.executable.target = #executable_target_amdaie_pdi_fb} {
...
+// CHECK:                 amdaie.dma_cpy_nd(%{{.*}}[0, 0, 0, 0, 0, 0] [1, 1, 8, 8, 8, 4] [2048, 2048, 32, 4, 256, 1], %[[LOF_RHS_L2_1]][%{{.*}}, 0, 0, 0] [1, 1, 64, 32] [16384, 2048, 32, 1]) :
+// CHECK:                 amdaie.dma_cpy_nd(%{{.*}}[0, 0, 0, 0, 0, 0] [1, 1, 8, 4, 8, 8] [2048, 2048, 32, 8, 256, 1], %[[LOF_LHS_L2]][0, 0, 0, 0] [1, 1, 32, 64] [16384, 2048, 64, 1]) :
+// CHECK:                 amdaie.core
+// CHECK:                 amdaie.core
+// CHECK:                 amdaie.dma_cpy_nd(%[[LOF_OUT_L2:.*]][0, 0, 0, 0] [1, 1, 32, 32] [8192, 1024, 32, 1], %{{.*}}[0, 0, 0, 0, 0, 0] [1, 1, 8, 4, 8, 4] [1024, 1024, 16, 4, 128, 1]) :
+// CHECK:             }
+// CHECK:             amdaie.dma_cpy_nd(%{{.*}}[0, 0] [256, 512] [4096, 1], %[[LOF_OUT_L2]][0, 0, 0, 0] [8, 32, 16, 32] [1024, 32, 8192, 1]) :
+// CHECK:          }
+#executable_target_amdaie_pdi_fb = #hal.executable.target<"amd-aie", "amdaie-pdi-fb", {num_cols = 8 : i32, num_rows = 4 : i32, target_device = "npu4", ukernels = "none"}>
+#translation = #iree_codegen.translation_info<pipeline = Custom>
```
yzhang93, line 527: I don't understand why this test works as in your description `num_cols=2 and num_rows=1`, while here it's `num_cols=8 and num_rows=4`.

yzhang93 (commented):
This revision is much more concise and cleaner. I have more comments on the test.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -218,17 +260,31 @@ LogicalResult collectSplittingDims(
       // Calculate the new source stride to be used for splitting the DMA.
       int64_t newSourceStride =
           splitStride != 1 ? splitDimSize / splitStride : 1;
+      FailureOr<int64_t> maybeNumUniqueConsumers =
+          fetchTotalUniqueLogicalObjFifoUsers<CopyOpOperateOn::Target>(
+              objFifo.getCopyLikeConsumers());
+      if (failed(maybeNumUniqueConsumers)) {
+        objFifo.emitOpError()
+            << "could not retrieve total unique L2<->L1 pairs";
```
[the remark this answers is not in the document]
jtuyls: Also, remove `L2<->L1 pairs` from the error message.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos.mlir
Abhishek-Varma: That's the intention : how much to split is not only dependent on the number of available columns but also the [unique L2<->L1 pairs](https://github.com/nod-ai/iree-amd-aie/blob/8c3f762348c9e3d756a19862d1fa412e51ad6063/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp#L127) -> so we use [gcd between the two](https://github.com/nod-ai/iree-amd-aie/blob/8c3f762348c9e3d756a19862d1fa412e51ad6063/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp#L270) (and in some cases we might need to take [gcd between source/target sizes](https://github.com/nod-ai/iree-amd-aie/blob/8c3f762348c9e3d756a19862d1fa412e51ad6063/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp#L274))

"`to keep the test case concise it demonstrates a similar splitting strategy for 1 row and 2 columns.`" means the actual compute is taking place on 1 row and 2 columns. I could change `num_cols=2` and `num_rows=1` but that'd defeat the purpose.

I've updated the comments.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -138,6 +177,9 @@ FailureOr<int64_t> getSplitStride(ArrayRef<AMDAIE::DmaCpyNdOp> dmaOps,
 /// that has product size larger than the other side's product size after
 /// splitting because that's the number of elements that should be
 /// produced/consumed on the respective sides before splitting.
+/// Towards the end fetch the count of unique L2<->L1 for the objectFifo which
```
jtuyls, line 180: Remove mentions related to L1/L2 from the comments, naming etc. This is not about L1/L2 connections but about producer/consumer connections.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos.mlir
```diff
@@ -436,38 +461,110 @@ module attributes {hal.executable.target = #executable_target_amdaie_pdi_fb} {
...
     memref.dealloc %alloc_0 : memref<4x1x32x32xi32, 1 : i32>
     memref.dealloc %alloc : memref<1x1x4x8x4x8xi32, 2 : i32>
     return
   }
 }
+
+// -----
+
+// A concise test for LHS/RHS/OUT of a Matmul on 4x8 for pack-peel-4-level-tiling. Although
+// we have 8 columns, L2 LHS buffers needs to be split into only 4, L2 RHS into 8 and L2 OUT into 8,
```
jtuyls, line 502: I don't think this is accurate? The LHS buffer doesn't seem to be split and the RHS seems to be split into 2?

Abhishek-Varma: The comment highlighted above is about the expected splits for an actual e2e IR - the previous expectation is that each L2 will be split 8 times, but that's not the case anymore.

Since I'm using 1 row and 2 columns for the actual compute : LHS is being split into 1 (therefore no split) and RHS into 2. So that is correct.

I've updated the comment to state the algorithm explicitly. Let me know if I can better the paraphrase wording.

jtuyls: This is confusing because people reading this will think it's about the test, not about some e2e example, and get confused as there is no splitting in 4/8 in this test. So, don't describe the use case, describe the actual test, with splitting into 2, and remove any mentions of the use case.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIESplitLogicalObjFifos.cpp
```diff
@@ -138,6 +178,9 @@ FailureOr<int64_t> getSplitStride(ArrayRef<AMDAIE::DmaCpyNdOp> dmaOps,
 /// that has product size larger than the other side's product size after
 /// splitting because that's the number of elements that should be
 /// produced/consumed on the respective sides before splitting.
+/// Towards the end fetch the count of unique producer (or consumers) for the
```
jtuyls, line 181:
```suggestion
/// Towards the end fetch the number of unique producers (or consumers) for the
```

jtuyls (approved):
LGTM, just one nit and also update the PR title to producers/consumers instead of L1/L2.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos.mlir
```diff
@@ -436,38 +461,108 @@ module attributes {hal.executable.target = #executable_target_amdaie_pdi_fb} {
...
     return
   }
 }
+
+// -----
+
+// Although we have 8 columns, L2 LHS buffers needs to be split into only 1, L2 RHS into 2 and
+// L2 OUT into 1.
+// This is because we decide the split factor for the L2 ObjectFifo depending on :-
+//     GCD(unique producer/consumer for the respective ObjectFifos being split, number of columns)
```
yzhang93, line 504: The description is much better. But I think if the split factor is 1, it's not splitting. Besides, you could emphasize the purpose of test. Something like:

This test demonstrates the case when the factor is not simply decided by the number of columns but the number of unique producers/consumers. In the example, although we are using 8 AIE columns, L2 LHS and output buffers are not split because there's only one producer/consumer, while L2 RHS buffer is split into 2 because there are 2 producers/consumers.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/split_logicalobjfifos.mlir
```diff
@@ -436,38 +461,108 @@ module attributes {hal.executable.target = #executable_target_amdaie_pdi_fb} {
...
+// CHECK:                 amdaie.dma_cpy_nd(%{{.*}}[0, 0, 0, 0, 0, 0] [1, 1, 8, 8, 8, 4] [2048, 2048, 32, 4, 256, 1], %[[LOF_RHS_L2_0]][%{{.*}}, 0, 0, 0] [1, 1, 64, 32] [16384, 2048, 32, 1]) :
+// CHECK:                 amdaie.dma_cpy_nd(%{{.*}}[0, 0, 0, 0, 0, 0] [1, 1, 8, 8, 8, 4] [2048, 2048, 32, 4, 256, 1], %[[LOF_RHS_L2_1]][%{{.*}}, 0, 0, 0] [1, 1, 64, 32] [16384, 2048, 32, 1]) :
+// CHECK:                 amdaie.dma_cpy_nd(%{{.*}}[0, 0, 0, 0, 0, 0] [1, 1, 8, 4, 8, 8] [2048, 2048, 32, 8, 256, 1], %[[LOF_LHS_L2]][0, 0, 0, 0] [1, 1, 32, 64] [16384, 2048, 64, 1]) :
+// CHECK:                 amdaie.dma_cpy_nd(%[[LOF_OUT_L2:.*]][0, 0, 0, 0] [1, 1, 32, 32] [8192, 1024, 32, 1], %{{.*}}[0, 0, 0, 0, 0, 0] [1, 1, 8, 4, 8, 4] [1024, 1024, 16, 4, 128, 1]) :
+// CHECK:             }
+// CHECK:             amdaie.dma_cpy_nd(%{{.*}}[0, 0] [256, 512] [4096, 1], %[[LOF_OUT_L2]][0, 0, 0, 0] [8, 32, 16, 32] [1024, 32, 8192, 1]) :
+// CHECK:          }
+#executable_target_amdaie_pdi_fb = #hal.executable.target<"amd-aie", "amdaie-pdi-fb", {num_cols = 8 : i32, num_rows = 4 : i32, target_device = "npu4", ukernels = "none"}>
+#translation = #iree_codegen.translation_info<pipeline = Custom>
```
yzhang93, line 523: Nit: #translation can be removed.

yzhang93 (approved):
Thanks for the changes! I have some final comments.

PR #1062, opened 2025-01-28
[LinalgFunctionOutlining] Add none, all and balanced outlining strategies

jtuyls:
To avoid overheads of non-outlined ops when using `outlining=true` with `--iree-amdaie-replace-outlined-functions-with-empty`, I think there should be an option to outline all ops if needed, so I created three outlining strategies:
- None: no outlining.
- All: outline all ops.
- Balanced (default): this is the same as what happens in main currently, outlining some ops and not others for a tradeoff in performance and program size.

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -800,7 +800,9 @@ class MatmulTruncf(BaseMatmul):
         """
         Currently without function outlining, we run out of program memory.
         """
-        self.add_aie_compilation_flags(["--iree-amdaie-enable-function-outlining"])
+        self.add_aie_compilation_flags(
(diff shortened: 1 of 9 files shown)


jtuyls:
> I disabled outlining of elementwise in #954 because (unintuitively) outlining elementwise was making us go OOM (PM). But I think enough has changed since then for this to possibly not be an issue anymore.

Yeah, I saw this as well, so  now I created outlining strategies to accommodate different use cases (see updated description). Could you have a look?

newling (approved):
Looks Good (TM)

newling:
I can't see why CI failed. If it's a PM issue, I should probably get back to finishing https://github.com/nod-ai/iree-amd-aie/pull/1006 off so that -O2 is viable again.

PR #1063, opened 2025-01-28
[CtrlPkt] Generate `connection` operations in the overlay

Yu-Zhewen:
Instead of generating `flow` operations, we now directly insert `connection` operations into the overlay. The motivation behind this change is that these `connection` operations will be useful for generating control packet DMAs in subsequent steps.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIETypes.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIETypes.cpp
@@ -32,11 +32,6 @@ void AMDAIEDialect::initializeAMDAIETypes() {
 LogicalResult LogicalObjectFifoType::verify(
     llvm::function_ref<mlir::InFlightDiagnostic()> emitError,
     mlir::MemRefType elementType, unsigned depth) {
-  if (llvm::any_of(elementType.getShape(), [](auto dimSize) {
-        return ShapedType::isDynamic(dimSize);
-      })) {
-    return emitError() << "should encapsulate static memref";
-  }
   if (depth < 1 || depth > 4) return emitError() << "depth should be in [1, 4]";
   return success();
 }
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
@@ -884,12 +884,10 @@ void addAMDAIEObjectFifoLoweringPasses(
 
   passManager.addPass(createAMDAIEObjFifoBufferizationPass());
   passManager.addPass(createAMDAIETemporaryAllocBufferizationPass());
-  passManager.addPass(createAMDAIEConnectionToFlowPass());
 
   passManager.addPass(createAMDAIEGenerateControlOverlayPass());
-  passManager.addPass(createCSEPass());
-  passManager.addPass(createCanonicalizerPass());
 
+  passManager.addPass(createAMDAIEConnectionToFlowPass());
   passManager.addPass(createAMDAIEAssignPacketIdsPass());
 
   passManager.addPass(createAMDAIENpuDmaToHalfDmaCpyNdPass());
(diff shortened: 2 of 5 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIETypes.cpp
```diff
@@ -32,11 +32,6 @@ void AMDAIEDialect::initializeAMDAIETypes() {
 LogicalResult LogicalObjectFifoType::verify(
     llvm::function_ref<mlir::InFlightDiagnostic()> emitError,
     mlir::MemRefType elementType, unsigned depth) {
-  if (llvm::any_of(elementType.getShape(), [](auto dimSize) {
```
Yu-Zhewen, line 35: The constraint has been removed because the control packet may require a dynamic shape, depending on the reconfiguration content. Additionally, if a device is reconfigured multiple times, the size of the control packet may vary with each reconfiguration.

reference: https://github.com/Xilinx/mlir-aie/pull/1776

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -884,12 +884,10 @@ void addAMDAIEObjectFifoLoweringPasses(
 
   passManager.addPass(createAMDAIEObjFifoBufferizationPass());
   passManager.addPass(createAMDAIETemporaryAllocBufferizationPass());
-  passManager.addPass(createAMDAIEConnectionToFlowPass());
 
   passManager.addPass(createAMDAIEGenerateControlOverlayPass());
-  passManager.addPass(createCSEPass());
-  passManager.addPass(createCanonicalizerPass());
```
jtuyls, line 891: Is there a specific reason for this removal?

Yu-Zhewen: They were initially introduced in #1012 solely to tidy up the generated overlay. However, with this PR, both the generated `connection` and `logicalobjectfifo.placeholder` operations (which are marked as pure) are required for the later DMA generation. Additionally, I don’t believe the CSE and Canonicalize here affect things like control code size.

jtuyls: > However, with this PR, both the generated connection and logicalobjectfifo.placeholder operations (which are marked as pure) are required for the later DMA generation. 

Right, but the placeholders shouldn't be removed because they are used by the connections? Just making sure, if you keep the cse pass, it would still work, or would it break?

Yu-Zhewen: Yeah, in this case, the placeholders are kept but the connections are still eliminated since they don't have a user yet.

Yu-Zhewen: Without canonicalize and cse, the generated code looks like this
```ll
      %channel = amdaie.channel(%tile_0_0, 0, port_type = DMA, direction = MM2S)
      %channel_0 = amdaie.channel(%tile_0_0, 0, port_type = CTRL, direction = S2MM)
      %0 = amdaie.logicalobjectfifo.placeholder{%tile_0_0} : !amdaie.logicalobjectfifo<memref<?xi32>>
      %1 = amdaie.logicalobjectfifo.placeholder{%tile_0_0} : !amdaie.logicalobjectfifo<memref<?xi32>>
      %2 = amdaie.connection(%1 {%channel_0}, %0 {%channel}) {connection_type = #amdaie<connection_type Packet>} : (!amdaie.logicalobjectfifo<memref<?xi32>>, !amdaie.logicalobjectfifo<memref<?xi32>>)

```

jtuyls: I see, we will have to fix that at some point to make it stable, but that needs some further thought with the connection ops being pure, so not needed in this PR for me.

jtuyls (approved):
LGTM

PR #1065, opened 2025-01-29
[ELF] Add PM size to debug log

jtuyls:


Abhishek-Varma (approved):
LGTM! Nice!

Just one nit (not blocking on it) - if this PM reading/writing to file is kept it its own standalone utility and invoked right before `addElfToTile`.

newling:
LGTM. If you think the python equivalent is redundant, i.e.  https://github.com/nod-ai/iree-amd-aie/blob/cc25a25eca1113bfbd69d77c69470e05d033329c/build_tools/ci/cpu_comparison/run.py#L897 we should remove it

PR #1067, opened 2025-01-29
[CI] Add larger shape e2e bfp16 Ukernel test

Abhishek-Varma:
-- This commit adds larger shape e2e bfp16 Ukernel test to CI for
   both `pack-peel` as well as `pack-peel-4-level-tiling` targeting
   4x8 array on Strix.
-- It wasn't working earlier (3 months ago) but now it does - hence
    enabling.

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -1751,7 +1751,6 @@ class Tests:
                 )
             )
 
-        # TODO: Failure is expected for the 128x128 case we don't yet understand why.
         self.register(
             Matmul(
                 64,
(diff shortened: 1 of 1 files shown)


newling:
Can you include a motivation in the description, and how is this related to the comment https://github.com/nod-ai/iree-amd-aie/blob/cc25a25eca1113bfbd69d77c69470e05d033329c/build_tools/ci/cpu_comparison/run.py#L1738 ?

newling (approved):

yzhang93:
The ukernel is for bfloat16 or block floating point?

build_tools/ci/cpu_comparison/run.py
```diff
@@ -1816,6 +1815,42 @@ def __init__(self):
...
+                tile_pipeline="pack-peel-4-level-tiling",
+                run_on_target=["npu4"],
+                aie_compilation_flags=[
+                    "--iree-amdaie-num-rows=4",
+                    "--iree-amdaie-num-cols=8",
+                ],
+                use_chess=True,
+            )
+        )
+        self.register(
```
jtuyls, line 1836: I think we should only add one of these tests as chess makes them really slow (both  these tests together add 15min to the strix CI time, from 7min on main to 22min now). I know there are issues with these, but should try to enable peano for the non-ukernel part to speed up the tests.

jtuyls:
> The ukernel is for bfloat16 or block floating point? Please clarify the type in the title.

It's bfloat16 with bfp16 under the hood.

jtuyls (approved):
LGTM

PR #1069, opened 2025-01-30
[CI] Add initial Strix matmul performance tests

jtuyls:


.github/workflows/ci-linux.yml
```diff
@@ -249,7 +249,7 @@ jobs:
           git config user.name "github-actions"
           git config user.email "github-actions@github.com"
           git add results_history.json results_history.html
```
Yu-Zhewen, line 251: Maybe add the suffix `_npu1` to these json/html/log files? I will also manually rename the existing files in the `gh-pages` branch before merging this PR to ensure we retain the collected history.

jtuyls: Yeah, good idea

Yu-Zhewen (approved):

PR #1071, opened 2025-01-30
Initial support for using peano with AIE2P (strix). 

newling:
The majority of this PR is refactoring to make it possible to support 2+ target ISAs. 

- aievec.matmul now verifies its operand shapes based on the target device in the module. 
- I've removed a bunch of intrinsic matmul shapes for phoenix that we're not using, to simplify the code 
- The XLLVM dialect ops now include the target device in their names. i.e. they name includes either `AIE2` or `AIE2P` now. 
- Only matmul can lower to XLLVM from aievec for AIE2P, all other aievec ops (UPS, etc) have an assert on them that the device is AIE2. 
- A few XLLVM ops are removed: we don't use broadcast or a few others, so I've trimmed the set of ops we support down. 
- This PR adds utils `isAie2` and `isAie2P` to AMDAIEUtils.h 

I have confirmed that a linalg matmul can compile all the way through peano for AIE2P, but only with -O0. Next step after this is to fix alignment issues in iree-amd-aie to get this to work for -On n>0: https://github.com/Xilinx/llvm-aie/issues/315


compiler/plugins/target/AMD-AIE/aievec/AIEVecOps.td
```diff
@@ -130,39 +129,43 @@ def AIEVec_SRSOp:
...
-         `vector<4x8xi8>`   | `vector<8x8xi8>`   | `vector<4x8xi32>`
-         `vector<4x4xi16>`  | `vector<4x8xi8>`   | `vector<4x8xi32>`
-         `vector<4x2xi16>`  | `vector<2x8xi16>`  | `vector<4x8xi32>`
-         `vector<2x8xi16>`  | `vector<8x8xi8>`   | `vector<2x8xi64>`
-         `vector<4x8xi16>`  | `vector<8x4xi8>`   | `vector<4x4xi64>`
-         `vector<2x4xi16>`  | `vector<4x8xi16>`  | `vector<2x8xi64>`
-         `vector<4x4xi16>`  | `vector<4x4xi16>`  | `vector<4x4xi64>`
-         `vector<4x2xi32>`  | `vector<2x4xi16>`  | `vector<4x4xi64>`
-         `vector<4x8xbf16>` | `vector<8x4xbf16>` | `vector<4x4xf32>`
+    Currently, this intrinsic supports the following type combinations
```
jtuyls, line 142: > A few XLLVM ops are removed: we don't use broadcast or a few others, so I've trimmed the set of ops we support down.

Why though? Can't we just keep supporting them?

newling: Just to check: are you referring to the reduced set of AIE2 matmuls supported, the removal of ops we don't use (like broadcast) or both? 

The code and ops I've removed is not tested (here or in mlir-aie afaict), not reachable through the compiler currently, and is kind of just a burden to maintain. I don't see any harm in removing it, we can add it back with tests when we need to.

compiler/plugins/target/AMD-AIE/aievec/AIEVecToLLVM.cpp
```diff
@@ -125,40 +142,167 @@ static VectorType getFlattenedVectorType(VectorType vecTy) {
...
-         ((unsigned)bmode << 3) | ((unsigned)variant << 5) |
-         (((unsigned)sgn_x << 9) | ((unsigned)sgn_y << 8)) |
-         ((unsigned)zero_acc << 0);
-}
+
+class DataPathConfiguration {
+  // Dynamic zero accumulation r[0]
+  // 0 – Use default first accumulator input to the postadder.
+  // 1 – Replace default first accumulator with zeros.
+  bool dynamcZeroAccumulation = 0;
```
jtuyls, line 174:
```suggestion
  bool dynamicZeroAccumulation = 0;
```

runtime/src/iree-amd-aie/aie_runtime/AMDAIEEnums.h
```diff
@@ -11,9 +11,27 @@
 #define IREE_AIE_RUNTIME_AMDAIE_ENUMS_H_
 
 #include "mlir/IR/BuiltinAttributes.h"
-
 // clang-format off
 #include "iree-amd-aie/aie_runtime/AMDAIEEnums.h.inc"
 // clang-format on
 
+// TODO(someone_who_knows) explain the connection between npu numbering and
+// aie numbering, and why npu jumps from '1' to '4'.
```
jtuyls, line 19: Because npu2/3 were already used in xdna-driver for some custom devices, one of them might be a 4x4 variation of Strix and not sure about the other one. Npu2 seems to have been the prototype for Strix (maybe the 4x4 variation)?

jtuyls (approved):
LGTM

runtime/src/iree-amd-aie/aie_runtime/AMDAIEEnums.h
newling: Thanks, incorporated

PR #1072, opened 2025-01-31
[CI] Add i8->i32 matmul pack-peel-4-level-tiling performance tests

jtuyls:
Adds new Strix performance tests with 4 levels of tiling, which should result in better performance. Note that this couldn't be added before as this needed the fix from https://github.com/nod-ai/iree-amd-aie/pull/1060.


Abhishek-Varma (approved):
LGTM.

Yu-Zhewen (approved):
LGTM, thanks!

PR #1073, opened 2025-01-31
[FoldLinearDims] Support folding with static non-zero offsets

jtuyls:
Add support for folding linear dimensions with non-zero, but static, offsets. For example:

```
Offsets: [8, 0], Size: [16, 8],  Strides: [8, 1]
```

can now be transformed into:

```
Offsets: [64], Size: [128],  Strides: [1]
```

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
@@ -334,16 +334,20 @@ LogicalResult foldLinearDims(
 
   for (int i = offsets.size() - 2; i >= 0; i--) {
     // Conditions for folding a dim.
-    // 1. Offsets[i] == 0.This is required because we are dropping the offset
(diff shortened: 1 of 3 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
```diff
@@ -334,28 +334,50 @@ LogicalResult foldLinearDims(
 
   for (int i = offsets.size() - 2; i >= 0; i--) {
```
newling, line 335:
```suggestion
  for (int i = static_cast<int>(offsets.size()) - 2; i >= 0; i--) {
```

newling: Fly by comment: I think the code is safe as is, but not obvious. `int i = (size_t)1 - 2` will produce compiler warnings (but `i` will be `-1`, so correct for the loop).

newling (approved):
Nice one.

PR #1077, opened 2025-02-04
Numerical test for strix compiled with peano  

newling:
The main change here is an update to the pass that unsets the load and store alignments. Now it unsets _all_ llvm.loads and llvm.stores (see comments in pass). Without this, peano tries to scalarize an llvm.load and gets into an infinite/quadratic loop. This PR adds an end-to-end numerical test of peano on strix. 

One unrelated change in AMDAIELowerToAIE.cpp : while experimenting with adding alignment/noalias attributes to the function signature in function outlining, noticed that the attributes were being dropped in this pass.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELoadStoreAlignmentReset.cpp
```diff
@@ -0,0 +1,104 @@
...
+/// iteration 2: load bytes 4-12.
+/// iteration 3: load bytes 8-16.
+///
+/// in this case using the width of the vector (8 bytes) as the alignment would
+/// be incorrect, as the loads are only 4-byte aligned. Future work: check if
+/// LLVMIR lowering correctly handles this case, if not implement an analysis.
+///
+/// See also https://jira.xilinx.com/projects/AIECC/issues/AIECC-589
+
+class AMDAIELoadStoreAlignmentReset
```
jtuyls, line 80: Could you add a lit test for this?

jtuyls (approved):
LGTM, just one nit below.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELoadStoreAlignmentReset.cpp
newling: Yes good idea

PR #1083, opened 2025-02-06
[Router] Move shim mux port mapping into `DeviceModel`

Yu-Zhewen:


compiler/plugins/target/AMD-AIE/aie/AMDAIECreatePathFindFlows.cpp
```diff
@@ -358,72 +372,57 @@ LogicalResult runOnPacketFlow(
     }
   }
 
-  // Add support for shimDMA
-  // From shimDMA to BLI: 1) shimDMA 0 --> North 3
-  //                      2) shimDMA 1 --> North 7
-  // From BLI to shimDMA: 1) North   2 --> shimDMA 0
-  //                      2) North   3 --> shimDMA 1
+  // Add special shim mux connections between DMA/NOC streams and BLI.
```
jtuyls, line 375: I know it was introduced earlier, but what's BLI?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -422,6 +422,45 @@ struct AMDAIEDeviceModel {
...
+  /// port.
+  llvm::SmallDenseMap<std::pair<StrmSwPortType, uint8_t>,
+                      std::pair<StrmSwPortType, uint8_t>>
+      s2mmDmaNocToSpecialShimPortMap = {
+          {{StrmSwPortType::DMA, 0}, {StrmSwPortType::NORTH, 2}},
+          {{StrmSwPortType::DMA, 1}, {StrmSwPortType::NORTH, 3}},
+          {{StrmSwPortType::NOC, 0}, {StrmSwPortType::NORTH, 2}},
+          {{StrmSwPortType::NOC, 1}, {StrmSwPortType::NORTH, 3}},
+          {{StrmSwPortType::NOC, 2}, {StrmSwPortType::NORTH, 4}},
+          {{StrmSwPortType::NOC, 3}, {StrmSwPortType::NORTH, 5}}};
```
jtuyls, line 447: These could be `static const`

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -422,6 +422,45 @@ struct AMDAIEDeviceModel {
...
+
+  /// Retrieves the speicail shim mux port that connects a given MM2S or S2MM
```
jtuyls, line 449:
```suggestion
  /// Retrieves the special shim mux port that connects a given MM2S or S2MM
```

compiler/plugins/target/AMD-AIE/aie/AMDAIECreatePathFindFlows.cpp
Yu-Zhewen: That doesn't seem relevant to us. Should I remove it?

According to the architecture spec: "_The BLI block (Boundary Logic Interface) connects COE blocks at the boundary to the FPGA fabric."_

jtuyls: Yeah, I don't think that's relevant to us.

jtuyls: Maybe good to remove indeed to avoid confusion.

jtuyls (approved):
LGTM

PR #1084, opened 2025-02-06
 [Pack-peel-4-level-tiling] Enable Matmul+Trunci : i8->i8 on Phoenix|Strix + Ukernel|Vectorization

Abhishek-Varma:
Add support for handling Matmul + Trunci : i8->i8 on Phoenix/Strix.

On Phoenix it works via both vectorization as well as ukernel path.
On Strix it currently works only via ukernel path.

Therefore, for now, three e2e CI tests have been added.

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -645,12 +645,6 @@ class MatmulTruncf(BaseMatmul):
         input_args = generate_inputs(
             filename, self.get_dir(config), 1, {1: self.lhs, 2: self.rhs}
         )
-        """
-        Currently without function outlining, we run out of program memory.
-        """
-        self.add_aie_compilation_flags(
-            ["--iree-amdaie-enable-function-outlining=balanced"]
-        )
         aie_vs_baseline(
             config=config,
             aie_compilation_flags=self.aie_compilation_flags,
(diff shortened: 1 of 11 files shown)


build_tools/ci/cpu_comparison/run.py
```diff
@@ -669,6 +671,76 @@ def _execute(self, config):
...
+        filename = self.get_filename(config)
+        input_args = generate_inputs(
+            filename, self.get_dir(config), 1, {1: self.lhs, 2: self.rhs}
+        )
+        """
+        Currently without function outlining, we run out of program memory.
+        """
+        self.add_aie_compilation_flags(
+            ["--iree-amdaie-enable-function-outlining=balanced"]
+        )
```
jtuyls, line 723:
```suggestion
```
This is the default, so flag should not be needed.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -669,6 +671,76 @@ def _execute(self, config):
         return True
 
 
+class MatmulTrunci(BaseMatmul):
+    """
+    A test of the form matmul(A,B) + trunci(C) where A:MxK, B:KxM and C:MxM
```
jtuyls, line 676: Why `MxKxM` and not `MxKxN`? The latter should work as well I would think, so it would be better to provide general matmul support.

jtuyls (approved):
LGTM

PR #1086, opened 2025-02-06
[CI] Merge Matmul class with its Benchmark class

yzhang93:
Currently, there are Matmul/MatmulTranspseB/MatmulTransposeA classes and their Benchmark classes separately. A lot of codes are duplicated between them. This also brings extra efforts and more codes if we introduce other classes, e.g., mmt4d.

This PR removes the Benchmark classes while adding an argument "run_benchmark" to the existing Matmul/MatmulTranspseB/MatmulTransposeA classes.

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -203,6 +203,7 @@ class BaseMatmul(BaseTest):
         use_chess=False,
         function_name="matmul",
         n_kernel_runs=1,
+        run_benchmark=False,
     ):
         """
         Base class for all variants of dispatches with a matmul, currently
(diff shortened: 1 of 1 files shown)


build_tools/ci/cpu_comparison/run.py
```diff
@@ -2183,13 +2052,13 @@ def __init__(self):
 
             if (transpose_a, transpose_b) == (False, False):
                 NumericTestClass = Matmul
```
Yu-Zhewen, line 2054: Maybe just use `TestClass` here, as `NumericTestClass` and `BenchmarkTestClass` are pointing to the same thing.

Yu-Zhewen (approved):

PR #1089, opened 2025-02-07
[e2e CI] Refactor e2e CI tests to make addition of new tests easier

Abhishek-Varma:
-- Currently there are lot of repeated codes which makes :-
   1. Maintaining the codebase
   2. Addition of new tests,
   quite cumbersome.
   
-- Since most tests are bound to go through a certain lowering pipeline,
   decide whether or not to use ukernel, chess, have additional compilation
   flags - it's better to separate those as RunningParams instances.

-- This way they're handled unanimously at the base class instead of each
   subclass taking the same onus.

-- This can further be refined but for now going ahead with the current
   design.

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -177,20 +177,14 @@ class BaseTest(ABC):
 class ConvolutionFromTemplate(BaseTest):
     def __init__(
         self,
-        params,
+        generator,
         running_params: RunningParams = RunningParams(),
     ):
         super().__init__(
-            name=f"{conv_type}_{N}_{IW}_{in_type}_{out_type}",
+            name=f"{generator.params['conv_type']}_{generator.params['N']}_{generator.params['IW']}_{generator.params['input_element_type']}_{generator.params['output_element_type']}",
             running_params=running_params,
         )
-        self.generator = ConvolutionMlirGenerator(**params)
-        params = self.generator.params
-        conv_type = params["conv_type"]
-        N = params["N"]
-        IW = params["IW"]
-        in_type = params["input_element_type"]
-        out_type = params["output_element_type"]
+        self.generator = generator
         # TODO(newling) Use all parameters in name, to avoid name collision.
         self.labels += ["Convolution"]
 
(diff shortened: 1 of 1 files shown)


build_tools/ci/cpu_comparison/run.py
```diff
@@ -38,6 +38,30 @@ def run_conv_test(config, aie_compilation_flags, filename, n_repeats):
     return True
 
 
+class RunningParams(ABC):
```
jtuyls, line 41:
```suggestion
class RunParams(ABC):
```
```suggestion
class LoweringParams(ABC):
```
```suggestion
class TestParams(ABC):
```
Not a native English speaker, but "RunningParams" sounds a bit weird to me, so some suggestions for alternatives.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -61,28 +85,50 @@ class BaseTest(ABC):
 
     def __init__(
         self,
-        run_on_target=["npu1_4col"],
-        aie_compilation_flags=None,
-        use_chess=False,
+        name="",
+        running_params: RunningParams = RunningParams(),
```
jtuyls, line 89: Here, and everywhere else below, use `None` as the default value instead to avoid potential mutability issues: https://www.geeksforgeeks.org/default-arguments-in-python/. Then inside the function do: 
```
running_params = running_params if running_params is not None else RunningParams()
```

jtuyls (changes requested):

build_tools/ci/cpu_comparison/run.py
Abhishek-Varma: Makes sense. I chose "TestParams" and have updated it.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -129,17 +176,17 @@ def get_filename(self, config):
...
-        in_type = params["input_element_type"]
-        out_type = params["output_element_type"]
+    def __init__(
+        self,
+        generator,
+        test_params=None,
+    ):
+        super().__init__(
+            name=f"{generator.params['conv_type']}_{generator.params['N']}_{generator.params['IW']}_{generator.params['input_element_type']}_{generator.params['output_element_type']}",
+            test_params=test_params if test_params is not None else TestParams(),
```
jtuyls, line 186: No need for this, you can just forward `test_params`. Same for all the other subclasses down below.
```suggestion
            test_params=test_params,
```

jtuyls (approved):
LGTM, except for nit above.

newling:
> -- Since most tests are bound to go through a certain lowering pipeline,

This was not my intention, for example what pipeline should `test_files/three_matmuls.mlir` have -- this test harness was designed to more general than single dispatch testing. The documentation on the class BaseClass that I wrote should have made that clear? This PR should have updated the class string for BaseTest to reflect the change...  I'll try a little change to update.

PR #1090, opened 2025-02-08
[LoweringStrategy] Use a more general method to fetch input dims and sizes

yzhang93:
There is a bug in the existing codes to get M/N/K size from a matmul-like op, i.e.,
```
const uint64_t M = initShape[0];
const uint64_t N = initShape[1];
const uint64_t K = lhsShape[1];
```
It's hard to infer the shape if the input matmul-like ops are transposed and in `linalg.generic` form. Or even the input has a higher number of dimensions such as mmt4d ops.

In addition, the indexing maps of matmul-like `linalg.generic` ops can be transposed during dispatch generation by default because of `TransposeGenericOpsPass`. 
Example:
`parallel parallel reduction parallel parallel reduction` will become `parallel parallel parallel parallel reduction reduction`.

And if we still put the pack size/tile size at the dim for the former indexing map, it will generate wrong size for tiling.

This PR uses the upstream method `linalg::inferContractionDims` to infer the dim indices of M/N/K for all contraction ops.


Abhishek-Varma (changes requested):
Nice! A few comments to address.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -84,6 +84,52 @@ FailureOr<std::array<uint32_t, 3>> getPackedSize(linalg::LinalgOp linalgOp,
...
+  SmallVector<unsigned, 2> nDims = contractionDims.n;
+  SmallVector<unsigned, 2> kDims = contractionDims.k;
+  if (mDims.empty() || nDims.empty() || kDims.empty()) {
+    return linalgOp.emitOpError("failed to fetch m/n/k dims.");
+  }
+
+  SmallVector<int64_t> shapes = linalgOp.getStaticLoopRanges();
+  if (mDims.size() + nDims.size() + kDims.size() > shapes.size()) {
+    return linalgOp.emitOpError(
+        "the total of m/n/k dims is larger than the number of loops.");
```
newling, line 114: emitOpError messages are of the form error: 'your_op_type' op <your string> so this error message will have odd grammar. Also please add additional information available (sizes of m, n, k, shape). Please check other error messages as well

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -842,6 +902,8 @@ static LogicalResult setRootConfig(mlir::FunctionOpInterface entryPointFn,
                                    uint32_t numCols) {
   assert(!getLoweringConfig<IREE::Codegen::LoweringConfigAttr>(genericOp) &&
          "expected lowering_config is not set");
+  unsigned numLoops = genericOp.getNumLoops();
+  assert(numLoops <= 7 && "expected input number of loops no more than 7");
```
newling, line 906: Are these asserts needed, where is this assumption of <= 7 used?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -84,6 +84,52 @@ FailureOr<std::array<uint32_t, 3>> getPackedSize(linalg::LinalgOp linalgOp,
...
+  if (failed(maybeContractionDims)) {
+    return linalgOp.emitOpError("failed to infer the contraction dimensions.");
+  }
+
+  linalg::ContractionDimensions contractionDims = *maybeContractionDims;
+  SmallVector<unsigned, 2> mDims = contractionDims.m;
+  SmallVector<unsigned, 2> nDims = contractionDims.n;
+  SmallVector<unsigned, 2> kDims = contractionDims.k;
+  if (mDims.empty() || nDims.empty() || kDims.empty()) {
+    return linalgOp.emitOpError("failed to fetch m/n/k dims.");
```
newling, line 108: When might this be empty after you've already not received failure from linalg::inferContractionDims ?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -84,6 +84,52 @@ FailureOr<std::array<uint32_t, 3>> getPackedSize(linalg::LinalgOp linalgOp,
   return instructionSize;
 }
 
+struct InputDimsAndSizes {
+  SmallVector<unsigned, 2> mDims;
```
newling, line 88: Personal preference: Use `SmallVector<T>` instead of `SmallVector<T, 2>`.

newling (changes requested):
Great improvement! 

Mostly nits, but a question about using ContractionDimensions's inferred batch dimension(s).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -84,6 +84,52 @@ FailureOr<std::array<uint32_t, 3>> getPackedSize(linalg::LinalgOp linalgOp,
...
+    return linalgOp.emitOpError("failed to fetch m/n/k dims.");
+  }
+
+  SmallVector<int64_t> shapes = linalgOp.getStaticLoopRanges();
+  if (mDims.size() + nDims.size() + kDims.size() > shapes.size()) {
+    return linalgOp.emitOpError(
+        "the total of m/n/k dims is larger than the number of loops.");
+  }
+
+  auto getSizesAt = [&shapes](const SmallVector<unsigned, 2> &idx) {
```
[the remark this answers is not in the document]
yzhang93: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -84,6 +84,61 @@ FailureOr<std::array<uint32_t, 3>> getPackedSize(linalg::LinalgOp linalgOp,
...
+
+  SmallVector<int64_t> shapes = linalgOp.getStaticLoopRanges();
+  size_t totalNumDims =
+      batchDims.size() + mDims.size() + nDims.size() + kDims.size();
+  assert(totalNumDims == shapes.size() &&
+         ("the total number of dims " + std::to_string(totalNumDims) +
+          " is not the same as the number of loops " +
+          std::to_string(shapes.size()) + ".")
+             .c_str());
+  (void)totalNumDims;
```
jtuyls, line 122:
```suggestion
  [[maybe_unused]] size_t totalNumDims =
      batchDims.size() + mDims.size() + nDims.size() + kDims.size();
  assert(totalNumDims == shapes.size() &&
         ("the total number of dims " + std::to_string(totalNumDims) +
          " is not the same as the number of loops " +
          std::to_string(shapes.size()) + ".")
             .c_str());
```

jtuyls (approved):
LGTM

MaheshRavishankar:
Nice!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -492,18 +551,24 @@ static LogicalResult setRootConfigForPackPeel4LevelTilingPipeline(
...
-  SmallVector<int64_t> tileSizeLevel2 = {0, 0, 1};
-  SmallVector<int64_t> tileSizeLevel3 = {1, 1, 0, 0, 0, 0};
 
+  SmallVector<int64_t> tileSizeLevel0(numLoops, 0);
   if (isa<linalg::BatchMatmulOp>(linalgOp)) {
-    tileSizeLevel0.insert(tileSizeLevel0.begin(), 1);
-    tileSizeLevel1.insert(tileSizeLevel1.begin(), 0);
-    tileSizeLevel2.insert(tileSizeLevel2.begin(), 0);
-    tileSizeLevel3.insert(tileSizeLevel3.begin(), 0);
+    tileSizeLevel0[batchDims[0]] = 1;
```
newling, line 557: Could assert that batchDims is not empty, like the others.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
newling: At the link above, it's doing the check just before calling back() on each of the vectors, so at a point where it knows that the intrinsic it's targeting has all 3. Your implicit assumption here is that you're doing some kind of matmul, but maybe we'll want to reuse this for matvec or something where there is no `n` dimension. 

Perhaps add a comment, change the class name, or move the checks later.

newling (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: Done.

Abhishek-Varma (approved):
LGTM!

PR #1094, opened 2025-02-11
[UKernel] Vectorize zeroization ukernel for npu4

jtuyls:
This should improve performance of matmul with ukernel on NPU4 and get it on par with the 'outline-empty' (no compute) performance.


Abhishek-Varma (approved):
LGTM!

jtuyls:
> Snap, I'm working on vectorizing zeroization on the peano pass.

This is not the same vectorization. Here, I adjusted the zeroization ukernel that was implemented with scalar operations to use vector operations instead.

newling:
Here's my PR https://github.com/nod-ai/iree-amd-aie/pull/1095

PR #1097, opened 2025-02-11
[UKernel] Add ukernel to be compiled through peano

jtuyls:
Adds a path to compile ukernels through peano. The compilation time is a lot faster this way, potentially at a decrease in performance, but potentially ukernels could be written in a way that enables good performance via peano as well.

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -47,6 +47,7 @@ class TestParams(ABC):
         lower_to_aie_pipeline="objectFifo",
         name_suffix="",
         use_chess=False,
+        use_chess_for_ukernel=True,
         use_ukernel=False,
         run_benchmark=False,
         n_repeats=1,
@@ -59,6 +60,7 @@ class TestParams(ABC):
         self.lower_to_aie_pipeline = lower_to_aie_pipeline
         self.name_suffix = name_suffix
         self.use_chess = use_chess
+        self.use_chess_for_ukernel = use_chess_for_ukernel
         self.use_ukernel = use_ukernel
         self.run_benchmark = run_benchmark
         self.n_repeats = n_repeats
@@ -104,6 +106,7 @@ class BaseTest(ABC):
         tile_pipeline = test_params.tile_pipeline
         lower_to_aie_pipeline = test_params.lower_to_aie_pipeline
         use_chess = test_params.use_chess
+        use_chess_for_ukernel = test_params.use_chess_for_ukernel
         use_ukernel = test_params.use_ukernel
         run_benchmark = test_params.run_benchmark
         n_repeats = test_params.n_repeats
@@ -128,8 +131,17 @@ class BaseTest(ABC):
             self.labels.append("Peano")
 
         if use_ukernel:
-            self.name += "_ukernel"
             self.labels.append("UKernel")
+            if use_chess_for_ukernel:
+                self.name += "_ukernel_chess"
+                self.add_aie_compilation_flags(
+                    [f"--iree-amd-aie-enable-chess-for-ukernel=1"]
+                )
+            else:
+                self.name += "_ukernel_peano"
+                self.add_aie_compilation_flags(
+                    [f"--iree-amd-aie-enable-chess-for-ukernel=0"]
+                )
 
         if run_benchmark:
             self.name += "_benchmark"
@@ -1707,7 +1719,9 @@ class Tests:
                 test_params=TestParams(
                     use_ukernel=True,
                     use_chess=False,
+                    use_chess_for_ukernel=False,
                     run_on_target=["npu4"],
+                    tile_pipeline="pack-peel-4-level-tiling",
                     aie_compilation_flags=[
                         "--iree-amdaie-num-rows=4",
                         "--iree-amdaie-num-cols=8",
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
@@ -519,7 +519,7 @@ static LogicalResult assembleFileUsingPeano(
   args.emplace_back("-o");
   args.emplace_back(outputFile);
   if (verbose) args.emplace_back("-v");
-  if (failed(runTool(peanoDir / "bin" / "clang", args, verbose))) {
+  if (failed(runTool((peanoDir / "bin" / "clang").string(), args, verbose))) {
     llvm::errs() << "Failed to assemble " << outputFile << ".o with peano";
     return failure();
   }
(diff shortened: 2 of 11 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -202,10 +194,14 @@ static const std::string _CHESS_INTRINSIC_WRAPPER_CPP{
 static const std::string _MM_NPU1_CC{
 #include "mm_npu1.cc"
 };
-// This is a string that contains a mm kernel for npu4.
+// This is a string that contains kernelS for npu4 for compilation by chess.
```
Abhishek-Varma, line 197: nit:
```suggestion
// This is a string that contains npu4 kernels for compilation by chess.
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -202,10 +194,14 @@ static const std::string _CHESS_INTRINSIC_WRAPPER_CPP{
...
 static const std::string _MM_NPU4_CC{
 #include "mm_npu4.cc"
 };
+// This is a string that contains kernels for npu4 for compilation by peano.
```
Abhishek-Varma, line 201: nit:
```suggestion
// This is a string that contains npu4 kernels for compilation by peano.
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/mm_npu4_peano.cc
```diff
@@ -0,0 +1,142 @@
...
+  static_assert(k / s > 0);
+  static_assert(n / t > 0);
+  return matmul_vectorized_i8_i32<m / r, k / s, n / t, r, s, t>
+      (pA, offsetA, pB, offsetB, pC, offsetC);
+}
+
+extern "C" {
+
+#define matmul_combos_i8(X, M, N, K)                                  \
+  X(int8, i8, int8, i8, int32, i32, M, N, K, 8, 8, 8)
```
Abhishek-Varma, line 116: Can we also add :-
```
#define matmul_combos(X, M, N, K)                                     \
  X(bfloat16, bf16, bfloat16, bf16, float, f32, M, N, K, 8, 8, 8)
```
like `mm_npu4.cc` ?

And same for different combos ?
```

matmul_combos(matmul_vectorized_c_func, 16, 8, 32)
matmul_combos(matmul_vectorized_c_func, 16, 8, 64)
matmul_combos(matmul_vectorized_c_func, 16, 16, 32)
matmul_combos(matmul_vectorized_c_func, 32, 32, 32)
matmul_combos(matmul_vectorized_c_func, 32, 32, 64)
matmul_combos(matmul_vectorized_c_func, 64, 64, 64)

matmul_combos_i8(matmul_vectorized_c_func, 16, 16, 32)
matmul_combos_i8(matmul_vectorized_c_func, 32, 16, 32)
matmul_combos_i8(matmul_vectorized_c_func, 32, 16, 64)
matmul_combos_i8(matmul_vectorized_c_func, 32, 32, 8)
matmul_combos_i8(matmul_vectorized_c_func, 32, 32, 32)
matmul_combos_i8(matmul_vectorized_c_func, 32, 32, 64)
matmul_combos_i8(matmul_vectorized_c_func, 64, 64, 64)
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/test/aie_elf_files_gen_test.cpp
```diff
@@ -111,6 +111,7 @@ int main(int argc, char **argv) {
           /*timing=*/false,
           /*tempDir=*/workDir.str().str(),
           /*useChess=*/false,
+          /*useChessForUKernel=*/false,
```
Abhishek-Varma, line 114: Wouldn't this also be `true` ?

build_tools/ci/cpu_comparison/run.py
```diff
@@ -1707,7 +1719,9 @@ def __init__(self):
                 test_params=TestParams(
                     use_ukernel=True,
                     use_chess=False,
```
Abhishek-Varma, line 1721: nit: This can be dropped.

Abhishek-Varma (changes requested):
Really nice @jtuyls ! A few comments to address.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/test/aie_elf_files_gen_test.cpp
jtuyls: It doesn't matter because it's not being called, but I explicitly don't want unit tests to depend on chess, so should be false.

build_tools/ci/cpu_comparison/run.py
jtuyls: I don't think it hurts to be explicit about it, and I also didn't touch this line.

Abhishek-Varma (approved):
LGTM!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -509,6 +505,27 @@ LogicalResult runTool(
...
+  args.reserve(args.size() + std::distance(extraArgs.begin(), extraArgs.end()));
+  args.insert(args.end(), extraArgs.begin(), extraArgs.end());
+  // TODO(jornt): O0 fails with peano, so we use O1 for now.
+  args.emplace_back("-O1");
+  args.emplace_back("-c");
+  args.emplace_back(inputFile);
+  args.emplace_back("-o");
+  args.emplace_back(outputFile);
+  if (verbose) args.emplace_back("-v");
+  if (failed(runTool((peanoDir / "bin" / "clang").string(), args, verbose))) {
```
newling, line 522: Wow, you can do this with peano? I must play around with this, see what the .ll generated looks like.

jtuyls: Indeed, this is seems quite useful for experimentation and speed of dev. However, I have run into a couple of issues already that I think you have seen as well with vectorization:
- O0 results in compilation errors
- Sometimes the stack size usage blows up and this results in incorrect results if it's larger than 1024 and I needed to manually increase it: https://github.com/nod-ai/iree-amd-aie/blob/8505ad73955b443b28d47c466604c026a42649bb/compiler/plugins/target/AMD-AIE/aie/AIEOps.td#L128. Ideally, we should know the stack size usage before allocating a stack buffer...

Also, it should work with the high level AIE_API as well, but then you need to include those header files and I think they are only available inside Vitis. But with Vitis installed on our CI machines, that should work as well in theory, although I haven't tried it yet.

PR #1098, opened 2025-02-12
Support matmul 4d inputs on pack-peel-4-level-tiling

yzhang93:
This PR adds basic support for matmul with 4d inputs and output on pack-peel-4-level-tiling.

In theory, the tiling strategy and pipeline should support different layouts of matmul 4d operations. However, the order of the input dims (both inner and outer dims) are crucial for the correct compilation and results. 

To ensure the correctness and for comparison purpose, this PR only adds a test that corresponds to the standard matmul, which means the order of the input dims for this operation corresponds to the L2 shapes of a matmul op after the first level packing, i.e.,

```C += matmul4d(A,B) where A:MxKxM0xK0, B:NxKxK0xN0, C:NxMxM0xN0```
  
The test class and instance added in run.py is preliminary and for experimental purpose. Generalization of the test class will be addressed as follow-ups.

### CODE REVIEW

**Comment by jtuyls** [comment]
> This is nice @yzhang93 !
> 
> But I think it'd be better to have a separate pipeline for targeting MMT4D - a notion similar to what [IREE LLVMCPU](https://github.com/iree-org/iree/blob/99304ff016d0ee0f742b620fef59ae1cd9706899/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.td#L26-L27) path takes.
> 
> I'm not too sure though if having an attribute that relays "number of loops" through an operation only to be used by the pipeline to decide which all set of passes to include is a nice design.

Yeah, I agree that these checks make the pipeline/tiling config hard to understand and update /debug in the future and a separate one, while trying to reuse as much as possible, would be better.

**Comment by yzhang93** [comment]
> > This is nice @yzhang93 !
> > But I think it'd be better to have a separate pipeline for targeting MMT4D - a notion similar to what [IREE LLVMCPU](https://github.com/iree-org/iree/blob/99304ff016d0ee0f742b620fef59ae1cd9706899/compiler/src/iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.td#L26-L27) path takes.
> > I'm not too sure though if having an attribute that relays "number of loops" through an operation only to be used by the pipeline to decide which all set of passes to include is a nice design.
> 
> Yeah, I agree that these checks make the pipeline/tiling config hard to understand and update /debug in the future and a separate one, while trying to reuse as much as possible, would be better.

I'm not sure if it's worth to use a new pipeline. This is directly derived from pack-peel/pack-peel-4-level-tiling pipeline (not like LLVMCPU mmt4d as a completely new pipeline). The only difference is it uses pad instead of pack for the dma copy https://github.com/nod-ai/iree-amd-aie/pull/1098/files#diff-42f0d0bb098689f25ee68e8f05ec6c2eaa89ce41a6394d7d99c1d1c912943b38R325-R346. The packing/tiling parameters are almost the same except that it omits the first level packing config. While using a separate pipeline may also hard to maintain, everytime we update the configs of pack-peel-4-level-tiling pipeline, we have to update the new pipeline.

File: compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEUtils.cpp [hunk 1]
```diff
@@ -341,7 +341,8 @@ bool isMatmulInDefChain(Value operand) {
 /// Utility to identify if `linalgOp` is an elementwise operation with a
 /// matmul-like op upstream in its computation tree.
 bool isMatmulProducerOfElementwise(linalg::LinalgOp linalgOp) {
```
newling, line 343: Confusing function name, should be isElementwiseWithMatmulProducer

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -460,55 +460,71 @@ static LogicalResult setRootConfigForPackPeel4LevelTilingPipeline(
...
-      context, packedSizesL0, transposePackIndices, unpackEmpty, innerPerm,
-      outerPerm);
+  SmallVector<int64_t> transposePackIndices;
+  SmallVector<bool> unpackEmpty;
+  SmallVector<SmallVector<int64_t>> innerPerm;
+  SmallVector<SmallVector<int64_t>> outerPerm;
+  SmallVector<PackingConfigPackingLevelAttr> packingConfigLevelsVal;
+
+  // Pack level => 1.
+  // The first level of packing is not needed if the number of loops is larger
```
newling, line 489: What is 'the number of loops' ?  How is this related to sizes of the SmallVector<unsigned, 2> vectors M, N, K, batch in your previous PR?

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEAttrs.td
```diff
@@ -162,4 +162,15 @@ def AMDAIE_PackingConfigAttr :
   }];
 }
 
+def AMDAIE_NumInputLoopsAttr :
```
newling, line 165: Documentation! Or maybe it can be removed... can the number of loops be inferred JIT?

build_tools/ci/cpu_comparison/run.py
```diff
@@ -473,6 +472,59 @@ def _execute(self, config):
         return self.vs_cpu(config)
 
 
+class Matmul4d(BaseMatmul):
+    """
+    A test of linalg.generic with 4d inputs and output implementing form:
+    C += matmul4d(A,B) where A:MxKxM0xK0, B:NxKxK0xN0, C:NxMxM0xN0
+
+    Note that the order of the input dims for this operation corresponds to
+    the L2 shapes of a standard matmul op after the first level packing.
```
newling, line 481: You have to be intimately familiar with the current pipeline to know what this sentence means, can you provide a more high-level description in terms of transposition? 

Actually is this sentence even needed, doesn't `A:MxKxM0xK0, B:NxKxK0xN0, C:NxMxM0xN0` describe the computation exactly (I assume that anyone reading this will know that K and K0 are reduction dimensions).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -460,55 +460,71 @@ static LogicalResult setRootConfigForPackPeel4LevelTilingPipeline(
     return linalgOp.emitOpError("failed to fetch m/n/k dims.");
   }
 
+  // Set attribute for number of input loops.
```
newling, line 463: I think I'm missing something here. You get the number of loops from linalgOp, and then set an attribute on the linalgOp. Why not just always call linalgOp.getNumLoops(); when you need it? 

Is it because  linalgOp.getNumLoops() changes as the op gets transformed, and you need to know what the original number of loops? If so the name should reflect that.

yzhang93: No, it's because we need this in Passes.cpp, but we can't directly access linalgOp. Ref https://github.com/nod-ai/iree-amd-aie/pull/1098/files#diff-42f0d0bb098689f25ee68e8f05ec6c2eaa89ce41a6394d7d99c1d1c912943b38R313.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEUtils.cpp
```diff
@@ -290,8 +290,8 @@ bool isMatmulTransposeA(linalg::LinalgOp linalgOp) {
 /// matmul_transpose_b op.
 bool isMatmulTransposeB(linalg::LinalgOp linalgOp) {
   // Step 0. Test if the op itself is a linalg.matmul_transpose_b op.
-  if (isa<linalg::MatmulTransposeBOp, linalg::BatchMatmulTransposeBOp>(
-          linalgOp))
+  if (isa<linalg::MatmulTransposeBOp, linalg::BatchMatmulTransposeBOp,
+          linalg::Mmt4DOp>(linalgOp))
```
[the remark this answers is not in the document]
yzhang93: I was originally directly working with Mmt4D op. Mmt4D op can compile through the same pipeline, but has numerical issue because the outer dimension of output has to be transposed. That's why I started to use linalg.generic ops. I can delete this if it brings confusion.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: As suggested above, I can bring a new pipeline if adding `num_loops` is hard to understand and debug,

newling: I see, thanks. 

I'm not really in favour of a new pipeline, I think it's not worth the duplication of code. Maybe it's possible to use the C++ language features to make this easy enough to understand (i.e. alternatives to if-else)? 

One option: instead of havinh the attribute num_loops, pass the `rootOp` into ` addPackPeel4LevelTilingBasedPassPipeline` directly? More direct and easier to understand IMO

Also... if we could have an identity linalg.pack there wouldn't even be any branching? Shouldn't we just request this change upstream?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEUtils.cpp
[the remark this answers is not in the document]
newling: Makes sense, thanks.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: > One option: instead of havinh the attribute num_loops, pass the `rootOp` into ` addPackPeel4LevelTilingBasedPassPipeline` directly? More direct and easier to understand IMO

Yeah, I think this could work. I'll do a revision with that.

> Also... if we could have an identity linalg.pack there wouldn't even be any branching? Shouldn't we just request this change upstream?

We could discuss other options in the sync meeting.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: It is the number of loops from the input linalg op. It equals to the total number of dims of batch,M,N,K.

build_tools/ci/cpu_comparison/run.py
yzhang93: Okay, updated the documentation.

yzhang93:
In the new revision, I've removed `num_loops` attribute and added a performance test on Strix. 

My local test shows the runtime as:
```matmul_512_4096_512_i8_i32 : 806us vs matmul4d_16_128_8_i8_i32: 491us```

File: compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp [hunk 7]
```diff
@@ -466,49 +466,64 @@ static LogicalResult setRootConfigForPackPeel4LevelTilingPipeline(
...
+  SmallVector<SmallVector<int64_t>> innerPerm;
+  SmallVector<SmallVector<int64_t>> outerPerm;
+  SmallVector<PackingConfigPackingLevelAttr> packingConfigLevelsVal;
+
+  // Pack level => 1.
+  // The initial number of loops from the linalg ops determines whether the
+  // first level of packing is needed. If the number of loops is larger
+  // than 4, it means the input is not a standard 2D matmul, then there is no
+  // need to pack operands from 2D to 4D.
+  unsigned numLoops = linalgOp.getNumLoops();
```
jtuyls, line 489: Wouldn't it be better to recognize a mmt4d pattern explicitly? Or alternatively, assert/throw an error if the number of loops is larger than 5? For example in case of a mmt6d, the second level shouldn't be packed either.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -550,14 +564,20 @@ static LogicalResult setRootConfigForPackPeel4LevelTilingPipeline(
...
+  if (isBatchMatmul) {
     assert(!batchDims.empty() && "expected batch dims not empty");
     tileSizeLevel0[batchDims[0]] = 1;
   }
-  tileSizeLevel0[mDims[0]] = packPeelTiling.M0 * scaleL0;
-  tileSizeLevel0[nDims[0]] = packPeelTiling.N0 * scaleL0;
+  if (numLoops > 4) {
+    m0Tile /= maybeInputDimsAndSizes.value().mSizes.back();
+    n0Tile /= maybeInputDimsAndSizes.value().nSizes.back();
+  }
```
jtuyls, line 578: Could you document this?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -323,18 +324,30 @@ void addPackPeel4LevelTilingBasedPassPipeline(
   funcPassManager.addPass(createCanonicalizerPass());
   funcPassManager.addPass(createCSEPass());
 
-  // First level packing
-  {
-    AMDAIEPackAndTransposeOptions packOptions;
-    packOptions.packLevel = 0;
-    funcPassManager.addPass(createAMDAIEPackAndTransposePass(packOptions));
+  // First level pack or pad operation depending on the number of input loops.
```
jtuyls, line 327: Could you add a similar comment as above about mmt4d etc:
> // The initial number of loops from the linalg ops determines whether the
  // first level of packing is needed. If the number of loops is larger
  // than 4, it means the input is not a standard 2D matmul, then there is no
  // need to pack operands from 2D to 4D.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -357,7 +370,7 @@ void addPackPeel4LevelTilingBasedPassPipeline(
   // Second level packing
   {
     AMDAIEPackAndTransposeOptions packOptions;
-    packOptions.packLevel = 1;
+    packOptions.packLevel = numInputLoops <= 4 ? 1 : 0;
```
jtuyls, line 373: Could you document why this is needed?

jtuyls (changes requested):

build_tools/ci/cpu_comparison/run.py
```diff
@@ -2181,6 +2251,24 @@ def __init__(self):
                 "tile_pipeline": "pack-peel-4-level-tiling",
                 "run_on_target": "npu4",
             },
+            # matmul4d test where the input M/N/K are outer dim values.
+            # The total input values correspond to a standard matmul
+            # from the above test are M:512, N:4096, K:512.
+            {
+                "M": 16,
+                "N": 128,
+                "K": 8,
```
jtuyls, line 2260: I think it would be better to provide total and inner and then derive the outer from it:
```
"M": 512,
"N": 4096,
"K": 512,
"M0": 32
"N0": 32
"K0": 64
```

jtuyls:
> In the new revision, I've removed `num_loops` attribute and added a performance test on Strix.
> 
> My local test shows the runtime as: `matmul_512_4096_512_i8_i32 : 806us vs matmul4d_16_128_8_i8_i32: 491us`

Very nice!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: That's a good point, will check the pattern directly.

build_tools/ci/cpu_comparison/run.py
yzhang93: I'd like to address this comment as a follow-up. Need more changes of Matmul4d and its base class.

jtuyls (approved):
LGTM

newling:
> matmul_512_4096_512_i8_i32 : 806us vs matmul4d_16_128_8_i8_i32: 491us

That's amazing, where does the speed-up come from?

PR #1099, opened 2025-02-12
[e2e CI] Add e2e CI tests for Matmul+Trunci with scaling

Abhishek-Varma:
-- This commit adds e2e CI tests for Matmul+Trunci with scaling.
-- In an actual quantized model, truncating from a higher bitwidth to a lower precision bitwidth 
    won't work and we need to scale.
-- Since the output of the Matmul here is an integer cannot be multiplied with a floating point
    scale factor, we need to represent the scale factor with a multiplier and a shift operator instead.
-- Eg: a float scale factor of 0.333 could become multiply by 357913941 and shift right 30.

### DIFF
--- a/build_tools/ci/cpu_comparison/matmul_template/matmul_generator.py
+++ b/build_tools/ci/cpu_comparison/matmul_template/matmul_generator.py
@@ -3,6 +3,14 @@ import re
 import os
 
 
+def get_higher_order_element_type(element_type):
+    if element_type[0] in ["i", "f"]:
+        assert element_type[1:].isdigit(), f"support for {element_type} is missing"
+        bit_width = int(element_type[1:])
+        return f"{element_type[0]}{bit_width*2}"
+    assert False, f"support for {element_type} is missing"
+
+
 def generate_matmul_test(output_fn, input_fn, m, n, k, lhs_rhs_type, acc_type, b=0):
     """
     Generate mlir file (output_fn) from the template file (input_fn).
@@ -14,6 +22,8 @@ def generate_matmul_test(output_fn, input_fn, m, n, k, lhs_rhs_type, acc_type, b
     replace["K"] = k
     replace["TYPE1"] = lhs_rhs_type
     replace["TYPE2"] = acc_type
+    # Only used for Matmul+Trunc via scaling.
+    replace["TYPE3"] = get_higher_order_element_type(acc_type)
 
     replace["B"] = b  # This is only used for batch matmul
     acc_is_int = acc_type[0] == "i"
--- a/build_tools/ci/cpu_comparison/matmul_template/matmul_generator.py
+++ b/build_tools/ci/cpu_comparison/matmul_template/matmul_generator.py
@@ -23,7 +23,7 @@ def generate_matmul_test(output_fn, input_fn, m, n, k, lhs_rhs_type, acc_type, b
     replace["TYPE1"] = lhs_rhs_type
     replace["TYPE2"] = acc_type
     # Only used for Matmul+Trunc via scaling.
-    replace["TYPE3"] = get_higher_order_element_type(acc_type)
+    replace["TYPE_MUL_RESULT"] = get_higher_order_element_type(acc_type)
 
     replace["B"] = b  # This is only used for batch matmul
     acc_is_int = acc_type[0] == "i"
--- a/build_tools/ci/cpu_comparison/matmul_template/matmul_trunci_scaling_MxK_KxN.mlir
+++ b/build_tools/ci/cpu_comparison/matmul_template/matmul_trunci_scaling_MxK_KxN.mlir
@@ -10,7 +10,7 @@ func.func @matmul_trunci(%arg0: tensor<${M}x${K}x${TYPE1}>, %arg1: tensor<${K}x$
 {
   %cst = arith.constant ${ZERO} : ${TYPE2}
   %cst_mul = arith.constant 10 : ${TYPE_MUL_RESULT}
-  %cst_div = arith.constant 7 : ${TYPE_MUL_RESULT}
+  %cst_shift = arith.constant 7 : ${TYPE_MUL_RESULT}
   %0 = tensor.empty() : tensor<${M}x${N}x${TYPE2}>
   %i8out = tensor.empty() : tensor<${M}x${N}x${TYPE1}>
   %1 = linalg.fill ins(%cst : ${TYPE2}) outs(%0 : tensor<${M}x${N}x${TYPE2}>) -> tensor<${M}x${N}x${TYPE2}>
@@ -25,7 +25,7 @@ func.func @matmul_trunci(%arg0: tensor<${M}x${K}x${TYPE1}>, %arg1: tensor<${K}x$
     ^bb0(%in: ${TYPE2}, %out: ${TYPE1}):
       %4 = arith.extsi %in : ${TYPE2} to ${TYPE_MUL_RESULT}
       %5 = arith.muli %4, %cst_mul : ${TYPE_MUL_RESULT}
-      %6 = arith.shrsi %5, %cst_div : ${TYPE_MUL_RESULT}
+      %6 = arith.shrsi %5, %cst_shift : ${TYPE_MUL_RESULT}
       %7 = arith.trunci %6 : ${TYPE_MUL_RESULT} to ${TYPE1}
       linalg.yield %7 : ${TYPE1}
     } -> tensor<${M}x${N}x${TYPE1}>
--- a/build_tools/ci/cpu_comparison/matmul_template/matmul_trunci_scaling_MxK_KxN.mlir
+++ b/build_tools/ci/cpu_comparison/matmul_template/matmul_trunci_scaling_MxK_KxN.mlir
@@ -9,8 +9,8 @@
 func.func @matmul_trunci(%arg0: tensor<${M}x${K}x${TYPE1}>, %arg1: tensor<${K}x${N}x${TYPE1}>) -> tensor<${M}x${N}x${TYPE1}>
 {
   %cst = arith.constant ${ZERO} : ${TYPE2}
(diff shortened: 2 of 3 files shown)


build_tools/ci/cpu_comparison/run.py
```diff
@@ -1590,6 +1594,78 @@ def __init__(self):
                 ),
             )
         )
+
+        # Tests Matmul + Trunci with Scaling.
+        # Phoenix : Ukernel + Peano.
+        self.register(
```
newling, line 1600: Not specific to this PR, but I think we should consider factorizing/standardizing tests/functions/templates where possible in this file, it's becoming too large and unwieldy IMO

build_tools/ci/cpu_comparison/matmul_template/matmul_trunci_scaling_MxK_KxN.mlir
```diff
@@ -0,0 +1,27 @@
+// input ${M}x${K}x${TYPE1}
+// input ${K}x${N}x${TYPE1}
+
+func.func @matmul_trunci(%arg0: tensor<${M}x${K}x${TYPE1}>, %arg1: tensor<${K}x${N}x${TYPE1}>) -> tensor<${M}x${N}x${TYPE1}>
```
newling, line 4: Needs a comment about what this template does

Seems like its 

C = ((A @ B) * cst_mul / cst_div) 

which a bit strange maybe, should the `*` be an `+`  (related to TYPE3 suggested by Jorn)

Function name misleading?

build_tools/ci/cpu_comparison/matmul_template/matmul_trunci_scaling_MxK_KxN.mlir
```diff
@@ -0,0 +1,27 @@
...
+{
+  %cst = arith.constant ${ZERO} : ${TYPE2}
+  %cst_mul = arith.constant 10 : ${TYPE2}
+  %cst_div = arith.constant 137 : ${TYPE2}
+  %0 = tensor.empty() : tensor<${M}x${N}x${TYPE2}>
+  %i8out = tensor.empty() : tensor<${M}x${N}x${TYPE1}>
```
newling, line 10: i8out, so TYPE1 is i8?

newling:
Nice to see this works! Can you please add the IR (aievec level and/or xllvm and/or .ll and/or .asm) that is generated in the PR description?

newling (changes requested):

build_tools/ci/cpu_comparison/matmul_template/matmul_trunci_scaling_MxK_KxN.mlir
jtuyls: No, it should be a multiplication + division to mimic a float scale to scale from i32 range to i8 range.

newling: Why are you trying to mimic that? (Documentation please Abhishek :-))

If the goal here is to get into a safe range where you don't overflow end-to-end test with large K with just a trunci, you could quite easily do that by choosing the initial values of A and B appropriately

build_tools/ci/cpu_comparison/matmul_template/matmul_trunci_scaling_MxK_KxN.mlir
Abhishek-Varma: Correct. It's a i8->i8 test.

build_tools/ci/cpu_comparison/matmul_template/matmul_trunci_scaling_MxK_KxN.mlir
```diff
@@ -0,0 +1,27 @@
...
+  %2 = linalg.matmul ins(%arg0, %arg1 : tensor<${M}x${K}x${TYPE1}>, tensor<${K}x${N}x${TYPE1}>)
+    outs(%1: tensor<${M}x${N}x${TYPE2}>) -> tensor<${M}x${N}x${TYPE2}>
+  %3 = linalg.generic {indexing_maps = [
+                              affine_map<(d0, d1) -> (d0, d1)>,
+                              affine_map<(d0, d1) -> (d0, d1)>
+                       ],
+                       iterator_types = ["parallel", "parallel"]
+                      } ins(%2 : tensor<${M}x${N}x${TYPE2}>) outs(%i8out : tensor<${M}x${N}x${TYPE1}>) {
+    ^bb0(%in: ${TYPE2}, %out: ${TYPE1}):
+      %26 = arith.muli %in, %cst_mul : ${TYPE2}
```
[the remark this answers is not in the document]
Abhishek-Varma: Done.

build_tools/ci/cpu_comparison/matmul_template/matmul_generator.py
```diff
@@ -14,6 +22,8 @@ def generate_matmul_test(output_fn, input_fn, m, n, k, lhs_rhs_type, acc_type, b
     replace["K"] = k
     replace["TYPE1"] = lhs_rhs_type
     replace["TYPE2"] = acc_type
+    # Only used for Matmul+Trunc via scaling.
+    replace["TYPE3"] = get_higher_order_element_type(acc_type)
```
jtuyls, line 26: Instead of TYPE1, TYPE2, TYPE3 etc, I think we should start using more descriptive  type names, like for example: `TYPE_MUL_RESULT`. This will scale better towards an increasing number of different tests. No need to change the existing types necessarily, but it would be good to replace `TYPE3` with something more descriptive.

Abhishek-Varma: Done.

jtuyls (approved):
LGTM

newling (changes requested):
> // Matmul + Trunci variant with scaling.
// In an actual quantized model, truncating from a higher bitwidth to a lower precision bitwidth
// won't work and we need to scale.
// Since the output of the Matmul here is an integer cannot be multiplied with a floating point
// scale factor, we need to represent the scale factor with a multiplier and a shift operator instead.

I still don't understand the motivation for this test. What new functionality is this testing? In practice dispatches will have casts to and from float, and float scale and bias. Why not do that here?

jtuyls:
> > // Matmul + Trunci variant with scaling.
> > // In an actual quantized model, truncating from a higher bitwidth to a lower precision bitwidth
> > // won't work and we need to scale.
> > // Since the output of the Matmul here is an integer cannot be multiplied with a floating point
> > // scale factor, we need to represent the scale factor with a multiplier and a shift operator instead.
> 
> I still don't understand the motivation for this test. What new functionality is this testing? In practice dispatches will have casts to and from float, and float scale and bias. Why not do that here?

Because then we would have to convert int to float and do scalar computations. With the logic here, we assume the floating point operation is converted into a fixed point operation and that's what we execute. This conversion can be done with a pattern rewriter in the frontend somewhere.

newling:
> Because then we would have to convert int to float and do scalar computations. With the logic here, we assume the floating point operation is converted into a fixed point operation and that's what we execute (we can use vector instructions for this). This conversion can be done with a pattern rewriter in the frontend somewhere.

Hmm, but isn't the scale factor less than 1.0, how can this be converted into an integer operation?

jtuyls:
> > Because then we would have to convert int to float and do scalar computations. With the logic here, we assume the floating point operation is converted into a fixed point operation and that's what we execute (we can use vector instructions for this). This conversion can be done with a pattern rewriter in the frontend somewhere.
> 
> Hmm, but isn't the scale factor less than 1.0, how can this be converted into an integer operation?

You convert it in a multiplier + shift. For example float scale 0.333... could become multiply by 357913941 and shift right 30.

newling:
My bad, I didn't realize the IR had already been updated to 

```
      %5 = arith.muli %4, %cst_mul : ${TYPE_MUL_RESULT}
      %6 = arith.shrsi %5, %cst_div : ${TYPE_MUL_RESULT}
      %7 = arith.trunci %6 : ${TYPE_MUL_RESULT} to ${TYPE1}
```

build_tools/ci/cpu_comparison/matmul_template/matmul_trunci_scaling_MxK_KxN.mlir
newling: If it's always an i8 test then TYPE1 shouldn't be templatized. 

Maybe this test shouldn't be templatized, for example: https://github.com/nod-ai/iree-amd-aie/blob/main/build_tools/ci/cpu_comparison/test_files/matmul_int32.mlir

newling (approved):

build_tools/ci/cpu_comparison/matmul_template/matmul_trunci_scaling_MxK_KxN.mlir
```diff
@@ -0,0 +1,33 @@
...
+// Matmul + Trunci variant with scaling.
+// In an actual quantized model, truncating from a higher bitwidth to a lower precision bitwidth
+// won't work and we need to scale.
+// Since the output of the Matmul here is an integer cannot be multiplied with a floating point
+// scale factor, we need to represent the scale factor with a multiplier and a shift operator instead.
+func.func @matmul_trunci(%arg0: tensor<${M}x${K}x${TYPE1}>, %arg1: tensor<${K}x${N}x${TYPE1}>) -> tensor<${M}x${N}x${TYPE1}>
+{
+  %cst = arith.constant ${ZERO} : ${TYPE2}
+  %cst_mul = arith.constant 10 : ${TYPE_MUL_RESULT}
+  %cst_div = arith.constant 7 : ${TYPE_MUL_RESULT}
```
jtuyls, line 13: Should be renamed to `%cst_shift`

Abhishek-Varma: Done.

PR #1101, opened 2025-02-13
Remove checks for Vitis when targeting ukernels

jtuyls:
This removes explicit checks for Vitis/chess and allows compiling and running an e2e test with ukernels with just peano installed.


Abhishek-Varma (approved):
Nice! LGTM.

PR #1111, opened 2025-02-14
[ControlOverlay] Introduce `dma_placeholder` to preserve `connection` operation from Dead-Code Elimination

Yu-Zhewen:
Following the discussion in #1063, the `connection` operation is defined as pure, meaning it will be eliminated by CSE and Canonicalization if it has no users. However, during control overlay generation, the generated `connection` operations do not initially have DMA users, as these are only generated later when the content of control packets is determined.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
@@ -586,6 +586,21 @@ def AMDAIE_NpuDmaCpyNdOp: AMDAIE_Op<"npu.dma_cpy_nd", [
   let hasCanonicalizer = 1;
 }
 
+def AMDAIE_NpuDmaPlaceHolderOp : AMDAIE_Op<"npu.dma_placeholder"> {
+  let summary = "Represents a placeholder for a DMA operation.";
+  let description = [{
+    This operation acts as a placeholder user for those `amdaie.connection` operations
+    that represent control flows, to prevent them from being dead-code eliminated.
+    This operation does not have any side effects on control code size.
+  }];
+
+  let arguments = (
+    ins Index:$connection
+  );
+
+  let assemblyFormat = [{ `(` $connection `)` attr-dict }];
+}
+
 def AMDAIE_NpuHalfDmaCpyNdOp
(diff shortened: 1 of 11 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp
```diff
@@ -601,8 +601,16 @@ LogicalResult AIEDeviceBuilder::connectionToAIE(
...
 
   std::optional<AMDAIE::FlowOp> maybeFlowOp = connectionOp.getFlowOp();
-  std::optional<uint8_t> packetId =
-      maybeFlowOp ? maybeFlowOp->getPacketId() : std::nullopt;
+  if (!maybeFlowOp) return connectionOp.emitOpError() << "has no flow op";
+
+  FailureOr<bool> isCtrlFlow = maybeFlowOp->isControlFlow();
+  if (failed(isCtrlFlow))
+    return connectionOp.emitOpError()
+           << "could not determine if flow is control";
```
jtuyls, line 609:
```suggestion
  if (failed(isCtrlFlow)) {
    return connectionOp.emitOpError()
           << "could not determine if flow is control";
  }
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
```diff
@@ -586,6 +586,21 @@ def AMDAIE_NpuDmaCpyNdOp: AMDAIE_Op<"npu.dma_cpy_nd", [
   let hasCanonicalizer = 1;
 }
 
+def AMDAIE_NpuDmaPlaceHolderOp : AMDAIE_Op<"npu.dma_placeholder"> {
+  let summary = "Represents a placeholder for a DMA operation.";
+  let description = [{
+    This operation acts as a placeholder user for those `amdaie.connection` operations
+    that represent control flows, to prevent them from being dead-code eliminated.
+    This operation does not have any side effects on control code size.
```
jtuyls, line 594: I would write this a bit more generally as there could be potential other uses for this placeholder in the future.
```suggestion
    This operation acts as a placeholder user for `amdaie.connection` operations to prevent
    them from being dead-code eliminated. This is used for control flow connections that are
    inserted before control packets are generated because they need to be taken into account
    together with data connections for routing. This operation does not have any side effects
    on control code size.
```

jtuyls (approved):
Nice!

PR #1112, opened 2025-02-14
[CI] Modify matmul4d to use total size as input args

yzhang93:
Before this PR, input args M/N/K are the outer dims for matmul4d ops. For easier comparison with standard matmul, this PR keeps the use of M/N/K as total matmul input sizes, while adding inner dim sizes as another parameter.


jtuyls (approved):
LGTM

PR #1114, opened 2025-02-17
[UKernel] Add bf16 ukernel for peano and move tests to peano

jtuyls:


build_tools/ci/cpu_comparison/run.py
```diff
@@ -1581,74 +1581,6 @@ def __init__(self):
         self.existing_names = []
         self.tests = []
 
-        # Tests Matmul + Trunci.
```
jtuyls, line 1584: Removing the matmul + trunci tests to reduce CI time as there are still matmul+scaling+trunci tests

Abhishek-Varma: We should then cleanup the `matmul_trunci*` template test and the base class ?

Abhishek-Varma (approved):
LGTM! Thanks for this @jtuyls ! Just one question.

build_tools/ci/cpu_comparison/run.py
jtuyls: Done

PR #1095, opened 2025-02-11
Vectorization of linalg.fill 

newling:
This PR contains multiple changes to get vectorized assembly through peano, I will split it into multiple PRs. 

Eyeballing the first test in the performance benchmark:

Before:

```
matmul_512_512_4096_bf16_f32_O2_npu1_4col_benchmark
--------------------------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations UserCounters...
--------------------------------------------------------------------------------------------------
BM_matmul/process_time/real_time_mean         2673 us         66.3 us            5 items_per_second=374.194/s
BM_matmul/process_time/real_time_median       2668 us         63.1 us            5 items_per_second=374.766/s
BM_matmul/process_time/real_time_stddev       19.3 us         16.7 us            5 items_per_second=2.69245/s
--------------------------------------------------------------------------------------------------
The largest program memory size (read from byte 72 of elf files) is 11184 bytes
```

After:

```
matmul_512_512_4096_bf16_f32_O2_npu1_4col_benchmark
--------------------------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations UserCounters...
--------------------------------------------------------------------------------------------------
BM_matmul/process_time/real_time_mean         2601 us         48.6 us            5 items_per_second=385.252/s
BM_matmul/process_time/real_time_median       2540 us         39.1 us            5 items_per_second=393.696/s
BM_matmul/process_time/real_time_stddev        132 us         21.0 us            5 items_per_second=18.3503/s
--------------------------------------------------------------------------------------------------
The largest program memory size (read from byte 72 of elf files) is 10112 bytes
```

So a nice saving on program memory, and maybe a marginal throughput boost. There a consistent saving of 1K memory across all (non-ukernel) benchmarks

PR #1117, opened 2025-02-19
[AMDAIESinkIntoCore] Generalize sinking for reuse with other regioned ops 

newling:
This is part of the PR to vectorize linalg.fill: https://github.com/nod-ai/iree-amd-aie/pull/1095

Basically one of the patterns introduced in https://github.com/nod-ai/iree-amd-aie/pull/1095 means that in one of the subsequent passes (lowering to llvm dialect) a cast operation is introduced outside of an aie.core, which needs to be inside the aie.core for core-to-standard to work.  i.e. we need to sink an operation into an aie.core. Before this PR, there is already a pass to sink operations into `amdaie.core`. This PR refactors that pass so that it can be reused to sink into `aie.core` (or any other regioned op).

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEUtils.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEUtils.cpp
@@ -16,7 +16,8 @@
 namespace mlir::iree_compiler::AMDAIE {
 
 std::string getConstantIntValuesString(ArrayRef<OpFoldResult> ofrs) {
-  auto maybeValues = mlir::getConstantIntValues(ofrs);
+  std::optional<SmallVector<int64_t>> maybeValues =
+      mlir::getConstantIntValues(ofrs);
   if (maybeValues.has_value())
     return getArrayString<int64_t>(maybeValues.value());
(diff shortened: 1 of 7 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEUtils.cpp
```diff
@@ -397,4 +397,47 @@ int detail::findLargestFactor(int num, int max, int multiple) {
...
+          // Skip if the dependency is already in the core.
+          if (parentOfRegion->isAncestor(dependencyOp)) continue;
+          if (!shouldSink(dependencyOp)) continue;
+          rewriter.setInsertionPointToStart(&block);
+          Operation *sunkOp = rewriter.clone(*dependencyOp);
+          nextIterationOps.push_back(sunkOp);
+          // Replace uses of the dependency op inside the block. Specifically,
+          // if `use` is in `block` then replace its operand with `sunkOp`.
+          auto isInBlock = [&block](OpOperand &use) {
+            auto op = use.getOwner();
```
jtuyls, line 425:
```suggestion
            Operation *op = use.getOwner();
```

jtuyls (approved):
LGTM, just one nit below.

PR #1119, opened 2025-02-19
Deprecate pad-pack pipeline and remove unused tests

yzhang93:
- pad-pack pipeline was only used with AIR, and now all the tests can be alternatively run with pack-peel pipeline. So deprecate this pipeline for better code maintenance. @erwei-xilinx 
- Also remove the pytests directory which are not used in ci currently because of https://github.com/nod-ai/iree-amd-aie/pull/924#issuecomment-2498827151.


erwei-xilinx (approved):

jtuyls (approved):
LGTM, thanks!

newling (approved):

PR #1123, opened 2025-02-21
[WIP][Do Not Reiew] Control packet end-to-end example

Yu-Zhewen:
Several places are currently hardcoded. Needs to be polished and should be split into multiple PRs.


Yu-Zhewen:
Checked in as: #1127, #1143, #1147, #1162

PR #1127, opened 2025-02-24
[CtrlPkt] Add pipeline to dump control packet binaries to files

Yu-Zhewen:
This PR implements a pipeline in `XCLBinGen` for the generation and conversion of control packets. The `aie.device` is converted into `ctrlpkt_sequence.txt` and `ctrlpkt_instructions.txt`, which store the contents and instructions for control packet DMAs, respectively. These files will later be used by the driver to perform the reconfiguration.

Changes related to the driver and a new on-device test will be submitted in follow-up PRs.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
@@ -1382,11 +1382,12 @@ namespace mlir::iree_compiler::AMDAIE {
 
 /// Pipeline to generate control packets from `xilinx::aie::device`, and dump
 /// them into files.
-LogicalResult generateControlPackets(
-    MLIRContext *context, AIE::DeviceOp deviceOp,
-    const IREE::HAL::ExecutableTargetAttr &targetAttr, const Path &tempDirPath,
-    bool printIRBeforeAll, bool printIRAfterAll, bool printIRModuleScope,
-    bool timing) {
+LogicalResult generateControlPackets(MLIRContext *context,
+                                     AIE::DeviceOp deviceOp,
+                                     const Path &tempDirPath,
+                                     bool printIRBeforeAll,
+                                     bool printIRAfterAll,
+                                     bool printIRModuleScope, bool timing) {
   assert(deviceOp->getParentOp() && isa<ModuleOp>(deviceOp->getParentOp()) &&
          "DeviceOp must be in a module parent");
   PassManager pm(context, ModuleOp::getOperationName());
@@ -1399,6 +1400,7 @@ LogicalResult generateControlPackets(
     options.pathToElfs = tempDirPath.string();
     pm.addPass(createAMDAIEConvertDeviceToControlPacketsPass(options));
   }
+  // TODO (zhewen): avoid regeneration?
   // Regenerate the overlay for sending control packets.
   {
     AMDAIEGenerateControlOverlayOptions options;
@@ -1407,6 +1409,7 @@ LogicalResult generateControlPackets(
     pm.addPass(createCSEPass());
     pm.addPass(createCanonicalizerPass());
   }
+  // TODO (zhewen): avoid regeneration?
   // Regenerate the flows and packet ids.
   pm.addPass(createAMDAIEConnectionToFlowPass());
   pm.addPass(createAMDAIEAssignPacketIdsPass());
@@ -1424,6 +1427,7 @@ LogicalResult generateControlPackets(
   pm.addPass(createAMDAIEControlCodeToTransactionPass());
 
   // Run the pipeline.
+  auto targetAttr = IREE::HAL::ExecutableTargetAttr::lookup(deviceOp);
   ModuleOp moduleOpCopy = cast<ModuleOp>(deviceOp->getParentOp()).clone();
   moduleOpCopy->setAttr("hal.executable.target", targetAttr);
   if (failed(pm.run(moduleOpCopy))) {
@@ -1505,8 +1509,7 @@ LogicalResult aie2xclbin(
     const std::string &xclBinInstanceName, const std::string &amdAIEInstallDir,
     const std::optional<std::string> &InputXCLBin,
     const std::optional<std::string> &ukernel,
-    const std::string &additionalPeanoOptFlags,
-    const IREE::HAL::ExecutableTargetAttr &targetAttr) {
+    const std::string &additionalPeanoOptFlags) {
   if (outputNPU.has_value() &&
       failed(emitDenseArrayAttrToFile(deviceOp, "npu_instructions",
                                       outputNPU.value()))) {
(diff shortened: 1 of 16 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -1375,69 +1376,116 @@ LogicalResult generateUnifiedObject(
...
-                                     bool printIRModuleScope, bool timing) {
+}  // namespace
+
+namespace mlir::iree_compiler::AMDAIE {
+
+/// Pipeline to generate control packets from `xilinx::aie::device`, and dump
+/// them into files.
+LogicalResult generateControlPackets(
+    MLIRContext *context, AIE::DeviceOp deviceOp,
+    const IREE::HAL::ExecutableTargetAttr &targetAttr, const Path &tempDirPath,
```
jtuyls, line 1387:
```suggestion
    IREE::HAL::ExecutableTargetAttr targetAttr, const Path &tempDirPath,
```
Attributes are typically passed by value, they are small wrappers around a pointer, so the same size.

Also, can't you just get the target attribute from the DeviceOp?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.h
```diff
@@ -26,10 +28,11 @@ mlir::LogicalResult aie2xclbin(
     const std::string &xclBinInstanceName, const std::string &amdAIEInstallDir,
     const std::optional<std::string> &InputXCLBin,
     const std::optional<std::string> &ukernel,
-    const std::string &additionalPeanoOptFlags);
+    const std::string &additionalPeanoOptFlags,
+    const IREE::HAL::ExecutableTargetAttr &targetAttr);
```
jtuyls, line 32:
```suggestion
    IREE::HAL::ExecutableTargetAttr targetAttr);
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -1375,69 +1376,116 @@ LogicalResult generateUnifiedObject(
...
-
-}  // namespace
+  // Assuming the ELF files have already been generated and are stored in
+  // `tempDirPath`, use aie-rt to generate control packets.
+  {
+    AMDAIEConvertDeviceToControlPacketsOptions options;
+    options.pathToElfs = tempDirPath.string();
+    pm.addPass(createAMDAIEConvertDeviceToControlPacketsPass(options));
+  }
+  // Regenerate the overlay for sending control packets.
```
jtuyls, line 1402: I can see that this is not possible right now, but it would be nice to avoid the need for this regeneration, so maybe add a TODO for that.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -1457,9 +1505,11 @@ LogicalResult aie2xclbin(
     const std::string &xclBinInstanceName, const std::string &amdAIEInstallDir,
     const std::optional<std::string> &InputXCLBin,
     const std::optional<std::string> &ukernel,
-    const std::string &additionalPeanoOptFlags) {
+    const std::string &additionalPeanoOptFlags,
+    const IREE::HAL::ExecutableTargetAttr &targetAttr) {
```
jtuyls, line 1509:
```suggestion
    IREE::HAL::ExecutableTargetAttr targetAttr) {
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/test/ctrlpkt_gen.mlir
```diff
@@ -0,0 +1,460 @@
+// RUN: aie_elf_files_gen_test %s %T true
+// RUN: FileCheck %s --check-prefix=CTRLPKT-SEQUENCE < %T/ctrlpkt_sequence.txt
+// RUN: FileCheck %s --check-prefix=CTRLPKT-INSTRUCTIONS < %T/ctrlpkt_instructions.txt
+
+// To check both files are not empty
```
jtuyls, line 5:
```suggestion
// To check that both files are not empty
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/test/ctrlpkt_gen.mlir
```diff
@@ -0,0 +1,460 @@
...
+        aie.rule(31, 2, %3) {packet_ids = array<i32: 2>}
+      }
+      aie.packet_rules(SOUTH : 7) {
+        aie.rule(30, 0, %2) {packet_ids = array<i32: 0, 1>}
+      }
+      aie.packet_rules(NORTH : 1) {
+        aie.rule(31, 0, %4) {packet_ids = array<i32: 0>}
+      }
+    }
+    %core_0_2 = aie.core(%tile_0_2) {
```
jtuyls, line 305: I think the size of the core code can be reduced to simplify the test and improve maintainability, like remove the scalar fill in the beginning and the store at the end.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEControlCodeLowering.cpp
```diff
@@ -23,34 +23,61 @@ struct HalfDmaCpyNdToNpuConverter final
...
+
+    int64_t argIdx;
+    int64_t elemWidthInBits;
+    uint8_t memSpace;
+    if (lowerCtrlpktDma) {
+      // Control packet DMAs require special handling.
+      // The `argIdx` must match the driver and is assumed to be 0.
+      // The `elemWidthInBits` is fixed at 32 for control packet data.
+      // The `memSpace` is set to 0, indicating storage in global memory.
+      argIdx = 0;
```
jtuyls, line 56: Is this reserving kernel argument 0 for control packets? But the normal data buffers are also offset from zero, so I think you're assuming here that there will be 2 runs? One to reconfigure and one to perform the actual operation?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
Yu-Zhewen: Done

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEControlCodeLowering.cpp
jtuyls: Thanks, makes sense

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEControlCodeLowering.cpp
```diff
@@ -23,34 +23,61 @@ struct HalfDmaCpyNdToNpuConverter final
...
-    auto subspanOp = dyn_cast_if_present<IREE::HAL::InterfaceBindingSubspanOp>(
-        logicalObjFifo.getMemref().getDefiningOp());
-    if (!subspanOp) {
-      return logicalObjFifo.emitOpError()
-             << "must operate on an `hal.interface.binding.subspan`";
+
+    int64_t argIdx;
+    int64_t elemWidthInBits;
+    uint8_t memSpace;
+    if (lowerCtrlpktDma) {
```
jtuyls, line 51: You could make these the default values for `argIdx`, `elemWidthInBits` and `memSpace` as they seem reasonable defaults as well (with the comments) or if you prefer it with the if/else maybe add default values for those variables above so they don't have undefined values and potentially cause future non deterministic bugs.

jtuyls (approved):
LGTM

PR #1135, opened 2025-02-26
MLIR-AIR pack-peel-4-level-tiling

erwei-xilinx:


compiler/plugins/target/AMD-AIE/iree-amd-aie/Test/samples/matmul_pack_peel_4_level_tiling_air_e2e.mlir
```diff
@@ -0,0 +1,15 @@
...
+  %0 = tensor.empty() : tensor<512x512xf32>
+  %1 = linalg.fill ins(%cst : f32) outs(%0 : tensor<512x512xf32>) -> tensor<512x512xf32>
+  %res = linalg.matmul ins(%lhs, %rhs: tensor<512x512xbf16>, tensor<512x512xbf16>)
+                    outs(%1: tensor<512x512xf32>) -> tensor<512x512xf32>
+  return %res : tensor<512x512xf32>
+}
+
+//   CHECK-LABEL: hal.executable.export public @matmul_bf16_dispatch_0_matmul_512x512x512_bf16xbf16xf32
+//         CHECK:    aie.device(npu1_4col)
+// CHECK-COUNT-3:    aie.shim_dma_allocation
```
newling, line 15: On the off chance you don't know: if you want confirm that there are _exactly_ 3, then this should be followed by a CHECK-NOT: aie.shim_dma_allocation

But I'm guessing you just want to confirm it compiles e2e.

newling (approved):
LGTM

yzhang93:
This is really nice @erwei-xilinx! Does it work e2e?

yzhang93 (approved):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Test/samples/matmul_pack_peel_4_level_tiling_air_e2e.mlir
erwei-xilinx: Yeah you are right. The goal is basically confirming that both the aie.device and the runtime sequence are successfully generated without any compilation failure in iree-amd-aie. We have some other tests in mlir-air which checks for the generation of shim_dma_allocation in greater details.

erwei-xilinx:
> This is really nice @erwei-xilinx! Does it work e2e?

Thanks! Currently it is passing for the same e2e test within mlir-air, but failed e2e in iree-amd-aie with some numerical error. Will take a closer look later.
 
> Could you please also check if this statement is still the case for AIR? https://github.com/nod-ai/iree-amd-aie/blob/main/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp#L583 Note that pack-peel-4-level-tiling is using the same outer perm as pack-peel pipeline for objectfifo.

Interesting. And I suspect that this could be the reason for the numerical error that I encountered? I will take a closer look at that.

yzhang93:
> Interesting. And I suspect that this could be the reason for the numerical error that I encountered? I will take a closer look at that.

Yes, I think the outer perm could be the reason for numerical error.

erwei-xilinx:
> > Interesting. And I suspect that this could be the reason for the numerical error that I encountered? I will take a closer look at that.
> 
> Yes, I think the outer perm could be the reason for numerical error.

I can confirm that this is the reason for the numerical error that I saw, and the mlir-air pipeline is able to pass e2e with this PR: https://github.com/nod-ai/iree-amd-aie/pull/1145

PR #1143, opened 2025-02-26
[Driver] `xrt-lite` support for device reconfiguration

Yu-Zhewen:
This PR introduces changes to kernel dispatch and flattenbuffer schemas to support device reconfiguration.

The steps of dispatch can now be described as:
1-1. load PDI;
1-2. get the NPU instructions and bindings;
1-3. run the original kernel;

2-1. get the NPU instructions for sending control packets;
2-2. get the sequence (content) of control packets;
2-3. execute reconfiguration;

3-1. get the NPU instructions and bindings for the new kernel;
3-2. run the new kernel, then loop back to 2-1 if needed.

In the case where the above process loops only once, it requires one PDI, three sets of NPU instructions (original kernel run, reconfiguration, new kernel run), plus one set of control packet sequences for the reconfiguration.


runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc
```diff
@@ -210,6 +189,97 @@ static iree_status_t iree_hal_xrt_lite_direct_command_buffer_dispatch(
...
+  IREE_RETURN_AND_END_ZONE_IF_ERROR(
+      z0, iree_hal_resource_set_insert(command_buffer->resource_set, 1,
+                                       &executable));
+  shim_xdna::hw_ctx context =
+      command_buffer->device->shim_device->create_hw_context(
+          kernel_params.pdi, kernel_params.kernel_name);
+  shim_xdna::hw_q* hwq = context.get_hw_queue();
+  shim_xdna::cuidx_t cu_idx =
+      context.open_cu_context(kernel_params.kernel_name);
+  // Dispatch the kernel.
```
jtuyls, line 258: High-level question, why is this first kernel dispatch/run needed?

Specifically:
>1-1. load PDI;
1-2. get the NPU instructions and bindings;
1-3. run the original kernel;

I don't see why 1-2 and 1-3 are done? Can we just do: load PDI -> reconfigure -> run ->reconfigure -> run -> etc? This would avoid a special first run.

Yu-Zhewen: It's mainly to accommodate all existing tests where no reconfiguration is present. Even if we wanted to switch all tests to the reconfiguration style in the future, it could impact performance and is not feasible for circuit flow tests due to routing resource constraints.

Yu-Zhewen: Another reason is that I made the driver support multiple reconfigurations, but the codegen currently only supports a single reconfiguration.

One blocker is that I'm unsure if passing control packets through file paths is a good approach. If we have 10 kernels, we would need to run iree-compile 10 times and specify 10 different file paths, which isn't ideal.

Example: #1147

jtuyls: > It's mainly to accommodate all existing tests where no reconfiguration is present.

Yeah, that makes sense.

> Even if we wanted to switch all tests to the reconfiguration style in the future, it could impact performance.

It should hopefully be a lot faster than loading the PDI (we're probably quite a bit away from that though).

>  is not feasible for circuit flow tests due to routing.

You could even use pkt flows to program most of the array and then use NPU instructions to override the pkt flow routes to circuit flow. Again, we're pretty far away from that as well.

> In my current approach, control packets are passed through file paths.

We can pass them through the flatbuffer instead?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -507,28 +543,70 @@ LogicalResult AIETargetBackend::serializeExecutable(
...
+    FailureOr<std::vector<uint32_t>> npuInstrs =
+        loadUInt32ArrayFromFile(npuInstPath);
+    if (failed(npuInstrs)) return failure();
+    asmInstrsRunlist.push_back(builder.createInt32Vec(npuInstrs.value()));
     asmInstrIndices[ordinal] = asmInstrRefs.size();
+    // TODO (zhewen): support multiple times of reconfiguration.
+    if (!options.pathToLoadCtrlPktFiles.empty()) {
+      // Load control packet instructions from file.
+      SmallString<128> ctrlpktInstPath(options.pathToLoadCtrlPktFiles);
+      llvm::sys::path::append(ctrlpktInstPath, "ctrlpkt_instructions.txt");
```
jtuyls, line 558: I think you could do this in the same way as the NPU instructions: `entryPointNamesFb[ordinal] + ".ctrlpkt_instructions.txt"`?

runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc
jtuyls: Also, can't we pass a boolean/enum through the flatbuffer to indicate whether this would be a normal run or a reconfigure run? Then we don't need to do the first run in case of reconfiguration.

Yu-Zhewen: > We can pass them through the flatbuffer instead?

Yes, they are already, but they are firstly dumped to files and then added to the flatbuffer.
```
iree-compile second_kernel.mlir >> second_kernel/ctrlpkts.txt
iree-compile first_kernel.mlir  << second_kernel/ctrlpkts.txt  >> all_kernels.vmfb
```
(the compilation order is also weird...)

I am not sure what's the current status for multi-dispatch?

runtime/src/iree-amd-aie/driver/xrt-lite/executable.cc
```diff
@@ -138,16 +138,106 @@ iree_amd_aie_hal_xrt_lite_native_executable_flatbuffer_verify(
...
+      iree_amd_aie_hal_xrt_lite_ExecutableDef_asm_instr_runlists_get(
+          executable_def);
+  iree_amd_aie_hal_xrt_lite_UI32Array2dDef_vec_t reconf_data_runlists_vec =
+      iree_amd_aie_hal_xrt_lite_ExecutableDef_reconf_data_runlists_get(
+          executable_def);
+  for (size_t i = 0; i < entry_point_count; ++i) {
+    int32_t reconf_data_runlist_index =
+        flatbuffers_int32_vec_at(reconf_data_runlist_indices_vec, i);
+    if (reconf_data_runlist_index >= 0) {
+      // Get ther number of reconfiguration data for the current entry point.
```
jtuyls, line 184:
```suggestion
      // Get the number of reconfiguration data for the current entry point.
```

runtime/src/iree-amd-aie/schemas/pdi_executable_def.fbs
```diff
@@ -15,41 +15,68 @@ table FileLineLocDef {
...
+  // - The type is `int32` instead of `uint32` to allow `-1` as a special value.
+  // - If an index is `-1`, it indicates that no reconfiguration is required for the corresponding entry point.
+  //   This is a workaround, as `reconf_data_runlists` does not support null elements.
+  reconf_data_runlist_indices:[int32];
+
+  // Device reconfiguration data.
+  // This forms a 3D uint32 array:
+  //   - The first dimension corresponds to entry points that require reconfiguration.
+  //     - Its size is smaller than `entry_points` because not all entry points require reconfiguration.
+  //     - Null elements are not supported in flattenbuffer, so only entries with reconfiguration are included.
```
jtuyls, line 76:
```suggestion
  //     - Null elements are not supported in flatbuffer, so only entries with reconfiguration are included.
```

Yu-Zhewen:
> Also, can't we pass a boolean/enum through the flatbuffer to indicate whether this would be a normal run or a reconfigure run? Then we don't need to do the first run in case of reconfiguration.

To determine whether reconfiguration is needed, we can simply check if `reconf_data_runlist` has a size of 0. 
https://github.com/nod-ai/iree-amd-aie/blob/cc9c7fa33d07bdb229a0cc4a0e9379197c9eb5f6/runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc#L263

However, we might want to introduce a different flag to explicitly control whether the first run is optional, rather than inferring it from whether reconfiguration is needed. This probably leaves us more room for future tuning.

Yu-Zhewen:
> > > Also, can't we pass a boolean/enum through the flatbuffer to indicate whether this would be a normal run or a reconfigure run? Then we don't need to do the first run in case of reconfiguration.
> > 
> > 
> > To determine whether reconfiguration is needed, we can simply check if `reconf_data_runlist` has a size of 0.
> > https://github.com/nod-ai/iree-amd-aie/blob/cc9c7fa33d07bdb229a0cc4a0e9379197c9eb5f6/runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc#L263
> > 
> > However, we might want to introduce a different flag to directly control whether the first run is optional, rather than inferring it from the reconfiguration condition. This probably leaves us more room for future tuning.
> 
> Yeah, this is kind of what I mean, but I don't see a need for doing this first run and then reconfiguring. So, for me it would be either doing a normal run or one with reconfiguring. In the latter, I don't see why we would do an (optional) first run.

I would assume both the PDI loading time and the ctrlpkt reconfiguration time include a constant overhead plus a linear (ideally) component that scales with the program size. If that's the case, and since we're already using PDI for control packet overlays anyway, it might be worth shipping the first-run kernel with the PDI (to overcome the constant cost of another reconfiguration). For now, I think it’s best to leave the option open so that we can easily investigate different options in the future?

jtuyls:
> > > > Also, can't we pass a boolean/enum through the flatbuffer to indicate whether this would be a normal run or a reconfigure run? Then we don't need to do the first run in case of reconfiguration.
> > > 
> > > 
> > > To determine whether reconfiguration is needed, we can simply check if `reconf_data_runlist` has a size of 0.
> > > https://github.com/nod-ai/iree-amd-aie/blob/cc9c7fa33d07bdb229a0cc4a0e9379197c9eb5f6/runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc#L263
> > > 
> > > However, we might want to introduce a different flag to directly control whether the first run is optional, rather than inferring it from the reconfiguration condition. This probably leaves us more room for future tuning.
> > 
> > 
> > Yeah, this is kind of what I mean, but I don't see a need for doing this first run and then reconfiguring. So, for me it would be either doing a normal run or one with reconfiguring. In the latter, I don't see why we would do an (optional) first run.
> 
> I would assume both the PDI loading time and the ctrlpkt reconfiguration time include a constant overhead plus a linear (ideally) component that scales with the program size. If that's the case, and since we're already using PDI for control packet overlays anyway, it might be worth shipping the first-run kernel with the PDI (to overcome the constant cost of another reconfiguration). For now, I think it’s best to leave the option open so that we can easily investigate different options in the future?

I think that's more complicated than what we need. What I am thinking that we will end up with is a single, largely empty PDI, that we load once per model and then a control packet - NPU instruction combination per dispatch. This way we never have the reload the PDI, and it can even be done when loading the model, so that it doesn't count towards the total time anyway.

Yu-Zhewen:
> > > > > Also, can't we pass a boolean/enum through the flatbuffer to indicate whether this would be a normal run or a reconfigure run? Then we don't need to do the first run in case of reconfiguration.
> > > > 
> > > > 
> > > > To determine whether reconfiguration is needed, we can simply check if `reconf_data_runlist` has a size of 0.
> > > > https://github.com/nod-ai/iree-amd-aie/blob/cc9c7fa33d07bdb229a0cc4a0e9379197c9eb5f6/runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc#L263
> > > > 
> > > > However, we might want to introduce a different flag to directly control whether the first run is optional, rather than inferring it from the reconfiguration condition. This probably leaves us more room for future tuning.
> > > 
> > > 
> > > Yeah, this is kind of what I mean, but I don't see a need for doing this first run and then reconfiguring. So, for me it would be either doing a normal run or one with reconfiguring. In the latter, I don't see why we would do an (optional) first run.
> > 
> > 
> > I would assume both the PDI loading time and the ctrlpkt reconfiguration time include a constant overhead plus a linear (ideally) component that scales with the program size. If that's the case, and since we're already using PDI for control packet overlays anyway, it might be worth shipping the first-run kernel with the PDI (to overcome the constant cost of another reconfiguration). For now, I think it’s best to leave the option open so that we can easily investigate different options in the future?
> 
> I think that's more complicated than what we need. What I am thinking that we will end up with is a single, largely empty PDI, that we load once per model and then a control packet - NPU instruction combination per dispatch. This way we never have the reload the PDI, and it can even be done when loading the model, so that it doesn't count towards the total time anyway.

I see, make sense

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
Yu-Zhewen: Done 
https://github.com/nod-ai/iree-amd-aie/blob/f59410c2f242e12a816816fb8da07540714709fe/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp#L480-L493

Yu-Zhewen:
> > I think that's more complicated than what we need. What I am thinking that we will end up with is a single, largely empty PDI, that we load once per model and then a control packet - NPU instruction combination per dispatch. This way we never have the reload the PDI, and it can even be done when loading the model, so that it doesn't count towards the total time anyway.

Removed the first kernel run in the case of reconfiguration.

https://github.com/nod-ai/iree-amd-aie/blob/f59410c2f242e12a816816fb8da07540714709fe/runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc#L259-L282

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -507,28 +560,71 @@ LogicalResult AIETargetBackend::serializeExecutable(
...
-        return moduleOp.emitOpError("Unable to parse instruction file");
-      }
-      npuInstrs.push_back(a);
-    }
-    flatbuffers_int32_vec_ref_t npuInstrsVec =
-        builder.createInt32Vec(npuInstrs);
+    SmallVector<flatbuffers_int32_vec_ref_t> asmInstrsRunlist;
+    SmallVector<flatbuffers_int32_vec_ref_t> reconfDataRunlist;
     asmInstrIndices[ordinal] = asmInstrRefs.size();
+    // TODO (zhewen): support multiple times of reconfiguration.
```
jtuyls, line 566:
```suggestion
    // TODO (zhewen): support multiple iterations of reconfiguration.
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.cpp
```diff
@@ -507,28 +560,71 @@ LogicalResult AIETargetBackend::serializeExecutable(
...
+    if (!options.dirToLoadCtrlPktFiles.empty()) {
```
jtuyls, line 567: Why not use `entryPointWorkDir`?

runtime/src/iree-amd-aie/schemas/pdi_executable_def.fbs
```diff
@@ -15,41 +15,68 @@ table FileLineLocDef {
...
+  // Assembly instructions for the LX6 processor to execute.
+  // This is a 3D uint32 array:
+  //   - The first dimension corresponds to different entry points (same size as `entry_points`).
+  //   - The second dimension represents the number of kernel runs per entry point.
+  //     Its size is either 1 (for a standard run) or `2 * num_reconfiguration` (if reconfiguration is required).
+  //     In the case of reconfiguration:
+  //       - Each reconfiguration consists of two kernel runs:
+  //         - One for the configuration.
+  //         - One for the actual execution after reconfiguration.
+  //   - The third dimension is a uint32 array containing the instruction stream for a single kernel run.
```
jtuyls, line 61: Isn't it a 2D array?

runtime/src/iree-amd-aie/schemas/pdi_executable_def.fbs
```diff
@@ -15,41 +15,68 @@ table FileLineLocDef {
...
+
+  // A map of entry point ordinals to the indices of the containing `reconf_data_runlists` (the following field).
+  // This list has the same size as the `entry_points` list.
+  // - The type is `int32` instead of `uint32` to allow `-1` as a special value.
+  // - If an index is `-1`, it indicates that no reconfiguration is required for the corresponding entry point.
+  //   This is a workaround, as `reconf_data_runlists` does not support null elements.
+  reconf_data_runlist_indices:[int32];
+
+  // Device reconfiguration data.
+  // This forms a 3D uint32 array:
```
jtuyls, line 73: Isn't it 2D?

runtime/src/iree-amd-aie/driver/xrt-lite/executable.cc
```diff
@@ -239,16 +335,32 @@ iree_status_t iree_hal_xrt_lite_native_executable_create(
...
+    uint32_t asm_instr_runlist_index =
+        flatbuffers_uint32_vec_at(asm_instr_runlist_indices_vec, entry_ordinal);
+    iree_amd_aie_hal_xrt_lite_UI32Array2dDef_table_t asm_inst_runlist_def =
+        iree_amd_aie_hal_xrt_lite_UI32Array2dDef_vec_at(
+            asm_instr_runlists_vec, asm_instr_runlist_index);
+    IREE_RETURN_AND_END_ZONE_IF_ERROR(
+        z0, iree_amd_aie_hal_xrt_lite_executable_parse_UI32Array2dDef(
+                asm_inst_runlist_def, params->asm_inst_runlist));
+
+    // Get the reconfiguratoin data runlist for the current entry point, and
```
jtuyls, line 350:
```suggestion
    // Get the reconfiguration data runlist for the current entry point, and
```

runtime/src/iree-amd-aie/schemas/pdi_executable_def.fbs
jtuyls: Right, I got confused with the `UI32Array2dDef`, but it's `[UI32Array2dDef]`

jtuyls (approved):
LGTM

PR #1144, opened 2025-02-26
[CanonicalizeDoublyStridedOp] Add expansion of dimensions exceeding max sizes

### LINKED ISSUE #1137: ALL ZEROES| 4x2 | 32x128x128 | i32->i32 | Strix
Repro:
```
aie_compilation_flags = [
            "--iree-amdaie-num-rows=4",
            "--iree-amdaie-num-cols=2",
        ]
self.register(
                Matmul(
                    32,
                    128,
                    128,
                    "i32",
                    "i32",
                    test_params=TestParams(
                        tile_pipeline="pack-peel-4-level-tiling",
                        run_on_target=["npu4"],
                        aie_compilation_flags=aie_compilation_flags,
                    ),
                )
            )
```

Issue: Results in ALL ZEROES

[e2e IR log for i32->i32](https://gist.github.com/Abhishek-Varma/72a4dee71808f79134928eb11379fa4a)

Even for i8->i32 it results in ALL ZEROES (after changing stack size [here](https://github.com/nod-ai/iree-amd-aie/blob/6eae2e1d6b7c2ceeb06e3956e3c4393a61603410/compiler/plugins/target/AMD-AIE/aie/AIEOps.td#L128) from `0x400` to `0x800`)

jtuyls:
This PR adds a dimension expansion rewrite to the doubly-strided op canonicalizer, which is effectively the reverse of the `FoldDmaOpLinearDims` transformation. This fixes: #1137 (see test being added).

Example:

```
offsets: [0], sizes: [8], strides: [1], maxSizes: [5]
```

is expanded into:

```
offsets: [0, 0], sizes: [2, 4], strides: [4, 1]
```

Here, all the sizes are smaller or equal to the max size. This helps with enabling more DMA composition transformations, and this is needed to avoid nested `npu.circular_dma_cpy_nd` ops after `DmaComposition`, which is why it fixes #1137. Note that we should probably add some checks that all `npu.circular_dma_cpy_nd`  ops are combined and subsumed correctly to help with debugging issues. I will address that in a follow-up.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
@@ -136,6 +136,18 @@ std::optional<int64_t> getGlobalOffsetDifference(
 
   return globalOffsetDifference;
 }
+
+int64_t findFactorResultingInSmallerSize(int64_t size, int64_t maxSize) {
+  assert(size >= 0 && "size must be non-negative");
+  assert(maxSize >= 0 && "maxSize must be non-negative");
+  if (size == 0 || maxSize == 0 || size == 1 || maxSize == 1) return 1;
+  if (size <= maxSize) return 1;
+  for (int64_t i = 2; i <= size / 2; i++) {
+    if (size % i == 0 && size / i <= maxSize) return i;
+  }
+  return size;
+}
+
 }  // namespace detail
 
 namespace {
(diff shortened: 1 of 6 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
```diff
@@ -292,6 +304,60 @@ LogicalResult combineAccessPatterns(
...
+    const SmallVector<OpFoldResult> &sizes,
+    const SmallVector<OpFoldResult> &strides,
+    SmallVector<OpFoldResult> &newOffsets, SmallVector<OpFoldResult> &newSizes,
+    SmallVector<OpFoldResult> &newStrides, ArrayRef<int64_t> maxSizes) {
+  assert(offsets.size() == sizes.size() && offsets.size() == strides.size() &&
+         "expected same number of offsets, sizes and strides");
+  assert(maxSizes.size() >= sizes.size() &&
+         "expected `maxSizes` to have at least as many elements as `sizes`");
+  bool expandableLinearDimsFound = false;
+  if (offsets.size() == 0) return success(expandableLinearDimsFound);
```
yzhang93, line 318: why not directly return success()?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
```diff
@@ -292,6 +304,60 @@ LogicalResult combineAccessPatterns(
   return success();
 }
 
+LogicalResult expandLargeDimIntoLinearDims(
```
yzhang93, line 307: It's better to add some documents for this function.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/AMDAIEDmaUtilsTest.cpp
```diff
@@ -652,6 +719,36 @@ TEST_F(FoldTest, FoldRepetitionCountFail) {
...
+TEST_F(FoldAndExpandTest, NoDimsExpand) {
+  checkExpandDimIntoLinearDims({}, {}, {}, {}, {}, {}, {}, false);
+  checkExpandDimIntoLinearDims({0}, {1}, {1}, {1}, {0}, {1}, {1}, false);
+  checkExpandDimIntoLinearDims({0}, {8}, {1}, {8}, {0}, {8}, {1}, false);
+  checkExpandDimIntoLinearDims({0}, {11}, {1}, {7}, {0}, {11}, {1}, false);
+  checkExpandDimIntoLinearDims({0, 8}, {16, 8}, {16, 1}, {16, 8}, {0, 8},
+                               {16, 8}, {16, 1}, false);
+}
+
+TEST_F(FoldAndExpandTest, ExpandLargeDim) {
```
yzhang93, line 731: Nice test cases!

yzhang93 (approved):
This is great! My only question is how did it happen with a dimension size larger than the max size? I thought there's max size check for all dma canonicalization patterns.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
jtuyls: Because no expansion took place, so that the user can avoid/abort a rewrite if not needed. This function is written in the same way as `foldLinearDims` and it's documented, but we could reconsider how we return that info:

> /// Expand dimensions within a strided access pattern that exceed a maximum
/// size. Returns `success` if expansion took place. The `maxSizes` array is
/// expected to be provided in the same order as `sizes` and they are compared
/// from right to left (innermost to outermost).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.cpp
jtuyls: Done

jtuyls:
> This is great! My only question is how did it happen with a dimension size larger than the max size? I thought there's max size check for all dma canonicalization patterns.

Here, it seemed to have been caused by `CombineStridedOps` + some dimension can have any size (changed that in this PR as well) and then they block further composition transformations. But it can also happen that the initial value of a dimension is too large, so that's why a more general solution is useful I think. But `CombineStridedOps` might indeed still have a bug as well, which I will investigate further. The problem is that the rules are quite cumbersome (different max size, dimensions etc), making these composition transformations hard to get right. This likely needs some further thought to improve the rule checks as a whole.

PR #1145, opened 2025-02-26
[MLIR-AIR] Apply `isObjectFifo` flag to the pack-peel-4-level-tiling pipeline

erwei-xilinx:
- This is needed to get numerically correct results for mlir-air pipeline when using `pack-peel-4-level-tiling` scheme.


yzhang93:
I've also seen the flaky failure on linux ci runner compiling third_party/mlir-air today. Could you please double check to see what caused the problem?

erwei-xilinx:
> I've also seen the flaky failure on linux ci runner compiling third_party/mlir-air today. Could you please double check to see what caused the problem?

`clang: error: clang frontend command failed with exit code 135 (use -v to see invocation)` seems to be the issue causing the failures. 135 is segfault. But it seems that the segfault is happening at varying places for different CI runs... The same issue hasn't occurred on the CI machine of mlir-air before.

yzhang93:
> > I've also seen the flaky failure on linux ci runner compiling third_party/mlir-air today. Could you please double check to see what caused the problem?
> 
> `clang: error: clang frontend command failed with exit code 135 (use -v to see invocation)` seems to be the issue causing the failures. 135 is segfault. But it seems that the segfault is happening at varying places for different CI runs... The same issue hasn't occurred on the CI machine of mlir-air before.

A rerun helps, but it did happen very frequently today on several PRs...

erwei-xilinx:
> A rerun helps, but it did happen very frequently today on several PRs...

I can try to go to mlir-air repo and see if I can remove some redundant code, to reduce the codebase size.

yzhang93:
> > A rerun helps, but it did happen very frequently today on several PRs...
> 
> I can try to go to mlir-air repo and see if I can remove some redundant code, to reduce the codebase size.

Thanks. You think it's OOM? Did you add a lot of codes compared to your previous bump?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -365,12 +365,15 @@ static SmallVector<int64_t> setOuterPermB(bool isMatmulTransposeB,
...
+    LowerToAIEPassPipeline useLowerToAIEPipeline, AMDAIEDevice targetDevice,
+    uint32_t numRows, uint32_t numCols) {
   // Scale the L1 K with a factor of 2 compared with the outer dimensions M and
   // N to increase the L1 memory usage.
+  bool isObjectFifo =
+      useLowerToAIEPipeline == LowerToAIEPassPipeline::ObjectFifo;
   auto maybePackPeelTiling =
-      ParameterSetting::create(linalgOp, /*isObjectFifo=*/true, targetDevice,
-                               numRows, numCols, /*kPackScaleL1=*/2);
+      ParameterSetting::create(linalgOp, isObjectFifo, targetDevice, numRows,
```
yzhang93, line 375: Are you sure this is enough to get correct numerics? You have to modify the outer perm similar to this https://github.com/nod-ai/iree-amd-aie/blob/8ded4b1a84445cf5367c5db8f41435659c2dc941/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp#L582 inside `setRootConfigForPackPeel4LevelTilingPipeline`.

yzhang93: If you like you can add a test in ci to make sure it's running.

erwei-xilinx:
> Thanks. You think it's OOM? Did you add a lot of codes compared to your previous bump?

Yeah I suspect it is OOM, as I notice that the segfault happens when more than one CI gets triggered at the same time. I didn't add a lot of codes recently, but the code base has been pretty big. AIRDependencyScheduleOpt.cpp for example is >5000 lines...

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
erwei-xilinx: Oh cool, that would be great! Will add a matmul test to make sure.

newling:
> clang: error: clang frontend command failed with exit code 135 (use -v to see invocation)

FWIW I saw this last week once, I'd never seen it before (ever ever).

yzhang93 (approved):

newling:
> > > clang: error: clang frontend command failed with exit code 135 (use -v to see invocation)
> > 
> > 
> > FWIW I saw this last week once, I'd never seen it before (ever ever).
> 
> Last week that's before mlir-air bump? 

I think so.  

> You saw the same linux build failure? 

Yes definitely saw this error message. I remember thinking `clang frontend command failed` sounding really bizarre. 

> We probably need to check the RAM on the runner?

I think I saw it on nuc50. It sounds like Erwei's hypothesis is that the memory requirement from AIR has been creeping up gradually, so this failure just becomes more and more likely. I think the best thing to do at this point, if we keep seeing this error, is to try to reduce C++ translation unit sizes.

PR #1146, opened 2025-02-27
Pass to add llvm annotations to avoid inlining (#2)

newling:
This is the same as https://github.com/nod-ai/iree-amd-aie/pull/1142 but it includes some extra llvm opt passes to cse-away some scalar operations. I'll add some more details if CI confirms this PR matches parity with our current main branch


newling:
Abandoning in favour of https://github.com/nod-ai/iree-amd-aie/pull/1149/files

PR #1147, opened 2025-02-27
[CtrlPkt] Reconfiguration example for Matmul with constant bias

Yu-Zhewen:
This PR introduces a proof of concept for control packet reconfiguration using a matmul operation with a constant bias. This example is chosen because the differences between kernels are minimal, only the constant bias varies, while other aspects, such as DMA connections and the rest of the core code, remain unchanged.

This test is currently enabled on `xrt-lite` (linux) only. Changes on `xrt` (windows) will be followed in another PR.


build_tools/ci/cpu_comparison/run.py
```diff
@@ -614,6 +614,139 @@ def _execute(self, config):
         return self.vs_cpu(config)
 
 
+class MatmulConstBiasCtrlpkt(BaseMatmul):
+    """
+    A test of the form matmul(A,B) where A:MxK, B:KxN
```
yzhang93, line 619: Modify this description.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -614,6 +614,139 @@ def _execute(self, config):
...
+            N=N,
+            K=K,
+            input_type=input_type,
+            acc_type=acc_type,
+            n_kernel_runs=n_kernel_runs,
+        )
+        self.labels.append("MatmulConstBiasCtrlPacket")
+        self.aie_compilation_flags += [
+            "--iree-amdaie-num-rows=1",
+            "--iree-amdaie-num-cols=1",
```
yzhang93, line 646: Add some comments on why 1 row and 1 col is needed? Will it work for multi-cores eventually?

build_tools/ci/cpu_comparison/run.py
```diff
@@ -614,6 +614,139 @@ def _execute(self, config):
...
+    def _execute(self, config):
+        # TODO (zhewen): Generalize this test flow.
+
+        matmul_template_dir = config.file_dir / "matmul_template"
+        template_name = matmul_template_dir / "matmul_constant_bias_MxK_KxN.mlir"
+        name_copy = self.name
+
+        # Compile the test file with an initial constant bias of 2.
+        # This will also generate the control packets needed for runtime reconfiguration.
+        self.name = name_copy + "_2"
```
yzhang93, line 669: Maybe good to name with "_bias_2".

build_tools/ci/cpu_comparison/run.py
```diff
@@ -614,6 +614,139 @@ def _execute(self, config):
...
+        test_file_1 = self.get_filename(config)
+        test_name_1 = name_from_mlir_filename(test_file_1)
+        aie_vmfb_1 = generate_aie_vmfb(
+            config,
+            self.aie_compilation_flags + aie_ctrlpkt_flags,
+            test_name_1,
+            self.tile_pipeline,
+            self.lower_to_aie_pipeline,
+            self.use_ukernel,
+            test_file_1,
```
yzhang93, line 700: I think this and the test above are better to be added as instances instead of directly in the class.

yzhang93 (changes requested):
Could you please summarize why the constant bias is needed?

Yu-Zhewen:
> I think this and the test above are better to be added as instances instead of directly in the class.

Updated the PR description

build_tools/ci/cpu_comparison/run.py
Yu-Zhewen: Done

build_tools/ci/cpu_comparison/run.py
Yu-Zhewen: Added a TODO. Only tested the single core case so far, I will test the whole array as the next step.

build_tools/ci/cpu_comparison/run.py
Yu-Zhewen: Thanks. Now wrapped into this new function
https://github.com/nod-ai/iree-amd-aie/blob/1b3516cf77f400cca4f2b7aa1bcf1df72ad300bb/build_tools/ci/cpu_comparison/run.py#L662

yzhang93 (approved):
LGTM, thanks! Another question, does the bias have to be constant?

Yu-Zhewen:
> LGTM, thanks! Another question, does the bias have to be constant?

Reconfiguring a constant is just the first step. The final goal is to modify the entire core code freely, allowing changes like switching between different sizes, ops, etc.

PR #1149, opened 2025-02-27
Add CSE to aie-opt pipeline to improve -O2 performance

newling:
This PR extends the aie-opt pipeline with Common Subexpression Elimination (CSE) to achieve better scalar arithmetic without increasing Program Memory (PM) usage.

Background:
- Currently, -O3 optimization provides better performance than -O2  (function outlining, phoenix, and matmul). 
- However, -O3 significantly increases program memory usage compared to -O2.
- The performance improvement in -O3 is largely due to unintended effects of function inlining, which rearranges integer arithmetic and reduces common subexpressions.

Problem:
- In -O2, the outlined function after optimization contains redundant operations, e.g.:

%45 = trunc i64 %38 to i20
...
%49 = trunc i64 %38 to i20

- -O3 eliminates these redundancies indirectly through inlining, but at the cost of increased memory usage.

Solution:
- Add a basic LLVM CSE pass after the -O* pass pipeline.
- This approach achieves most/all of the performance benefits of -O3 without the memory overhead.

Implementation:
- Extended aie-opt pipeline with CSE.
- Considered adding Global Value Numbering (GVN) but found it merged vector loads, causing register spills and performance degradation.

References:
- Early CSE information: https://llvm.org/doxygen/structllvm_1_1EarlyCSEPass.html#details

Below I've scraped all the benchmark numbers from CI, for this PR and top of main (ToM). Table summary: it looks like we can match the performance of -O3 (line 4) now with -O2 (line 2), but with much less PM usage (8474 bytes -> 2608 bytes). 

### Time Comparison
| Benchmark | Before (us) | After (us) |
| --- | --- | --- |
| matmul_512_512_4096_bf16_f32_O2_npu1_4col | 2585.0 | 1817.0 |
| matmul_512_512_4096_bf16_f32_O2_npu1_4col_outline | 2714.0 | **1807.0** |
| matmul_512_512_4096_bf16_f32_O3_npu1_4col | 2625.0 | 1801.0 |
| matmul_512_512_4096_bf16_f32_O3_npu1_4col_outline | **1775.0** | 1804.0 |
| matmul_512_512_4096_bf16_f32_O3_npu1_4col_outline_ukernel_chess | 1771.0 | 1832.0 |

### Memory Comparison
| Benchmark | Before (bytes) | After (bytes) |
| --- | --- | --- |
| matmul_512_512_4096_bf16_f32_O2_npu1_4col | 10112 | 8800 |
| matmul_512_512_4096_bf16_f32_O2_npu1_4col_outline | 2784 | **2608** |
| matmul_512_512_4096_bf16_f32_O3_npu1_4col | 10112 | 8800 |
| matmul_512_512_4096_bf16_f32_O3_npu1_4col_outline | **8464** | 8464 |
| matmul_512_512_4096_bf16_f32_O3_npu1_4col_outline_ukernel_chess | 3328 | 3328 |

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
@@ -248,39 +248,13 @@ FailureOr<std::vector<std::string>> makePeanoOptArgs(
     }
   }
 
-  // Some background to this opt pass finetuning/hack.
-  // currently performance (function outlining, phoenix, matmul) is much better
-  // with -O3 than -O2. But the memory used with -O3 is much higher than with
-  // -O2. This is because with -O3, the outlined function is inlined at most
-  // call sites and in the the process of inlining for some reason the integer
-  // arithmetic is rearranged, and in particular there are fewer common
-  // sub-expressions. For example in the -O2 case, the outlined function after
-  // optimization contains
-  //
-  //    ```
-  //    ...
-  //    %45 = trunc i64 %38 to i20
-  //    ...
-  //    %49 = trunc i64 %38 to i20
-  //    ...
-  //    ```
-  //
-  // whereas the inlined (with -O3) does not. So inlining has found a roundabout
-  // way to improve the code which has nothing to do with any actual traditional
-  // advantages of inlining (fewer function calls etc).
-  //
-  // Can we get this improvement without the memory overhead of -O3? I have
-  // found that running the -O2 pass pipeline, followed by a basic llvm cse pass
-  // gets us almost all the way there.
-  //
-  // early-cse information:
-  //    https://llvm.org/doxygen/structllvm_1_1EarlyCSEPass.html#details
-  //
-  // I have experimented with adding gvn too (gvn is a more powerful version of
-  // cse), but it results in vector loads being merged, which results in spills
-  // from registers, which results in performance degredation.
-  //
-  for (auto &flag : args) {
+  // Adding cse after the default O2 pipeline eliminates repeated
+  // ```
+  // %49 = trunc i64 %38 to i20
+  // ```
+  // for certain matmuls (outlining, phoenix), and results in dramatic
+  // improvements in performance.
+  for (std::string &flag : args) {
     if (isOptLevelFlag(flag)) {
       auto optLevel = flag.substr(1);
       auto passes = "default<" + optLevel + ">,early-cse,dce";
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
@@ -247,6 +247,46 @@ FailureOr<std::vector<std::string>> makePeanoOptArgs(
       *iter = flag;
     }
   }
+
+  // Some background to this opt pass finetuning/hack.
+  // currently performance (function outlining, phoenix, matmul) is much better
+  // with -O3 than -O2. But the memory used with -O3 is much higher than with
+  // -O2. This is because with -O3, the outlined function is inlined at most
+  // call sites and in the the process of inlining for some reason the integer
+  // arithmetic is rearranged, and in particular there are fewer common
+  // sub-expressions. For example in the -O2 case, the outlined function after
+  // optimization contains
+  //
+  //    ```
+  //    ...
+  //    %45 = trunc i64 %38 to i20
+  //    ...
+  //    %49 = trunc i64 %38 to i20
+  //    ...
+  //    ```
+  //
+  // whereas the inlined (with -O3) does not. So inlining has found a roundabout
+  // way to improve the code which has nothing to do with any actual traditional
+  // advantages of inlining (fewer function calls etc).
+  //
+  // Can we get this improvement without the memory overhead of -O3? I have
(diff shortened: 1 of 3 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -247,6 +247,46 @@ FailureOr<std::vector<std::string>> makePeanoOptArgs(
       *iter = flag;
     }
   }
+
+  // Some background to this opt pass finetuning/hack.
+  // currently performance (function outlining, phoenix, matmul) is much better
+  // with -O3 than -O2. But the memory used with -O3 is much higher than with
+  // -O2. This is because with -O3, the outlined function is inlined at most
+  // call sites and in the the process of inlining for some reason the integer
```
yzhang93, line 255:
```suggestion
  // call sites and in the process of inlining for some reason the integer
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -247,6 +247,46 @@ FailureOr<std::vector<std::string>> makePeanoOptArgs(
...
+  // found that running the -O2 pass pipeline, followed by a basic llvm cse pass
+  // gets us almost all the way there.
+  //
+  // early-cse information:
+  //    https://llvm.org/doxygen/structllvm_1_1EarlyCSEPass.html#details
+  //
+  // I have experimented with adding gvn too (gvn is a more powerful version of
+  // cse), but it results in vector loads being merged, which results in spills
+  // from registers, which results in performance degredation.
+  //
```
yzhang93, line 282: Nit: It's better to make the document more concise.

yzhang93 (approved):

yzhang93:
One more request: could you truncate the table when you squash and merge the PR?

newling:
> One more request: could you truncate the table when you squash and merge the PR?

Sure, but I'm not sure why? Are you worried it'll look bad when you do `git log`? Because if so it's not as bad as PRs like https://github.com/nod-ai/iree-amd-aie/pull/1115 which look really untidy in terminal

newling:
> This is great! Have you tried to compare O2 and O3 results using other shapes?

I haven't, no

yzhang93:
> > One more request: could you truncate the table when you squash and merge the PR?
> 
> Sure, but I'm not sure why? Are you worried it'll look bad when you do `git log`? Because if so it's not as bad as PRs like #1115 which look really untidy in terminal

Yes, the log would be quite long. Besides I think the first 5 rows are enough to demonstrate your contribution.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -247,6 +247,46 @@ FailureOr<std::vector<std::string>> makePeanoOptArgs(
...
+  for (auto &flag : args) {
```
jtuyls, line 283: Nit: could you use explicit types here?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
```diff
@@ -247,6 +247,46 @@ FailureOr<std::vector<std::string>> makePeanoOptArgs(
...
+  //    %49 = trunc i64 %38 to i20
+  //    ...
+  //    ```
+  //
+  // whereas the inlined (with -O3) does not. So inlining has found a roundabout
+  // way to improve the code which has nothing to do with any actual traditional
+  // advantages of inlining (fewer function calls etc).
+  //
+  // Can we get this improvement without the memory overhead of -O3? I have
+  // found that running the -O2 pass pipeline, followed by a basic llvm cse pass
```
jtuyls, line 273: I am likely missing something, but where exactly are we changing `O3` to `O2`? I still see `O3` here: https://github.com/nod-ai/iree-amd-aie/blob/d79bb60b662d8f1d79e5bbd63b673b9e401eee13/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp#L215

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
newling: O3 is still the default, I haven't changed that in this PR. This PR just makes O2 better. There are tests in run.py which use O2 by explicitly opting in for it. 

I could make O2 the default in this PR, but I'd rather do it in a separate PR.

jtuyls: Ok, I see, I thought that the point of this PR was to make a case to switch to O2.

jtuyls (approved):
LGTM

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp
newling: On second thoughts, a test isn't really feasible -- it would be string check on that relies on peano's opt pass not changing too much. So I think I'll keep the testing in this PR as is

PR #1151, opened 2025-02-27
[CI] Logging tidy up 

newling:
This PR attempts to make the logging generated via run.py (mainly coming from aie2xclbin) tidier, by

1) controlling verbosity (don't print the same logging 16 times when the same program is run for each core)
2) better spacing (where to \n)
3) fix the time logging (it was printing time 0.000 [s] for every step of compilation)
4) remove mlir warning 

p.s. I think adding a test that the logging 'looks good' isn't worth the effort at this point, so you'll just have to trust me that it's nicer!


jtuyls (approved):
LGTM

PR #1153, opened 2025-02-28
Make O2 the default optimization level for aie-opt 

newling:


yzhang93:
Maybe leave this test `matmul_512_4096_512_bf16_f32_npu1_4col_outline_4_level_tiling` using O3 for now?

newling:
> Maybe leave this test `matmul_512_4096_512_bf16_f32_npu1_4col_outline_4_level_tiling` using O3 for now?

I think the O3 code is better than O2 for all the tests, it's just that this is only test where the data movement is fast enough for it show up. So I'm going to dig deeper to see what else I can do to make O2 match O3. Ideally the code generated with the 2 should be identical.

PR #1154, opened 2025-02-28
Do not explicitly use `--iree-amdaie-num-rows` if it is default 

newling:
I expect/hope the C++ code will use defaults if not provided

{3FDAD5D8-CA6B-4571-9B97-868FD1A79F90}

My guess is that we need to update the default for strix, might not be 8 columns 

UPDATE:

My guess was wrong. The issue was related to this: https://github.com/nod-ai/iree-amd-aie/pull/955#discussion_r1870234195

The default number of rows and columns wasn't based on the provided device, it was based on the default device! I suggested this to Vivian (see above), but in retrospect I don't think it was a good design choice. In this PR I've improved it, so that the default number of rows and columns is just 'all', where the actual number is determined after the device is known.

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -1583,10 +1583,6 @@ class Tests:
                     name_suffix="scaling",
                     tile_pipeline="pack-peel-4-level-tiling",
                     run_on_target=["npu1_4col"],
-                    aie_compilation_flags=[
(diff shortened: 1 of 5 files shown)


Yu-Zhewen (approved):
LGTM. Thanks

yzhang93 (approved):
This is a great finding! Thanks for the improvement!

PR #1158, opened 2025-02-28
Performance testing of core code 'in isolation'

newling:
To check the performance of the data movement, we added the flag 'replaceOutlinedFunctionsWithEmpty' which when used replaces calls to all compute (matmul/elementwise) functions with calls a function that does no compute. This means all the wall time reported must be spent moving data to and from the core from host memory. We currently don't have a way of answering the related question: what is the performance of the core code? In other words, if data movement was infinitely fast, how long would our matmul take to run? 

This PR makes it possible to answer that question. 

It's not obvious to me how we can eliminate _all_ data movement. There is probably a way to do this in isolation, from the input.o or input.asm files that peano generates. Maybe something like is [here](https://github.com/nod-ai/iree-amd-aie/pull/732/files). 

This PR gives the user the option, when a function is outlined, to wrap the new call site in a loop of user-chosen size. So if the loop count is n=100, the peano/ukernel code will run 100x, but without any more L3->L2->L1 data movement. In the limit n->infinity we get the core code performance

This PR subsumes the `replacOutlinedFunctionsWithEmpty' logic. 

The benchmark added in this PR would take  100'000 us if running at 100 %. With O3 it takes 3.5x that (see below). With O2 it takes 4.2x that (see below).  i.e. there is lots of room for improvement. 

```
matmul_512_512_512_bf16_f32_O2_npu1_4col_callrepl_100_outline_benchmark
--------------------------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations UserCounters...
--------------------------------------------------------------------------------------------------
BM_matmul/process_time/real_time_mean       424285 us         1996 us            3 items_per_second=2.35691/s
BM_matmul/process_time/real_time_median     424356 us         1992 us            3 items_per_second=2.35651/s
BM_matmul/process_time/real_time_stddev        147 us          108 us            3 items_per_second=818.927u/s
--------------------------------------------------------------------------------------------------
The largest program memory size (read from byte 72 of elf files) is 6096 bytes



matmul_512_512_512_bf16_f32_O3_npu1_4col_callrepl_100_outline_benchmark
--------------------------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations UserCounters...
--------------------------------------------------------------------------------------------------
BM_matmul/process_time/real_time_mean       354238 us         1843 us            3 items_per_second=2.82297/s
BM_matmul/process_time/real_time_median     354047 us         1865 us            3 items_per_second=2.82449/s
BM_matmul/process_time/real_time_stddev        835 us         72.6 us            3 items_per_second=6.64549m/s
--------------------------------------------------------------------------------------------------
The largest program memory size (read from byte 72 of elf files) is 9568 bytes
```


newling:
>  So if 100 calls run 354238 us, it means 3542.38 us on average for 1 call?

That is correct. 3542.38 us to do a 512x512x512 matmul on a single core. 

>  It seems much worse than a regular test result without 100 calls?

I'm not sure what you mean here. Let me explain the number from first principles 

The test runs a 512x512x512 matmul on a single AIE core, 100 times. 

That is 512*512*512*2 flops per call, so 268'435'456 flops per call. 

So that's a total of 26'843'545'600 flops for the 100 calls. 

It does this in  354238  us. 

That is  0.354238 seconds (1'000'000 microseconds in a second). 

So it's chugging out  26'843'545'600 /  0.354238 per second. 

That's 75778277880.97269 flops per second, which is 

 0.07577827788097269 teraflops per second (divide by 1e12). 
 
 That's one core. We know that phoenix can do 4 tflops with 16 cores, so the theoretical peak for 1 core is 4/16 = 0.25 tflops. 
 
Dividing observed tflops by theoretical peak tflops,  this test shows we're getting 

0.07577827788097269 / 0.25 = 0.303  of peak theoretical performance. That is 30.3 % of theoretical peak

yzhang93:
> > So if 100 calls run 354238 us, it means 3542.38 us on average for 1 call?
> 
> That is correct. 3542.38 us to do a 512x512x512 matmul on a single core.
> 
> > It seems much worse than a regular test result without 100 calls?
> 
> I'm not sure what you mean here. Let me explain the number from first principles

Thanks for the explanation, I didn't realize it's on a single core.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETile.cpp
```diff
@@ -5,6 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include "iree-amd-aie/Transforms/Passes.h"
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h"
```
yzhang93, line 8: Why is this header added?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETileAndFuse.cpp
```diff
@@ -5,6 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include "iree-amd-aie/Transforms/Passes.h"
+#include "iree/compiler/Codegen/Dialect/Codegen/IR/IREECodegenAttrs.h"
```
yzhang93, line 8: Same question for all the added headers, are they needed?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
```diff
@@ -97,7 +99,8 @@ class AMDAIELinalgFunctionOutliningPass
       : AMDAIELinalgFunctionOutliningBase(opts) {}
 
   void getDependentDialects(DialectRegistry &registry) const override {
-    registry.insert<linalg::LinalgDialect>();
+    registry.insert<linalg::LinalgDialect, scf::SCFDialect, LLVM::LLVMDialect,
+                    func::FuncDialect>();
```
yzhang93, line 103: Do you need to add all these dialects dependency?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td
```diff
@@ -306,14 +306,16 @@ def AMDAIELinalgFunctionOutlining :
...
   let options = [
-    Option<"emptyFunctions", "empty-functions", "bool", /*default=*/"false",
-      "A developer only option that results in incorrect numerics. "
-      "Replace all outlined functions with a function that does nothing, "
-      "i.e. it just returns. Useful for measuring the performance of data "
-      "movement to/from the device -- by doing zero compute, all time is spent "
-      "moving data to/from the AIE cores.">,
+    Option<"callReplication", "call-replication", "int64_t", /*default=*/"1",
+      "The number of times the outlined function is called inside a loop. "
+      "This is a 'developer only' only that results in incorrect numerics. "
```
yzhang93, line 315:
```suggestion
      "This is a 'developer-only' feature that may result in incorrect numerics. "
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIETile.cpp
newling: My editor gave a warning about 

```
#include "iree/compiler/Codegen/Common/TileSizeSelection.h"
```

being unused in Passes.h 

Which makes sense: in general you should only include headers in the file that they're needed. This is a well established rule in C++, because it can save unnecessary recompilation when headers change. 

So I removed that header, and added the appropriate headers in the source files they're needed.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELinalgFunctionOutlining.cpp
newling: I use all these dialects, so I think so. The linalg and func dialects are obvious, the scf dialect and llvm dialect are used now too: for creating scf.for ops that don't get unrolled (llvm attribute to prevent unrolling).

yzhang93 (approved):
This is a nice feature, thanks!

konstantinschwarz:
Drive-by comment: have you tried using performance counters (events) to measure the performance of the compute code?

newling:
> Drive-by comment: have you tried using performance counters (events) to measure the performance of the compute code?

Not really. Do you mean something AIE/npu specific?

Yu-Zhewen:
> > Drive-by comment: have you tried using performance counters (events) to measure the performance of the compute code?
> 
> Not really. Do you mean something AIE/npu specific?

Yes, it's an AIE feature that enables cycle-accurate performance measurement using timers within the cores.
https://github.com/Xilinx/mlir-aie/tree/main/programming_guide/section-4/section-4b

To support this, it requires some routing resources and also sets up special registers. There is also existing code in `mlir-air` to automate this process:

https://github.com/Xilinx/mlir-air/blob/7301aa0b6be83806554e5f739c55a37e6bb87d56/mlir/lib/Conversion/AIRToAIEPass.cpp#L3189

https://github.com/Xilinx/mlir-air/blob/7301aa0b6be83806554e5f739c55a37e6bb87d56/mlir/lib/Conversion/AIRRtToNpuPass.cpp#L1324

The only issue is that setting up such performance tracing for a full-array design can be challenging due to routing congestion (unless the design itself also uses packet flows). However, it should work fine for a single core, a single column, or 2×2 cases.

newling:
Thanks @Yu-Zhewen I'll take a look. It'd be better to have a cycle accurate evaluation.

PR #1159, opened 2025-03-01
[AssignBufferAddresses] Add bank-aware scheme

yzhang93:
The scheme mostly adheres to the implementation in MLIR-AIE. Most changes are within pre-specified address/bank part, and more test cases are added to prove a better logic.


compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
```diff
@@ -46,6 +50,204 @@ static LogicalResult basicAllocation(
...
+          deviceModel.getLocalMemorySize(tile.getCol(), tile.getRow());
+    }
+
+    int numBanks = deviceModel.getNumBanks(tile.getCol(), tile.getRow());
+    int bankSize = maxDataMemorySize / numBanks;
+
+    // Each entry of `nextAddrInBanks` is the next address available for use
+    // in that bank, and the index is the bank number.
+    int stackSize = 0;
+    std::vector<int64_t> nextAddrInBanks;
```
Yu-Zhewen, line 196:
```suggestion
    SmallVector<int64_t> nextAddrInBanks;
```

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
```diff
@@ -46,6 +50,204 @@ static LogicalResult basicAllocation(
...
+    for (int i = 0; i < numBanks; i++) nextAddrInBanks.push_back(bankSize * i);
+    // Leave room at the bottom of the address range for stack.
+    if (CoreOp core = tile.getCoreOp()) {
+      stackSize = core.getStackSize();
+      nextAddrInBanks[0] += stackSize;
+    }
+
+    // Each entry of `bankLimits` contains pairs of start and end addresses for
+    // that bank.
+    std::vector<BankLimits> bankLimits;
```
Yu-Zhewen, line 206:
```suggestion
    SmallVector<BankLimits> bankLimits;
```

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
```diff
@@ -46,6 +50,204 @@ static LogicalResult basicAllocation(
...
+
+    // Each entry of `nextAddrInBanks` is the next address available for use
+    // in that bank, and the index is the bank number.
+    int stackSize = 0;
+    std::vector<int64_t> nextAddrInBanks;
+    for (int i = 0; i < numBanks; i++) nextAddrInBanks.push_back(bankSize * i);
+    // Leave room at the bottom of the address range for stack.
+    if (CoreOp core = tile.getCoreOp()) {
+      stackSize = core.getStackSize();
+      nextAddrInBanks[0] += stackSize;
```
Yu-Zhewen, line 201: Would it make sense to return an error if stackSize > bankSize?

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
```diff
@@ -46,6 +50,204 @@ static LogicalResult basicAllocation(
...
+
+    // Set addresses for remaining buffers.
+    SmallVector<BufferOp> allocatedBuffers;
+    int bankIndex = 0;
+    for (BufferOp buffer : buffersToAlloc) {
+      // If the buffer doesn't fit in any of the bank space, it emits an error
+      // and then deallocates all the buffers.
+      if (!setBufferAddress(buffer, numBanks, bankIndex, nextAddrInBanks,
+                            bankLimits)) {
+        deAllocateBuffers(allocatedBuffers);
```
Yu-Zhewen, line 241: Would it be necessary to keep `deAllocateBuffers` here? From what I see, `mlir-aie` includes it primarily for this fallback mode (https://github.com/Xilinx/mlir-aie/pull/1625), which we aren’t planning to support?

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
```diff
@@ -46,6 +50,204 @@ static LogicalResult basicAllocation(
...
+
+// Function that checks whether the given buffer already has a set address
+// attribute. If it does, it finds in which bank the buffer is and checks
+// whether there is enough space left for it. If there is the function returns
+// true and if not, the function emits a warning that the address will be
+// overwritten and returns false (which will cause the buffer to be added to
+// the list of buffers without addresses, to be completed later on).
+FailureOr<bool> checkAndAddBufferWithAddress(
+    BufferOp buffer, int numBanks, std::vector<int64_t> &nextAddrInBanks,
+    std::vector<BankLimits> &bankLimits) {
```
newling, line 90:
```suggestion
    const std::vector<BankLimits> &bankLimits) {
```

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
```diff
@@ -46,6 +50,204 @@ static LogicalResult basicAllocation(
...
+  auto addrAttr = buffer->getAttrOfType<IntegerAttr>("address");
+  if (!addrAttr) return false;
+
+  int addr = addrAttr.getInt();
+  for (int i = 0; i < numBanks; i++) {
+    // If the address is not within the bank, continue.
+    if (addr < bankLimits[i].startAddr || addr >= bankLimits[i].endAddr)
```
newling, line 97: Nit: you can get the (first) bank that the address is on in O(1) instead of O(numBanks).

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
```diff
@@ -46,6 +50,204 @@ static LogicalResult basicAllocation(
...
+
+// Function that checks whether the given buffer already has a set mem_bank
+// attribute. If it does, it checks whether there is enough space left for
+// it. If there is, it sets the buffer's address field and if not, the function
+// emits a warning that the mem_bank will be overwritten and returns false
+// (which will cause the buffer to be added to the list of buffers without
+// addresses, to be completed later on).
+FailureOr<bool> checkAndAddBufferWithMemBank(
+    BufferOp buffer, int numBanks, std::vector<int64_t> &nextAddrInBanks,
+    std::vector<BankLimits> &bankLimits) {
```
newling, line 119:
```suggestion
    const std::vector<BankLimits> &bankLimits) {
```

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
```diff
@@ -46,6 +50,204 @@ static LogicalResult basicAllocation(
...
+    return buffer->emitOpError("would override existing mem_bank");
+  setAndUpdateAddressInBank(buffer, startAddr, endAddr, nextAddrInBanks);
+  return true;
+}
+
+// Function that given a buffer will iterate over all the memory banks starting
+// from the given index to try and find a bank with enough space. If it does,
+// it will set the buffer's address and mem_bank attributes and update the
+// nextAddrInBanks vector. If it does not find one with enough space, it will
+// throw an error. Returns true if the buffer was successfully allocated, false
```
newling, line 136:
```suggestion
// emit an error. Returns true if the buffer was successfully allocated, false
```

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -536,6 +536,10 @@ struct AMDAIEDeviceModel {
 
   FailureOr<std::array<uint32_t, 3>> getAIEMatmulInstructionSize(
       Type elTypeLhs, Type elTypeRhs, Type elTypeAcc) const;
+
```
newling, line 539: Where did you find these values?

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
newling, line 206: Neater IMO would be to have a constructor for the BankLimits class, 

BankLimits(int numBanks, int bankSize);

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
```diff
@@ -46,6 +50,204 @@ static LogicalResult basicAllocation(
...
+          deviceModel.getMemTileSize(tile.getCol(), tile.getRow());
+    } else {
+      maxDataMemorySize =
+          deviceModel.getLocalMemorySize(tile.getCol(), tile.getRow());
+    }
+
+    int numBanks = deviceModel.getNumBanks(tile.getCol(), tile.getRow());
+    int bankSize = maxDataMemorySize / numBanks;
+
+    // Each entry of `nextAddrInBanks` is the next address available for use
```
newling, line 193: Is this code duplicated with the linear algorithm?

newling (commented):
An alternative design that might be 

1) Try bank aware allocation. 
2) If bank aware allocation failed, emit warning and use linear allocation. 

Where in bank aware allocation, automatically fail if 
1) any buffer has a prespecified address
2) any buffer has a prespecified memory bank
3) any buffer is allocated during the algorithm to more than 1 bank (i.e. overflows). 

My thinking is that 

1) if a user prespecifies the address/bank if a buffer, they probably don't need/want us to do anything more than just place the remaining buffers as simply as possible.  
2) the above will hopefully simplify the logic a bit
3) the above will result in compilation failure less frequently

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
yzhang93: I think the stack size is usually small compared to the bank size.

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
yzhang93: Sorry, I didn't get this.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
yzhang93: From mlir-aie: https://github.com/Xilinx/mlir-aie/blob/043267ec235b1d2a1dc957ff2e786a535c52c0ce/include/aie/Dialect/AIE/IR/AIETargetModel.h#L399

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
yzhang93: The sequential scheme doesn't have the `mem_bank` concept and doesn't have each buffer assigned to a particular bank. So I don't see the code duplicated.

yzhang93: Or maybe you mean the codes to get `maxDataMemorySize` for each tile? That part is duplicated I think.

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
newling: You can statically determine which `i` won't be a `continue`, and so eliminate the for loop:

auto bankIndex = bankLimits[i].startAddr / bankSize;


Or you can have a function:

// Return the bank that `address` is in, where banks are non-interleaved and of size `bankSize`.
int getBankOfAddress(int address, int bankSize) {
  return address / bankSize;
}

yzhang93: Oh, I see. Thanks!

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
yzhang93: I have simplified these codes.

Yu-Zhewen (approved):
Thanks. LGTM

newling (commented):
Looks good, my remaining comments are quite minor. 

Please let me know later if you'd like me to take another look, or discuss some comments further.

compiler/plugins/target/AMD-AIE/aie/test/bank_aware_buffer_alloc.mlir
```diff
@@ -0,0 +1,93 @@
...
+
+// -----
+
+module @prealloc_membank {
+  aie.device(npu1) {
+    %tile44 = aie.tile(4, 4)
+    %buf0 = aie.buffer(%tile44) : memref<200xi32>
+    %buf1 = aie.buffer(%tile44) : memref<100xi32>
+    %buf2 = aie.buffer(%tile44) { sym_name = "b", address = 4096 : i32 } : memref<1024xi32>
+    // expected-error@+1 {{'aie.buffer' op would override the existing mem_bank}}
```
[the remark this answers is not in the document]
yzhang93: Here it doesn't mean overflow. The check is 
```
if (addr < nextAddrInBanks[bankIndex])
    return buffer->emitOpError("would override the allocated address");
```
If the condition is satisfied, it means there's already some buffer that is allocated at `addr`. So if we try to allocate this buffer, it will overwrite the existing address.

I think for pre-allocation case there is misunderstanding (to be honest, I don't know why the codes were written in that way, but I'm just trying to follow the logic in mlir-aie).

It seems like it does not only preserve the address for the pre-allocation, but also assign a new address for the buffer.(https://github.com/Xilinx/mlir-aie/blob/0526d7edece3359f73f7b2b1ba684eeb436e5d3d/lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp#L207)

So for example, if the input is 
```
%buf2 = aie.buffer(%tile44) { sym_name = "b", address = 4096 : i32 } : memref<1024xi32>
```
the output will be 
```
%buffer_4_4_1 = aie.buffer(%tile_4_4) {address = 8192 : i32, mem_bank = 0 : i32, sym_name = "b"} : memref<1024xi32>
```
And for other buffers it will assign address after `12288`.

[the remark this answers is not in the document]
yzhang93: Some comments which may help https://github.com/Xilinx/mlir-aie/blob/0526d7edece3359f73f7b2b1ba684eeb436e5d3d/lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp#L398

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
```diff
@@ -46,6 +52,188 @@ static LogicalResult basicAllocation(
...
+      FailureOr<bool> has_bank =
+          checkAndAddBufferWithMemBank(buffer, nextAddrInBanks, bankLimits);
+      if (failed(has_addr) || failed(has_bank)) return failure();
+      if (!has_addr.value() && !has_bank.value())
+        buffersToAlloc.push_back(buffer);
+      else
+        preAllocatedBuffers.push_back(buffer);
+    }
+
+    // Note: This is currently disabled to avoid numerical error in ci for
```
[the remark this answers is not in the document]
yzhang93: Not fully understand the issue. I will investigate more.

compiler/plugins/target/AMD-AIE/aie/test/bank_aware_buffer_alloc.mlir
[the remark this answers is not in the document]
newling: >  I'm not sure what's the use case in mlir-aie.

I agree. that's my initial suggestion was to fail early with the bank aware function in this case, to keep it simpler

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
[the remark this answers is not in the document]
newling: Yeah, it's probably not directly related to this PR, my guess is another alignment edge case. But maybe worth confirming that it's not uncovering an easy to fix issue in this PR which we're not seeing

compiler/plugins/target/AMD-AIE/aie/test/bank_aware_buffer_alloc.mlir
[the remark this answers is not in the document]
yzhang93: @newling sorry, I misread your commented line. I agree for the `mem_bank error` part you pointed, it should be over run. I've made changes accordingly.

In addition, I've modified the checking logic for pre-specified bank and address. I also added some new tests. I think now the code logic makes more sense to me.

compiler/plugins/target/AMD-AIE/aie/test/bank_aware_buffer_alloc.mlir
```diff
@@ -0,0 +1,93 @@
+
```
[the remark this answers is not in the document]
yzhang93: Done.

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
```diff
@@ -46,6 +52,188 @@ static LogicalResult basicAllocation(
...
+
+    // Set addresses for remaining buffers.
+    SmallVector<BufferOp> allocatedBuffers;
+    int bankIndex = 0;
+    for (BufferOp buffer : buffersToAlloc) {
+      // If the buffer doesn't fit in any of the bank space, it emits an error
+      // and then deallocates all the buffers.
+      if (!setBufferAddress(buffer, numBanks, bankIndex, nextAddrInBanks,
+                            bankLimits)) {
+        deAllocateBuffers(allocatedBuffers);
```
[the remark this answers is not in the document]
yzhang93: I've modified the logic for pre-specified bank and address and added new test cases.

compiler/plugins/target/AMD-AIE/aie/test/bank_aware_buffer_alloc.mlir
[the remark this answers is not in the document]
newling: Ok great, thanks!

newling:
@yzhang93 

Just rebased on this and seeing 

```
<unknown>:0: warning: Buffer assignment scheme is unrecognized. Defaulting to bank-aware scheme.
```

Hmm I guess this shouldn't be a warning, seems a bit noisy?

yzhang93:
> @yzhang93
> 
> Just rebased on this and seeing
> 
> ```
> <unknown>:0: warning: Buffer assignment scheme is unrecognized. Defaulting to bank-aware scheme.
> ```
> 
> Hmm I guess this shouldn't be a warning, seems a bit noisy?

Oh, yeah, I modified the scheme to `None` by default and that's why it has the warning all the time. I'll remove it.

PR #1164, opened 2025-03-05
[MLIR-AIR] Make `outerPerm` compatible to both `objectfifo` and `air` in `pack-peel-4-level-tiling`

erwei-xilinx:


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -427,12 +427,13 @@ static LogicalResult setRootConfigForPackPeel4LevelTilingPipeline(
...
-    // Add outer permutation for unpack. NOTE: This currently fails for some
-    // tests in the AIR pipeline.
+    // Add outer permutation for unpack.
     if (isBatchMatmul) {
       outerPerm.push_back({0, 2, 1});
     } else {
-      outerPerm.push_back({1, 0});
+      if (isObjectFifo) {
+        outerPerm.push_back({1, 0});
+      }
```
yzhang93, line 436: So for AIR the outerPerm needs to be {0, 1}? Does it work if you don't push anything?

yzhang93 (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: The title is a bit confusing... Are you using the same outerPerm or not?

erwei-xilinx: The matmul test on the CI seems to be passing for both with and without the outerPerm... I haven't been able to investigate deeply why both configurations would pass the same test. This PR was intended to make it follow the same code structure as the pack-peel pipeline.

erwei-xilinx: Make sense, will definitely start sweeping the GEMM sizes and will keep the outerPerm consideration in mind! I will close the PR for now, and follow up afterwards.

PR #1167, opened 2025-03-06
Pass to 'unroll and jam' the scf.for loops around aievec.matmul

newling:
Currently: the scf for loops become cf blocks, which get unrolled in llvm's opt (after leaving iree-amd-aie). The unrolling results in an order of matmuls that is linear: iterate `k` then iterate `m` then iterate `n`. This isn't optimal for AIE -- I don't know the details but peano somehow can't orchestrate the pipelining of loads to registers with the mac/mfma instructions. 

The basic idea with this PR is to 'tile' the scf.for loops, so that multiple m and n's can be running in parallel. See lit tests for what  jam and unroll actually does. The PR allows us to experiment with different approaches and manually find the best performing approach. 

The improves performance (measured as wall clock time to run a matmul in a tight loop) by between 25% and 35%. With this PR, the bf16 on strix kernel performance is about 55% overall (see CI and will appear in https://nod-ai.github.io/iree-amd-aie/results_history_npu1.html when landed). 

FOLLOW-UP work: this PR only improves performance for 1 case: bf16 on phoenix. It also relies on the tiling strategy not changing (expects 64x64x64 matmul on core). This is a bit brittle, and can hopefully be made more robust. 

NOTE: currently this only hits the pack-peel pipeline, see also https://github.com/nod-ai/iree-amd-aie/pull/1177


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollJamAIEVecMatmul.cpp
```diff
@@ -0,0 +1,448 @@
...
+#include "mlir/Transforms/DialectConversion.h"
+
+#define DEBUG_TYPE "iree-amdaie-unroll-jam-aievec-matmul"
+
+namespace mlir::iree_compiler::AMDAIE {
+
+namespace {
+
+/// Get the loop count of \p forOp if its loop bounds and step are constant.
+/// otherwise return std::nullopt. The loop count is 'end - start / step'.
```
Yu-Zhewen, line 30:
```suggestion
/// Otherwise return std::nullopt. The loop count is '(end - start) / step'.
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/unroll_jam_aievec_matmul.mlir
```diff
@@ -0,0 +1,295 @@
...
+// CHECK: transfer_write {{.*}} %arg0[%c1, %c3]
+// CHECK: transfer_write {{.*}} %arg0[%c2, %c2]
+// CHECK: transfer_write {{.*}} %arg0[%c3, %c2]
+// CHECK: transfer_write {{.*}} %arg0[%c2, %c3]
+// CHECK: transfer_write {{.*}} %arg0[%c3, %c3]
+
+
+// -----
+
+// Test of the 'none' sequence, which specifies that no this pass effectively does nothing.
```
Yu-Zhewen, line 183:
```suggestion
// Test of the 'none' sequence, which specifies that this pass effectively does nothing.
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/unroll_jam_aievec_matmul_failures.mlir
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// -----
+
+func.func @sequence_of_length_remainder_2(%arg0: memref<100xf32>, %arg1: vector<4x8xbf16>, %arg2: vector<8x4xbf16>, %arg3: vector<4x4xf32>) {
+  %c0 = arith.constant 0 : index
+  %c1 = arith.constant 1 : index
+  %c6 = arith.constant 6 : index
+  scf.for %arg4 = %c0 to %c6 step %c1 {
+     //expected-error @below {{'aievec.matmul' op has an unroll sequence "uj_0" whose length is is 3*n + 2 for some n.}}
```
Yu-Zhewen, line 98:
```suggestion
     //expected-error @below {{'aievec.matmul' op has an unroll sequence "uj_0" whose length is 3*n + 2 for some n.}}
```

Yu-Zhewen (approved):
LGTM! Really interesting results. Looking forward to seeing how it performs on different devices and pipelines.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEUtils.h
```diff
@@ -135,6 +135,10 @@ std::string getConstantIntValuesString(ArrayRef<OpFoldResult> opFoldResults);
 bool sinkInto(Region &, IRRewriter &,
               std::function<bool(Operation *)> shouldSink);
 
+/// Annotate `forOp` with the llvm.loop_annotation attribute specifying that
+/// it should never be unrolled.
+void addNoUnrollAttribute(scf::ForOp forOp, OpBuilder &builder);
```
jtuyls, line 140: What should happen if the loop already has a 'NoUnroll' loop annotation? And unless I overlooked, you don't have a testcase for that?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollJamAIEVecMatmul.cpp
```diff
@@ -0,0 +1,448 @@
...
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Transforms/DialectConversion.h"
+
+#define DEBUG_TYPE "iree-amdaie-unroll-jam-aievec-matmul"
+
+namespace mlir::iree_compiler::AMDAIE {
+
+namespace {
+
+/// Get the loop count of \p forOp if its loop bounds and step are constant.
```
jtuyls, line 29: This `\p` notation doesn't seem to be used yet in the codebase and neither in `IREE`, can we just use `` for consistency within the codebase?

jtuyls (changes requested):
> The basic idea with this PR is to 'tile' the scf.for loops, so that multiple m and n's can be running in parallel

What do you mean with this? How can `m` en `n`'s run in parallel?

> The improves performance (measured as wall clock time to run a matmul in a tight loop) by between 25% and 35%. With this PR, the bf16 on strix kernel performance is about 55% overall (see CI and will appear in https://nod-ai.github.io/iree-amd-aie/results_history_npu1.html when landed).

I think you mean on Phoenix and not Strix? And how do you calculate the 55% number?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollJamAIEVecMatmul.cpp
yzhang93: +1 It's not very readable.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollJamAIEVecMatmul.cpp
```diff
@@ -0,0 +1,448 @@
...
+        return matmul->emitOpError()
+               << errStart() << "with an unknown unroll/no-unroll at end, '"
+               << noUnrollAtEndStr << "'. Expected 'NOUNROLL' or 'UNROLL'.";
+      }
+      splits.pop_back();
+    }
+
+    else if (splitsMod == 2) {
+      return matmul->emitOpError()
+             << errStart() << "whose length is is 3*n + 2 for some n.";
```
yzhang93, line 274:
```suggestion
             << errStart() << "whose length is 3*n + 2 for some n.";
```
Also I'm not quite understand what this error means.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollJamAIEVecMatmul.cpp
```diff
@@ -0,0 +1,448 @@
...
+    }
+
+    else {
+      assert(splitsMod == 0 && "case of 1 and 2 already handled");
+    }
+    assert(
+        splits.size() % 3 == 0 &&
+        "case of 1 reduced to case of 0, case of 2 already handled as error.");
```
yzhang93, line 282: I don't think this assert is still needed?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollJamAIEVecMatmul.cpp
```diff
@@ -0,0 +1,448 @@
...
+  // the start of the function's block.
+  auto hoistConstants = [&](scf::ForOp scfFor) {
+    Operation &firstOp = block->getOperations().front();
+    SmallVector<arith::ConstantOp> constantsToHoist;
+    scfFor.walk([&](arith::ConstantOp constantOp) {
+      constantsToHoist.push_back(constantOp);
+    });
+    for (arith::ConstantOp constant : constantsToHoist) {
+      builder.moveOpBefore(constant, &firstOp);
+    }
```
yzhang93, line 333: I think this can go inside the scfFor.walk?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollJamAIEVecMatmul.cpp
newling: Yeah I guess so, especially as we're not generating any doxygen comments. Will update. 

> +1 It's not very readable.

Looks quite nice with the right vim plugin :)

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUnrollJamAIEVecMatmul.cpp
newling: Remainder when divided by 3 is 2. There are 2 supported formats. If remainder is 0, it's something like `uj_0_2_u_1_4`. If remainder is 1, it's like the remainder 0 case but with `_UNROLL` or `_NOUNROLL` in the end. If remainder is 2: invalid.

jtuyls:
> By 'in parallel' I mean that we don't process a full k-reduction for (m0, n0) before starting the k-reduction of (m1, n1)

I see. I would avoid 'in parallel' as it conveys multiple parallel macs imo, which is not what is happening. Rather, it would describe it as tiling/processing across the parallel dimensions first, before the reduction dimension.

> flops = 100*512*512*512*2
time = 0.172 seconds
observed tflops = 1e-12 * 100*512*512*512*2 / 0.172 = 0.156
Phoenix 16 cores can to 4 tflops, so 1 core does 0.25 tflops (the above test is a single core).
observed / peak = 62.4 % (so 55% was a slight underestimate) O2 took 181897 us which is 59%.

One caveat is that this CI machine is a Hawk point version I believe, which runs at higher frequency (1.6GHz I believe).

newling:
> One caveat is that this CI machine is a Hawk point version I believe, which runs at higher frequency (1.6GHz I believe).

The "4 tflops" number for phoenix, what frequency is that based on?

newling:
> > > One caveat is that this CI machine is a Hawk point version I believe, which runs at higher frequency (1.6GHz I believe).
> > 
> > 
> > The "4 tflops" number for phoenix, what frequency is that based on?
> 
> 1GHz
> 
> 128 MACs (bf16) * 2 (flops) * 16 cores * 1e9 = 4tflops

Oh I guess I could have solved for the 1 unknown myself in that equation! But darn, the 62% drops to 39%.

PR #1170, opened 2025-03-07
[PktFlow] Disable packet flows on kernel output to allow loop subsumption

Yu-Zhewen:
The DMA loop subsumption optimization is crucial for unlocking packet flow performance.

Without  loop subsumption:
```ll
for i = 0 to 16
  dma_cpy arg_0[i]
  dma_cpy arg_1[i]
  sync
```
When unrolled, this generates a large number of DMA operations in the control code.

However, with loop subsumption:
```ll
 dma_cpy arg_0
 dma_cpy arg_1
 sync
```
Here, the loops are absorbed by strides in `dma_cpy`, significantly reducing control code size.

However, enabling this optimization for packet flows may introduce deadlocks if `arg_0` and `arg_1` share an arbiter at any point along their routed path.  Since arbitration order is not guaranteed, a deadlock can occur if two tiles of `arg_0` are sent from shim, but zero tiles of `arg_1`, while the core is waiting for `arg_1` before accepting the second tile of `arg_0`.

#806 introduced the concept of `packetGroupId` for this deadlock issue, which tries to separate the channels for `arg_0` and `arg_1` to reduce the chance for them to go through the same arbiter. However, this solution does not fully guarantee the exclusion of deadlocks, especially when we need to merge the usage of arbiters [here](https://github.com/nod-ai/iree-amd-aie/blob/b720c20b1e61759b8fd30261c5c7430c06354a17/runtime/src/iree-amd-aie/aie_runtime/amsel_generator.cc#L173). 

Solving this deadlock issue in a robust way may require significant effort to modify the router logic or exploit partial reconfiguration in the future. 

As a temporary workaround, this PR limits the packet flow usage only for transferring kernel input (L3->L2->L1), while disabling that for the outputs. This change helps to free up some arbiters, so that deadlocks won't happen after loop subsumption, at least not on the following matmul test cases.

Another (kind of) rationale for this workaround is, the current need for using packet flows is mainly on the kernel input side, as we also want to send control packets into the device. 

Finally, performance results:

M: 512, N: 512, K: 4096

|                       | Latency(us) | Instructions |
|-----------------------|-------------|--------------|
| circuit               | 2816        | 1096         |
| packet before this PR | 40761      | 503176       |
| packet (only input) after this PR  | 2901       | 1096       |

M: 512, N: 4096, K: 512

|                       | Latency(us) | Instructions |
|-----------------------|-------------|--------------|
| circuit               | 3640        | 736         |
| packet before this PR | 39025      | 511400       |
| packet (only input) after this PR  | 3705       | 736       |

M: 4096, N: 512, K: 512

|                       | Latency(us) | Instructions |
|-----------------------|-------------|--------------|
| circuit               | 2805        | 4124         |
| packet before this PR | 45004      | 513032       |
| packet (only input) after this PR  | 2849       | 4124       |

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -649,7 +649,7 @@ class MatmulConstBiasCtrlpkt(BaseMatmul):
         self.aie_compilation_flags += [
             "--iree-amdaie-num-rows=1",
             "--iree-amdaie-num-cols=1",
-            "--iree-amdaie-enable-packet-flow=true",
+            "--iree-amdaie-enable-input-packet-flow=true",
             "--iree-amdaie-emit-control-packet=true",
             "--mlir-disable-threading",
         ]
--- a/build_tools/ci/run_matmul_test.sh
+++ b/build_tools/ci/run_matmul_test.sh
@@ -371,7 +371,7 @@ function run_matmul_test() {
                       --iree-amdaie-tile-pipeline=${tile_pipeline} \
                       --iree-amd-aie-peano-install-dir=${peano_install_path} \
                       --iree-amd-aie-install-dir=${amd_aie_install_path} \
-                      --iree-amdaie-enable-packet-flow=${enable_packet_flow} \
+                      --iree-amdaie-enable-input-packet-flow=${enable_packet_flow} \
                       --iree-hal-dump-executable-files-to=$PWD \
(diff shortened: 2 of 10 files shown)


yzhang93 (approved):
Great improvement! Nice work :)

jtuyls (approved):
LGTM

Issue #1130, opened 2025-02-25
Matmul4d + Trunci to improve performance on Strix


yzhang93:
As individual work of matmul4d and trunci are working e2e on Strix, we would like to combine them for better performance. This task is currently blocked by the following aspects:

- https://github.com/iree-org/iree/issues/20019
- The Trunci codes need to be vectorized or added to ukernel.

PR #1172, opened 2025-03-07
[AMDAIEPad] Refactor the pass to pad elementwise op

yzhang93:
Before this PR, if we use pad operation for data movement, only the operands from root computation op were padded/copied. However, for matmul-elementwise dispatch, we want to generate copies for the matmul inputs and elementwise output but not matmul output or elementwise input.

This PR adds option `pad-elementwise` to deal with matmul and elementwise operations separately. In addition, it removes option `padding-level`, instead use a more direct name `pad-operand` to be clear about what operands are padded. These option settings and logic of applying them are similar to [BufferizeToAllocationPass](https://github.com/nod-ai/iree-amd-aie/blob/b569da497fba0c2840bae42fd0c3bb8380c4953f/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEBufferizeToAllocation.cpp#L166).

This is a fix for matmul4d + Trunci task https://github.com/nod-ai/iree-amd-aie/issues/1130.


newling (approved):
LGTM

PR #1176, opened 2025-03-11
[CtrlPkt] Add DMA channels reset for memtiles

Yu-Zhewen:


runtime/src/iree-amd-aie/aie_runtime/iree_aie_configure.cc
```diff
@@ -242,14 +242,8 @@ LogicalResult addElfToTile(const AMDAIEDeviceModel &deviceModel,
   }
   // Ensure the core is disabled before loading the elf.
   TRY_XAIE_API_LOGICAL_RESULT(XAie_CoreDisable, devInst, tileLoc);
-  // Reset All Channels of DMA to make sure they are at idle state.
-  TRY_XAIE_API_LOGICAL_RESULT(XAie_DmaChannelResetAll, devInst, tileLoc,
```
Yu-Zhewen, line 246: Previously, the reset is only done for the core tiles, but it is actually also needed for the mem tiles.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIERT.cpp
```diff
@@ -282,6 +282,9 @@ LogicalResult addInitConfig(const AMDAIEDeviceModel &deviceModel,
       continue;
     }
 
+    // Reset and unreset all DMA channels before configuring BDs.
+    if (failed(resetUnResetDmaChannels(deviceModel, tileLoc))) return failure();
```
Yu-Zhewen, line 286: Previously, only core tiles were reset, leading to numerical errors after reconfiguration. This reset function is now relocated  to ensure that DMA channels on both core tiles and memory tiles are properly reset.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -2182,6 +2182,7 @@ def __init__(self):
 
         # Control packet test with constant biases 1 and 2.
         self.register(MatmulConstBiasCtrlpkt(8, 8, 8, "i8", "i32", 1, 2))
+        self.register(MatmulConstBiasCtrlpkt(1024, 1024, 1024, "i8", "i32", 1, 2))
```
Yu-Zhewen, line 2185: This test was failing without this PR.

jtuyls (approved):
LGTM

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AMDAIERT.cpp
jtuyls: Wow, great find!

PR #1180, opened 2025-03-13
[CI] Add benchmark test for reconfiguration time measurement

Yu-Zhewen:
This PR adds support for reconfiguration time measurement, with results being logged on the performance tracking webpage once merged:
https://nod-ai.github.io/iree-amd-aie/results_history_npu1.html

Observations on my local machine: The current reconfiguration time for the whole array is quite poor (~110ms), whereas the ideal target is < 100µs. After discussing with @jtuyls, the top optimization priority should be reducing the control code size for control packet DMAs.

Note: Due to the current bad performance, this newly added benchmark test will increase phoenix CI time by ~10s, but this should get optimized soon.

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -726,7 +726,7 @@ class MatmulConstBiasCtrlpkt(BaseMatmul):
                 self.n_repeats,
                 self.n_kernel_runs,
                 self.n_reconfigure_runs,
-                time_unit="ms",
+                time_unit="us",
             )
             return True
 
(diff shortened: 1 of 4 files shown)


build_tools/ci/cpu_comparison/run.py
```diff
@@ -708,6 +714,22 @@ def _execute(self, config):
...
+            benchmark_aie_kernel_time(
+                config,
+                aie_vmfb_C,
+                input_args,
+                self.function_name,
+                test_name_C,
+                self.n_repeats,
+                self.n_kernel_runs,
+                self.n_reconfigure_runs,
+                time_unit="ms",
```
yzhang93, line 729: I suggest to use "us" to keep the same unit as the other tests. Otherwise you need to change the unit when plotting this result.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -2217,6 +2245,22 @@ def __init__(self):
...
+        # Benchmark reconfiguration time only, do not run the kernel.
+        self.register(
+            MatmulConstBiasCtrlpkt(
+                1024,
+                1024,
+                1024,
+                "i8",
+                "i32",
+                1,
+                2,
```
yzhang93, line 2257: If someone is not familiar with `MatmulConstBiasCtrlpkt` class members, it's hard to recognize what these values mean. I'd suggest to use `parameter = ` format.

build_tools/ci/cpu_comparison/run.py
```diff
@@ -1224,8 +1247,9 @@ def benchmark_aie_kernel_time(
         *input_args,
         f"--device={config.device_hal}",
         f"--benchmark_repetitions={n_repeats}",
-        f"--batch_size={n_kernel_runs}",
+        f"--batch_size={max(n_kernel_runs, n_reconfigure_runs)}",
```
jtuyls, line 1250: I don't think it makes sense in any test to have both `n_kernel_runs` and `n_reconfigure_runs` runs set and that would result in incorrect numbers I think, so maybe we check for invalid combinations and throw an error.

jtuyls (approved):
LGTM, just one nit.

build_tools/ci/cpu_comparison/run.py
Yu-Zhewen: Thanks! Done

PR #1185, opened 2025-03-14
[CtrlPkt] Lower control packets to `DmaCpNd` instead of  `HalfDmaCpyNd`

Yu-Zhewen:
This change enables the usage of `AssignNpuDmaBdIds pass for control packet DMAs, which previously had their BD ID assigned to a constant 0. It further allows for the controlcode reduction through BD chaining and sync folding.

As a result, the reconfiguration time for the CI test`matmul_const_bias_ctrlpkt_1024_1024_1024_i8_i32_benchmark`  has been reduced from 110ms to 15ms on my local machine.


jtuyls (approved):
LGTM, very nice speed ups!

PR #1189, opened 2025-03-18
[WIP][DistributeL1Allocations] Add a fix for handling subviews in case of 1x1 core

Abhishek-Varma:
-- This commit adds a fix for handling memref.subviews in case of 1x1 core and for `pack-peel-4-level-tiling` which resolves a compilation issue for `64 x 128 x 128` and thus an e2e test has been added for the same.

-- The pass considers only those memref.subview ops as candidate for L1 alloc substitution which are contained inside `scf.forall` with GPU thread IDs.

-- That assumption is correct for multiple core ops, but in case of 1x1 and for `pack-peel-4-level-tiling`  the subview op need not  follow that and can well be hoisted in the preceding `scf.forall` with NO GPU thread IDs.

-- The logic being tried to add via this PR can be made cleaner and more robust. Will address as part of this PR itself.


Abhishek-Varma:
IR input to `iree-amdaie-distribute-l1-allocations` : https://gist.github.com/Abhishek-Varma/420ffeadcd883559d67783bc96ff0a5d

build_tools/ci/cpu_comparison/run.py
```diff
@@ -2236,6 +2236,25 @@ def __init__(self):
...
+                128,
+                "i32",
+                "i32",
+                test_params=TestParams(
+                    tile_pipeline="pack-peel-4-level-tiling",
+                    aie_compilation_flags=[
+                        "--iree-amdaie-num-rows=1",
+                        "--iree-amdaie-num-cols=1",
+                    ],
+                    name_suffix="1row_1col",
```
Yu-Zhewen, line 2252: To be consistent with the tests above, can we use `OneCore` for `name_suffix` and `additional_labels`?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeL1Allocations.cpp
```diff
@@ -43,42 +44,21 @@ FailureOr<DenseSet<Value>> getThreadIndVars(ModuleOp moduleOp) {
...
-/// allocation `alloc` does not look like it's distributed across threads/tiles,
-/// return an empty memref type. Otherwise, return the memref type that the
-/// subviews are viewing.
+/// For a given alloc the distributed type is fetched by looking at each of its
+/// subview users. For each subview, we check if their users are within the
+/// innermost scf.forall(s). We look at the uses in innermost scf.forall ops
+/// because they comprise the entire computation that each core will perform.
+/// We return empty memref type if :-
+///   a. Any subview's user is NOT within the innermost scf.forall.
+///   b. The result type of two subviews do not match.
```
Yu-Zhewen, line 53:
```suggestion
/// For a given alloc, the distributed type is fetched by looking at each of its
/// subview users. For each subview, we check if their users are within the
/// innermost scf.forall(s). We look at the uses in innermost scf.forall ops
/// because they comprise the entire computation that each core will perform.
/// We return an empty memref type if :-
///   a. Any subview's user is NOT within the innermost scf.forall.
///   b. The result types of two subviews do not match.
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeL1Allocations.cpp
```diff
@@ -114,6 +94,17 @@ LogicalResult distributeLocalMemory(ModuleOp moduleOp) {
...
+
+  // Fetch all innermost scf.forall op.
+  DenseSet<Operation *> innermostForallOps;
+  moduleOp->walk([&](scf::ForallOp forallOp) {
+    bool isInnermost = true;
+    forallOp.walk([&](scf::ForallOp innerForallOp) {
+      if (innerForallOp != forallOp) isInnermost = false;
+    });
+    if (isInnermost) innermostForallOps.insert(forallOp);
+  });
```
jtuyls, line 106: We should look for the `scf.forall` ops with thread mapping, because that's what we consider to be mapped to the cores.

build_tools/ci/cpu_comparison/run.py
Abhishek-Varma: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeL1Allocations.cpp
Abhishek-Varma: Done.

Yu-Zhewen (approved):
LGTM. Thanks!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEDistributeL1Allocations.cpp
```diff
@@ -25,60 +26,38 @@ using namespace mlir;
...
 namespace {
 
-/// Find all induction variables of all `scf.forall` ops that are mapped to
-/// gpu thread dimensions (as opposed to gpu block dimensions etc).
-FailureOr<DenseSet<Value>> getThreadIndVars(ModuleOp moduleOp) {
-  DenseSet<Value> threadIndVars;
+/// Find all `scf.forall` ops that are mapped to gpu thread dimensions (as
+/// opposed to gpu block dimensions etc) i.e. the ones which will be mapped to
+/// cores.
+DenseSet<Operation *> getCoreForallOps(ModuleOp moduleOp) {
```
jtuyls, line 32: I think you can return `DenseSet<scf::ForallOp>` instead

Abhishek-Varma: Done.

jtuyls (approved):
LGTM

Issue #1192, opened 2025-03-19
Buffer sorting in address assignment caused numerical error for depthwise conv2d op


yzhang93:
Sorting the buffer by largest allocation and assigning those larger buffers first would help for better bank assignment and avoid conflicts. However, the `depthwise_conv2d` op had numerical error after sorting. Using sequential buffer assignment doesn't have such issue.

PR #1193, opened 2025-03-19
[AssignBufferAddresses] Enable buffer sorting and disable bank-aware scheme for Conv

yzhang93:
Sorting the buffer by largest allocation and assigning those larger buffers first would help for better bank assignment and avoid conflicts. However, this was disabled previously because `depthwise_conv2d` op had numerical error after sorting. I haven't found the reason behind that and couldn't find a better solution. This PR temporarily disable the bank-aware scheme for conv ops as I continue looking into the [issue](https://github.com/nod-ai/iree-amd-aie/issues/1192) for depthwise ops.


Yu-Zhewen (approved):
Thanks! LGTM

PR #1194, opened 2025-03-19
[CtrlPkt] Improve reconfiguration time by broadcasting core configurations

Yu-Zhewen:
During device reconfiguration, instead of performing separate memory copies for each core, this PR broadcasts the configuration sequence to all cores via the control overlay.

This optimization further reduces reconfiguration time from 15ms (#1185) to 1.3ms.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertDeviceToControlPackets.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertDeviceToControlPackets.cpp
@@ -21,21 +21,67 @@ using Path = std::filesystem::path;
 namespace mlir::iree_compiler::AMDAIE {
 
 namespace {
+
+/// Recursively erase an operation and all its users.
+void recursiveErase(Operation *op) {
+  // Save the users in a separate vector to avoid iterator invalidation.
+  SmallVector<Operation *> users(op->getUsers());
+  // Recursively erase all users first.
+  for (Operation *userOp : users) recursiveErase(userOp);
+  // Now erase this operation.
+  op->erase();
+}
+
+/// Use aie-rt to generate transactions for the given device operation.
+LogicalResult generateTransactions(const AMDAIEDeviceModel &deviceModel,
+                                   xilinx::AIE::DeviceOp deviceOp,
+                                   const std::string &pathToElfs,
+                                   bool broadcastCoreConfig) {
+  if (broadcastCoreConfig) {
+    // Find all core tiles.
+    SmallVector<xilinx::AIE::TileOp> coreTileOps;
+    llvm::copy_if(
+        deviceOp.getOps<xilinx::AIE::TileOp>(), std::back_inserter(coreTileOps),
+        [&](xilinx::AIE::TileOp tileOp) {
+          return deviceModel.isCoreTile(tileOp.getCol(), tileOp.getRow());
+        });
+    // Broadcast means we only need to generate transactions for one core tile.
+    // Keep one core tile and erase the rest.
+    if (coreTileOps.empty()) {
+      return deviceOp.emitOpError()
+             << "no core tiles found in the device operation";
+    }
+    coreTileOps.erase(coreTileOps.begin());
+    for (xilinx::AIE::TileOp tileOp : coreTileOps) recursiveErase(tileOp);
+  }
+
+  if (failed(addAllAieElfs(deviceModel, deviceOp, Path{pathToElfs},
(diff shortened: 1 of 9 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertDeviceToControlPackets.cpp
```diff
@@ -21,21 +21,67 @@ using Path = std::filesystem::path;
...
+                                   const std::string &pathToElfs,
+                                   bool broadcastCoreConfig) {
+  if (broadcastCoreConfig) {
+    // Find all core tiles.
+    SmallVector<xilinx::AIE::TileOp> coreTileOps;
+    llvm::copy_if(
+        deviceOp.getOps<xilinx::AIE::TileOp>(), std::back_inserter(coreTileOps),
+        [&](xilinx::AIE::TileOp tileOp) {
+          return deviceModel.isCoreTile(tileOp.getCol(), tileOp.getRow());
+        });
```
jtuyls, line 47: You can use `llvm::filter_to_vector` here.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEGenerateControlOverlay.cpp
```diff
@@ -46,9 +46,66 @@ LogicalResult initializeChannelsGenerators(
...
+/// target tile CTRL ports. If multiple target tiles are provided, the
+/// connection is configured as a broadcast.
+LogicalResult buildShimTileToCtrlConnection(
+    IRRewriter &rewriter, AMDAIE::ControlCodeOp controlCodeOp,
+    AMDAIE::TileOp srcTileOp, ArrayRef<TileOp> targetTileOps,
+    DenseMap<Value, ChannelGenerator> &shimTileToGeneratorMap) {
+  // Get the available DMA channel for the shim tile, and assign it for the
+  // packet flow.
+  std::optional<uint8_t> maybeChannel =
+      shimTileToGeneratorMap[srcTileOp.getResult()]
```
jtuyls, line 59: I think it would be good to check here that `srcTileOp.getResult()` exists in `shimTileToGeneratorMap` and/or that `srcTileOp` is a Shim tile and return an error otherwise.

jtuyls (approved):
Very nice! Just a few nits.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEConvertDeviceToControlPackets.cpp
```diff
@@ -21,21 +21,67 @@ using Path = std::filesystem::path;
 namespace mlir::iree_compiler::AMDAIE {
 
 namespace {
+
+/// Recursively erase an operation and all its users.
```
newling, line 25: This seems dangerous. In general it's a bad idea to call erase directly on ops, in this case there is no signaling that an op's users have been destroyed. 

An example, consider the triangle where 
d has no users
b and c are used by d
a is used by b and c


I think if you call recursiveErase on a, you'll be dereferencing nullptrs...

newling: Also IMO I think using recursion is ok (but not great) for moving up-and-down parent ops (op->block->region->op->block->region) but that when traversing ops within a block, we shouldn't use recursion (stack size argument: generally an op has fewer ancestors than preceding ops in a block).

newling: I think the following is safe in a general situation:

+/// erase an operation and all its direct and indirect users.
+void eraseWithAllUsers(Operation *op) {
+  DenseSet<Operation *> toErase;
+  SmallVector<Operation *> toExplore = {op};
+  while (!toExplore.empty()) {
+    Operation *currentOp = toExplore.pop_back_val();
+    toErase.insert(currentOp);
+    for (Operation *userOp : currentOp->getUsers()) {
+      if (toErase.count(userOp) == 0) {
+        toErase.insert(userOp);
+        toExplore.push_back(userOp);
+      }
+    }
+  }
+  for (Operation *op : toErase) {
+    op->erase();
+  }

Yu-Zhewen: Thanks!

PR #1195, opened 2025-03-19
[CI] Fix: Python dict not copying as expected

newling:
- Revert attempt to reuse python dict for different tests, it was not copying as I expected (even with myDict.copy()). This was causing O2 and O3 flags to do strange things, which I spotted in https://nod-ai.github.io/iree-amd-aie/results_history_npu1.html

- Also: add some more information in https://nod-ai.github.io/iree-amd-aie/results_history_npu1.html : number of operations performaned, and optional note for each plot.

### DIFF
--- a/build_tools/ci/cpu_comparison/performance_publish.py
+++ b/build_tools/ci/cpu_comparison/performance_publish.py
@@ -32,15 +32,17 @@ def get_total_ops(name):
     2) return the product of all values in the list that are integers.
     """
     fragments = name.split("_")
-    if (fragments[0] is not "matmul") return None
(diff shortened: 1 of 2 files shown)


Yu-Zhewen:
> * Revert attempt to reuse python dict for different tests, it was not copying as I expected (even with myDict.copy()). This was causing O2 and O3 flags to do strange things, which I spotted in https://nod-ai.github.io/iree-amd-aie/results_history_npu1.html

I think a deep copy is needed here instead of shallow copy
https://stackoverflow.com/a/3975388

PR #1196, opened 2025-03-20
[CI][Python] Deepcopy base dictionary, so list vals get copied 

newling:
O2 vs O3 testing wasn't working correctly before this change.


Yu-Zhewen (approved):
Thanks, LGTM

PR #1197, opened 2025-03-20
[CI] test for script that generates html and add #ops notes

newling:


build_tools/ci/cpu_comparison/performance_publish.py
```diff
@@ -25,6 +26,24 @@ def append_history(results_json_path: str, results_history_path: str):
         json.dump(results_history, f, indent=2)
 
 
+def get_total_ops(name: str):
```
Yu-Zhewen, line 29: Instead of inferring `total_ops` from the test name, can we make it an explicit attribute of the Test class, and then store it in  `results.json`?  

For example, add `self.total_ops = M * N * K` here:
https://github.com/nod-ai/iree-amd-aie/blob/8cd5a574e15c371fff444990654588efa010d64b/build_tools/ci/cpu_comparison/run.py#L240-L264

We could also update `self.total_ops` to account for `call_replication` at this point:

https://github.com/nod-ai/iree-amd-aie/blob/8cd5a574e15c371fff444990654588efa010d64b/build_tools/ci/cpu_comparison/run.py#L2587

newling: Yeah, that'll be better. I'll make this change.

Yu-Zhewen (approved):
LGTM. Thanks!

PR #1202, opened 2025-03-24
[AlignTransferRead] Properly handle cases where offset is a constant zero

Yu-Zhewen:
getValue() returns the attribute type rather than the actual value, causing the comparison to always evaluate to false previously.


newling (approved):
Nice catch, thanks.

PR #1203, opened 2025-03-25
Bump mlir-air to e19dd32108a1d55132bbc115384dc493e3126af4

erwei-xilinx:
- Remove more assertions to potentially fix the random CI failures.


jtuyls:
@erwei-xilinx @yzhang93 I reverted this because it resulted in the same clang error when applied on main. Not sure why it doesn't show up in the PR CI itself..

PR #1204, opened 2025-03-25
[FuseProducerIntoLoop] Fix materializing the operand

Abhishek-Varma:
To be merged after https://github.com/nod-ai/iree-amd-aie/pull/1189

- This commit adds a fix for materializing the operand of the compute
   op inside `iree-amdaie-fuse-producer-into-loop`.
-- Whenever the defining op of an operand is a linalg.pack/copy and is
   not defined in the same block, we simply move that inside the block.
-- Else we try to find a tensor.extract_slice through the use-def chain
   and tile/fuse it.
-- This makes 32x32x128 i32->i32 work on 1x1 for `pack-peel-4-level-tiling`.

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -2238,6 +2238,26 @@ class Tests:
 
         self.register(
             Matmul(
+                32,
+                32,
+                128,
+                "i32",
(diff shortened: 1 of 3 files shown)


jtuyls (approved):
LGTM

PR #1206, opened 2025-03-27
[NFC][e2e CI] Refactor e2e CI Matmul tests

Abhishek-Varma:
-- This commit refactors e2e CI Matmul tests to :-
i.  Enable streamlined addition of new tests
ii. Track all test configs for Phoenix vs Strix.

NOTE: Take a look at [these two tests](https://github.com/nod-ai/iree-amd-aie/blob/d68ec8b21a6b8bfa184825891d796f2fe6539ba9/build_tools/ci/cpu_comparison/run.py#L2099-L2129) which are in `main`- they're simply duplicates and are not checking anything different. By this refactor the goal is to ensure whenever we're adding new tests in the config file we do so based on the number of cores as well as in increasing order of shapes. This way there's lesser chance of test duplication as well as easier to track the config.

### DIFF
--- a/build_tools/ci/cpu_comparison/matmul_test_config.py
+++ b/build_tools/ci/cpu_comparison/matmul_test_config.py
@@ -1,0 +1,209 @@
+# NPU1_4COL matmul test(s):
+npu1_4col_matmul_tests = [
+    # 1x1 core tests.
+    {
+        "M": 128,
+        "N": 128,
+        "K": 128,
+        "input_type": "i8",
+        "acc_type": "i32",
+        "name_suffix": "OneCore_npu1_4col",
+        "additional_labels": ["OneCore"],
+        "aie_compilation_flags": [
+            "--iree-amdaie-num-rows=1",
+            "--iree-amdaie-num-cols=1",
+        ],
+    },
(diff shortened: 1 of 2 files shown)


Yu-Zhewen (commented):
Thanks for cleaning this up. Just a high level question: any particular reason why we are not storing these test configs as json/toml files?

Abhishek-Varma:
> Thanks for cleaning this up. Just a high level question: any particular reason why we are not storing these test configs as json/toml files?

Hi Zhewen.

That's a fair ask. I did think about them but chose to go ahead with just the basic Python interoperability.
Here's why :-
1. For `.toml` I guess we will need `pip install toml` - I didn't want to include any such dependency.
2. And for `.json` it'll involve file reading using `with ...` followed by `json.load(file)` - I didn't want to include any further interoperability logic for these data structures than what Python offers by default. Also the config file contains layers of configs from which the PR aims to only fetch `matmul_tests_for_individual_devices` - so just helps being in the default Python-ways and simply importing the required thing. Can also be done similarly for performance config - could be something for a future PR work.

jtuyls (approved):
LGTM

PR #1208, opened 2025-03-27
Move objFifo depth assignment pass up in pipeline

jtuyls:
Moves `AMDAIEAssignLogicalObjectFifoDepthPass` up in the pipeline so that the depth (double buffering) information is available as soon as possible and before the `AMDAIEAssignTiles` pass. Also fixes a bug in `AMDAIEFlattenLogicalObjectFifo` that would remove the depth.


yzhang93 (approved):
LGTM

PR #1209, opened 2025-03-27
[CI] Add Strix reconfiguration tests

Yu-Zhewen:


build_tools/ci/cpu_comparison/run.py
```diff
@@ -2016,47 +2023,61 @@ def __init__(self):
...
-                test_params=TestParams(
-                    aie_compilation_flags=[
-                        "--iree-amdaie-num-rows=1",
-                        "--iree-amdaie-num-cols=1",
-                    ],
-                    name_suffix="OneCore",
-                ),
+        for target, in_type, out_type in [
+            ["npu1_4col", "i8", "i32"],
+            ["npu4", "i32", "i32"],
```
Yu-Zhewen, line 2028: Switching to i32 inputs for Strix tests since i8 Matmul tests (without uKernel) currently exceed the stack limit.

Error message:
`upper bound for the stack size of the core (col=0, row=2), inferred from the object file, is 1408 bytes. The assigned memory for the stack is 1024 bytes, which is insufficient (1408 > 1024).`

This issue is not related to reconfiguration itself, as plain Matmul tests also hit the same error.

jtuyls (approved):
LGTM

PR #1212, opened 2025-03-28
[CI] Clean up Windows temp files older than 30 days

Yu-Zhewen:
During the compilation process, some log files are dumped to the system temp directory by default. 
https://github.com/nod-ai/iree-amd-aie/blob/357698b70dd329239cbf7d2c28e2683560972c2a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/XCLBinGen.cpp#L564
However, on Windows machines, this directory is not automatically cleaned up, leading to accumulated files that can cause CI failures or significantly slow down runs.


jtuyls (approved):
LGTM

Issue #985, opened 2024-12-12
Issues when increasing L1 K tile size


yzhang93:
I have observed several issues when increasing the L1 K tile size from 32 to 64.

1. Vectorization path slows down significantly (3ms -> 10ms) with default settings, i.e., peano flag O3 and function outlining.
2. Vectorization path has program memory issue if function outlining is disabled.
3. Ukernel path has error `ld.lld: error: undefined symbol: matmul_bf16_bf16_f32_32x32x64_4x8x4` with function outlining enabled.
4. Ukernel path doesn't have the above error if function outlining is disabled, however the performance doesn't change.

CC: @newling @jtuyls

newling:
@yzhang93 can you please point me to the line to change (32 to 64) to experiment with this?

PR #1214, opened 2025-04-01
[LoweringStrategy] Refactor the logic of L1 tile size selection

yzhang93:
This is a WIP. Refactor the logic to make the most use of L1 memory. The result of the change is using a larger k tile size, however, this causes many failure of tests. Also increasing k tile size can cause performance regression and we haven't found the cause, ref issue https://github.com/nod-ai/iree-amd-aie/issues/985. 

Update:
1. Tested vectorization path with k tile size as 128. 
- For bf16->f32 tests, all could compile with outline, however the performance degrades about 20% compared to k = 32. 
- For i8 to i32, all could compile, but some tests generates intermittent or consistent numeric error. 
- So for now using max k tile size with ukernel path, while keeping smaller size for vectorization path.

2. For ukernel with larger tile sizes, I see around 15~30% performance improvement on Phoenix and around 10%~23% improvement on Strix. For example,

Before

```
matmul_512_4096_512_bf16_f32_O3_npu1_4col_outline_ukernel_chess_benchmark
--------------------------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations UserCounters...
--------------------------------------------------------------------------------------------------
BM_matmul/process_time/real_time_mean         2694 us         52.3 us            5 items_per_second=371.513/s
BM_matmul/process_time/real_time_median       2696 us         50.9 us            5 items_per_second=370.985/s
BM_matmul/process_time/real_time_stddev       92.9 us         3.26 us            5 items_per_second=12.9349/s
--------------------------------------------------------------------------------------------------

```
After

```
matmul_512_4096_512_bf16_f32_O3_npu1_4col_outline_ukernel_chess_benchmark
--------------------------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations UserCounters...
--------------------------------------------------------------------------------------------------
BM_matmul/process_time/real_time_mean         1877 us         55.0 us            5 items_per_second=532.731/s
BM_matmul/process_time/real_time_median       1876 us         56.1 us            5 items_per_second=532.923/s
BM_matmul/process_time/real_time_stddev       11.9 us         2.46 us            5 items_per_second=3.39164/s
--------------------------------------------------------------------------------------------------
--------------------------------------------------------------------------------------------------
```


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -224,96 +224,88 @@ FailureOr<ParameterSetting> ParameterSetting::create(
...
+
+  // Get the level 1 pack size, i.e., vector instruction size.
+  auto maybePackedSize = getPackedSize(linalgOp, M, N, K, targetDevice);
+  if (failed(maybePackedSize)) return failure();
+  auto [m1Pack, n1Pack, k1Pack] = maybePackedSize.value();
 
-  // Consider trunci op that converts i32 to i8 after matmul ops, then the scale
-  // factor can be doubled to make the most use of L1 memory.
+  // MLIR-AIR disable double buffering in L1 memory.
+  unsigned isDoubleBufferA = isObjectFifo ? 2 : 1;
```
Yu-Zhewen, line 250: Just from the name, `isDoubleBufferA` implies it’s a boolean type. Since it's actually representing the buffer depth (e.g., 1 or 2), consider renaming it to something like `bufferDepthA` for clarity.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIETileSizeSelectionUtils.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include "AMDAIETileSizeSelectionUtils.h"
+
+#include "AMDAIEUtils.h"
+
+namespace mlir::iree_compiler::AMDAIE {
+
+void findLargestTileSizes(uint32_t m, uint32_t n, uint32_t k, uint32_t& curMax,
+                          const TileParams& params, TileSize& best) {
+  if (m < 16 || n < 16 || k < 16) return;
```
Yu-Zhewen, line 15:
```suggestion
  if (m < best.M || n < best.N || k < best.K) return;
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -224,96 +224,88 @@ FailureOr<ParameterSetting> ParameterSetting::create(
...
-  uint32_t maxL1SizeN = 16 * scaleFactor;
-  uint32_t M1 = findLargestFactor(M / m1Pack, maxL1SizeM / m1Pack, m1Pack);
-  uint32_t N1 = findLargestFactor(N / n1Pack, maxL1SizeN / n1Pack, n1Pack);
-
-  uint32_t maxL0SizeM = numRows * maxL1SizeM;
-  uint32_t maxL0SizeN = numCols * maxL1SizeM;
-  uint32_t M0 = findLargestFactor(M, maxL0SizeM, m1Pack * M1);
-  uint32_t N0 = findLargestFactor(N, maxL0SizeN, n1Pack * N1);
+  // Get the largest tile sizes that can fit in L1 memory.
+  TileParams tileParams{/*memoryLimit=*/65536,
```
Yu-Zhewen, line 270: Instead of hardcoding 65536 here, can we replace it with `AMDAIEDeviceModel::getCoreTileLocalMemorySize()`?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIETileSizeSelectionUtils.cpp
yzhang93: I don't think these can be replaced. Here we'd want a fixed minimum threshold for exiting the recursion.

Yu-Zhewen: I see. I missed the recursive part of this function. Just I noticed the value 16 appears both here and in `selectL1TileSizes`.  If they are related, it might be worth defining them as macros or const variables, for better maintainability and readability?

Yu-Zhewen (approved):
Thanks! LGTM, only one more comment.

PR #1215, opened 2025-04-02
[Router] Replace `PathEndPoint` with `PhysPort` everywhere

Yu-Zhewen:
`PathEndPoint` and `PhysPort` are two very similar data structures, with the only difference being that `PhysPort` includes an additional `direction` member. 

Note this PR does affect routing results in some unit tests due to the inclusion of the extra `direction` field, which impacts sorting-related behavior.


jtuyls:
> Note this PR does affect routing results in some unit tests due to the inclusion of the extra `direction` field, which impacts sorting-related behavior.

Why does that field impact the sorting and is that an improvement or it doesn't matter?

Yu-Zhewen:
> > Note this PR does affect routing results in some unit tests due to the inclusion of the extra `direction` field, which impacts sorting-related behavior.
> 
> Why does that field impact the sorting and is that an improvement or it doesn't matter?

For "sorting-related," I was mainly referring to the use of `std::sort` and `std::map` in the router. I also believe the inclusion of the direction field is an improvement. Otherwise, two ports with the same bundle name and channel ID but different directions cannot be distinguished.

PR #1210, opened 2025-03-27
[InsertDmaBdChain] Do not insert interleaved BD chains to maintain control packet order

Yu-Zhewen:
Interleaved BD chains may alter the issuing order of control packets, potentially leading to incorrect configurations. Disabled for now to ensure correctness.

PR #1218, opened 2025-04-07
[e2e CI] Add 4096x128x32 | 1x1| Matmul | i32 | pack-peel-4-level-tiling

Abhishek-Varma:
**NEW**
-- With increase and better usage of L1 memory's usage the issue with DMA BD
   chain is not visible with 4096x128x32. The input to `insert-dma-bd-chain` pass
   has changed!
-- Still adding this to the e2e CI and would track the next shape which leads to the same
   issue.

**OUTDATED (won't be part of commit history when merging - this is just to keep a track!)**
-- This commit adds a fix to `iree-amdaie-insert-dma-bd-chain` to cap the
   DMA BD chain's length with maximum queue size.
-- This is done because later in the stack `iree-amdaie-fold-dma-waits` makes
   use of the maximum queue size to fold away DMA waits. This causes an issue
   because of misaligned DMA chain + DMA wait.
-- With this fix e2e `4096x128x32` Matmul on `1x1` on `Strix` works for
   `pack-peel-4-level-tiling`.


Yu-Zhewen:
From my understanding, the maximum DMA BD chain length and the maximum queue size are independent. Regardless of how many BDs are in a chain, at any given time, the hardware queue only holds one BD per chain. Only when that BD is executed (popped from the queue), the hardware will check if `next_bd` is set. If so, the next_bd is then pushed to the queue to continue the chain. 

That means the queue size determines how many chains we can have, not the length of each chain.

If this fix is related to the numerical errors, one alternative that might help is setting the `enableInterleave` option to false in the pass, which would prevent having multiple chains running in parallel.
https://github.com/nod-ai/iree-amd-aie/blob/1d1d06e8944318cb6efccfa382def298a3076b48/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.td#L410

The risk with having multiple chains is that for each chain, we need to link all the BDs together before execution can begin. If two chains have different lengths, the execution of the longer chain will be delayed. I previously encountered a similar issue with control packets (see [#1210](https://github.com/nod-ai/iree-amd-aie/pull/1210)). For matmuls, I was expecting it wouldn't be an issue (_since the BD is assigned before the control code loop unroll, which should potentially align the chain lengths_), though this might not always be the case...

Issue #700, opened 2024-08-23
Multi-kernel tests (plus a few others) don't work with driver date 20240819 


makslevental:
The Windows E2E PR #689 is blocked because in order to get Windows to work I had to use a particular commit of XRT (required by XRT-MCDM). At that commit, the current linux driver (date 20240819) fails some tests while the older driver (such as the one on the sharkbox phoenix nuc) passes. 

Possibly there's some connection to https://github.com/nod-ai/iree-amd-aie/blob/main/runtime/src/iree-amd-aie/driver/xrt/native_executable.cc#L270, which I had to `ifndef _WIN32` away on Windows (which @jtuyls added when battling numerics issues) but it's not clear since it's not only the multi-kernel tests that fail.

MaheshRavishankar:
Could you call out specific tests... They might not all get resolved in one go.

makslevental:
> Could you call out specific tests... They might not all get resolved in one go.

```python
  for name in [
      "two_matmul_switching",
      "matmul_f32_8_8_4",
      "matmul_f32_8_4_8",
  ]:
      aie_vs_llvm_cpu(config, test_files_dir / f"{name}.mlir")

  aie_vs_llvm_cpu(
      config,
      test_files_dir / "three_matmuls.mlir",
      function_name="three_$mm$",
  )

  # Test(s) of the form matmul(A,B) where A:MxK, B:KxN
  test_name = output_dir / "test_from_template.mlir"
  template_name = matmul_template_dir / "matmul_MxK_KxN.mlir"
  generate_matmul_test(test_name, template_name, 32, 32, 64, "bf16", "f32")
  aie_vs_llvm_cpu(config, test_name)

  # Test(s) of the form matmul(A,B) + C where A:MxK, B:KxN, C:N
  test_name = output_dir / "test_from_template_bias_N.mlir"
  template_name = matmul_template_dir / "matmul_bias_MxK_KxN_N.mlir"
  generate_matmul_test(
      test_name, template_name, 1024, 1024, 512, "bf16", "f32"
  )
  if config.vitis_dir:
      aie_vs_llvm_cpu(
          config, test_name, tile_pipeline="pack-peel", use_ukernel=True
      )
  aie_vs_llvm_cpu(
      config, test_name, tile_pipeline="pack-peel", use_ukernel=False
  )
```

makslevental:
[...]

PR #1220, opened 2025-04-07
Bump IREE to iree-org/iree@5e2e37d4

yzhang93:

PR #1223, opened 2025-04-10
[CI] Add multi-dispatch tests back

Yu-Zhewen:
Reverts #1110 and fix multi-dispatch tests on linux (both phenoix and strix are supported). Tests remain disabled on windows due to #700.


runtime/src/iree-amd-aie/driver/xrt-lite/device.cc
```diff
@@ -202,6 +202,17 @@ static iree_status_t iree_hal_xrt_lite_device_queue_alloca(
   return iree_ok_status();
 }
 
+static iree_status_t iree_hal_xrt_lite_device_queue_dealloca(
```
yzhang93, line 205: Heads up, this should be updated by adding another argument `iree_hal_alloca_flags_t flags` after the IREE bump merged. Ref https://github.com/nod-ai/iree-amd-aie/pull/1220/files#diff-33d7a4a056b0c413c9edc400cc4a454dc156c59783f6606c6b60e449a093f456R267

jtuyls (approved):
LGTM, thanks.

PR #1226, opened 2025-04-11
[AssignTiles] Update algo to take into account tile memory limitations

jtuyls:
Resubmitting https://github.com/nod-ai/iree-amd-aie/pull/1222. I haven't been able to repro the issue locally, but it seems to be related to peano, so submitting the tile assignment changes with test updated to chess for now.

This PR increases the usage of L2 memory by increasing the L2 tile sizes within memory bounds and by improving the algorithm to assign logical objectFifos across tiles. This updated algorithm now assigns larger logical objectFifos first and takes into account the actual memory usage on tiles, with leads to better usage of available memory.


Yu-Zhewen (approved):
LGTM. Thanks!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lowering_strategy_objectfifo_npu1.mlir
```diff
@@ -48,7 +48,7 @@ module {
 // CHECK-4x4{LITERAL}: #config = #iree_codegen.lowering_config<tile_sizes = [[128, 128, 0], [0, 0, 1], [1, 1, 0]]>
 // CHECK-4x4{LITERAL}: #amdaie.packing_config<packing_config = [{packedSizes = [32, 32, 64], transposePackIndices = [0, 1, 2], unpackEmpty = [false, false, true], innerPerm = [[0, 1], [1, 0], [0, 1]], outerPerm = [[0, 1], [1, 0], [1, 0]]}, {packedSizes = [0, 0, 0, 4, 4, 8], transposePackIndices = [0, 1, 2], unpackEmpty = [false, false, true], innerPerm = [[0, 1], [1, 0], [0, 1]], outerPerm = [[0, 1, 3, 2], [0, 1, 3, 2], [0, 1, 3, 2]]}]>
 
-// PACK-PEEL-4-LEVEL{LITERAL}: #config = #iree_codegen.lowering_config<tile_sizes = [[256, 256, 0], [4, 4, 0], [0, 0, 1], [1, 1, 0]]>
+// PACK-PEEL-4-LEVEL{LITERAL}: #config = #iree_codegen.lowering_config<tile_sizes = [[128, 128, 0], [4, 4, 0], [0, 0, 1], [1, 1, 0]]>
```
yzhang93, line 51: Why are these tile sizes reduced? If the outer tile sizes are reduced to 128, I think the second loop is gone, and it will be the same as the pack-peel pipeline?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -202,13 +204,14 @@ FailureOr<ParameterSetting> ParameterSetting::create(
     uint32_t kPackScaleL1) {
   auto initType =
       llvm::cast<ShapedType>(linalgOp.getDpsInitOperand(0)->get().getType());
-  unsigned nBytesInit = initType.getElementTypeBitWidth() / 8;
+  uint32_t nBytesInit = initType.getElementTypeBitWidth() / 8;
+  uint32_t nBytesOut = nBytesInit;
```
yzhang93, line 208: I'm not clear why this change is made. I think previously `nBytesInit` is for matmul output, and `nBytesElemOut` is not zero only when there are elementwise ops. If you use both `nBytesInit` and `nBytesOut` for pure matmul, it will count extra memory when calculating L1 memory usage?

yzhang93 (dismissed):
I have some questions regarding your changes on tile size. (sorry, I didn't looked into the details previously).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -486,20 +489,30 @@ static LogicalResult setRootConfigForPackPeel4LevelTilingPipeline(
...
+      (l2SizeA + l2SizeOut);
+  if (largestM0ScaleFactor % 2 != 0)
+    largestM0ScaleFactor = llvm::PowerOf2Ceil(largestM0ScaleFactor) / 2;
+  int64_t scaleL0M =
+      (isObjectFifo && !isBatchMatmul) ? largestM0ScaleFactor : 1;
+  int64_t scaleL0N = (isObjectFifo && !isBatchMatmul && fitsInL2) ? 2 : 1;
+  int64_t m0Tile = std::min(static_cast<uint32_t>(packPeelTiling.M0 * scaleL0M),
+                            packPeelTiling.M);
+  int64_t n0Tile = std::min(static_cast<uint32_t>(packPeelTiling.N0 * scaleL0N),
+                            packPeelTiling.N);
```
yzhang93, line 515: I couldn't follow the logic here. Could you add some explanations? Also I don't think you need to check `isBatchMatmul` since the tile size for batch dimension in L2 is always 1.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
jtuyls: Thanks for catching that! I think we can solve it by initializing `nBytesOut` to `0`, so I updated it.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
jtuyls: I added a comment for this. As for `isBatchMatmul`, this was added because batch matmul failed with 'pack-peel-4-level-tiling' and as I think enabling that is out-of-scope of this PR, so I added a TODO for it.

yzhang93:
Is it possible to split this PR into two? One is for changes in `AssignTiles` pass, and the other is for those tile size selection in `KernelDispatch`? I also have a draft PR that refactors the L2 tile size selection and was based on the main. I think we should collaborate on this effort and take small steps once a time.

jtuyls:
> Is it possible to split this PR into two? One is for changes in `AssignTiles` pass, and the other is for those tile size selection in `KernelDispatch`? I also have a draft PR that refactors the L2 tile size selection and was based on the main. I think we should collaborate on this effort and take small steps one at a time.

I think these changes need to go together as this actually makes use of the assign tiles changes to achieve best L2 usage. Without it, I can't ensure that this works e2e. I think you could rebase your changes on top of this and make sure that this case I am adding stays functional.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
jtuyls: Actually, there was no double count, so I reverted it back as I need `nBytesOut` to compute L2 usage.

yzhang93: Why there's no double count? The calculation is as
```
uint32_t A = params.numBytesA * m * k * params.bufferDepthA;
uint32_t B = params.numBytesB * n * k * params.bufferDepthB;
uint32_t C = params.numBytesC * m * n * params.bufferDepthC;
uint32_t Acc = params.numBytesAcc * m * n * params.bufferDepthAcc;
int64_t memoryUsage = A + B + C + Acc;
```
If `params.bufferDepthC` is not zero for matmul, I think it will count more memory usage.

yzhang93: I know you need `nBytesOut` for L2 memory calculation.  I think you can keep `nBytesElemOut` while adding a new variable as `nBytesOut` after L1 memory usage calculation as
`uint32_t nBytesOut = isMatmulWithElementwiseConsumer(linalgOp) ? nBytesElemOut : nBytesInit`.

jtuyls: Oh, I see what you mean. Yeah, there is some potential double count. I updated it to keep nBytesElemOut and pass that for L1 memory calculation.

yzhang93 (approved):
Thanks for the change! LGTM

PR #1228, opened 2025-04-15
[CtrlPkt] Support control packets for both single and multi-dispatch tests

Yu-Zhewen:
This PR adds support for using control packets in multi-dispatch tests, where the PDI now contains only routing-related configuration and needs to be loaded only once.  The original flow with multiple PDI reloading is still supported, and these two flows can be switched using the `enableCtrlPkt` flag.

- New Flow (Control Packet Enabled)
`Load base PDI → Control packet for Kernel A → Launch Kernel A → Control packet for Kernel B → Launch Kernel B → ...`

- Old Flow (Control Packet Disabled)
`Load PDI for Kernel A → Launch Kernel A → Load PDI for Kernel B → Launch Kernel B → ...`

Note the single-dispatch tests also support both flows, and potentially, everything could switch to the new flow in the future.

This PR also removes the hand-crafted matmul constant bias tests for control packets. These tests are no longer needed, as the new flow already covers both the numerical correctness check, and reconfiguration time benchmarking.


build_tools/ci/cpu_comparison/run.py
```diff
@@ -1267,22 +1124,41 @@ def benchmark_aie_kernel_time(
...
+        )
+
+    """
+    Check consistency only for values > 0 (or >1 for PDI loads)
+    Valid Examples:
+    - n_kernel_runs=0, n_reconfigure_runs=0, n_pdi_loads=10 (measure PDI load time)
+    - n_kernel_runs=10, n_reconfigure_runs=0, n_pdi_loads=1 (measure kernel time)
+    - n_kernel_runs=0, n_reconfigure_runs=10, n_pdi_loads=1 (measure reconfigure time)
+    - n_kernel_runs=10, n_reconfigure_runs=10, n_pdi_loads=1 (measure kernel and reconfigure time)
+    """
```
jtuyls, line 1148: Prefer `#` for block comments: https://peps.python.org/pep-0008/#block-comments
```suggestion
    # Check consistency only for values > 0 (or > 1 for PDI loads)
    # Valid Examples:
    # - n_kernel_runs=0, n_reconfigure_runs=0, n_pdi_loads=10 (measure PDI load time)
    # - n_kernel_runs=10, n_reconfigure_runs=0, n_pdi_loads=1 (measure kernel time)
    # - n_kernel_runs=0, n_reconfigure_runs=10, n_pdi_loads=1 (measure reconfigure time)
    # - n_kernel_runs=10, n_reconfigure_runs=10, n_pdi_loads=1 (measure kernel and reconfigure time)
```

runtime/src/iree-amd-aie/driver/xrt-lite/direct_command_buffer.cc
```diff
@@ -259,14 +259,23 @@ static iree_status_t iree_hal_xrt_lite_direct_command_buffer_dispatch(
...
-      context.open_cu_context(kernel_params.kernel_name);
 
   size_t num_reconfigurations = kernel_params.reconf_data_runlist.size();
+  // Only create the context when either the context does not exist or no
+  // control packet reconfiguration is present.
+  bool create_context =
+      (executable->context == nullptr || num_reconfigurations == 0);
+  shim_xdna::cuidx_t cu_idx{.index = 0};
+  if (create_context) {
+    for (size_t i = 0; i < kernel_params.n_pdi_loads; i++) {
```
jtuyls, line 270: Did you perform any experiments on how much time this takes on average?

Yu-Zhewen: On Strix, PDI loading takes ~330 µs, while control packet-based reconfiguration is faster (~80–150 µs). However, note that even with control packets, a small PDI is still required to configure the switches—this step alone takes ~305 µs.

On Phoenix, PDI loading takes ~810 µs, and control packet-based reconfiguration is even slower (~1200 µs).

My conclusion is: For all multi-dispatch tests on Strix, it's preferable to use control packets by default. We only need to load the PDI once, and subsequent reconfigurations using control packets are fast enough to justify this approach.

build_tools/ci/cpu_comparison/run.py
Yu-Zhewen: Done.

jtuyls (approved):
LGTM

PR #1229, opened 2025-04-15
[Driver] Fix propagating binding_table

jpy794:
`iree_hal_xrt_lite_device_queue_execute` always executes with an empty `binding_table`, which breaks kernel argument passing. This change would fix it.


jpy794:
Sure, applied to `xrt` too.

Yu-Zhewen (approved):
Thanks for the contribution! LGTM

PR #1231, opened 2025-04-17
Refactor L2 tile size selection for 4-level-tiling pipeline

yzhang93:
The new logic automatically selects the largest L2 tile sizes that can fit in the memTile memory space. I have checked that the tile sizes generated by the new logic are the same as the ones used in https://github.com/nod-ai/iree-amd-aie/pull/1226.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -290,15 +282,28 @@ FailureOr<ParameterSetting> ParameterSetting::create(
...
   uint32_t n0Pack = (N0 / numCols) % n1Pack == 0 ? (N0 / numCols) : N0;
 
-  // In pack-peel pipeline there is only one level of tiling for K dimension,
-  // so set K1 = 0. The packed outer K dimension needs to be 1, so set K0 = 1.
+  // For pack-peel-4-level-tiling pipeline, we use the largest tile sizes that
+  // can fit in the MemTile memory.
+  if (kPackScaleL1 == 2 && isObjectFifo) {
+    tileParams.memoryLimit = deviceModel.getMemTileSizeInBytes() * numCols;
+    tileParams.bufferDepthAcc =
+        isMatmulWithElementwiseConsumer(linalgOp) ? 0 : 2;
```
jtuyls, line 299: Why is `bufferDepthAcc` equal to `0` if we have an elementwise consumer? We should still need an accumulator in that case so should be `1`?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/AMDAIETileSizeSelectionUtilsTest.cpp
```diff
@@ -67,6 +67,36 @@ TEST(SelectTileSizeTest, L1TileSizeTest) {
...
+          {524288, 2, 2, 0, 4, 2, 2, 0, 2, 512, 512, 64, 4, 4, 8}, 32, 32)),
+      (TileSize{256, 128, 64}));
+  EXPECT_EQ(
+      (selectL2TileSizes(
+          {524288, 1, 1, 0, 4, 2, 2, 0, 2, 1024, 1024, 64, 4, 4, 8}, 32, 32)),
+      (TileSize{256, 128, 64}));
+  EXPECT_EQ((selectL2TileSizes({524288 * 8, 1, 1, 1, 2, 2, 2, 2, 0, 1024,
+                                4096 * 4, 512, 4, 4, 8},
+                               64, 64)),
+            (TileSize{1024, 1024, 512}));
```
jtuyls, line 97: Could you add some comment to each test case to explain what case it is exactly trying to test and make sure we're covering most cases as well as add some edge cases? For example: are we testing non-divisible inputs, what if memory limit is `0` etc?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
yzhang93: This is just a trick to remove the memory computation for matmul output since the real output is from elementwise op in this case.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/AMDAIETileSizeSelectionUtilsTest.cpp
jtuyls: You do check for it here: https://github.com/nod-ai/iree-amd-aie/pull/1231/files#diff-88fce979e5c13b4309da836589d63398f0d479e819ed81e18a251418b0f5132bR75 ?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
jtuyls: > output buffer from L1 to L2 is only from elementwise op
Even though that's correct, there is still an internal accumulator buffer used (depth 1), so I don't think this is correct and you should still take into account that buffer as well.

yzhang93: I understand there's an accumulator buffer in L1, but why there's also one in L2? There's some preserved memory space for that?

yzhang93: Okay, I think it's easy to get confused because I was reusing the same `TileParams` struct when calculating L2 memory usage. 

`tileParams.bufferDepthAcc` should not always be 0, because we have to use it for pure matmul. In the pure matmul case, because `nBytesElemOut` = 0, it makes this to be zero regardless of `bufferDepthC` value.
`uint32_t C = params.numBytesC * m * n * params.bufferDepthC;`

While for `matmul + elementwise` case, I have to use the trick to make this zero:
`uint32_t Acc = params.numBytesAcc * m * n * params.bufferDepthAcc;
`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/AMDAIETileSizeSelectionUtilsTest.cpp
yzhang93: I added new tests for that.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
jtuyls: > tileParams.bufferDepthAcc should not always be 0, because we have to use it for pure matmul. In the pure matmul case, because nBytesElemOut = 0, it makes uint32_t C = params.numBytesC * m * n * params.bufferDepthC to be zero regardless of bufferDepthC value. So we have to keep bufferDepthAcc for the output buffer.

The use of `Acc` for L2 is confusing, I think it would be more clear to always use `nBytesElemOut` / `bufferDepthC ` for both L1 and L2 and set `bufferDepthAcc` to `0` for L1 if not followed by an elementwise operation.

yzhang93: I've made changes as suggested. Also removed `Acc` in L2 memory calculation.

PR #1239, opened 2025-04-22
[IR] Unify the definition of `repeat_count`

Yu-Zhewen:
Previously,  we have the following semantics of `repeat_count` across different ops/dialects

- `aie.dma_start`, used for memtile and coretile memory DMA, where `repeat_count = 0` means "do it once".
- `aiex.npu.push_queue`, used in shim DMA control code generated from the AIR pipeline. Also treats `repeat_count = 0` as "do it once".
- `amdaie.npu.push_to_queue`, used in shim DMA control code generated from the objectfifo pipeline. In this case, `repeat_count = 1` means "do it once".

Also, when using aie-rt APIs, they interpret `repeat_count = 1` as "do it once", and then internally subtract 1 for hardware encoding purposes.

In this PR, we standardize on `repeat_count = 1` to mean "do it once". This simplifies our logic by avoiding the need for scattered +1/-1 adjustments throughout the codebase. The only remaining case that requires special handling is `aie.dma_start` ops generated from MLIR-AIR, and the new `AMDAIEIncrementRepeatCount` pass is designed for handling this semantic mismatch.


compiler/plugins/target/AMD-AIE/aie/AMDAIEDmaToNpu.cpp
```diff
@@ -244,7 +244,7 @@ struct PushToNpuPattern : OpConversionPattern<NpuPushQueueOp> {
 
     // the value to write
     uint32_t bd_id = op.getBdId();
-    uint32_t repeat_cnt = op.getRepeatCount();
+    uint32_t repeat_cnt = op.getRepeatCount() - 1;
```
Yu-Zhewen, line 247: The "-1" can be hidden if we can also use `aie-rt` here to generate these transactions. Planning to address that in a separate PR.

Yu-Zhewen, line 247: The "-1" here can be hidden if we can also use aie-rt to generate these transactions. Planning to address that in a separate PR.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/test/add_12_i8_using_2d_dma_op_with_padding.mlir
```diff
@@ -38,7 +38,7 @@ module {
 // CHECK: XAIE API: XAie_DmaSetNextBd with args: &dmaDesc=ptr, nextBdId.value()=0, enableNextBd=1
 // CHECK: XAIE API: XAie_DmaEnableBd with args: &dmaDesc=ptr
 // CHECK: XAIE API: XAie_DmaWriteBd with args: devInst=ptr, &dmaDesc=ptr, tileLoc=TileLoc(col: 0, row: 1), bdId=0
-// CHECK: XAIE API: XAie_DmaChannelSetStartQueue with args: devInst=ptr, tileLoc=TileLoc(col: 0, row: 1), chNum=0, direction=0, bdId=0, repeatCount=2, enTokenIssue=0
+// CHECK: XAIE API: XAie_DmaChannelSetStartQueue with args: devInst=ptr, tileLoc=TileLoc(col: 0, row: 1), chNum=0, direction=0, bdId=0, repeatCount=1, enTokenIssue=0
```
Yu-Zhewen, line 41: I believe the intention for these unit tests was to have the DMA BD run only once, which means the previous checks were incorrect?

jtuyls: Yeah, shows how the encoding shouldn't be exposed at this level as it's hard to interpret, introduces inconsistencies and bugs...

jtuyls (approved):
LGTM, very nice clean up!

compiler/plugins/target/AMD-AIE/aie/AMDAIEDmaToNpu.cpp
Yu-Zhewen: Addressed by #1240.

PR #1242, opened 2025-04-24
[AssignNpuDmaBdIds] Equally distribute BD IDs amongst DMAs in same scf.for block/tile

Abhishek-Varma:
-- This commit updates `iree-amdaie-npu-dma-bd-ids` pass to ensure
   equal distribution of BD IDs amongst DMAs within same scf.for block and
   operating on same tile.
-- This fixes the issue of misalignment of DMA ops which causes issues
    further down the stack in `insert-dma-chain` optimization pass.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
@@ -176,17 +176,19 @@ class TileDmaBatchGraph {
   /// This API is invoked when we bump into a DMA wait op. That means we have
   /// reached the end of the current DmaBatch and can start a new DmaBatch in
   /// the chain for the corresponding tile.
-  void updateLatestTileDmaBatch(AMDAIE::TileOp tile) {
-    std::unique_ptr<DmaBatch> &oldLastDmaBatch = getLastDmaBatch(tile);
+  void appendNewBatch(AMDAIE::TileOp tile) {
+    // Get parent for op of the last DMA batch in order to set the same for the
+    // newly created DMA batch.
+    scf::ForOp forOpParent = getLastDmaBatch(tile)->forOpParent;
     tileDmaBatchGraph[tile].push_back(std::make_unique<DmaBatch>());
     std::unique_ptr<DmaBatch> &newLastDmaBatch = getLastDmaBatch(tile);
-    newLastDmaBatch->forOpParent = oldLastDmaBatch->forOpParent;
+    newLastDmaBatch->forOpParent = forOpParent;
   };
 
   /// Given another `TileDmaBatchGraph` instance, we consider that as the nested
   /// graph to be added to the current graph instance. We do this by updating
   /// the `immediateInnerBatches` of the corresponding tile's DmaBatch.
-  void updateInnerBatchesOfCurrentTileDmaBatchGraph(
+  void updateInnerBatchesOfLastBatch(
       TileDmaBatchGraph &innerTileDmaBatchGraph) {
     for (auto &[tile, dmaBatchVec] : innerTileDmaBatchGraph.getGraph()) {
       for (std::unique_ptr<DmaBatch> &dmaBatch : dmaBatchVec) {
@@ -199,7 +201,7 @@ class TileDmaBatchGraph {
 
   /// Traverse each DmaBatch in a given (Tile -> list_of_DmaBatches) and infer
   /// Bd Ids required for it.
-  void inferBdIdsRequiredInBatches() {
+  void inferRequiredBdIds() {
     for (auto &[tile, dmaBatchVec] : getGraph()) {
       for (std::unique_ptr<DmaBatch> &dmaBatch : dmaBatchVec) {
         if (dmaBatch->isEmpty()) continue;
@@ -247,7 +249,7 @@ class BdIdAssignmentUtil {
  public:
   BdIdAssignmentUtil(
       DenseMap<Value, ChannelBdIdGenerator> shimTileToGeneratorMap)
-      : shimTileToGeneratorMap(shimTileToGeneratorMap) {}
+      : shimTileToGeneratorMap(std::move(shimTileToGeneratorMap)) {}
 
   DenseMap<AMDAIE::NpuDmaCpyNdOp, SmallVector<AMDAIE::BdIdOp, 2>> &
   getDmaOpToBdIdMap() {
@@ -548,8 +550,7 @@ static TileDmaBatchGraph createTileDmaBatchGraph(
   auto processOpWithNestedRegion = [&](Operation *op) {
     TileDmaBatchGraph innerTileDmaBatchGraph =
         createTileDmaBatchGraph(workgroupOp, op, shimTileToGeneratorMap);
-    tileDmaBatchGraph.updateInnerBatchesOfCurrentTileDmaBatchGraph(
-        innerTileDmaBatchGraph);
+    tileDmaBatchGraph.updateInnerBatchesOfLastBatch(innerTileDmaBatchGraph);
   };
 
   auto processNpuDmaCpyNdOp = [&](AMDAIE::NpuDmaCpyNdOp dmaOp) {
@@ -578,15 +579,13 @@ static TileDmaBatchGraph createTileDmaBatchGraph(
           FailureOr<AMDAIE::TileOp> tile =
               getGeneratorTileOp<CopyOpOperateOn::Source>(
                   npuDmaOp, shimTileToGeneratorMap);
-          if (succeeded(tile))
-            tileDmaBatchGraph.updateLatestTileDmaBatch(*tile);
+          if (succeeded(tile)) tileDmaBatchGraph.appendNewBatch(*tile);
         }
         if (npuDmaOp.getTarget()) {
           FailureOr<AMDAIE::TileOp> tile =
               getGeneratorTileOp<CopyOpOperateOn::Target>(
                   npuDmaOp, shimTileToGeneratorMap);
-          if (succeeded(tile))
-            tileDmaBatchGraph.updateLatestTileDmaBatch(*tile);
+          if (succeeded(tile)) tileDmaBatchGraph.appendNewBatch(*tile);
         }
       }
     }
@@ -631,7 +630,7 @@ LogicalResult assignNpuDmaBdIds(AMDAIE::WorkgroupOp workgroupOp) {
   DenseMap<AMDAIE::NpuDmaCpyNdOp, SmallVector<AMDAIE::BdIdOp>> dmaOpToBdIdMap;
   TileDmaBatchGraph tileDmaBatchGraph = createTileDmaBatchGraph(
       workgroupOp, controlCodeOp, shimTileToGeneratorMap);
-  tileDmaBatchGraph.inferBdIdsRequiredInBatches();
+  tileDmaBatchGraph.inferRequiredBdIds();
   if (failed(bdIdAssignmentUtil.assignRequiredBdIdsInBatch(
           rewriter, tileDmaBatchGraph))) {
     return failure();
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
@@ -75,9 +75,11 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
   }
 }
 
-/// Given a list of DMA Ops and the number of required BD IDs by the first DMA
-/// op in the list, split the available BD IDs equally amongst all and assign it
-/// to the DMA ops.
+/// A DmaBatch contains the following :-
+///   1. DmaOps within the current batch.
+///   2. Required Bd Ids by the current batch.
+///   3. Nested DmaBatch chain.
+///   4. Pointer to the sibling DmaBatch.
 typedef struct DmaBatchStruct {
   SmallVector<AMDAIE::NpuDmaCpyNdOp> currentDmaOps = {};
   int32_t requiredBdIds = 0;
@@ -85,39 +87,46 @@ typedef struct DmaBatchStruct {
   DmaBatchStruct *nextDmaBatch = nullptr;
 } DmaBatch;
 
-using ControlCodeGraph = llvm::MapVector<AMDAIE::TileOp, DmaBatch *>;
-static ControlCodeGraph initControlCodeGraph(AMDAIE::WorkgroupOp workgroupOp) {
-  ControlCodeGraph controlCodeGraph;
+using TileDmaBatchGraph = llvm::MapVector<AMDAIE::TileOp, DmaBatch *>;
+
+/// Create a new TileDmaBatchGraph by traversing over each tiles in a workgroup.
+static TileDmaBatchGraph initTileDmaBatchGraph(
+    AMDAIE::WorkgroupOp workgroupOp) {
+  TileDmaBatchGraph tileDmaBatchGraph;
   workgroupOp.walk([&](AMDAIE::TileOp tile) {
-    if (controlCodeGraph.contains(tile)) return WalkResult::skip();
-    controlCodeGraph[tile] = new DmaBatch();
+    if (tileDmaBatchGraph.contains(tile)) return WalkResult::skip();
+    tileDmaBatchGraph[tile] = new DmaBatch();
     return WalkResult::advance();
   });
-  return controlCodeGraph;
+  return tileDmaBatchGraph;
 }
 
-static void updateCurrentControlCodeGraph(
+/// A DmaBatch with no DmaOps and no nested DmaBatches is an empty DmaBatch.
+static bool isEmptyDmaBatch(DmaBatch *dmaBatch) {
+  return (dmaBatch->currentDmaOps.empty() &&
+          dmaBatch->immediateInnerBatches.empty());
+}
+
+static void updateInnerBatchesOfCurrentTileDmaBatchGraph(
     DenseMap<AMDAIE::TileOp, DmaBatch *> &currentTileBatch,
-    ControlCodeGraph &innerControlCodeGraph) {
-  for (auto [tile, dmaBatch] : innerControlCodeGraph) {
-    if (dmaBatch->currentDmaOps.empty() &&
-        dmaBatch->immediateInnerBatches.empty())
-      continue;
+    TileDmaBatchGraph &innerTileDmaBatchGraph) {
+  for (auto [tile, dmaBatch] : innerTileDmaBatchGraph) {
+    if (isEmptyDmaBatch(dmaBatch)) continue;
     currentTileBatch[tile]->immediateInnerBatches.push_back(dmaBatch);
   }
 }
 
-static ControlCodeGraph formControlCodeGraph(
+static TileDmaBatchGraph formTileDmaBatchGraph(
     AMDAIE::WorkgroupOp &workgroupOp, Operation *parentOp,
     DenseMap<Value, ChannelBdIdGenerator> &shimTileToGeneratorMap) {
-  ControlCodeGraph controlCodeGraph = initControlCodeGraph(workgroupOp);
+  TileDmaBatchGraph tileDmaBatchGraph = initTileDmaBatchGraph(workgroupOp);
 
   DenseMap<AMDAIE::TileOp, DmaBatch *> currentTileBatch;
-  for (auto [tile, dmaBatch] : controlCodeGraph) {
+  for (auto [tile, dmaBatch] : tileDmaBatchGraph) {
     currentTileBatch[tile] = dmaBatch;
   }
   auto addDmaToBatch = [&](AMDAIE::TileOp tile, AMDAIE::NpuDmaCpyNdOp dmaOp) {
-    assert(controlCodeGraph.contains(tile) && "Tile op not found");
+    assert(tileDmaBatchGraph.contains(tile) && "Tile op not found");
     currentTileBatch[tile]->currentDmaOps.push_back(dmaOp);
   };
 
@@ -131,9 +140,10 @@ static ControlCodeGraph formControlCodeGraph(
   // tileParentOpDmaBatchMap.
   for (Operation &op : parentOp->getRegion(0).getOps()) {
     if (isa<scf::ForOp, scf::ForallOp>(op)) {
-      ControlCodeGraph innerControlCodeGraph =
-          formControlCodeGraph(workgroupOp, &op, shimTileToGeneratorMap);
-      updateCurrentControlCodeGraph(currentTileBatch, innerControlCodeGraph);
+      TileDmaBatchGraph innerTileDmaBatchGraph =
+          formTileDmaBatchGraph(workgroupOp, &op, shimTileToGeneratorMap);
+      updateInnerBatchesOfCurrentTileDmaBatchGraph(currentTileBatch,
+                                                   innerTileDmaBatchGraph);
     } else if (auto dmaOp = dyn_cast<AMDAIE::NpuDmaCpyNdOp>(op)) {
       if (dmaOp.getSource()) {
         FailureOr<AMDAIE::TileOp> tile =
@@ -170,9 +180,10 @@ static ControlCodeGraph formControlCodeGraph(
       }
     }
   }
-  return controlCodeGraph;
+  return tileDmaBatchGraph;
 }
 
+/// Given a non-empty DmaBatch find the immediate surrounding parent op.
 static Operation *findParentOpOfBatch(DmaBatch *dmaBatch) {
   if (!dmaBatch->currentDmaOps.empty())
     return dmaBatch->currentDmaOps[0]->getParentOp();
@@ -183,18 +194,17 @@ static Operation *findParentOpOfBatch(DmaBatch *dmaBatch) {
   return parentOp->getParentOp();
 }
 
+/// Given a DmaBatch `outerDmaBatch`, traverse each nested DmaBatch in it and
+/// infer the required Bd Ids for them. If the nsted DmaBatch is surrounded with
+/// a scf.for, we account for that while sending the required Bd Ids to the
+/// caller of this API.
 static int32_t getRequiredBdIdsForInnerBatches(DmaBatch *outerDmaBatch) {
   if (outerDmaBatch->immediateInnerBatches.empty()) return 0;
   int32_t requiredBdIds = 0;
   for (DmaBatch *dmaBatch : outerDmaBatch->immediateInnerBatches) {
-    if (dmaBatch->currentDmaOps.empty() &&
-        dmaBatch->immediateInnerBatches.empty())
-      continue;
     DmaBatch *currDmaBatch = dmaBatch;
     do {
-      if (currDmaBatch->currentDmaOps.empty() &&
-          currDmaBatch->immediateInnerBatches.empty())
-        break;
+      if (isEmptyDmaBatch(currDmaBatch)) break;
       int32_t totalDmaOpsInCurrentBatch = currDmaBatch->currentDmaOps.size();
       int32_t requiredBdIdsForInnerBatches =
           getRequiredBdIdsForInnerBatches(currDmaBatch);
@@ -216,13 +226,13 @@ static int32_t getRequiredBdIdsForInnerBatches(DmaBatch *outerDmaBatch) {
   return requiredBdIds;
 }
 
-static void inferBdIdsRequiredInBatches(ControlCodeGraph &controlCodeGraph) {
-  for (auto [tile, dmaBatch] : controlCodeGraph) {
-    if (dmaBatch->currentDmaOps.empty() &&
-        dmaBatch->immediateInnerBatches.empty())
-      continue;
+/// Traverse each DmaBatch in a given (Tile -> DmaBatch) and infer Bd Ids
+/// required for it.
+static void inferBdIdsRequiredInBatches(TileDmaBatchGraph &tileDmaBatchGraph) {
+  for (auto [tile, dmaBatch] : tileDmaBatchGraph) {
     DmaBatch *currDmaBatch = dmaBatch;
     do {
+      if (isEmptyDmaBatch(currDmaBatch)) break;
       int32_t totalDmaOpsInCurrentBatch = currDmaBatch->currentDmaOps.size();
       int32_t requiredBdIdsForInnerBatches =
           getRequiredBdIdsForInnerBatches(currDmaBatch);
@@ -233,11 +243,15 @@ static void inferBdIdsRequiredInBatches(ControlCodeGraph &controlCodeGraph) {
   }
 }
 
+/// A struct that maintains the channel and source/target information for a
+/// given DmaOp and TileOp.
 typedef struct DmaTileDataStruct {
   uint32_t channel;
   uint32_t bdIdMapIndex;
 } DmaTileData;
 
+/// Given a DmaOp and a TileOp - extract whether the source (or target) operates
+/// on the tile and also extract the corresponding channel.
 static FailureOr<DmaTileData> getDmaTileData(
     AMDAIE::NpuDmaCpyNdOp npuDmaOp, AMDAIE::TileOp tileOp,
     DenseMap<Value, ChannelBdIdGenerator> &shimTileToGeneratorMap) {
@@ -268,6 +282,10 @@ static FailureOr<DmaTileData> getDmaTileData(
   return dmaTileData;
 }
 
+/// Assign required Bd Ids to the DmaOps of the current DmaBatch. This
+/// assignment is tracked by maintaining `dmaOpToBdIdMap`, which essentially
+/// maps a DmaOp to its source/target Bd Ids. Also, the API splits the available
+/// BD IDs equally amongst all DmaOps in the DmaBatch when assigning
 static LogicalResult assignRequiredBdIdsInCurrentBatch(
     IRRewriter &rewriter, AMDAIE::TileOp tileOp, DmaBatch *dmaBatch,
     DenseMap<Value, ChannelBdIdGenerator> &shimTileToGeneratorMap,
@@ -353,6 +371,7 @@ static LogicalResult assignRequiredBdIdsInCurrentBatch(
   return success();
 }
 
+/// Release a given BdId op.
 LogicalResult releaseBdId(
     AMDAIE::BdIdOp bdIdOp,
     DenseMap<Value, ChannelBdIdGenerator> &shimTileToGeneratorMap,
@@ -386,6 +405,10 @@ LogicalResult releaseBdId(
   return success();
 }
 
+/// DmaOps are assigned Bd Ids prior to invoking this function and a map
+/// `dmaOpToBdIdMap` is maintained that maps a DmaOp to its source/target Bd
+/// Ids. For each DmaOp in the list `dmaOps`, this API will check the
+/// `dmaOpToBdIdMap` and release Bd Ids if assigned.
 static LogicalResult releaseAssignedBdIdsInCurrentBatch(
     SmallVector<AMDAIE::NpuDmaCpyNdOp> dmaOps,
     DenseMap<Value, ChannelBdIdGenerator> &shimTileToGeneratorMap,
@@ -409,9 +432,52 @@ static LogicalResult releaseAssignedBdIdsInCurrentBatch(
   return success();
 }
 
-static bool isEmptyDmaBatch(DmaBatch *dmaBatch) {
-  return (dmaBatch->currentDmaOps.empty() &&
-          dmaBatch->immediateInnerBatches.empty());
+/// Declaration of an API which will work on assigning Bd Ids to the inner
+/// DmaBatch of the current DmaBatch `parentDmaBatch`. And maintain a mapping of
+/// DmaOp -> source/target Bd Id which would be used later during new DmaOp
+/// creation/replacement.
+static LogicalResult assignRequiredBdIdsInInnerBatch(
+    IRRewriter &rewriter, AMDAIE::TileOp tileOp, DmaBatch *parentDmaBatch,
+    DenseMap<Value, ChannelBdIdGenerator> &shimTileToGeneratorMap,
+    DenseMap<AMDAIE::BdIdOp, SmallVector<uint32_t>> &bdIdOpToBdIdsMap,
+    DenseMap<AMDAIE::NpuDmaCpyNdOp, SmallVector<AMDAIE::BdIdOp>>
+        &dmaOpToBdIdMap);
+
+/// Since a DmaBatch contains the following :-
+///   1. DmaOps within the current batch.
+///   2. Nested DmaBatch chain.
+///   3. Pointer to the sibling DmaBatch.
+/// This API processes a given DmaBatch by :-
+///   a. Assigning Bd Ids to 1.
+///   b. Assigning Bd Ids to 2.
+///   c. Releasing Bd Ids assigned to 1.
+///   d. Moving on to 3 and repeating steps a-to-d for it.
+static LogicalResult processDmaBatch(
+    IRRewriter &rewriter, AMDAIE::TileOp tileOp, DmaBatch *currDmaBatch,
+    DenseMap<Value, ChannelBdIdGenerator> &shimTileToGeneratorMap,
+    DenseMap<AMDAIE::BdIdOp, SmallVector<uint32_t>> &bdIdOpToBdIdsMap,
+    DenseMap<AMDAIE::NpuDmaCpyNdOp, SmallVector<AMDAIE::BdIdOp>>
+        &dmaOpToBdIdMap) {
+  do {
+    if (isEmptyDmaBatch(currDmaBatch)) break;
+    if (failed(assignRequiredBdIdsInCurrentBatch(
+            rewriter, tileOp, currDmaBatch, shimTileToGeneratorMap,
+            bdIdOpToBdIdsMap, dmaOpToBdIdMap)))
+      return failure();
+
+    if (failed(assignRequiredBdIdsInInnerBatch(
+            rewriter, tileOp, currDmaBatch, shimTileToGeneratorMap,
+            bdIdOpToBdIdsMap, dmaOpToBdIdMap)))
+      return failure();
+
+    if (failed(releaseAssignedBdIdsInCurrentBatch(
+            currDmaBatch->currentDmaOps, shimTileToGeneratorMap,
+            bdIdOpToBdIdsMap, dmaOpToBdIdMap)))
+      return failure();
+
+    currDmaBatch = currDmaBatch->nextDmaBatch;
+  } while (currDmaBatch != nullptr);
+  return success();
 }
 
 static LogicalResult assignRequiredBdIdsInInnerBatch(
@@ -421,63 +487,75 @@ static LogicalResult assignRequiredBdIdsInInnerBatch(
     DenseMap<AMDAIE::NpuDmaCpyNdOp, SmallVector<AMDAIE::BdIdOp>>
         &dmaOpToBdIdMap) {
   for (DmaBatch *dmaBatch : parentDmaBatch->immediateInnerBatches) {
-    DmaBatch *currDmaBatch = dmaBatch;
-    do {
-      if (isEmptyDmaBatch(currDmaBatch)) break;
-      if (failed(assignRequiredBdIdsInCurrentBatch(
-              rewriter, tileOp, currDmaBatch, shimTileToGeneratorMap,
-              bdIdOpToBdIdsMap, dmaOpToBdIdMap)))
-        return failure();
-
-      if (failed(assignRequiredBdIdsInInnerBatch(
-              rewriter, tileOp, currDmaBatch, shimTileToGeneratorMap,
-              bdIdOpToBdIdsMap, dmaOpToBdIdMap)))
-        return failure();
-
-      if (failed(releaseAssignedBdIdsInCurrentBatch(
-              currDmaBatch->currentDmaOps, shimTileToGeneratorMap,
-              bdIdOpToBdIdsMap, dmaOpToBdIdMap)))
-        return failure();
-
-      // Now release assigned BdIds.
-      currDmaBatch = currDmaBatch->nextDmaBatch;
-    } while (currDmaBatch != nullptr);
+    if (failed(processDmaBatch(rewriter, tileOp, dmaBatch,
+                               shimTileToGeneratorMap, bdIdOpToBdIdsMap,
+                               dmaOpToBdIdMap)))
+      return failure();
   }
   return success();
 }
 
+/// This API will work on assigning Bd Ids to each DmaBatch belonging to a
+/// particular Tile. And maintain a mapping of DmaOp -> source/target Bd Id
+/// which would be used later during new DmaOp creation/replacement.
 static LogicalResult assignRequiredBdIdsInBatch(
-    IRRewriter &rewriter, ControlCodeGraph &controlCodeGraph,
+    IRRewriter &rewriter, TileDmaBatchGraph &tileDmaBatchGraph,
     DenseMap<Value, ChannelBdIdGenerator> &shimTileToGeneratorMap,
     DenseMap<AMDAIE::BdIdOp, SmallVector<uint32_t>> &bdIdOpToBdIdsMap,
     DenseMap<AMDAIE::NpuDmaCpyNdOp, SmallVector<AMDAIE::BdIdOp>>
         &dmaOpToBdIdMap) {
-  for (auto [tileOp, dmaBatch] : controlCodeGraph) {
-    // Find required BD ID by the DMA Ops in current batch.
-    // Required Bd Id = Total Dma Ops in current Batch + Bd Ids required by
-    // nested batches.
-    DmaBatch *currDmaBatch = dmaBatch;
-    do {
-      if (isEmptyDmaBatch(currDmaBatch)) break;
-      if (failed(assignRequiredBdIdsInCurrentBatch(
-              rewriter, tileOp, currDmaBatch, shimTileToGeneratorMap,
-              bdIdOpToBdIdsMap, dmaOpToBdIdMap)))
-        return failure();
-
-      if (failed(assignRequiredBdIdsInInnerBatch(
-              rewriter, tileOp, currDmaBatch, shimTileToGeneratorMap,
-              bdIdOpToBdIdsMap, dmaOpToBdIdMap)))
-        return failure();
-
-      if (failed(releaseAssignedBdIdsInCurrentBatch(
-              currDmaBatch->currentDmaOps, shimTileToGeneratorMap,
-              bdIdOpToBdIdsMap, dmaOpToBdIdMap)))
-        return failure();
-
-      // Now release assigned BdIds.
-      currDmaBatch = currDmaBatch->nextDmaBatch;
-    } while (currDmaBatch != nullptr);
+  for (auto [tileOp, dmaBatch] : tileDmaBatchGraph) {
+    if (failed(processDmaBatch(rewriter, tileOp, dmaBatch,
+                               shimTileToGeneratorMap, bdIdOpToBdIdsMap,
+                               dmaOpToBdIdMap)))
+      return failure();
   }
+  return success();
+}
+
+/// Traverse each DmaOp inside ControlCode and replace it with new new DmaOp
+/// that has Bd Ids assigned using `dmaOpToBdIdMap`.
+static LogicalResult createNewDmaOpsAndReplaceOldDmaOps(
+    IRRewriter &rewriter, AMDAIE::ControlCodeOp controlCodeOp,
+    DenseMap<AMDAIE::NpuDmaCpyNdOp, SmallVector<AMDAIE::BdIdOp>>
+        &dmaOpToBdIdMap) {
+  WalkResult res = controlCodeOp->walk([&](AMDAIE::NpuDmaCpyNdOp npuDmaOp) {
+    if (npuDmaOp.getSource()) {
+      assert(dmaOpToBdIdMap.contains(npuDmaOp) && "No BD ID mapping found");
+      assert((dmaOpToBdIdMap[npuDmaOp][/*sourceBdIdIndex=*/0] != nullptr) &&
+             "No source BD ID mapping found");
+      AMDAIE::BdIdOp bdIdOp = dmaOpToBdIdMap[npuDmaOp][/*sourceBdIdIndex=*/0];
+      rewriter.setInsertionPoint(npuDmaOp);
+      AMDAIE::NpuDmaCpyNdOp oldNpuDmaOp = npuDmaOp;
+      npuDmaOp = rewriter.replaceOpWithNewOp<AMDAIE::NpuDmaCpyNdOp>(
+          npuDmaOp, npuDmaOp.getResultTypes(), npuDmaOp.getConnection(),
+          npuDmaOp.getTarget(), npuDmaOp.getTargetMixedOffsets(),
+          npuDmaOp.getTargetMixedSizes(), npuDmaOp.getTargetMixedStrides(),
+          npuDmaOp.getTargetBdId(), npuDmaOp.getSource(),
+          npuDmaOp.getSourceMixedOffsets(), npuDmaOp.getSourceMixedSizes(),
+          npuDmaOp.getSourceMixedStrides(), bdIdOp);
+      // Since we have created a new NPU DMA op by assigning BD ID to the
+      // source, we need to copy the BD ID assignment to this new NPU DMA op
+      // in order to preserve the information about target BD ID candidate.
+      dmaOpToBdIdMap[npuDmaOp] = dmaOpToBdIdMap[oldNpuDmaOp];
+    }
+    if (npuDmaOp.getTarget()) {
+      assert(dmaOpToBdIdMap.contains(npuDmaOp) && "No BD ID mapping found");
+      assert((dmaOpToBdIdMap[npuDmaOp][/*targetBdIdIndex=*/1] != nullptr) &&
+             "No target BD ID mapping found");
+      AMDAIE::BdIdOp bdIdOp = dmaOpToBdIdMap[npuDmaOp][/*targetBdIdIndex=*/1];
+      rewriter.setInsertionPoint(npuDmaOp);
+      (void)rewriter.replaceOpWithNewOp<AMDAIE::NpuDmaCpyNdOp>(
+          npuDmaOp, npuDmaOp.getResultTypes(), npuDmaOp.getConnection(),
+          npuDmaOp.getTarget(), npuDmaOp.getTargetMixedOffsets(),
+          npuDmaOp.getTargetMixedSizes(), npuDmaOp.getTargetMixedStrides(),
+          bdIdOp, npuDmaOp.getSource(), npuDmaOp.getSourceMixedOffsets(),
+          npuDmaOp.getSourceMixedSizes(), npuDmaOp.getSourceMixedStrides(),
+          npuDmaOp.getSourceBdId());
+    }
+    return WalkResult::advance();
+  });
+  if (res.wasInterrupted()) return failure();
   return success();
 }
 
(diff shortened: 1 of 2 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -146,7 +155,9 @@ template <CopyOpOperateOn OperateOn>
 FailureOr<AMDAIE::BdIdOp> getBdIdOp(
     IRRewriter &rewriter, AMDAIE::NpuDmaCpyNdOp &npuDmaOp,
     DenseMap<Value, ChannelBdIdGenerator> &shimTileToGeneratorMap,
-    DenseMap<AMDAIE::BdIdOp, SmallVector<uint32_t>> &bdIdOpToBdIdsMap) {
+    DenseMap<AMDAIE::BdIdOp, SmallVector<uint32_t>> &bdIdOpToBdIdsMap,
+    DenseMap<AMDAIE::NpuDmaCpyNdOp, AMDAIE::BdIdOp> &dmaOpToBdIdMap) {
+  if (dmaOpToBdIdMap.contains(npuDmaOp)) return dmaOpToBdIdMap[npuDmaOp];
```
jtuyls, line 160: As you're assigning all BD ids for all `NpuDmaCpyNdOp`'s in one go I think, this function should really operate on a `Block` instead of a `NpuDmaCpyNdOp`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -177,40 +188,63 @@ FailureOr<AMDAIE::BdIdOp> getBdIdOp(
...
+    uint32_t totalDmaOps = dmaOps.size();
     uint32_t size = std::max(numAvailable / numRequired, 1u);
-
-    // Only create expression if more than 1 BD ID is needed,
-    // otherwise, fall back to constant BD ID.
-    if (size > 1) {
-      // Assigning BD IDs for all iterations in the loop.
-      SmallVector<uint32_t> bdIds;
-      for (uint32_t i = 0; i < size; i++) {
+    // In case an equal split of BD IDs is not possible, return failure.
```
jtuyls, line 202: This seems not about an equal split, but in case the there are not enough BD ids available?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -76,11 +76,12 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
 
 /// Computes the number of BD IDs required between the current
-/// DMA copy operation and its corresponding DMA wait operation.
-/// If a sub-loop is encountered, it assumes that a BD ID expression will be
-/// used within the sub-loop. This approach ensures that the inner loop has
-/// access to a greater number of BD IDs, which is favorable for enabling
-/// efficient BD chaining in subsequent passes.
+/// DMA copy operation and its corresponding DMA wait operation; returns the DMA
+/// ops it traverses within same block and on same tile. If a sub-loop is
+/// encountered, it assumes that a BD ID expression will be used within the
```
jtuyls, line 81: Could you point to where we make this assumption and explain this more?

jtuyls (commented):
Some preliminary comments/questions.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
Abhishek-Varma: So I'm assigning BD ids for all `NpuDmaCpyNdOp` within a `scf.for` block that share the same `tile`.

The following is not really being done in one go :-
```
controlcode {
   %1 = npu_dma_cpy
   dma_wait(%1)
   %2 = npu_dma_cpy
   dma_wait(%2)
   %3 = npu_dma_cpy
   dma_wait(%3)
}
```
In the above case the API is not assigning BD IDs for `%2` and `%3` when dealing with `%1`.

Similarly, even in case of scf.for the following is not done in one go :-
```
controlcode {
   scf.for {
        %1 = npu_dma_cpy
        dma_wait(%1)
        %2 = npu_dma_cpy
        dma_wait(%2)
        %3 = npu_dma_cpy
        dma_wait(%3)
  }
}
```
In the above case as well the API is not assigning BD IDs for `%2` and `%3` when dealing with `%1`.

I guess the confusion of "same block" occurred because of the way I phrased it. My bad!
I've rephrased the corresponding comments.

Could you check now?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
Abhishek-Varma: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
Abhishek-Varma: Actually [this comment](https://github.com/nod-ai/iree-amd-aie/blob/5319071511a701243680ff0be2068daeda03fd60/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp#L80-L81) seems to be related to [inner scf.for traversal here](https://github.com/nod-ai/iree-amd-aie/blob/5319071511a701243680ff0be2068daeda03fd60/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp#L127-L137).

I believe it was meant as a way to recursively traverse the inner loop and get corresponding BD ID requirement in case there are DMA ops operating on same tile as the "current DMA op".

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
jtuyls: I see, but I still think that calling this function again for NpuDmaCpyNdOps that have already been assigned BD ids makes it harder to understand what's going on. It would simplify the logic if you split it up in two phases:
1. Collect all NpuDmaCpyNdOps on the same tiles
2. Assign BD ids for those as a batch (vector of `NpuDmaCpyNdOp`)

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
jtuyls: Oh, ok, I see that this is not new, I thought you changed this.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
Abhishek-Varma: I understand.
The thing is only if the parent if a `scf.for` are we performing/capturing other `NpuDmaCpyNdOp`s through the traversal.

Else, even for the following case :-
```
controlcode {
   %1 = npu_dma_cpy
   %2 = npu_dma_cpy
   %3 = npu_dma_cpy
   dma_wait(%1)
   dma_wait(%2)
   dma_wait(%3)
}
```
Even if `%2` and `%3` are between `%1` and the DMA wait - but they'll be processed individually and are NOT assigned BD IDs.

Only if the surrounding `NpuDmaCpyNdOp` has a `scf.for` then for the following :-
```
controlcode {
 scf.for {
   %1 = npu_dma_cpy
   %2 = npu_dma_cpy
   %3 = npu_dma_cpy
   dma_wait(%1)
   dma_wait(%2)
   dma_wait(%3)
 }
}
```
We capture `%2` and `%3` between `%1` and the DMA wait - and assign BD IDs to them.

What I can instead do, to make the logic easier to understand, is -> before invoking the API `getBdIdOp`, I'd make a call to the map `dmaOpToBdIdMap` at the caller side itself and work through that.

This way I'll have the following :-
```
AMDAIE::BdIdOp bdIdOp;
if (dmaOpToBdIdMap.contains(npuDmaOp)) {
   bdIdOp = dmaOpToBdIdMap[npuDmaOp];
} else {
   bdIdOp  = getBdIdOp(....);
}
```
I think this should work just fine and is better.

Let me know if that sounds good to you and I can address it accordingly.

Yu-Zhewen: Thanks for working on this!

Yeah, I also agree it becomes difficult to understand what's inside `getBdIdOp`. Maybe also consider restructuring like so?
```
if (!dmaOpToBdIdMap.contains(npuDmaOp)) {
 (void) getBdIdOp(dmaOpToBdIdMap);
}
assert (dmaOpToBdIdMap.contains(npuDmaOp))
bdIdOp = dmaOpToBdIdMap[npuDmaOp];
```

This [loop](https://github.com/nod-ai/iree-amd-aie/blob/1e1abb4120ea68fb32ebb86133db1e2c7a096875/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp#L209) might be better placed in a standalone utility function. [This special case](https://github.com/nod-ai/iree-amd-aie/blob/1e1abb4120ea68fb32ebb86133db1e2c7a096875/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp#L252) can also be merged into that loop?

Also, consider renaming `getBdIdOp` and updating its description [here](https://github.com/nod-ai/iree-amd-aie/blob/1e1abb4120ea68fb32ebb86133db1e2c7a096875/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp#L150)

jtuyls: > The thing is only if the parent if a scf.for are we performing/capturing other NpuDmaCpyNdOps through the traversal.

I don't see why the same traversal wouldn't work without an `scf.for`?

> Even if %2 and %3 are between %1 and the DMA wait - but they'll be processed individually and are NOT assigned BD IDs.

They don't have to be processed individually right? You could iterate through them, collect all on the same tiles and then process those together.

Abhishek-Varma: Hi @jtuyls @Yu-Zhewen - I've updated this revision and also included one more lit test which wasn't being addressed by my changes previously.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -95,96 +96,115 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
 /// - The sub-loop executes 2 iterations, each requiring 1 BD ID.
 /// - Therefore, the required number of BD IDs is:
 ///  3 = 1 (current) + 2 * 1(sub-loop).
-void getNumRequiredBdIds(
-    scf::ForOp loop, AMDAIE::NpuDmaCpyNdOp currDmaOp, AMDAIE::TileOp tileOp,
+SmallVector<AMDAIE::NpuDmaCpyNdOp> getNumRequiredBdIdsAndDMAOps(
```
Yu-Zhewen, line 99: Might be more consistent with the function name to return both `numRequiredBdIds` and `dmaOps` as a pair, or pass both by reference.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -95,96 +96,115 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+  // first BD ID op assigned.
+  rewriter.setInsertionPoint(dmaOps[0]);
+  for (AMDAIE::NpuDmaCpyNdOp dmaOp : dmaOps) {
+    uint32_t bdIdMapIndex = 0;
+    if (dmaOp.getSource()) {
+      FailureOr<AMDAIE::TileOp> tile =
+          getGeneratorTileOp<CopyOpOperateOn::Source>(dmaOp,
+                                                      shimTileToGeneratorMap);
+      if (succeeded(tile) && *tile == tileOp) {
+        bdIdMapIndex = 0;
```
Yu-Zhewen, line 196: `CopyOpOperateOn` is already an enum?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,132 +75,280 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
-      if (npuDmaOp == currDmaOp) startCounting = true;
-      if (!startCounting) continue;
-      if (npuDmaOp.getSource()) {
+/// A DmaBatch contains the following :-
+///   1. DmaOps within the current batch.
+///   2. Required Bd Ids by the current batch.
+///   3. Nested DmaBatch chain.
+///   4. Pointer to the sibling DmaBatch.
+///   4. A pointer to the parent op if the parent of is a scf.for.
+typedef struct DmaBatchStruct {
```
Yu-Zhewen, line 84: I think `typedef`  is not necessary in modern C++.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,132 +75,280 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+static void updateInnerBatchesOfCurrentTileDmaBatchGraph(
+    DenseMap<AMDAIE::TileOp, DmaBatch *> &currentTileBatch,
+    TileDmaBatchGraph &innerTileDmaBatchGraph) {
+  for (auto [tile, dmaBatch] : innerTileDmaBatchGraph) {
+    if (isEmptyDmaBatch(dmaBatch)) continue;
+    currentTileBatch[tile]->immediateInnerBatches.push_back(dmaBatch);
+  }
+}
+
+static TileDmaBatchGraph formTileDmaBatchGraph(
```
Yu-Zhewen, line 121: I found the logic in this function a bit hard to follow at first. From what I understand, `currentTileBatch `is used to build the linked lists, and `tileDmaBatchGraph` holds the heads of these lists? If true, any specific reason to use the linked list here instead of just using vectors?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,132 +75,280 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+
+using TileDmaBatchGraph = llvm::MapVector<AMDAIE::TileOp, DmaBatch *>;
+
+/// Create a new TileDmaBatchGraph by traversing over each tiles in a workgroup.
+static TileDmaBatchGraph initTileDmaBatchGraph(
+    AMDAIE::WorkgroupOp workgroupOp) {
+  TileDmaBatchGraph tileDmaBatchGraph;
+  workgroupOp.walk([&](AMDAIE::TileOp tile) {
+    if (tileDmaBatchGraph.contains(tile)) return WalkResult::skip();
+    tileDmaBatchGraph[tile] = new DmaBatch();
```
Yu-Zhewen, line 100: Where are they getting deleted?

jtuyls: I think we should be using `unique_ptr` instead of raw pointers here and everywhere we use `DmaBatch`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,132 +75,280 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+    return WalkResult::advance();
+  });
+  return tileDmaBatchGraph;
+}
+
+/// A DmaBatch with no DmaOps and no nested DmaBatches is an empty DmaBatch.
+static bool isEmptyDmaBatch(DmaBatch *dmaBatch) {
```
jtuyls, line 107: You can add this to the struct instead: `DmaBaths::isEmpty()`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,132 +75,280 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+  return (dmaBatch->currentDmaOps.empty() &&
+          dmaBatch->immediateInnerBatches.empty());
+}
+
+static void updateInnerBatchesOfCurrentTileDmaBatchGraph(
```
jtuyls, line 112: Could you document in what way you update it?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,132 +75,280 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+    currentTileBatch[tile]->forOpParent = dyn_cast<scf::ForOp>(parentOp);
+  };
+
+  auto updateInnerBatchesOfCurrentTileDmaBatchGraph =
+      [&](TileDmaBatchGraph &innerTileDmaBatchGraph) {
+        for (auto [tile, dmaBatch] : innerTileDmaBatchGraph) {
+          if (isEmptyDmaBatch(dmaBatch)) continue;
+          currentTileBatch[tile]->immediateInnerBatches.push_back(dmaBatch);
+        }
+      };
```
jtuyls, line 148: It looks useful to create a class/struct to maintain `currentTileBatch` and add these methods to the class instead.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
jtuyls, line 121: You would prefer `createTileDmaBatchGraph`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,132 +75,280 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+/// Given a non-empty DmaBatch find the immediate surrounding parent op.
+static Operation *findParentOpOfBatch(DmaBatch *dmaBatch) {
+  if (!dmaBatch->currentDmaOps.empty())
+    return dmaBatch->currentDmaOps[0]->getParentOp();
+  assert(!dmaBatch->immediateInnerBatches.empty() &&
+         "DmaBatch found with no DmaOps and no immediate inner batches");
+  Operation *parentOp = findParentOpOfBatch(dmaBatch->immediateInnerBatches[0]);
+  assert(parentOp && "Found inner DmaBatch with no parent");
+  return parentOp->getParentOp();
+}
```
jtuyls, line 206: This should be a struct/class method imo as it only operates on members in `DmaBatch`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,132 +75,280 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
 
-  // Get the channel.
+/// Given a DmaBatch `outerDmaBatch`, traverse each nested DmaBatch in it and
+/// infer the required Bd Ids for them. If the nsted DmaBatch is surrounded with
+/// a scf.for, we account for that while sending the required Bd Ids to the
+/// caller of this API.
+static int32_t getRequiredBdIdsForInnerBatches(DmaBatch *outerDmaBatch) {
```
jtuyls, line 212: Same as above, could also be a class method.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,132 +75,280 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+  assert(!dmaBatch->immediateInnerBatches.empty() &&
+         "DmaBatch found with no DmaOps and no immediate inner batches");
+  Operation *parentOp = findParentOpOfBatch(dmaBatch->immediateInnerBatches[0]);
+  assert(parentOp && "Found inner DmaBatch with no parent");
+  return parentOp->getParentOp();
+}
 
-  // Get the channel.
+/// Given a DmaBatch `outerDmaBatch`, traverse each nested DmaBatch in it and
+/// infer the required Bd Ids for them. If the nsted DmaBatch is surrounded with
```
jtuyls, line 209:
```suggestion
/// infer the required Bd Ids for them. If the nested DmaBatch is surrounded with
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,132 +75,280 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+///   4. A pointer to the parent op if the parent of is a scf.for.
+typedef struct DmaBatchStruct {
+  SmallVector<AMDAIE::NpuDmaCpyNdOp> currentDmaOps = {};
+  int32_t requiredBdIds = 0;
+  SmallVector<DmaBatchStruct *> immediateInnerBatches = {};
+  DmaBatchStruct *nextDmaBatch = nullptr;
+  scf::ForOp forOpParent = nullptr;
+} DmaBatch;
+
+using TileDmaBatchGraph = llvm::MapVector<AMDAIE::TileOp, DmaBatch *>;
```
jtuyls, line 92: You might want to encapsulate this as well with a class/struct to add appropriate utility methods to access and update data, like `inferBdIdsRequiredInBatches` below.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,132 +75,280 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+    DmaBatch *currDmaBatch = dmaBatch;
+    do {
+      if (isEmptyDmaBatch(currDmaBatch)) break;
+      int32_t totalDmaOpsInCurrentBatch = currDmaBatch->currentDmaOps.size();
+      int32_t requiredBdIdsForInnerBatches =
+          getRequiredBdIdsForInnerBatches(currDmaBatch);
+      currDmaBatch->requiredBdIds =
+          requiredBdIdsForInnerBatches + totalDmaOpsInCurrentBatch;
+      requiredBdIds += totalDmaOpsInCurrentBatch;
+      // Operation *parentOp = findParentOpOfBatch(currDmaBatch);
```
jtuyls, line 225:
```suggestion
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,132 +75,280 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+    } while (currDmaBatch != nullptr);
+  }
+}
+
+/// A struct that maintains the channel and source/target information for a
+/// given DmaOp and TileOp.
+typedef struct DmaTileDataStruct {
+  uint32_t channel;
+  uint32_t bdIdMapIndex;
+} DmaTileData;
```
jtuyls, line 263: Same as above comment from @Yu-Zhewen, typedef not needed as this is all C++.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -257,6 +416,144 @@ LogicalResult releaseBdId(
...
+/// `dmaOpToBdIdMap` and release Bd Ids if assigned.
+static LogicalResult releaseAssignedBdIdsInCurrentBatch(
+    SmallVector<AMDAIE::NpuDmaCpyNdOp> dmaOps,
+    DenseMap<Value, ChannelBdIdGenerator> &shimTileToGeneratorMap,
+    DenseMap<AMDAIE::BdIdOp, SmallVector<uint32_t>> &bdIdOpToBdIdsMap,
+    DenseMap<AMDAIE::NpuDmaCpyNdOp, SmallVector<AMDAIE::BdIdOp>>
+        &dmaOpToBdIdMap) {
+  // Release BD ID used by input DMA op.
+  for (AMDAIE::NpuDmaCpyNdOp npuDmaOp : dmaOps) {
+    if (dmaOpToBdIdMap[npuDmaOp][0]) {
```
jtuyls, line 431: Is this guaranteed that index `[0]` and `[1]` exist in `SmallVector<AMDAIE::BdIdOp>`? I think this is always a pair, so `std::pair` would be more appropriate?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -257,6 +416,144 @@ LogicalResult releaseBdId(
...
+    }
+  }
+  return success();
+}
+
+/// Declaration of an API which will work on assigning Bd Ids to the inner
+/// DmaBatch of the current DmaBatch `parentDmaBatch`. And maintain a mapping of
+/// DmaOp -> source/target Bd Id which would be used later during new DmaOp
+/// creation/replacement.
+static LogicalResult assignRequiredBdIdsInInnerBatch(
```
jtuyls, line 450: You could just move the definition?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -257,6 +416,144 @@ LogicalResult releaseBdId(
...
+    Value sourceBdId = nullptr;
+    Value targetBdId = nullptr;
+    if (dmaOpToBdIdMap[npuDmaOp][/*sourceBdIdIndex=*/0]) {
+      sourceBdId = dmaOpToBdIdMap[npuDmaOp][/*sourceBdIdIndex=*/0].getResult();
+    }
+    if (dmaOpToBdIdMap[npuDmaOp][/*targetBdIdIndex=*/1]) {
+      targetBdId = dmaOpToBdIdMap[npuDmaOp][/*targetBdIdIndex=*/1].getResult();
+    }
+    rewriter.setInsertionPoint(npuDmaOp);
+    (void)rewriter.replaceOpWithNewOp<AMDAIE::NpuDmaCpyNdOp>(
```
jtuyls, line 544:
```suggestion
    rewriter.replaceOpWithNewOp<AMDAIE::NpuDmaCpyNdOp>(
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/assign_npu_dma_bd_ids.mlir
```diff
@@ -474,3 +474,305 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
...
+    return
+  }
+}
+
+// -----
+
+// Expect all DMA ops, between the first DMA op and its corresponding DMA wait op, operating
+// within same scf.for's block and on same tile to have equal BD ID distribution for the
+// corresponding source/target. This test demonstrates how even with different tile distribution
+// amongst the DMA ops a well distributed BD ID assignment split is ensured.
```
jtuyls, line 553:
```suggestion
// among the DMA ops a well distributed BD ID assignment split is ensured.
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
Abhishek-Varma: Yes, although this is guaranteed to always have index `[0]` and `[1]` because of [this line](https://github.com/nod-ai/iree-amd-aie/blob/f74e7ddb4819e3f17f6064c3e34b00bcf9e6db4f/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp#L361) - but using `std::pair` here would render [this one-line update](https://github.com/nod-ai/iree-amd-aie/blob/f74e7ddb4819e3f17f6064c3e34b00bcf9e6db4f/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp#L365) useless.

Is there anything else we could consider using here that has the essence of that one-line capture?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
Yu-Zhewen: Thanks for adding those comments!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,185 +75,494 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+  /// Release a given BdId op.
+  static LogicalResult releaseBdId(
+      AMDAIE::BdIdOp bdIdOp,
+      DenseMap<Value, ChannelBdIdGenerator> &shimTileToGeneratorMap,
+      DenseMap<AMDAIE::BdIdOp, SmallVector<uint32_t>> &bdIdOpToBdIdsMap) {
+    auto tileOp =
+        dyn_cast_if_present<AMDAIE::TileOp>(bdIdOp.getTile().getDefiningOp());
+    if (!tileOp)
+      return bdIdOp.emitOpError()
+             << "doesn't operate on a `amdaie.tile` operation";
```
Yu-Zhewen, line 411:
```suggestion
    if (!tileOp) {
      return bdIdOp.emitOpError()
             << "doesn't operate on an `amdaie.tile` operation";
    }
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,185 +75,494 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+
+    if (!shimTileToGeneratorMap.contains(tileOp.getResult()))
+      return bdIdOp.emitOpError()
+             << "no BD ID generator found for this BD ID op's tile";
```
Yu-Zhewen, line 415:
```suggestion
    if (!shimTileToGeneratorMap.contains(tileOp.getResult())) {
      return bdIdOp.emitOpError()
             << "no BD ID generator found for this BD ID op's tile";
    }
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,185 +75,494 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+
+    ChannelBdIdGenerator &generator =
+        shimTileToGeneratorMap[tileOp.getResult()];
+    Value value = bdIdOp.getValue();
+    if (auto op = value.getDefiningOp<affine::AffineApplyOp>()) {
+      // If the BD ID is a semi-affine expression.
+      if (bdIdOpToBdIdsMap.contains(bdIdOp)) {
+        for (uint32_t bdId : bdIdOpToBdIdsMap[bdIdOp]) {
+          generator.releaseBdId(bdId);
+        }
```
Yu-Zhewen, line 425:
```suggestion
        for (uint32_t bdId : bdIdOpToBdIdsMap[bdIdOp])
          generator.releaseBdId(bdId);
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/assign_npu_dma_bd_ids.mlir
```diff
@@ -406,9 +406,9 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
 // Expect all three DMA copy operations have BD IDs as expressions. #map0: 0~1, #map1: 2~8, #map2: 9~15.
```
Yu-Zhewen, line 406:
```suggestion
// Expect all three DMA copy operations have BD IDs as expressions. #map0: 0~4, #map1: 5~9, #map2: 10~14.
```

Yu-Zhewen (approved):
LGTM. Thanks!

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
Abhishek-Varma: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,185 +75,505 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+};
+
+struct TileDmaBatchGraph {
+ private:
+  /// The main tile->DmaBatch graph.
+  llvm::MapVector<AMDAIE::TileOp, std::unique_ptr<DmaBatch>> tileDmaBatchGraph;
+  /// Helps keep track of the current DmaBatch of a tile that's being processed.
+  /// It holds the reference to the last DmaBatch for every tile on the main
+  /// graph.
+  DenseMap<AMDAIE::TileOp, DmaBatch *> currentTileBatch;
```
jtuyls, line 184: If you really need multiple references to the same `DmaBatch`, you should use a `shared_ptr` instead.

jtuyls (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,185 +75,505 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+    AMDAIE::WorkgroupOp &workgroupOp, Operation *parentOp,
+    DenseMap<Value, ChannelBdIdGenerator> &shimTileToGeneratorMap) {
+  TileDmaBatchGraph tileDmaBatchGraph =
+      TileDmaBatchGraph(workgroupOp, parentOp);
+
+  AMDAIE::NpuDmaCpyNdOp currDmaOp = nullptr;
+  // Traverse the parent operation's immediate child ops and form the
+  // tileParentOpDmaBatchMap.
+  for (Operation &op : parentOp->getRegion(0).getOps()) {
+    if (isa<scf::ForOp, scf::ForallOp>(op)) {
```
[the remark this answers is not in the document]
Abhishek-Varma: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,185 +75,505 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+            requiredBdIdsForInnerBatches + totalDmaOpsInCurrentBatch;
+        currDmaBatch = currDmaBatch->nextDmaBatch.get();
+      } while (currDmaBatch != nullptr);
+    }
+  }
+
+  /// This API will work on assigning Bd Ids to each DmaBatch belonging to a
+  /// particular Tile. And maintain a mapping of DmaOp -> source/target Bd Id
+  /// which would be used later during new DmaOp creation/replacement.
+  LogicalResult assignRequiredBdIdsInBatch(
```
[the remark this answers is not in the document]
Abhishek-Varma: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,185 +75,505 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+      // In case the DMA ops are not surrounded by scf.for, we will assign
+      // only one BD ID.
+      size = 1;
+    }
+
+    rewriter.setInsertionPoint(dmaBatch->currentDmaOps[0]);
+    for (AMDAIE::NpuDmaCpyNdOp dmaOp : dmaBatch->currentDmaOps) {
+      maybeDmaTileData = getDmaTileData(dmaOp, tileOp, shimTileToGeneratorMap);
+      if (failed(maybeDmaTileData)) return failure();
+      dmaTileData = *maybeDmaTileData;
```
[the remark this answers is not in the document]
Abhishek-Varma: Yes, made changes to do this once now.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,185 +75,505 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+  /// This API will work on assigning Bd Ids to each DmaBatch belonging to a
+  /// particular Tile. And maintain a mapping of DmaOp -> source/target Bd Id
+  /// which would be used later during new DmaOp creation/replacement.
+  LogicalResult assignRequiredBdIdsInBatch(
+      IRRewriter &rewriter,
+      DenseMap<Value, ChannelBdIdGenerator> &shimTileToGeneratorMap,
+      DenseMap<AMDAIE::BdIdOp, SmallVector<uint32_t>> &bdIdOpToBdIdsMap,
+      DenseMap<AMDAIE::NpuDmaCpyNdOp, SmallVector<AMDAIE::BdIdOp>>
+          &dmaOpToBdIdMap) {
+    for (auto &[tileOp, dmaBatch] : *getGraph()) {
```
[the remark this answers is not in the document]
Abhishek-Varma: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,185 +75,505 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+  if (npuDmaOp.getSource()) {
+    FailureOr<AMDAIE::TileOp> tile =
+        getGeneratorTileOp<CopyOpOperateOn::Source>(npuDmaOp,
+                                                    shimTileToGeneratorMap);
+    if (succeeded(tile) && *tile == tileOp) {
+      maybeChannelOp = npuDmaOp.getSourceChannelOp();
+      bdIdMapIndex = 0;
+    }
+  }
+  if (npuDmaOp.getTarget()) {
```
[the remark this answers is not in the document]
Abhishek-Varma: Correct. But do we have a case where the source and target of the same DMA op works on the SAME tile ?
And if so wouldn't that mean that the DMA Op is redundant as it's attempting to transfer a data that's on one tile back to the SAME tile ?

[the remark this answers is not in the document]
jtuyls: Right, that makes sense, but then we should catch that and return a failure instead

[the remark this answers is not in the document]
Abhishek-Varma: I think this will be better suited as a part in the verifier of DMA ops itself - this way we catch this case early on in the stack while creating the ops' creation.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -268,79 +572,27 @@ LogicalResult assignNpuDmaBdIds(AMDAIE::WorkgroupOp workgroupOp) {
...
-          if (!bdIdOp) return WalkResult::advance();
-          if (failed(releaseBdId(bdIdOp, shimTileToGeneratorMap,
-                                 bdIdOpToBdIdsMap)))
-            return WalkResult::interrupt();
-        }
+  // Since a DMA op can have source and target, therefore we can have two BD IDs
+  // for any DMA op. Hence we maintain a map from DMA op to a vector of BD IDs.
+  DenseMap<AMDAIE::NpuDmaCpyNdOp, SmallVector<AMDAIE::BdIdOp>> dmaOpToBdIdMap;
+  TileDmaBatchGraph tileDmaBatchGraph = createTileDmaBatchGraph(
+      workgroupOp, controlCodeOp, bdIdAssignmentUtil.shimTileToGeneratorMap);
```
jtuyls, line 586: It's a little bit strange that we need to pass the `shimTileToGeneratorMap` from `BdIdAssignmentUtil` here. I think it's better to decouple this interdependency by for example extracting `createShimTileToGeneratorMap` from `BdIdAssignmentUtil` to a standalone function, call that to initialize the map and pass it both to `createTileDmaBatchGraph` and to `BdIdAssignmentUtil`.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,186 +75,490 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+  LogicalResult replaceDmaOps(IRRewriter &rewriter,
+                              AMDAIE::ControlCodeOp controlCodeOp) {
+    WalkResult res = controlCodeOp->walk([&](AMDAIE::NpuDmaCpyNdOp npuDmaOp) {
+      assert(dmaOpToBdIdMap.contains(npuDmaOp) && "No BD ID mapping found");
+      Value sourceBdId = nullptr;
+      Value targetBdId = nullptr;
+      if (dmaOpToBdIdMap[npuDmaOp][/*sourceBdIdIndex=*/0]) {
+        sourceBdId =
+            dmaOpToBdIdMap[npuDmaOp][/*sourceBdIdIndex=*/0].getResult();
+      }
```
jtuyls, line 467:
```suggestion
      if (BdIdOp bdIdOp = dmaOpToBdIdMap[npuDmaOp][/*sourceBdIdIndex=*/0]; bdIdOp) {
        sourceBdId = bdIdOp.getResult();
      }
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,186 +75,490 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+    }
+    return requiredBdIds;
+  }
+};
+
+/// `TileDmaBatchGraph` is a struct that maintains a mapping from a tile to its
+/// corresponding `DmaBatch`. It contains utilities that help form/update the
+/// `DmaBatch` graph for a particular tile.
+struct TileDmaBatchGraph {
+ private:
```
jtuyls, line 134:
```suggestion
class TileDmaBatchGraph {
```
I think a class is more appropriate as you're making use of private member variables. Then you can remove `private:` here as well.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,186 +75,490 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+                                                   currDmaBatch)))
+        return failure();
+
+      if (failed(
+              assignRequiredBdIdsInInnerBatch(rewriter, tileOp, currDmaBatch)))
+        return failure();
+
+      if (failed(
+              releaseAssignedBdIdsInCurrentBatch(currDmaBatch->currentDmaOps)))
+        return failure();
```
jtuyls, line 292:
```suggestion
      if (failed(assignRequiredBdIdsInCurrentBatch(rewriter, tileOp,
                                                   currDmaBatch))) {
        return failure();
      }
      if (failed(
              assignRequiredBdIdsInInnerBatch(rewriter, tileOp, currDmaBatch))) {
        return failure();
      }
      if (failed(
              releaseAssignedBdIdsInCurrentBatch(currDmaBatch->currentDmaOps))) {
        return failure();
      }
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,186 +75,490 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+
+  /// Assign Bd Ids to the inner DmaBatch of the current DmaBatch
+  /// `parentDmaBatch`. And maintain a mapping of DmaOp -> source/target Bd Id
+  /// which would be used later during new DmaOp creation/replacement.
+  LogicalResult assignRequiredBdIdsInInnerBatch(
+      IRRewriter &rewriter, AMDAIE::TileOp tileOp,
+      std::shared_ptr<DmaBatch> parentDmaBatch) {
+    for (std::shared_ptr<DmaBatch> &dmaBatch :
+         parentDmaBatch->immediateInnerBatches) {
+      OpBuilder::InsertionGuard guard(rewriter);
```
jtuyls, line 307: It's better to put the insertion guard in `processBatch` itself, then it would work in all cases in which the function is called.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,186 +75,490 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+    return success();
   }
-  return success();
+
+  /// DmaOps are assigned Bd Ids prior to invoking this function and a map
+  /// `dmaOpToBdIdMap` is maintained that maps a DmaOp to its source/target Bd
+  /// Ids. For each DmaOp in the list `dmaOps`, this API will check the
+  /// `dmaOpToBdIdMap` and release Bd Ids if assigned.
+  LogicalResult releaseAssignedBdIdsInCurrentBatch(
+      SmallVector<AMDAIE::NpuDmaCpyNdOp> dmaOps) {
```
jtuyls, line 442:
```suggestion
      SmallVectorImpl<AMDAIE::NpuDmaCpyNdOp> &dmaOps) {
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,186 +75,490 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
-    generator.releaseBdId(bdId);
+    return success();
   }
-  return success();
+
+  /// DmaOps are assigned Bd Ids prior to invoking this function and a map
+  /// `dmaOpToBdIdMap` is maintained that maps a DmaOp to its source/target Bd
+  /// Ids. For each DmaOp in the list `dmaOps`, this API will check the
+  /// `dmaOpToBdIdMap` and release Bd Ids if assigned.
+  LogicalResult releaseAssignedBdIdsInCurrentBatch(
```
jtuyls, line 441:
```suggestion
  LogicalResult releaseAssignedBdIdsInDmaOps(
```
This is not really a batch, either pass a ``DmaBatch` or rename.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,186 +75,490 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+      SmallVector<AMDAIE::NpuDmaCpyNdOp> dmaOps) {
+    // Release BD ID used by input DMA op.
+    for (AMDAIE::NpuDmaCpyNdOp npuDmaOp : dmaOps) {
+      if (dmaOpToBdIdMap[npuDmaOp][0]) {
+        if (failed(releaseBdId(dmaOpToBdIdMap[npuDmaOp][0]))) return failure();
+      }
+
+      if (dmaOpToBdIdMap[npuDmaOp][1]) {
+        if (failed(releaseBdId(dmaOpToBdIdMap[npuDmaOp][1]))) return failure();
+      }
```
jtuyls, line 451:
```suggestion
      if (BdIdOp bdIdOp = dmaOpToBdIdMap[npuDmaOp][0]; bdIdOp) {
        if (failed(releaseBdId(bdIdOp))) return failure();
      }
      if (BdIdOp bdIdOp = dmaOpToBdIdMap[npuDmaOp][1]; bdIdOp) {
        if (failed(releaseBdId(bdIdOp ))) return failure();
      }
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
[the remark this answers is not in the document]
jtuyls: I think we should add it here as well as it's a big assumption in this logic. Also, you could potentially route to the same tile as a copy-like operation. I don't think we will see is soon, but it is technically possible, so I am not even sure about adding it to the verifier.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
Abhishek-Varma: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,186 +75,507 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+/// `TileDmaBatchGraph` is a struct that maintains a mapping from a tile to its
+/// corresponding `DmaBatch`. It contains utilities that help form/update the
+/// `DmaBatch` graph for a particular tile.
+class TileDmaBatchGraph {
+  /// The main tile->DmaBatch graph.
+  llvm::MapVector<AMDAIE::TileOp, std::shared_ptr<DmaBatch>> tileDmaBatchGraph;
+  /// Helps keep track of the current DmaBatch of a tile that's being processed.
+  /// It holds the reference to the last DmaBatch for every tile on the main
+  /// graph.
+  DenseMap<AMDAIE::TileOp, std::shared_ptr<DmaBatch>> currentTileBatch;
```
jtuyls, line 139: Do we really need both? I fail to see where `currentTileBatch` is used in the processing in `BdIdAssignmentUtil`?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,186 +75,507 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+      FailureOr<AMDAIE::TileOp> tile =
+          getGeneratorTileOp<CopyOpOperateOn::Target>(npuDmaOp,
+                                                      shimTileToGeneratorMap);
+      if (succeeded(tile) && *tile == tileOp) {
+        if (bdIdMapIndex == -1) {
+          maybeChannelOp = npuDmaOp.getTargetChannelOp();
+          bdIdMapIndex = 1;
+        } else {
+          return failure();
+        }
```
jtuyls, line 256: Add a comment on why we return failure here.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,186 +75,507 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
-    } else if (auto npuWaitOp = dyn_cast<AMDAIE::NpuDmaWaitOp>(op)) {
-      for (AMDAIE::NpuDmaCpyNdOp npuDmaOp : npuWaitOp.getDmaOps()) {
-        // Reached the DMA wait operation, stop counting.
-        if (npuDmaOp == currDmaOp) return;
+    }
+
+    if (failed(maybeChannelOp)) return failure();
+
+    DmaTileData dmaTileData = {maybeChannelOp.value().getValue(), bdIdMapIndex};
+    return dmaTileData;
```
jtuyls, line 263:
```suggestion
    return DmaTileData{maybeChannelOp.value().getValue(), bdIdMapIndex};
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,186 +75,507 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+        if (failed(releaseBdId(bdIdOp))) return failure();
+      }
+      if (AMDAIE::BdIdOp bdIdOp = dmaOpToBdIdMap[npuDmaOp][1]; bdIdOp) {
+        if (failed(releaseBdId(bdIdOp))) return failure();
+      }
+    }
+    return success();
+  }
+
+ public:
```
jtuyls, line 454: Typically, you put the public member functions and variables at the top and private at the bottom: for example https://github.com/nod-ai/iree-amd-aie/blob/00995f9e21022b1dc9d40a9eb5788a8eb145aa94/runtime/src/iree-amd-aie/aie_runtime/amsel_generator.h#L112

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,186 +75,507 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+  BdIdAssignmentUtil(
+      DenseMap<Value, ChannelBdIdGenerator> &shimTileToGeneratorMap) {
+    this->shimTileToGeneratorMap = shimTileToGeneratorMap;
```
jtuyls, line 457: We should move it or `this->shimTileToGeneratorMap` a reference. I prefer the former because `shimTileToGeneratorMap` never has to be updated externally, so we can own it, see suggestion.
```suggestion
      DenseMap<Value, ChannelBdIdGenerator> shimTileToGeneratorMap) : shimTileToGeneratorMap(shimTileToGeneratorMap) {
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,186 +75,507 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+      IRRewriter &rewriter, TileDmaBatchGraph &tileDmaBatchGraph) {
+    for (auto &[tileOp, dmaBatch] : tileDmaBatchGraph.getGraph()) {
+      if (failed(processDmaBatch(rewriter, tileOp, dmaBatch))) return failure();
+    }
+    return success();
+  }
+
+  /// Traverse each DmaOp inside ControlCode and replace it with new new DmaOp
+  /// that has Bd Ids assigned using `dmaOpToBdIdMap`.
+  LogicalResult replaceDmaOps(IRRewriter &rewriter,
```
jtuyls, line 473: I feel like this function belongs outside the class and you could add a helper function to the class to return the `dmaOpToBdIdMap`, ideally as a const reference, as this the only info you need inside this function.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
Abhishek-Varma: `currentTileBatch` was used to make the creation of tileDmaBatchGraph fairly simple as it involved node addition to a linkedlist, thus maintaining the pointer to the end.

But, I have now updated the algo to not use LinkedList paradigm anymore. Can you check now?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,188 +75,537 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+
+  void addDmaToBatch(AMDAIE::TileOp tile, AMDAIE::NpuDmaCpyNdOp dmaOp) {
+    std::unique_ptr<DmaBatch> &lastDmaBatch = getLastDmaBatch(tile);
+    lastDmaBatch->currentDmaOps.push_back(dmaOp);
+  };
+
+  /// This API is invoked when we bump into a DMA wait op. That means we have
+  /// reached the end of the current DmaBatch and can start a new DmaBatch in
+  /// the chain for the corresponding tile.
+  void updateLatestTileDmaBatch(AMDAIE::TileOp tile) {
```
jtuyls, line 179:
```suggestion
  void appendNewBatch(AMDAIE::TileOp tile) {
```
I think this would be more descriptive of what the function does.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,188 +75,537 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+    std::unique_ptr<DmaBatch> &oldLastDmaBatch = getLastDmaBatch(tile);
+    tileDmaBatchGraph[tile].push_back(std::make_unique<DmaBatch>());
+    std::unique_ptr<DmaBatch> &newLastDmaBatch = getLastDmaBatch(tile);
+    newLastDmaBatch->forOpParent = oldLastDmaBatch->forOpParent;
+  };
+
+  /// Given another `TileDmaBatchGraph` instance, we consider that as the nested
+  /// graph to be added to the current graph instance. We do this by updating
+  /// the `immediateInnerBatches` of the corresponding tile's DmaBatch.
+  void updateInnerBatchesOfCurrentTileDmaBatchGraph(
```
jtuyls, line 189:
```suggestion
  void updateInnerBatchesOfLastBatch(
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,188 +75,537 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
-      if (subIterations) {
-        bdIdCount += subLoopBdIdCount * subIterations.value();
-      } else {
-        bdIdCount += subLoopBdIdCount;
+    }
+  };
+
+  /// Traverse each DmaBatch in a given (Tile -> list_of_DmaBatches) and infer
+  /// Bd Ids required for it.
+  void inferBdIdsRequiredInBatches() {
```
jtuyls, line 202:
```suggestion
  void inferBdIdsRequired() {
```
```suggestion
  void inferRequiredBdIds() {
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
```diff
@@ -75,188 +75,537 @@ std::optional<uint32_t> getNumberIterations(scf::ForOp loop) {
...
+  // A mapping from a BD ID Op to the BD IDs it covers.
+  DenseMap<AMDAIE::BdIdOp, SmallVector<uint32_t>> bdIdOpToBdIdsMap;
+  // A mapping from DMAOp to its corresponding source/target BD IDs.
+  DenseMap<AMDAIE::NpuDmaCpyNdOp, SmallVector<AMDAIE::BdIdOp, 2>>
+      dmaOpToBdIdMap;
+
+ public:
+  BdIdAssignmentUtil(
+      DenseMap<Value, ChannelBdIdGenerator> shimTileToGeneratorMap)
+      : shimTileToGeneratorMap(shimTileToGeneratorMap) {}
```
jtuyls, line 250:
```suggestion
      : shimTileToGeneratorMap(std::move(shimTileToGeneratorMap)) {}
```
This should avoid a copy.

PR #1247, opened 2025-04-28
Bump IREE to iree-org/iree@b66819fe

yzhang93:
Update `Flow` dependencies to `TensorExt` according to IREE changes https://github.com/iree-org/iree/pull/20564.


Abhishek-Varma (approved):
LGTM.

PR #1248, opened 2025-04-29
[CtrlPkt] Fix `three_matmuls` test

Yu-Zhewen:
Follow-up to #1210 and #1243.

This PR introduces a new  `barrier` op to provide finer-grained control over the ordering of control packets.

It serves as an optimization boundary, preventing any DMA wait/sync optimizations (such as folding or chaining) from crossing it. 

Currently, the barrier is heuristically set when control packets switch between destination tile types.

Issue #1148, opened 2025-02-27
Matmul packet flow performance


Yu-Zhewen:
{
        "M": 512,
        "N": 512,
        "K": 4096,
        "use_ukernel": True,
        "peano_opt_level": 3,
        "outline": "balanced",
        "transpose_a": False,
        "transpose_b": False,
        "tile_pipeline": "pack-peel",
    }

|         | Latency(us) | Instructions |
|---------|-------------|--------------|
| circuit | 2695        | 1096         |
| packet  | 225164      | 559236       |

PR #1251, opened 2025-04-30
[CI] Enable control packets by default

Yu-Zhewen:
Depends on #1248.

This PR updates most tests in the python script to use control packets and also packet flows by default, except:
- air pipeline tests: due to missing passes related to control packets
- convolution tests + two strix matmul tests: due to the deadlock issue described in #1148

### DIFF
--- a/build_tools/ci/cpu_comparison/matmul_test_config.py
+++ b/build_tools/ci/cpu_comparison/matmul_test_config.py
@@ -189,6 +189,7 @@ npu4_matmul_tests = [
         "use_chess_for_ukernel": False,
         "tile_pipeline": "pack-peel-4-level-tiling",
         "name_suffix": "4rows_8cols_npu4",
+        "enable_ctrlpkt": False,
     },
     {
         "M": 512,
@@ -209,6 +210,7 @@ npu4_matmul_tests = [
         "acc_type": "i32",
         "tile_pipeline": "pack-peel-4-level-tiling",
         "name_suffix": "4rows_8cols_npu4_pack_peel_4_level_tiling",
+        "enable_ctrlpkt": False,
     },
     {
         "M": 1024,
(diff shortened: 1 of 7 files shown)


build_tools/ci/cpu_comparison/matmul_test_config.py
```diff
@@ -189,6 +189,7 @@
         "use_chess_for_ukernel": False,
         "tile_pipeline": "pack-peel-4-level-tiling",
         "name_suffix": "4rows_8cols_npu4",
+        "enable_ctrlpkt": False,
```
Yu-Zhewen, line 192: Disabled only for these two specific Strix Matmul tests, as the sorting logic in [assignTiles](https://github.com/nod-ai/iree-amd-aie/blob/3b3a953e3a0eccadc4212128f5860c2fc4cd3bda/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp#L268) results in different routing behavior compared to other Matmul tests, ultimately leading to a packet flow deadlock.

Yu-Zhewen, line 192: Disabled only for these two specific Strix Matmul tests, as the sorting logic in [assignTiles](https://github.com/nod-ai/iree-amd-aie/blob/3b3a953e3a0eccadc4212128f5860c2fc4cd3bda/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignTiles.cpp#L268) results in different routing behavior compared to other Matmul tests, ultimately leading to a packet flow deadlock.

build_tools/ci/cpu_comparison/matmul_test_config.py
```diff
@@ -209,6 +210,7 @@
         "acc_type": "i32",
         "tile_pipeline": "pack-peel-4-level-tiling",
         "name_suffix": "4rows_8cols_npu4_pack_peel_4_level_tiling",
+        "enable_ctrlpkt": False,
```
Yu-Zhewen, line 213: Same reason as above.

jtuyls (approved):
LGTM, very nice!

PR #1252, opened 2025-05-02
[AssignChannels] Prioritize circuit over packet connections

Yu-Zhewen:
Due to the round-robin assignment mode for packet flows, they attempt to utilize all available channels for load balancing. Therefore, they should be assigned only after circuit flows. Note that this change does not affect control flows, as their channels are pre-assigned in a separate pass ([GenerateControlOverlay](https://github.com/nod-ai/iree-amd-aie/blob/8546f47d34c62ca9b2f796b192f06b901141834d/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEGenerateControlOverlay.cpp#L67)).

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignChannels.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignChannels.cpp
@@ -90,10 +90,20 @@ LogicalResult assignChannels(AMDAIE::WorkgroupOp workgroupOp) {
     return failure();
   }
   // Get all `amdaie.connection` ops.
-  SmallVector<AMDAIE::ConnectionOp> connectionOps;
-  workgroupOp->walk([&](AMDAIE::ConnectionOp connectionOp) {
-    connectionOps.push_back(connectionOp);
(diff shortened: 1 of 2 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignChannels.cpp
```diff
@@ -90,10 +90,20 @@ LogicalResult assignChannels(AMDAIE::WorkgroupOp workgroupOp) {
...
+  workgroupOp->walk([&](AMDAIE::ConnectionOp op) {
+    if (op.getConnectionType() == AMDAIE::ConnectionType::Packet) {
+      packetConnections.push_back(op);
+    } else {
+      circuitConnections.push_back(op);
+    }
   });
+  SmallVector<AMDAIE::ConnectionOp> connectionOps;
+  connectionOps.reserve(circuitConnections.size() + packetConnections.size());
+  // Append cicuit connections first, so that they are also assigned first.
```
jtuyls, line 103:
```suggestion
  // Append circuit connections first, so that they are also assigned first.
```

jtuyls (approved):
LGTM

PR #1256, opened 2025-05-06
[AssignBufferAddresses] Improve bank-aware scheme to support large  buffer across multiple contiguous banks

yzhang93:
This PR mainly contains two improvements:
1. Switch different schemes per tile, so if a particular tile failed on `bank-aware` scheme, it will fall back to the `sequential` scheme but not affect any buffer in other tiles.
2. For large buffers that cannot fit in a single bank, it will try to allocate a group of contiguous banks and avoid bank conflicts.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
+++ b/compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
@@ -28,26 +28,23 @@ static uint32_t getMaxMemorySize(AMDAIEDeviceModel deviceModel, TileOp tile) {
 }
 
 //===----------------------------------------------------------------------===//
-// BasicAllocation : sequential alloc from largest to smallest
+// BasicAllocation : sequential allocation for all buffers
 //===----------------------------------------------------------------------===//
-static LogicalResult basicAllocation(
-    DenseMap<TileOp, SetVector<BufferOp>> &tileToBuffers,
-    AMDAIEDeviceModel deviceModel) {
-  for (auto &&[tile, buffers] : tileToBuffers) {
-    // Leave room at the bottom of the address range for stack.
-    int64_t address = 0;
-    if (CoreOp core = getCoreOp(tile)) address += core.getStackSize();
+static LogicalResult basicAllocation(TileOp tile, SetVector<BufferOp> buffers,
+                                     AMDAIEDeviceModel deviceModel) {
+  // Leave room at the bottom of the address range for stack.
+  int64_t address = 0;
+  if (CoreOp core = getCoreOp(tile)) address += core.getStackSize();
 
-    for (BufferOp buffer : buffers) {
-      buffer.setAddress(address);
-      address += getAllocationSize(buffer);
-    }
+  for (BufferOp buffer : buffers) {
+    buffer.setAddress(address);
+    address += getAllocationSize(buffer);
+  }
 
-    uint32_t maxDataMemorySize = getMaxMemorySize(deviceModel, tile);
-    if (address > maxDataMemorySize) {
-      return tile.emitOpError("allocated buffers exceeded available memory (")
-             << address << ">" << maxDataMemorySize << ")\n";
-    }
+  uint32_t maxDataMemorySize = getMaxMemorySize(deviceModel, tile);
+  if (address > maxDataMemorySize) {
+    return tile.emitOpError("allocated buffers exceeded available memory (")
+           << address << ">" << maxDataMemorySize << ")\n";
   }
   return success();
 }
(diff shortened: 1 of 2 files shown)


compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
```diff
@@ -176,73 +220,70 @@ void deAllocateBuffers(SmallVector<BufferOp> &buffers) {
...
+  // in that bank, and the index is the bank number.
+  SmallVector<int64_t> nextAddrInBanks;
+  for (int i = 0; i < numBanks; i++) nextAddrInBanks.push_back(bankSize * i);
+  // Leave room at the bottom of the address range for stack.
+  if (CoreOp core = tile.getCoreOp()) {
+    int64_t stackSize = core.getStackSize();
+    if (stackSize > bankSize)
+      return tile.emitOpError("stack size: ")
+             << stackSize
+             << " should not be larger than the bank size: " << bankSize;
```
Yu-Zhewen, line 239:
```suggestion
    if (stackSize > bankSize) {
      return tile.emitOpError("stack size: ")
             << stackSize
             << " should not be larger than the bank size: " << bankSize;
    }
```

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
```diff
@@ -176,73 +220,70 @@ void deAllocateBuffers(SmallVector<BufferOp> &buffers) {
...
+  for (BufferOp buffer : buffers) {
+    FailureOr<bool> has_bank =
+        checkAndAddBufferWithMemBank(buffer, nextAddrInBanks, bankLimits);
+    FailureOr<bool> has_addr = checkAndAddBufferWithAddress(
+        buffer, bankSize, nextAddrInBanks, bankLimits);
+    if (failed(has_addr) || failed(has_bank)) return failure();
+    if (!has_addr.value() && !has_bank.value())
+      buffersToAlloc.push_back(buffer);
+    else
+      preAllocatedBuffers.push_back(buffer);
```
Yu-Zhewen, line 265:
```suggestion
    if (!has_addr.value() && !has_bank.value()) {
      buffersToAlloc.push_back(buffer);
    }
    else {
      preAllocatedBuffers.push_back(buffer);
    }
```

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
```diff
@@ -176,73 +220,70 @@ void deAllocateBuffers(SmallVector<BufferOp> &buffers) {
...
-    SmallVector<BankLimits> bankLimits;
-    for (int i = 0; i < numBanks; i++) {
-      bankLimits.emplace_back(i * bankSize, (i + 1) * bankSize);
-    }
+  // Each entry of `bankLimits` contains pairs of start and end addresses for
+  // that bank.
+  SmallVector<BankLimits> bankLimits;
+  for (int i = 0; i < numBanks; i++) {
+    bankLimits.emplace_back(i * bankSize, (i + 1) * bankSize);
+  }
```
Yu-Zhewen, line 248:
```suggestion
  for (int i = 0; i < numBanks; i++)
    bankLimits.emplace_back(i * bankSize, (i + 1) * bankSize);
```

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
```diff
@@ -131,41 +128,88 @@ FailureOr<bool> checkAndAddBufferWithMemBank(
...
+// contiguous group of banks, starting from the given index and wrapping around
+// if needed. It finds the minimal number of contiguous banks required to hold
+// the buffer and checks if sufficient space is available. On success, it sets
+// the buffer's memory bank attribute and address, and updates
+// `nextAddrInBanks`. If no suitable banks are found, it emits an error and
+// returns false.
+bool setBufferAddress(BufferOp buffer, uint32_t numBanks, uint32_t bankSize,
+                      int &startBankIndex,
                       SmallVector<int64_t> &nextAddrInBanks,
                       const SmallVector<BankLimits> &bankLimits) {
```
Yu-Zhewen, line 141: Not specific to this PR, but I believe the following holds:
`nextAddrInBanks.size() == bankLimits.size() == numBanks`, and
`bankLimits[bankIndex].endAddr - bankLimits[bankIndex].startAddr == bankSize`.

If that's the case, it might be worth consolidating the related data into a single struct, perhaps expanding the current `struct BankLimits` to something like:
```cc
struct Bank {
  int64_t startAddrLimit;
  int64_t endAddrLimit;
  int64_t nextAddrAvail;
  ...
};
```
so that we only need to do
```cc
bool setBufferAddress(BufferOp buffer, int &startBankIndex, SmallVector<Bank> &banks)
```

Just a thought for potential future refactoring.

Yu-Zhewen (approved):
LGTM, only a few format-related comments.

compiler/plugins/target/AMD-AIE/aie/AMDAIEAssignBufferAddresses.cpp
yzhang93: Thanks, since there will be quite a lot renaming if I update it to a single struct, I'll add it as a follow-up.

PR #1258, opened 2025-05-07
[AssignConnectionTypes][CI] Use packet flows only under congestion

Yu-Zhewen:
Previously, when control packets were enabled, all kernel input data flows were forced to use packet flows as well. This increased the likelihood of router arbiter allocation failures due to potential deadlocks #1254.

This PR introduces an "auto" connection type assignment strategy, where packet flows are used only when congestion is detected. With this more selective approach, we reduce the risk of arbiter allocation failure.

As a result, control packets can now be enabled for the convolution tests and the two Strix MatMul tests referenced in #1251.


Yu-Zhewen (commented):
Previously, routing issues in the convolution test led to a deadlock.

IR after the router:
https://gist.github.com/Yu-Zhewen/94ae54141a6f4c6739927a94d8f1e64b

Specifically, 
```ll
%switchbox_0_1 = aie.switchbox(%tile_0_1) {
  %0 = aie.amsel<0> (0)
  %1 = aie.amsel<1> (0)
  %2 = aie.amsel<2> (0)
  %3 = aie.amsel<3> (0)
  %4 = aie.amsel<4> (0)
  %5 = aie.amsel<5> (0)
  %6 = aie.amsel<0> (1)
  %7 = aie.amsel<1> (1)
  %8 = aie.amsel<0> (2)
```

build_tools/ci/run_matmul_test.sh
```diff
@@ -591,14 +591,16 @@ run_matmul_test_on_shapes ${i32_shapes_small[@]} \
...
+# TODO (zhewen): Support kernel outputs.
 run_matmul_test_on_shapes ${i32_shapes_small[@]} \
     --name_prefix "small" \
     --lower_to_aie_pipeline "objectFifo" \
     --tile_pipeline "pack-peel" \
     --lhs_rhs_type "i32" \
     --acc_type "i32" \
     --num_corruption_repeat_runs "10" \
-    --enable_packet_flow "true"
+    --enable_packet_flow "inputs"
```
jtuyls, line 603: `auto` won't work here? Because of the outputs?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Passes.cpp
```diff
@@ -601,10 +601,10 @@ void buildAMDAIETransformPassPipeline(
     uint32_t numCols, TilePassPipeline useTilePipeline,
     LowerToAIEPassPipeline useLowerToAIEPipeline, bool matmulElementwiseFusion,
     bool enableVectorizationPasses, std::string enableAMDAIEUkernels,
-    const std::string &pathToUkernels, bool enableInputPacketFlow,
-    bool enableOutputPacketFlow, bool enableCoalescingLoops,
-    bool enableCollapsingUnitDims, OutliningStrategy enableFunctionOutlining,
-    int callReplication, bool insertLoopAroundCoreBlock, bool enableCtrlPkt) {
+    const std::string &pathToUkernels, PacketFlowStrategy enablePacketFlow,
```
jtuyls, line 604: Maybe rename `enablePacketFlow` to `packetFlowStrategy`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignConnectionTypes.cpp
```diff
@@ -29,11 +31,216 @@ class AMDAIEAssignConnectionTypesPass
...
+            (b.direction == AMDAIE::DMAChannelDir::MM2S)
+                ? deviceModel.getNumSourceSwitchBoxConnections(
+                      b.tileLoc.col, b.tileLoc.row, b.portType)
+                : deviceModel.getNumDestSwitchBoxConnections(
+                      b.tileLoc.col, b.tileLoc.row, b.portType);
+      }
+    }
+    return maxCircuitUsage[b];
+  };
+  for (size_t i = 0; i < connectionOps.size(); ++i) {
```
jtuyls, line 169: You could use enumerate here to iterate through both index and connection op.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignConnectionTypes.cpp
```diff
@@ -29,11 +31,216 @@ class AMDAIEAssignConnectionTypesPass
   void runOnOperation() override;
 };
 
-void AMDAIEAssignConnectionTypesPass::runOnOperation() {
-  Operation *parentOp = getOperation();
-  IRRewriter rewriter(parentOp->getContext());
+/// Utility function to retrieve the endpoint infomation (tile location, port
```
jtuyls, line 34:
```suggestion
/// Utility function to retrieve the endpoint information (tile location, port
```

build_tools/ci/run_matmul_test.sh
Yu-Zhewen: I think the intention for these specific tests are to validate packet flow functionality; therefore, they should utilize as many packet flows as possible.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.h
```diff
@@ -71,6 +71,26 @@ struct SwitchSetting {
 
 using SwitchSettings = std::map<TileLoc, SwitchSetting>;
 
+struct PhysBundle {
```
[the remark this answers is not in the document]
Yu-Zhewen: In our codebase, I believe the term `port/endpoint` typically refers to `bundle/port type + channel ID`. The key distinction here is that this new struct doesn't carry a channel ID, which is assigned only in a later pass.

build_tools/ci/run_matmul_test.sh
jtuyls: Ok, nice, could you add this as a comment?

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.h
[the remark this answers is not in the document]
Yu-Zhewen: Then maybe something like `PhysStrmSwPortType/PhysPortType` ? Since, `PhysPort=TileLoc+Port+Direction` and the new struct is `PhysStrmSwPortType=TileLoc+StrmSwPortType+Direction`

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignConnectionTypes.cpp
```diff
@@ -29,11 +31,219 @@ class AMDAIEAssignConnectionTypesPass
...
+        auto channelOp = dyn_cast<AMDAIE::ChannelOp>(channel.getDefiningOp());
+        if (!channelOp) {
+          return connectionOp.emitOpError()
+                 << "expected an `amdaie.channel` op";
+        }
+        AMDAIE::TileOp tileOp = channelOp.getTileOp();
+        int32_t col = getConstantIndexOrAssert(tileOp.getCol());
+        int32_t row = getConstantIndexOrAssert(tileOp.getRow());
+        physBundles.push_back({{col, row}, channelOp.getPortType(), direction});
+      }
```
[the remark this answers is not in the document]
Yu-Zhewen: Yes, we're seeing both cases. Control packet related connections already have channels assigned, and they may use port types other than DMA (e.g., CTRL ports), so it's better to retrieve the info directly from the channels in those cases.

For other connections where channels haven't been assigned yet, we fall back to using the LogicalObjFifo to get the tile information and assume the port type is always DMA.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.h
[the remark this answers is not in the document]
jtuyls: Yeah, makes sense to me.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignConnectionTypes.cpp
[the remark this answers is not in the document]
jtuyls: Yeah, makes sense, I got a bit confused on when the channels were assigned already.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignConnectionTypes.cpp
```diff
@@ -29,11 +31,219 @@ class AMDAIEAssignConnectionTypesPass
...
+    return failure();
+  }
+  return physBundles;
+}
+
+/// Checks if the given `PhysBundle` is contained in the provided list of
+/// connection operations.
+FailureOr<bool> containsPhysBundle(
+    const PhysBundle &key, ArrayRef<AMDAIE::ConnectionOp> connectionOps) {
+  for (AMDAIE::ConnectionOp op : connectionOps) {
```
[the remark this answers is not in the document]
Yu-Zhewen: Now the information is cached here
https://github.com/nod-ai/iree-amd-aie/blob/351ad18f82b4041f521342c28f946957c1efde58/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignConnectionTypes.cpp#L159-L167

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignConnectionTypes.cpp
```diff
@@ -29,11 +31,212 @@ class AMDAIEAssignConnectionTypesPass
...
+                               ? deviceModel.getNumSourceSwitchBoxConnections(
+                                     b.tileLoc.col, b.tileLoc.row, b.portType)
+                               : deviceModel.getNumDestSwitchBoxConnections(
+                                     b.tileLoc.col, b.tileLoc.row, b.portType);
+    }
+    return maxCircuitUsage[b];
+  };
+
+  // Record the final connection index associated with each physical port type.
+  DenseMap<PhysPortType, uint32_t> physPortTypeToFinalConnectionIndex;
```
jtuyls, line 160:
```suggestion
  DenseMap<PhysPortType, uint32_t> physPortTypeToLastConnectionIndex;
```
Nit, but you could consider renaming 'final' to 'last' as I think it's slightly more clear, but this is just a suggestion.

jtuyls (approved):
LGTM

PR #1260, opened 2025-05-09
Add flag to specify stack size to be allocated in AIE cores

jtuyls:
This is needed to start enabling O2 in the peano ukernel compilation as the stack size needs increase. This doesn't enable O2 yet by default as that results in out-of-program memory issues in other tests, so leaving that to a follow up.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
@@ -62,7 +62,7 @@ def AMDAIE_CoreOp: AMDAIE_Op<"core", [SingleBlock, AttrSizedOperandSegments]>, R
     relay to AIE core op for linking with Ukernel.
 
     It has a `stack_size` attribute with default value to specify the size of stack
-    buffer that has to be allocated on the cores' local memory.
+    buffer that has to be allocated on the core's local memory.
   }];
 
   let arguments = (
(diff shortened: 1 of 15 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
```diff
@@ -60,12 +60,16 @@ def AMDAIE_CoreOp: AMDAIE_Op<"core", [SingleBlock, AttrSizedOperandSegments]>, R
 
     It has an optional attribute `link_with` which will carry the information to
     relay to AIE core op for linking with Ukernel.
+
+    It has a `stack_size` attribute with default value to specify the size of stack
+    buffer that has to be allocated on the cores' local memory.
```
newling, line 65:
```suggestion
    buffer that has to be allocated on the core's local memory.
```

newling: I assume it's a single a single core referred to here?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/AIETarget.h
```diff
@@ -91,6 +91,9 @@ struct AMDAIEOptions {
   enum class DeviceHAL { XRT, XRT_LITE };
   DeviceHAL deviceHal{DeviceHAL::XRT_LITE};
 
+  // The default stack size for all cores is 1024.
```
newling, line 94:
```suggestion
  // The default stack size for all cores is 1024 bytes.
```

newling (approved):
LGTM. Long-term still better to automatically set the stack-size I guess?

PR #1267, opened 2025-05-14
[CI] Do not run benchmark tests twice on strix

Yu-Zhewen:
This PR updates the CI workflow to skip the Performance (benchmark) tests during the `E2E comparison of AIE to llvm-cpu` step on Strix. The corresponding numerical correctness tests will still run as expected.

No changes are needed for Phoenix tests, as the `--skip_tests` flag is already set correctly for them.

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -313,6 +313,7 @@ jobs:
             --reset_npu_between_runs \
             --xrt_lite_n_core_rows=$XRT_LITE_N_CORE_ROWS \
             --xrt_lite_n_core_cols=$XRT_LITE_N_CORE_COLS \
+            --skip_tests=Performance \
             -v
 
       # Run the 'Performance' tests. These do not check numerical correctness,


.github/workflows/ci-linux.yml
```diff
@@ -313,6 +313,7 @@ jobs:
             --reset_npu_between_runs \
             --xrt_lite_n_core_rows=$XRT_LITE_N_CORE_ROWS \
             --xrt_lite_n_core_cols=$XRT_LITE_N_CORE_COLS \
+            --skip_tests=Performance \
             -v
 
       # Run the 'Performance' tests. These do not check numerical correctness,
```
newling, line 319: Is this comment relevant? 

Can you please double check that the performance tests are tested numerically. 

Also worth considering is if the fact that the output is piped i.e. 

```
   --tests=Performance > performance_npu4.log
```

in the performance benchmarks means we shouldn't do the numerical testing there (because the diagnostics errors aren't as visible). 

I know the code complexity here is a bit high, so I'm sorry to ask for a double check, but it'd be pretty bad if we accidentally stopped checking the numerics of these tests!

Yu-Zhewen: Looking into the python script: the numerical tests have the label "PerformanceCorrectness", and the benckmark tests have the label "Performance".

https://github.com/nod-ai/iree-amd-aie/blob/00995f9e21022b1dc9d40a9eb5788a8eb145aa94/build_tools/ci/cpu_comparison/run.py#L2340-L2386

newling: Thanks, all good then 🟢

newling (approved):

jtuyls (approved):
Nice!

PR #1268, opened 2025-05-15
[CI] Remove Strix chess tests except one for benchmarking

jtuyls:
Removes most chess tests on Strix, except one for benchmarking, to bring down CI time. Leaves them on Phoenix for now.


Yu-Zhewen (approved):
LGTM. Thanks!

PR #1273, opened 2025-05-21
Enable O2 by default for Peano ukernels

jtuyls:
Enabling O2 by default for compiling peano ukernels on all tests in CI needs following changes:
- Avoid out-of-memory errors (program memory)  by adding `-ffunction-sections` and `-fdata-sections` flags to put each function and data item into its own section to enable them to be discarded later during linking with `-Wl,--gc-sections`.
- Avoid numerical issues due to not enough stack size being allocated in matmul tests, leading to override of data buffers on L1. This is solved for now by passing large enough stack sizes for those tests. The exact stack sizes being passed are found as experimentally as the smallest multiple of 1024 that is large enough.

### DIFF
--- a/build_tools/ci/cpu_comparison/run.py
+++ b/build_tools/ci/cpu_comparison/run.py
@@ -54,6 +54,7 @@ class TestParams(ABC):
         run_benchmark=False,
         n_repeats=1,
         enable_ctrlpkt=True,
+        stack_size=1024,
     ):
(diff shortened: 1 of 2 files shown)


newling (approved):
LGTM
For future reference, please add a comment on the PR description why stack size 3072 is chosen, and what the error is with stack size 1024 for these tests

PR #1274, opened 2025-05-21
[aievec][nfc] Make test names consistent and remove unused code 

newling:


Yu-Zhewen (commented):
Regarding the test names, I believe the rest of the repository consistently uses underscores, while only the aievec tests were mixing underscores and hyphens?

Yu-Zhewen (approved):
LGTM. Thanks!

PR #1276, opened 2025-05-22
[KernelDispatch] Add a temporary hack for 1x1 shapes with tile sizes M,N

Abhishek-Varma:
-- This commit adds a temporary hack for 1x1 AIE array to make those
   2D matmul shapes work for which all of the operands get pulled in to L2
   buffer. Once reprogramming of DMA ops is supported, we can get rid of this
   workaround. We need to add this only for pack-peel-4-level-tiling NOT
   pack-peel. The workaround just ensures that the tile size of first level is
   NOT equal to M,N by halving the n0Tile.
-- Also adds e2e 32x512x64 for 1x1 npu4 test.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
@@ -515,6 +515,28 @@ static LogicalResult setRootConfigForPackPeel4LevelTilingPipeline(
     m0Tile /= maybeInputDimsAndSizes.value().mSizes.back();
     n0Tile /= maybeInputDimsAndSizes.value().nSizes.back();
   }
+  // TODO(avarma): This is currently a workaround for 1x1 AIE array to make
+  // those 2D matmul shapes work for which all of the operands get pulled in to
+  // L2 buffer. Once reprogramming of DMA ops is supported, we can get rid of
+  // this workaround. We need to add this only for pack-peel-4-level-tiling NOT
+  // pack-peel. The workaround just ensures that the tile size of first level is
+  // NOT equal to M,N by halving the n0Tile.
+  if (is2DMatmulLike && numRows == 1 && numCols == 1) {
+    auto getTotalSize = [](ArrayRef<int64_t> sizes) {
+      return std::accumulate(sizes.begin(), sizes.end(), 1,
+                             std::multiplies<int64_t>());
+    };
+
+    // Get the shape (M, N) of the full Matmul operation.
+    auto maybeInputDimsAndSizes = getInputDimsAndSizes(linalgOp);
+    int64_t M = getTotalSize(maybeInputDimsAndSizes.value().mSizes);
+    int64_t N = getTotalSize(maybeInputDimsAndSizes.value().nSizes);
+    // Check if the tile size generated is exactly same as operand size. If yes,
+    // halve n0Tile.
+    if (m0Tile == M && n0Tile == N) {
+      n0Tile /= 2;
+    }
+  }
   tileSizeLevel0[mDims[0]] = m0Tile;
   tileSizeLevel0[nDims[0]] = n0Tile;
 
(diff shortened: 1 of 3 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lowering_strategy_objectfifo_npu4.mlir
```diff
@@ -40,3 +41,54 @@ module {
     return
   }
 }
+
+// -----
+
+// Tests a matmul shape for pack-peel-4-level-tiling in which the tile size generated
+// should ideally be equal to the M,N,K dimensions of the matmul - but it won't work
```
yzhang93, line 48:
```suggestion
// should ideally be equal to the M, N, K dimension size of the matmul - but it won't work
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -441,6 +450,30 @@ static LogicalResult setRootConfigForPackPeel4LevelTilingPipeline(
...
+      auto maybeInputDimsAndSizes = getInputDimsAndSizes(linalgOp);
+      int64_t M = getTotalSize(maybeInputDimsAndSizes.value().mSizes);
+      int64_t N = getTotalSize(maybeInputDimsAndSizes.value().nSizes);
+      // Check if the tile size generated is exactly same as operand size. If
+      // yes, halve n0Tile.
+      if (m0Tile == M && n0Tile == N) {
+        n0Tile /= 2;
+        if (n0Tile < packedSizesL0[nDims.back()])
+          packedSizesL0[nDims.back()] /= 2;
+      }
```
yzhang93, line 475: I think this part should be added to `ParameterSetting::create` function, so you don't have to duplicate the codes for total M/N size.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lowering_strategy_objectfifo_npu4.mlir
```diff
@@ -40,3 +41,54 @@ module {
...
+// until support for DMA ops' reconfiguration is added. The workaround therefore halves
+// the tile size for N as N/2.
+
+// Pack-peel-4-level tiling on 4x8 cores : the tile size remains maximum in this case.
+// PACK-PEEL-4-LEVEL{LITERAL}: #config = #iree_codegen.lowering_config<tile_sizes = [[32, 512, 0], [4, 4, 0], [0, 0, 1], [1, 1, 0]]>
```
yzhang93, line 53: This seems you are still using 4 rows and 4 cols, so the above description that tiling on 4x8 cores is not accurate.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/lowering_strategy_objectfifo_npu4.mlir
```diff
@@ -40,3 +41,54 @@ module {
...
+  %6 = linalg.fill ins(%c0_i32 : i32) outs(%5 : tensor<32x512xi32>) -> tensor<32x512xi32>
+  %7 = linalg.matmul ins(%3, %4 : tensor<32x64xi32>, tensor<64x512xi32>) outs(%6 : tensor<32x512xi32>) -> tensor<32x512xi32>
+  iree_tensor_ext.dispatch.tensor.store %7, %2, offsets = [0, 0], sizes = [32, 512], strides = [1, 1] : tensor<32x512xi32> -> !iree_tensor_ext.dispatch.tensor<writeonly:tensor<32x512xi32>>
+  return
+}
+
+// -----
+
+// Based on above workaround this test shows the packing size of N also being halved
+// in case the tile size for N dimension becomes lesser than the corresponding packing size.
```
yzhang93, line 77:
```suggestion
// in case the tile size for N dimension becomes less than the corresponding packing size.
```

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -441,6 +450,30 @@ static LogicalResult setRootConfigForPackPeel4LevelTilingPipeline(
     packedSizesL0[nDims.back()] = packPeelTiling.n0Pack;
     packedSizesL0[kDims.back()] = packPeelTiling.k0Pack;
 
+    // TODO(avarma): This is currently a workaround for 1x1 AIE array to make
+    // those 2D matmul shapes work for which all of the operands get pulled in
+    // to L2 buffer. Once reprogramming of DMA ops is supported, we can get rid
```
yzhang93, line 455: I didn't follow what's the problem here. What's the error message?

yzhang93 (changes requested):

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
Abhishek-Varma: Without this we get `'aie.memtile_dma' op could not find and assign a valid BD id` - because of exploding `memtile_dma` issue, which should be solved by having DMA ops reconfigured. But since the support is not yet available, this is a temporary workaround.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp
```diff
@@ -303,6 +303,21 @@ FailureOr<ParameterSetting> ParameterSetting::create(
...
     M0 = maxL0Size.M;
     N0 = maxL0Size.N;
+
+    // TODO(avarma): This is currently a workaround for 1x1 AIE array to make
+    // those 2D matmul shapes work for which all of the operands get pulled in
+    // to L2 buffer. Once reprogramming of DMA ops is supported, we can get rid
+    // of this workaround. We need to add this only for pack-peel-4-level-tiling
+    // NOT pack-peel. The workaround just ensures that the tile size of first
+    // level is NOT equal to M, N by halving the N0 tile.
+    if (numRows == 1 && numCols == 1) {
```
jtuyls, line 313: Do you only see this issue for 1x1? I would expect it for any number of cores?

PR #1279, opened 2025-05-27
Remove untested and unused aievec dialect ops

newling:


yzhang93 (approved):
I'm wondering what's the use case for `FMAElemOp` in mlir-aie. Will it be useful if we have a matvec op?

PR #1285, opened 2025-05-29
[DeviceModel] Raise an error when DMA property is not available

Yu-Zhewen:
This occurs when targeting npu1 (5 columns), where tile (0, 0) does not have a DMA channel.


runtime/src/iree-amd-aie/aie_runtime/iree_aie_runtime.h
```diff
@@ -206,6 +206,89 @@ inline ::XAie_TxnOpcode txnToTxn(XAie_TxnOpcode t) {
 // mlir-air legacy
 enum class AIEArch : uint8_t { AIE1 = 1, AIE2 = 2, AIE2p = 3 };
 
+/// ============================= BEGIN ==================================
```
Yu-Zhewen, line 209: Simply move these lines up, as they are required by `getDmaProp`. Note that `getDmaProp` cannot be moved to the .cc file since it's a template function.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.h
```diff
@@ -322,24 +322,20 @@ struct DmaDimConfig {
...
-      nbIntraDims = deviceModel.getDmaProp<uint8_t>(
-          AMDAIE::AMDAIETileType::AIETILE, AMDAIE::AMDAIEDmaProp::NumAddrDim);
       tileType = AMDAIE::AMDAIETileType::AIETILE;
-      maxNbDims = nbIntraDims;
     } else {
       assert(false && "unsupported memspace: ");
     }
+    FailureOr<uint8_t> maybeNbIntraDims = deviceModel.getDmaProp<uint8_t>(
+        tileType, AMDAIE::AMDAIEDmaProp::NumAddrDim);
+    assert(succeeded(maybeNbIntraDims));
```
jtuyls, line 336: Maybe it's better to set `nbIntraDims` to `0` in case of failure?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/assign_channels.mlir
```diff
@@ -45,6 +45,29 @@ module attributes {hal.executable.target = #executable_target_amdaie_xclbin_fb}
 
 // -----
 
+// Target device `npu1` whose tile (0, 0) has no DMA channels at all. Use `npu1_4col` instead.
```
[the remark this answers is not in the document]
jtuyls: I think these two are good enough for now.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIEDmaUtils.h
Yu-Zhewen: `getDmaProp` will only return failure if the tileType is invalid (`AMDAIETileType::SHIMPL` or `AMDAIETileType::MAX`). However, the if-else logic above already guarantees that tileType is valid by construction. Therefore, using an assertion here should be sufficient?

PR #1289, opened 2025-06-02
[Softmax] Support run on a column of cores

yzhang93:
Currently, the softmax ukernel can run on a single column of 4 cores. Since there is only one parallel dimension, we'll have to modify the logic of `AssignTiles` to utilize more AIE columns.


jtuyls (approved):
LGTM, but let's wait for the Strix CI as well

PR #1295, opened 2025-06-06
Include ukernel in stack sizes check for each core

jtuyls:
Previously, the ukernel stack sizes where not included in the stack size calculation and check. This PR adds it by passing the `-fstack-size-section` flag to the ukernel compilation and refactors the logic to check each core file separately.


newling (approved):
LGTM, just 1 clarification question and 1 nit. 

Oh and is there some stack-size related code in run.py that can be deleted?

PR #1303, opened 2025-06-18
Use pip install in peano download script

Yu-Zhewen:
Previously, we used `pip download` and then extracted the .whl manually. However, this approach fails when there are multiple versions of llvm_aie wheels present in the same directory, as the script cannot reliably identify the correct .whl to unzip, and will raise an error.

To make the process more simple and robust, we now use `pip install` directly with the `--target` option to install/overwrite llvm_aie to a specified directory. 

However, a drawback of using `--target` is that it disables pip's cache mechanism, causing the .whl file (~140MB) to be downloaded every time the script runs, even if the version hasn't changed.


jtuyls (approved):
LGTM, this is great!

> However, a drawback of using --target is that it disables pip's cache mechanism, causing the .whl file (~140MB) to be downloaded every time the script runs, even if the version hasn't changed.

I think that's fine. Doesn't seem to have too much impact on the CI times.

PR #1304, opened 2025-06-18
[Softmax] Add chess uKernel for npu4

Yu-Zhewen:
Note that the accuracy for npu1 and npu4 are not the same, due to a different implementation of the exponential function.


newling:
>  rtol=6e-2,

So... 1.000 == 1.055 😮

yzhang93 (approved):
LGTM. Just one question, why is the lookup table not needed for npu4?

newling:
Thanks @Yu-Zhewen for the explanations

newling:
> So far, I am just copying whatever mlir-aie has

Please mention this is PR descriptions

Issue #1306, opened 2025-06-20
Direct codegen vectorized lowering of reduction operation


Sameeranjoshi:

Sameeranjoshi:
> To reproduce the above IR, modify this [line](https://github.com/nod-ai/iree-amd-aie/blob/d4436019e7f9001a7c4834c9c6632d06e69851fd/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/KernelDispatch.cpp#L860) to

Along with the above change need `f32->bf16` in the input test as `f32` doesn't reproduce this issue as it can't vectorize.

```
// These lines are required for e2e numerical testing:
// input 128x128xbf16
// output 128xbf16

!in_ty = tensor<128x128xbf16>
!out_ty = tensor<128xbf16>

func.func @reduction_sum(%arg0: !in_ty) -> !out_ty {
  %cst = arith.constant 0.0 : bf16
  %3 = tensor.empty() : !out_ty
  %4 = linalg.fill ins(%cst : bf16) outs(%3 : !out_ty) -> !out_ty
  %5 = linalg.generic {indexing_maps = [affine_map<(d0, d1) -> (d0, d1)>, affine_map<(d0, d1) -> (d0)>], iterator_types = ["parallel", "reduction"]} ins(%arg0 : !in_ty) outs(%4 : !out_ty) {
  ^bb0(%in: bf16, %out: bf16):
    %6 = arith.addf %in, %out : bf16
    linalg.yield %6 : bf16
  } -> !out_ty
  return %5 : !out_ty
}

```

Command to reproduce:
```
iree-compile reduction_sum.mlir --iree-hal-target-backends=amd-aie --iree-amdaie-target-device=npu1_4col --iree-amdaie-tile-pipeline=general-copy   --iree-amdaie-num-rows=4 --iree-amdaie-num-cols=4  -
-mlir-print-ir-after-all &> after_all.mlir
```

Sameeranjoshi:
> What does our test-lower-vector-to-aievec do with this?

**Here's the summary**: No change [diff](https://www.diffchecker.com/IyORb4Zr/) 
**Command used**: `iree-opt test_files/amdaievectorization.mlir  --test-lower-vector-to-aievec`
**Input:** I am assuming this flag is applied after `AMDAIEVectorization (iree-amdaie-vectorization)` optimization as that's the first one where I can see `vector.*` references in the IR, I might be wrong.

[...]

PR #1312, opened 2025-06-26
Apply pattern for `populateVectorMultiReductionLoweringPatterns` from…

Sameeranjoshi:
… mlir-opt in iree-opt

For issue #1306, converting vector.multi_reduction to vector.reduction is the first step as `vector.reduction` can be converted to Peano instructions which can vectorize. `mlir-opt` supports it with the flag `lower-vector-multi-reduction` and the `populateVectorMultiReductionLoweringPatterns` pattern, reuse it in IREE, as a last pass to kick in `AMDAIEVectorizationPass` pipeline.


newling:
My understanding is that this PR is **all** that is needed to have vectorization through peano. There is currently no dialect conversion from vector::MultiDimReductionOp to llvm, so if we were to not use vector::ReductionOp, we'd have to implement that conversion. The current expectation is vector.multi_reduction has been lowered before conversion to llvm.

newling:
wrt linting error:

I think something like
```
pip install pre-commit
pre-commit run --all-files
```

will fix all linting errors (python, c++, whitespace removal, etc)

PR #1314, opened 2025-06-27
[CI] Refactor benchmark functions to support more operations

Yu-Zhewen:
This PR refactors the code to improve reusability and clarity in the benchmarking infrastructure:

- Moves `vs_cpu` and `benchmark` functions from the `BaseMatmul` class to `BaseTest`, enabling them to be shared across different operation types (e.g., convolution, softmax, ...).

- Uses the `@dataclass` decorator for `TestParams`, reducing boilerplate and improving readability.

- Previously, parameters like `n_repeats` were passed indirectly through `TestParams`, while others like `n_kernel_runs` were passed directly to test class constructors. This inconsistency sometimes led to confusion and unintended overrides.  The PR now embeds more parameters into `TestParams` instead.

Note: This PR does not actually turn on benchmarks for other operations in order to avoid increasing CI workload. Instead, benchmarks can be enabled as needed or run locally by setting `run_benchmark=True`.

### DIFF
--- a/build_tools/ci/cpu_comparison/performance_publish.py
+++ b/build_tools/ci/cpu_comparison/performance_publish.py
@@ -15,14 +15,15 @@ def append_history(results_json_path: str, results_history_path: str):
     # Append the results to the history.
     results_history = []
     max_history = 100
-    with open(results_history_path, "r+") as f:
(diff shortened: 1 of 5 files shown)


newling (approved):
Thanks 🙏

PR #1319, opened 2025-07-07
[Softmax] Select tile size based on op shape

Yu-Zhewen:


compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/uKernels/npu1/chess/zero_fill.cc
```diff
@@ -42,6 +42,7 @@ zero_fill_combos(zero_fill_vectorized_c_func, 16, 16)
 zero_fill_combos(zero_fill_vectorized_c_func, 32, 32)
 zero_fill_combos(zero_fill_vectorized_c_func, 64, 32)
 zero_fill_combos(zero_fill_vectorized_c_func, 64, 64)
+zero_fill_vectorized_c_func(bfloat16, bf16, 4, 1024, 32)
```
Yu-Zhewen, line 45: Can't use `zero_fill_combos` here, as it would set the vector size r to be N/2, which causes numerical issues when N is too large (1024 in this case).

jtuyls (approved):
Nice!

PR #1320, opened 2025-07-08
Bump IREE to iree-org/iree@ce656af8

yzhang93:
Upstream has some changes of bufferization and vectorization APIs.


Yu-Zhewen (approved):
Thanks!

Issue #1287, opened 2025-05-30
Support reconfiguration of DMAs


Abhishek-Varma:
The following describes the set of tasks and their context for supporting reconfiguration of DMAs.

### OPTION 1 : Circular DmaCpyNd -> Non-circular DmaCpyNd
This means all conversions around Circular DmaCpyNdOp should be based around Non-circular DmaCpyNd op.
Circular DmaCpyNdOp runs infinitely while Non-circular DmaCpyNdOp runs once for any given configuration; and the latter can be then made to run for some other configuration. This is the reconfiguration option we're currently choosing to go ahead with because of an important "UNKNOWN" factor in option 2 (mentioned towards the end).

Current state of the IR so far with option 1 : [IR log](https://gist.github.com/Abhishek-Varma/bfc3ac0225da69cb40f4494df58191c1)

Phoenix test's log : [reprogram_runtime_zero_phoenix.txt](https://github.com/user-attachments/files/20778026/reprogram_runtime_zero_phoenix.txt)

### OPTION 2 : Include a new op to reset BD IDs in Memtile_Dma
> Context for `numIters`:

In `lower-to-aie` pass [numIters](https://github.com/nod-ai/iree-amd-aie/blob/ba8541a1bf8a9fac09b8fa7c5bd17b1f5d924104/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp#L236) essentially gets hold of the canonicalized repeat counts needed to execute a DMA block.
This when multiplied with total [bufferOps](https://github.com/nod-ai/iree-amd-aie/blob/ba8541a1bf8a9fac09b8fa7c5bd17b1f5d924104/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIELowerToAIE.cpp#L246C44-L246C53) gives the total number of DMA blocks that are going to be invoked as part of a DMA chain started by `aie.dma_start`.

-- (HP) Create a separate/common utility that gets hold of `numIters` : we would want to make use of this information in one of the earlier pass and then again at `lower-to-aie`.
-- (LP) Perhaps create an attribute as part of `amdaie.npu.circular_dma_cpy_nd` to make use of this information later at `lower-to-aie` instead of re-invoking/recomputing the same information.

> Context of `aie.memtile_dma` + Transcation Binary: (TO BE USED IN OPTION 1 AS WELL)

[...]

PR #1325, opened 2025-07-14
[Reprogramming][BufferAddress] Add Assign Buffer Address + BD IDs pass to AMDAIE

Abhishek-Varma:
-- This commit adds duplicated pass `assign-buffer-address` and `assign-bd-ids`
   (already there in AIE dialect) to work on AMDAIE dialect ops.
-- Updates Tile/Buffer op's .td to add a few utilities.
-- This is being added to AMDAIE dialect to make [DMA reprogramming](https://github.com/nod-ai/iree-amd-aie/issues/1287) work.


compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignBufferAddress.cpp
```diff
@@ -0,0 +1,365 @@
+// Copyright 2025 The IREE Authors
```
Yu-Zhewen, line 1: I understand that duplicating these passes simplifies things, but from a maintenance perspective, I wonder if it might make sense to create some templated utility functions instead. Not necessarily for this PR.

jtuyls: Yeah, I agree we should see how we can avoid duplication, but I think it's ok initially if we add a todo as there are quite a few differences.

Yu-Zhewen (approved):
Nice! LGTM

PR #1333, opened 2025-07-18
[Reprogram] Add fix for handling Lof.from_memref + amdaie.buffer address attr as i32

Abhishek-Varma:
-- This commit adds a fix for handling logicalobjectfifo.from_memref to have a common logicalobjectfifo defined for any given connection as part of `create-aie-workgroup` pass.

-- This also changes `amdaie.buffer` op to have `address` attribute as `i32` instead of `ui32` to be compatible with `aie.buffer`.

-- This is being added to AMDAIE dialect to make [DMA reprogramming](https://github.com/nod-ai/iree-amd-aie/issues/1287) work.


Yu-Zhewen:
Just curious, since address is conceptually an unsigned value, why are we changing `amdaie.buffer` to use i32 instead of updating `aie.buffer` to use ui32 instead? Is that because air tests would fail in the latter case?

PR #1335, opened 2025-07-21
[Router] Fix packet group id assignment

Yu-Zhewen:
Background: The group ID is used inside the router, where only packet flows that share any source or destination port are assigned the same group ID. The router then restricts channel sharing to occur only within the same group, which helps prevent arbiter deadlocks.

Consider three packet flows: a, b, and c, where a and c have overlapping source or destination ports, and b is independent. The expected group IDs should be: 0, 1, 0.

However, the previous assignment logic was incorrect: `packetGroupId = std::max(packetGroupId, existingId);`, which produced the wrong results: 0, 1, 1.

This issue was observed when using out-of-order mode DMA for MatmulThinBias.

### DIFF
--- a/runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
+++ b/runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
@@ -273,33 +273,25 @@ void Router::addFlow(TileLoc srcCoords, Port srcPort, TileLoc dstCoords,
     }
   }
 
-  // Assign a group ID for packet flows
-  // any overlapping in source/destination will lead to the same group ID
-  // channel sharing will happen within the same group ID
-  // for circuit flows, group ID is always -1, and no channel sharing
-  int packetGroupId = -1;
+  // Assign a group ID to packet flows:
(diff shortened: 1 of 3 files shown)


runtime/src/iree-amd-aie/aie_runtime/iree_aie_router.cc
```diff
@@ -273,32 +273,23 @@ void Router::addFlow(TileLoc srcCoords, Port srcPort, TileLoc dstCoords,
...
+          })) {
+        return existingId;
       }
-      packetGroupId = std::max(packetGroupId, existingId);
-    }
-    if (!found) {
-      packetGroupId++;
     }
-  }
+    return nextAvailablePacketGroupId++;
```
jtuyls, line 291: As this is int8_t (which should be fine), maybe it's still good to put an assert here that it should not exceed max(int8_t) to better support the maximum number of groups changing in future generations. Or you could make this int32_t/int64_t.

jtuyls (approved):
LGTM, thanks!

PR #1336, opened 2025-07-21
[Reduction] Verify different sizes and types work for reduction.

Sameeranjoshi:
Patch adds tests to make sure all the tests with `vector.multi_reduction` generate successfully pass Peano legalizer and generate efficient vectorized code. This patch checks only the IREE side to keep the dependency minimun on Peano. (Depends on Peano:
1. https://github.com/Xilinx/llvm-aie/pull/548
2. https://github.com/Xilinx/llvm-aie/pull/557 )

1. `reassociateFpReductions=true` is must else code is scalarized. This flag could be added into the IREE vectorization pipeline to trigger automatically.
2. bf16/i32/f32 all types with different sizes work now.


newling (commented):
These tests are nice, but I think we should see what our compiler flow is once we've done the unrolling of transfer_reads. 

Ideally we'd be able to have FileCheck commands for the generated assembly, to verify that the legalization changes are working as expected too. But that'd presumably add a dependency on peano which we don't want in our lit testing

PR #1340, opened 2025-07-28
[AIERT] Configure DMA channel enable only for AIE1

Yu-Zhewen:
On later-generation devices (aie2, aie2p, ...), the channel enable bit is no longer used. As a result, invoking this API may unintentionally overwrite the channel control register, as a MaskWrite64 with an all-zero mask is "incorrectly" treated as a Write64 operation by aie-rt.


jtuyls (approved):
Nice!

PR #1341, opened 2025-07-29
[MatmulThinBias] Enable out-of-order DMA mode

Yu-Zhewen:
MatmulThinBias was previously only supported through the AIR pipeline, not the ObjectFifo pipeline. The main challenge is that each core requires three input tensors, while the hardware provides only two DMA S2MM (input) channels.

In the AIR pipeline, this limitation was addressed by concatenating the bias tensor with another input to reduce channel usage.

This PR introduces support for the ObjectFifo pipeline by leveraging out-of-order DMA mode instead, which allows multiple MM2S channels to feed data into a single S2MM channel.  When an MM2S channel sends data, it embeds the corresponding BD ID, which it wants the destination S2MM to trigger, directly into the packet header of the data stream.

A single-core example has been added to the CI. Support for larger shapes and MatmulFullBias is still blocked at the ObjFifoSplitting stage and will be addressed in follow-up PRs.

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIETransactionBuilder.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIETransactionBuilder.cpp
@@ -198,8 +198,8 @@ LogicalResult TransactionBuilder::appendDmaStartOp(
   // Configure push to BD queue.
   // TODO: Generalize it as this is currently hardcoded to only shim side for
   // now.
-  AMDAIE::DMABDOp dmaBdOp =
-      *dmaStartOp.getBody().getOps<AMDAIE::DMABDOp>().begin();
+  if (dmaBdOps.empty()) return dmaStartOp.emitError("has no DMA BD ops");
+  AMDAIE::DMABDOp dmaBdOp = *dmaBdOps.begin();
   bool issueToken = tileLoc.Row == 0 && channelDir == DMAChannelDir::MM2S;
   if (failed(configurePushToBdQueue(
           deviceModel, tileLoc, chNum, channelDir, dmaBdOp.getBdId().value(),
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_dma_out_of_order_block.mlir
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/test/insert_dma_out_of_order_block.mlir
@@ -21,21 +21,23 @@ module {
 // CHECK:       amdaie.workgroup {
 // CHECK:         %[[TILE_0_1:.+]] = amdaie.tile(%[[C0]], %[[C1]])
 // CHECK:         %[[TILE_0_2:.+]] = amdaie.tile(%[[C0]], %[[C2]])
-// CHECK:         %[[CHANNEL_0:.+]] = amdaie.channel(%[[TILE_0_1]], 0, port_type = DMA, direction = MM2S)
-// CHECK:         %[[CHANNEL_1:.+]] = amdaie.channel(%[[TILE_0_1]], 1, port_type = DMA, direction = MM2S)
-// CHECK:         %[[CHANNEL_2:.+]] = amdaie.channel(%[[TILE_0_2]], 0, port_type = DMA, direction = S2MM)
-// CHECK:         amdaie.flow({%[[CHANNEL_0]]} -> {%[[CHANNEL_2]]}) {is_packet_flow = true, keep_pkt_header = true, packet_id = 0 : ui8}
-// CHECK:         amdaie.flow({%[[CHANNEL_1]]} -> {%[[CHANNEL_2]]}) {is_packet_flow = true, keep_pkt_header = true, packet_id = 1 : ui8}
+// CHECK:         %[[CHANNEL:.+]] = amdaie.channel(%[[TILE_0_1]], 0, port_type = DMA, direction = MM2S)
+// CHECK:         %[[CHANNEL_15:.+]] = amdaie.channel(%[[TILE_0_1]], 1, port_type = DMA, direction = MM2S)
+// CHECK:         %[[CHANNEL_16:.+]] = amdaie.channel(%[[TILE_0_2]], 0, port_type = DMA, direction = S2MM)
+// CHECK:         amdaie.flow({%[[CHANNEL]]} -> {%[[CHANNEL_16]]}) {is_packet_flow = true, keep_pkt_header = true, packet_id = 0 : ui8}
+// CHECK:         amdaie.flow({%[[CHANNEL_15]]} -> {%[[CHANNEL_16]]}) {is_packet_flow = true, keep_pkt_header = true, packet_id = 1 : ui8}
 // CHECK:         amdaie.controlcode {
-// CHECK:           %[[DMA_START_0:.+]] = amdaie.dma_start(%[[CHANNEL_0]], {%{{.+}}}) {
+// CHECK:           %[[DMA_START_0:.+]] = amdaie.dma_start(%[[CHANNEL]], {%{{.+}}}) {
 // CHECK:             amdaie.dma_bd_packet {out_of_order_bd_id = 0 : i32, packet_id = 0 : i32, packet_type = 0 : i32}
+// CHECK:             amdaie.dma_bd_packet {out_of_order_bd_id = 1 : i32, packet_id = 0 : i32, packet_type = 0 : i32}
 // CHECK:           }
-// CHECK:           %[[DMA_START_1:.+]] = amdaie.dma_start(%[[CHANNEL_2]], {%{{.+}}}) {
-// CHECK-COUNT-2:     amdaie.dma_bd(
+// CHECK:           %[[DMA_START_1:.+]] = amdaie.dma_start(%[[CHANNEL_16]], {%{{.+}}}) {
+// CHECK-COUNT-4:     amdaie.dma_bd(
 // CHECK-NOT:         amdaie.dma_bd(
-// CHECK:           } {enable_out_of_order = true, repeat_count = 2 : i8}
-// CHECK:           %[[DMA_START_2:.+]] = amdaie.dma_start(%[[CHANNEL_1]], {%{{.+}}}) {
-// CHECK:             amdaie.dma_bd_packet {out_of_order_bd_id = 1 : i32, packet_id = 1 : i32, packet_type = 0 : i32}
+// CHECK:           } {enable_out_of_order = true, repeat_count = 4 : i8}
+// CHECK:           %[[DMA_START_2:.+]] = amdaie.dma_start(%[[CHANNEL_15]], {%{{.+}}}) {
+// CHECK:             amdaie.dma_bd_packet {out_of_order_bd_id = 2 : i32, packet_id = 1 : i32, packet_type = 0 : i32}
+// CHECK:             amdaie.dma_bd_packet {out_of_order_bd_id = 3 : i32, packet_id = 1 : i32, packet_type = 0 : i32}
 // CHECK:           }
 #executable_target_amdaie_pdi_fb = #hal.executable.target<"amd-aie", "amdaie-pdi-fb", {target_device = "npu1_4col", ukernels = "none"}>
 module attributes {hal.executable.target = #executable_target_amdaie_pdi_fb} {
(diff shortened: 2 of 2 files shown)


compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/test/add_12_i8_using_2d_dma_op_with_padding.mlir
```diff
@@ -38,7 +38,7 @@ module {
 // CHECK: XAIE API: XAie_DmaSetNextBd with args: &dmaDesc=ptr, nextBdId.value()=0, enableNextBd=1
 // CHECK: XAIE API: XAie_DmaEnableBd with args: &dmaDesc=ptr
 // CHECK: XAIE API: XAie_DmaWriteBd with args: devInst=ptr, &dmaDesc=ptr, tileLoc=TileLoc(col: 0, row: 1), bdId=0
-// CHECK: XAIE API: XAie_DmaChannelSetStartQueue with args: devInst=ptr, tileLoc=TileLoc(col: 0, row: 1), chNum=0, direction=0, bdId=0, repeatCount=1, enTokenIssue=0
+// CHECK: XAIE API: XAie_DmaChannelSetStartQueueGeneric with args: devInst=ptr, tileLoc=TileLoc(col: 0, row: 1), chNum=0, direction=0, &dmaQueueDesc=ptr
```
Yu-Zhewen, line 41: We had to switch to a different API in order to configure out-of-order DMA. As a result, parameters like bd_id, repeatCount, and enTokenIssue are no longer directly visible in the lit checks, since they are now passed via a dmaQueueDesc pointer.

This shouldn't be a major issue, as these aie-rt API calls are already validated indirectly through the generated binary in the controlcode_to_transaction checks.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertDmaOutOfOrderBlock.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+LogicalResult insertDmaOutOfOrder(
+    IRRewriter &rewriter, AMDAIE::ChannelOp s2mmChannelOp,
+    SmallVector<AMDAIE::DMAStartOp> &s2mmDmaStartOps,
+    DenseMap<AMDAIE::ChannelOp, SmallVector<AMDAIE::DMAStartOp>>
+        &mm2sChannelToDmaStarts) {
+  // If there is only one DMA start op, do nothing.
+  if (s2mmDmaStartOps.size() <= 1) return success();
+  // Merge the given DMA start ops into a single out-of-order block.
+  int8_t newRepeatCount = 0;
```
jtuyls, line 48: I think repeat count can in principle go up to 256? I think it would be better to use an uint32_t and then check for overflow down below before creating the DMAStartOp.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertDmaOutOfOrderBlock.cpp
```diff
@@ -0,0 +1,202 @@
...
+             << "could not find corresponding MM2S DMA start op";
+    }
+    // Get the blocks in the S2MM and MM2S DMA start ops.
+    SmallVector<Block *> s2mmBlocks = llvm::to_vector(llvm::map_range(
+        s2mmDmaStartOp.getBody().getBlocks(), [](Block &b) { return &b; }));
+    SmallVector<Block *> mm2sBlocks = llvm::to_vector(
+        llvm::map_range(maybeMm2sDmaStartOp.value().getBody().getBlocks(),
+                        [](Block &b) { return &b; }));
+    int8_t numBDs = 0;
+    for (size_t i = 0; i < s2mmBlocks.size(); i++) {
```
jtuyls, line 83: I think you could just iterate over the blocks?
```suggestion
    for (Block *block : s2mmBlocks) {
```
And could you add a comment explaining this block of logic?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertDmaOutOfOrderBlock.cpp
```diff
@@ -0,0 +1,202 @@
...
+      return s2mmDmaStartOp.emitOpError()
+             << "could not find corresponding MM2S DMA start op";
+    }
+    // Get the blocks in the S2MM and MM2S DMA start ops.
+    SmallVector<Block *> s2mmBlocks = llvm::to_vector(llvm::map_range(
+        s2mmDmaStartOp.getBody().getBlocks(), [](Block &b) { return &b; }));
+    SmallVector<Block *> mm2sBlocks = llvm::to_vector(
+        llvm::map_range(maybeMm2sDmaStartOp.value().getBody().getBlocks(),
+                        [](Block &b) { return &b; }));
+    int8_t numBDs = 0;
```
jtuyls, line 82: Same here, use larger and unsigned int and check for overflow before using. The max number of BDs should ideally come from the device model.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertDmaOutOfOrderBlock.cpp
```diff
@@ -0,0 +1,202 @@
...
+           << "source channel does not have corresponding MM2S DMA start op";
+  }
+  if (mm2sChannelToDmaStarts[channelOp].size() != 1) {
+    return connectionOp.emitOpError()
+           << "expected exactly one MM2S DMA start op";
+  }
+  return mm2sChannelToDmaStarts[channelOp][0];
+}
+
+LogicalResult insertDmaOutOfOrder(
```
jtuyls, line 40: Could you add a docstring?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertDmaOutOfOrderBlock.cpp
```diff
@@ -0,0 +1,202 @@
...
+      for (Operation &op : *s2mmBlocks[i]) {
+        // Find the DMA BDs and locks in the S2MM block.
+        if (isa<AMDAIE::DMABDOp, AMDAIE::UseLockOp>(op))
+          lockAndBdOps.back().push_back(&op);
+        if (auto bdOp = dyn_cast<AMDAIE::DMABDOp>(op)) {
+          numBDs++;
+          maybeS2mmBdId = bdOp.getBdId();
+        }
+      }
+      if (lockAndBdOps.back().size() == 0) {
```
jtuyls, line 95: Maybe add a comment that no `DMABDOp`s are found, so we continue to the next block?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertDmaOutOfOrderBlock.cpp
```diff
@@ -0,0 +1,202 @@
...
+               << "BD ID is not assigned in S2MM block";
+      }
+      // Embed the S2MM BD ID in the corresponding MM2S packer header.
+      auto maybeMm2sDmaBdPacketOps =
+          mm2sBlocks[i]->getOps<AMDAIE::DmaBdPacketOp>();
+      if (llvm::range_size(maybeMm2sDmaBdPacketOps) != 1) {
+        return maybeMm2sDmaStartOp.value().emitOpError()
+               << "expected exactly one `amdaie.dma_bd_packet` op in each MM2S "
+                  "block";
+      }
```
jtuyls, line 110: Maybe this could be put in the DMAStartOp verifier? And maybe the DMAStartOp could then have a utility to retrieve the `amdaie.dma_bd_packet` ops? This could potentially simplify the logic here and above. WDYT?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertDmaOutOfOrderBlock.cpp
```diff
@@ -0,0 +1,202 @@
...
+           << "expected exactly one MM2S DMA start op";
+  }
+  return mm2sChannelToDmaStarts[channelOp][0];
+}
+
+LogicalResult insertDmaOutOfOrder(
+    IRRewriter &rewriter, AMDAIE::ChannelOp s2mmChannelOp,
+    SmallVector<AMDAIE::DMAStartOp> &s2mmDmaStartOps,
+    DenseMap<AMDAIE::ChannelOp, SmallVector<AMDAIE::DMAStartOp>>
+        &mm2sChannelToDmaStarts) {
```
jtuyls, line 44: I think it would be good to add an InsertionGuard on the rewriter at the beginning of this function.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertDmaOutOfOrderBlock.cpp
```diff
@@ -0,0 +1,202 @@
...
+      rewriter.getUnknownLoc(), s2mmChannelOp, newConnections, newRepeatCount,
+      /*enOutOfOrder=*/rewriter.getBoolAttr(true));
+  IRMapping mapper;
+  for (SmallVector<Operation *> ops : lockAndBdOps) {
+    rewriter.setInsertionPointToStart(
+        &newS2mmDmaStartOp.getRegion().emplaceBlock());
+    for (Operation *op : ops) {
+      rewriter.clone(*op, mapper);
+      // Break the existing chain.
+      if (auto bdOp = dyn_cast<AMDAIE::DMABDOp>(op))
```
jtuyls, line 131: Why is this needed?

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIETransactionBuilder.cpp
```diff
@@ -97,90 +97,112 @@ LogicalResult TransactionBuilder::appendDmaStartOp(
...
-    if (useLockOp.getAction() == AMDAIE::LockAction::AcquireGreaterOrEqual ||
-        useLockOp.getAction() == AMDAIE::LockAction::Acquire) {
-      acqValue = useLockOp.getValue();
-      if (useLockOp.getAction() == AMDAIE::LockAction::AcquireGreaterOrEqual)
-        acqValue.value() = -acqValue.value();
-      acqLockId = lockOp.getValue();
-    } else if (useLockOp.getAction() == AMDAIE::LockAction::Release) {
-      relValue = useLockOp.getValue();
-      relLockId = lockOp.getValue();
+  WalkResult res = dmaStartOp->walk([&](AMDAIE::DMABDOp dmaBdOp) {
```
jtuyls, line 110: You could also just walk to collect the DMABDOps in a vector and then iterate over the vector. Then you can keep returning failures instead of the advance and interrupt.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_configure.cc
```diff
@@ -198,22 +199,44 @@ LogicalResult configureDMABD(
...
+  }
+
   auto devInst = const_cast<XAie_DevInst *>(&deviceModel.devInst);
   TRY_XAIE_API_LOGICAL_RESULT(XAie_DmaWriteBd, devInst, &dmaDesc, tileLoc,
                               bdId);
   return success();
 }
 
+LogicalResult configureOutofOrderMode(const AMDAIEDeviceModel &deviceModel,
+                                      const TileLoc &tileLoc, uint8_t chNum,
```
jtuyls, line 214: It's safer to make the channel number a larger int because above `int` is used and then check here whether it exceeds the maximum before passing to aiert.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertDmaOutOfOrderBlock.cpp
Yu-Zhewen: Need to iterate on both the S2MM and MM2S sides, so changed to `for (auto &&[s2mmBlock, mm2sBlock] : llvm::zip(s2mmBlocks, mm2sBlocks))` instead

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertDmaOutOfOrderBlock.cpp
Yu-Zhewen: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertDmaOutOfOrderBlock.cpp
Yu-Zhewen: Done.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertDmaOutOfOrderBlock.cpp
Yu-Zhewen: Assume there are two `dma_start` ops targeting the same channel, and each contains a DMA chain of length 2, resulting in a total of 4 BDs.

When these are merged into a single out-of-order block, all individual chains should be erased, and we can simply set repeat_count = 4 (assuming the transfer is non-cyclic).

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIETransactionBuilder.cpp
Yu-Zhewen: Done.

runtime/src/iree-amd-aie/aie_runtime/iree_aie_configure.cc
Yu-Zhewen: Add the deviceModel check.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
```diff
@@ -1762,7 +1762,8 @@ def AMDAIE_DMAStartOp: AMDAIE_Op<"dma_start", [
     ins Index:$channel,
         Variadic<Index>:$connections,
         // `repeat_count==1` means "do it once".
-        DefaultValuedAttr<I8Attr, "1">:$repeat_count
+        DefaultValuedAttr<I8Attr, "1">:$repeat_count,
```
jtuyls, line 1765: I know this was there before but this should be at least UI8Attr to accommodate repeat counts up to 256 and better even UI32Attr to accommodate potential future hardware changes. Not necessary to address in this PR though.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertDmaOutOfOrderBlock.cpp
jtuyls: There is no lit test case for this right? I think it would be good to add to make sure this logic is tested.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/Utils/AMDAIETransactionBuilder.cpp
```diff
@@ -97,90 +97,113 @@ LogicalResult TransactionBuilder::appendDmaStartOp(
...
-                            dmaBdOp.getLenInBytes(), dmaBdOp.getOffsetInBytes(),
-                            dmaBdOp.getBufferElementTypeWidthInBytes(),
-                            maybeDims, maybePadDims, maybeIter))) {
-    return failure();
   }
 
   // Configure push to BD queue.
   // TODO: Generalize it as this is currently hardcoded to only shim side for
   // now.
+  AMDAIE::DMABDOp dmaBdOp =
```
jtuyls, line 201: It would be good to check that `dmaStartOp.getBody().getOps<AMDAIE::DMABDOp>()` is not empty.

compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEInsertDmaOutOfOrderBlock.cpp
Yu-Zhewen: Change the original lit test to the double-buffered version instead.

compiler/plugins/target/AMD-AIE/iree-amd-aie/IR/AMDAIEOps.td
Yu-Zhewen: Yeah, a proper cleanup needs to be done at some point :)

jtuyls (approved):
LGTM, very nice!

PR #1362, opened 2025-10-06
Bump IREE to iree-org/iree@7cdfc13

Abhishek-Varma:


Abhishek-Varma:
The issue is that the path being formed on Windows is INCORRECT :-
```
D:/a/iree-amd-aie/iree-amd-aie/iree-build/test_tmpdir/iree/target/amd-aie/Target/test/add_12_i8_using_2d_dma_op_with_padding.mlir.test_test_tmpdir/compiler/plugins/target/AMD-AIE/iree-amd-aie/Target/test/Output/add_12_i8_using_2d_dma_op_with_padding.mlir.tmpdir\xyz.bin 
```
Observe `\xyz.bin` whereas it should've been `/xyz.bin`
Using `std::filesystem::path` seems to not form POSIX compliant paths. :/

I've added a `llvm::sys::path::convert_to_slash` to address that in the latest push.

yzhang93:
@Abhishek-Varma Could you remove all debug prints in your PR and merge it? You can continue debugging the windows issue afterwards. We don't want the windows failure to keep blocking the linux bump. CC: @jtuyls

jtuyls (changes requested):
I think we should first fix Windows. There's no point in bumping if we drop support for it.

PR #1400, opened 2026-05-08
[xrt-lite] Move HAL queue ops to async submission and pass remaining CTS tests

jtuyls:
The xrt-lite driver wrapped the AMD XDNA NPU kernel synchronously: every queue_alloca / queue_execute / queue_dispatch blocked the calling thread on iree_hal_semaphore_list_wait before doing work, several HAL command-buffer ops were missing entirely (fill_buffer, signal_event, reset_event, wait_events), the device had no event impl, and query_queue_pool_backend was a NULL slot in the vtable. Together these caused four CTS regressions: core_tests EventTest cases failed, command_buffer_tests segfaulted on the first fill, queue_tests segfaulted on BasicAlloca, and QueueAllocaTest.AllocaWithWaitSemaphores deadlocked because the driver synchronously waited on a semaphore that the test signaled later from the same thread.

This change moves xrt-lite to async submission (host-side defer of queue ops via a per-device async work queue + worker thread, mirroring the iree/hal/drivers/amdgpu pending-op pattern at smaller scope), fills in the missing HAL plumbing, and closes the CTS gaps.

HAL plumbing
- direct_command_buffer.cc: implement fill_buffer (memset fast path for pattern_length=1, byte-loop otherwise; matches the update_buffer mapping pattern). Wire signal_event, reset_event, wait_events to unimplemented_ok_status since the HAL is synchronous and command-buffer ordering is implicit at submission time.
- nop_event.{cc,h}: minimal iree_hal_event_t (resource + allocator only, no signal/reset state), mirroring iree/hal/drivers/local_task/task_event.c. device.cc create_event delegates to it.
- allocator.cc: add import_buffer / export_buffer = unimplemented stubs so the memory_file fallback path no longer dereferences a NULL fn pointer. The XDNA kernel ABI has no userptr / wrap-host-pointer ioctl, so the fallback (iree_hal_memory_file_wrap → HOST_LOCAL heap buffer) is the correct semantics for HOST_ALLOCATION imports.
- device.cc: file_import delegates to iree_hal_file_from_handle (the shared iree_hal_memory_file path), profiling_begin/flush/end return UNIMPLEMENTED so CTS profiling-dependent tests skip via IsProfilingUnsupported instead of treating the no-op as success, queue_alloca tags buffers with IREE_HAL_BUFFER_PLACEMENT_FLAG_ASYNCHRONOUS + a single-bit queue_affinity selection, multi-shot command buffers no longer return UNIMPLEMENTED.

Async work queue
- async_queue.{h,cc}: per-device task scheduler. Arena-allocated pending_op_t with one iree_async_semaphore_acquire_timepoint per wait semaphore; an atomic wait_count decrements per fired callback and pushes the op to a single-thread worker when it hits zero. Worker runs op_fn, signals or fails signal_list, optionally advances a frontier-tracker epoch. Shutdown walks an inflight list and cancels still-pending timepoints via iree_async_semaphore_cancel_timepoint, force-failing their signal lists with CANCELLED so destroy never hangs.
- device.cc queue_alloca now allocates the buffer synchronously (caller needs it back) but defers signal_list via the async queue. queue_dealloca enqueues an op that marks the buffer deallocated when its waits fire. query_queue_pool_backend is wired to a CPU slab provider + proactor notification + frontier-tracker-backed epoch query; assign_topology_info registers the device's queue axis with the runtime's frontier tracker.
- buffer.{h,cc}: atomic deallocated flag with mark/check helpers; queue_fill rejects ops on deallocated buffers with INVALID_ARGUMENT so QueueAllocaTest.DeallocaReleasesMemory passes.

CTS-suite registration
- cts/backends.cc: register the four cross-queue / wait-frontier QueueAllocaTest cases as unsupported_tests with a single-queue rationale (matches iree/hal/drivers/local_sync's list). TODO points to the multi-queue benchmark plan in iree-amd-aie-workspace/todo.md.

Tests
- async_queue_test.cc: direct unit tests covering empty waits, the wait-then-signal-on-same-thread pattern, multi-wait / multi-signal, op_fn errors, shutdown with an inflight op (cancel path), and a many-enqueue stress test. 6/6 PASS.

CTS deltas (xrt_lite backend, before this branch / after):

  buffer_tests:           19 PASS, 2 SKIP                 / unchanged
  core_tests:             24 PASS, 3 FAIL                 / 27 PASS, 0 FAIL
  command_buffer_tests:   SEGFAULT after 4 tests          / 36 PASS, 0 FAIL
  queue_tests:            SEGFAULT on BasicAlloca + 1 hang/ 39 PASS, 0 FAIL,
                                                            5 SKIP (4 multi-queue
                                                            + 1 profiling),
                                                            AllocaWithWaitSemaphores
                                                            PASSES in isolation
  file_tests:             0 by design                     / unchanged
  xrt_lite_executable_cache_test: 3 PASS                  / unchanged
  xrt_lite_dispatch_test: 4 PASS (npu1 only)              / unchanged

### DIFF
--- a/runtime/src/iree-amd-aie/driver/xrt-lite/cts/backends.cc
+++ b/runtime/src/iree-amd-aie/driver/xrt-lite/cts/backends.cc
@@ -1,4 +1,4 @@
-// Copyright 2025 The IREE Authors
+// Copyright 2026 The IREE Authors
 //
 // Licensed under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
@@ -48,7 +48,44 @@ static iree_status_t CreateXrtLiteDevice(
 static bool xrt_lite_registered_ =
     (CtsRegistry::RegisterBackend({
          "xrt_lite",
-         {"xrt_lite", CreateXrtLiteDevice},
+         {"xrt_lite",
+          CreateXrtLiteDevice,
+          /*executable_format=*/nullptr,
+          /*executable_data=*/nullptr,
+          RecordingMode::kDirect,
+          /*unsupported_tests=*/
+          {
+              // xrt-lite has a single XDNA hwctx and a single hardware
+              // queue. The four tests below assume cross-queue release/
+              // alloca interleaving or pool-notification retry semantics
+              // that are only meaningful for drivers with multiple physical
+              // queues. local_sync skips them for the same reason.
+              //
+              // TODO(1401): revisit when a real workload measurably
+              // benefits from parallel hwctx submission on disjoint AIE
+              // column groups. The XDNA kernel and IREE HAL both support
+              // multi-queue today; xrt-lite would need ~500-800 LOC
+              // ported from the amdgpu HAL pattern to enable it. The
+              // linked issue tracks the benchmark plan and decision
+              // criteria.
+              {"QueueAllocaTest.ExplicitFixedBlockPoolCrossQueueWaitFrontier",
+               "single-queue driver: pool's OK_NEEDS_WAIT path is only "
+               "meaningful when peer queues release while the freed work is "
+               "still in flight on another queue."},
+              {"QueueAllocaTest.ExplicitFixedBlockPoolRequiresWaitFrontierFlag",
+               "single-queue driver: cannot distinguish async queue-owned "
+               "hidden frontier waits from pool-notification retries when "
+               "all alloca/dealloca ordering is on one physical queue."},
+              {"QueueAllocaTest.ExplicitTLSFPoolCrossQueueStaleBlockGrows",
+               "single-queue driver: stale cross-queue block growth requires "
+               "a peer queue still holding the released reservation, which "
+               "this backend cannot model."},
+              {"QueueAllocaTest.ExplicitFixedBlockPoolNotificationRetry",
+               "single-queue driver: cannot submit the dealloca that "
+               "releases the first block while the second alloca is waiting "
+               "on pool notification on the same physical queue."},
+          },
+          /*expected_failures=*/{}},
          /*tags=*/{"allocator", "buffer_mapping", "driver"},
      }),
      true);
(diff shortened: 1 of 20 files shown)


runtime/src/iree-amd-aie/driver/xrt-lite/cts/backends.cc
```diff
@@ -48,7 +48,44 @@ static iree_status_t CreateXrtLiteDevice(
...
+              // alloca interleaving or pool-notification retry semantics
+              // that are only meaningful for drivers with multiple physical
+              // queues. local_sync skips them for the same reason.
+              //
+              // TODO(multi-queue): revisit when a real workload measurably
+              // benefits from parallel hwctx submission on disjoint AIE
+              // column groups. The XDNA kernel and IREE HAL both support
+              // multi-queue today; xrt-lite would need ~500-800 LOC ported
+              // from the amdgpu HAL pattern to enable it. See
+              // iree-amd-aie-workspace/todo.md for the benchmark plan and
```
Yu-Zhewen: `iree-amd-aie-workspace/todo.md` looks like a leftover reference from an agent run :) Could you file a GitHub issue for the multi-queue plan and link that instead?

runtime/src/iree-amd-aie/driver/xrt-lite/device.cc
```diff
@@ -313,37 +412,106 @@ static iree_status_t iree_hal_xrt_lite_device_queue_alloca(
...
+  // Tag the buffer with async placement so callers can route dealloca back to
+  // this device/queue.
+  //
+  // The CTS BufferMetadata test asserts placement.queue_affinity has exactly
+  // one bit set (so callers can use it as a target queue id directly). The
+  // caller passes IREE_HAL_QUEUE_AFFINITY_ANY which is all-bits, so we
+  // collapse to a single bit here. Today xrt-lite reports a single physical
+  // queue; the lowest-set-bit pick is therefore both deterministic and
+  // correct. A multi-queue redesign (see
+  // iree-amd-aie-workspace/todo.md "Multi-queue (multi-hwctx) benchmark")
```
Yu-Zhewen: Same here. The reference to `iree-amd-aie-workspace/todo.md` shall be removed

runtime/src/iree-amd-aie/driver/xrt-lite/cts/backends.cc
jtuyls: Thanks for catching! Updated it and filed an issue here: https://github.com/nod-ai/iree-amd-aie/issues/1401

Yu-Zhewen (approved):
LGTM

PR #1407, opened 2026-05-29
[InsertDmaBdChain][AssignNpuDmaBdIds] Fix packet-flow chain ordering + BD-id wrap

jtuyls:
Two independent correctness fixes exposed by hand-authored / loop-subsumed controlcode (the LOF pipeline), but general to both passes:

1. AMDAIEInsertDmaBdChain: a BD chain defers every queue-push except the chain tail's (ControlCodeLowering), so building chains interleaved reorders their pushes. Packet flows share stream-switch arbiters and their global push order is observable, so interleaved packet-flow chains corrupt the output. Treat a packet-flow DMA as a sequence point: close all open chains when one is reached (circuit flows, which have dedicated routes, still interleave freely). This re-establishes the non-interleaving invariant from #1210 that #1248 relaxed in favor of `barrier` ops -- but those are only inserted for control packets, so data packet flows regressed. Tests: @two_connections_packet, @circuit_chain_broken_by_packet.

2. AMDAIEAssignNpuDmaBdIds: the affine BD-id expression `iv % size + offset` indexes loop iterations into the contiguous id block [offset, offset+size); it reserved them with getAndAssignBdId(Incremental), which can wrap the channel's valid range (e.g. [15,0,1,2,3] on a shim's 0..15 pool) and emit out-of-range ids (XAie "Invalid Next Bd"/"Invalid BD number"). Add ChannelBdIdGenerator::getAndAssignConsecutiveBdIds to reserve a consecutive run (shorter under fragmentation -> fewer distinct ids, still correct). Unit tests incl an anti-wrap regression case.

HW-validated on npu4 across i32 + bf16 fused matmul-chain sweeps (subsume-loops on and off, grids up to 8x4 incl the SmolLM FFN [64,576,1536,576]).

### DIFF
--- a/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
+++ b/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEAssignNpuDmaBdIds.cpp
@@ -392,25 +392,40 @@ class BdIdAssignmentUtil {
       // Only create expression if more than 1 BD ID is needed and if,
       // otherwise, fall back to constant BD ID.
       if (size > 1) {
-        // Assigning BD IDs for all iterations in the loop.
-        SmallVector<uint32_t> bdIds;
-        for (uint32_t i = 0; i < size; i++) {
-          std::optional<uint32_t> bdId = generator.getAndAssignBdId(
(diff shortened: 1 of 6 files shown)


Abhishek-Varma (approved):
LGTM!

PR #1416, opened 2026-06-03
[LogicalObjectFifoFromMemrefOp ] Fix canonicalize's convergence

Abhishek-Varma:
The canonicalize pattern for LogicalObjectFifoFromMemrefOp was returning success() in two no-op cases (empty tile list, already-sorted-and-deduped tiles) without performing any IR rewrite. The MLIR greedy driver treats a success() return as "rewrite happened" and re-enqueues the op forever, causing applyPatternsGreedily to either burn through its iteration cap or silently diverge (observed as iree-util-fold-globals never terminating on placed-LOF modules).
Fix both cases to return failure(). Add a lit test with test-convergence to catch regressions for the empty-tiles, already-canonical, unsorted, and duplicate-tiles cases.


jtuyls (commented):
LGTM, nice!

